PR #13, opened 2017-10-10
test #10: add database testing infrastructure; tests and fixes for Base.

### LINKED ISSUE #10: Testing scheme for database calls
Clint writes:

> - Devise a testing scheme for &hellip; database calls that elude straightforward unit testing.
> - Unit test more/all of the surface area we have imported already.

You could also split this into two issues if that's easier&hellip;

clint-tseng:
* the DatabaseShim is the main harness for testing knex’s database
  chaining calls.
  * our Model (Base-derived) objects all use dependency injection to
    actually receive their database objects to work with. What the
    DatabaseShim does is create a mock knex object with all the knex and
    Promise methods mocked out for chaining; but instead of actually
    performing any database actions, they simply record what they were
    called with and move on.
  * the Shim itself provides access to these recorded method calls, and
    a bunch of convenience getters and assertions for pulling out the
    relevant calls and parameters for testing.
* it is a bit of a fine balance between testing the appropriate pieces
  of logic for the database calls and over-testing fairly obvious code,
  but in general expected behaviour should be tested (returning objects
  of the correct type, performing various transformations on the data,
  etc) so that if/as the project grows more complex, our basic contracts
  are not violated.
* closes #10; although more work is needed to expand coverage on extant
  model code.

### DIFF
--- a/test/model/base.js
+++ b/test/model/base.js
@@ -1,0 +1,211 @@
+const should = require('should');
+const { DatabaseShim } = require('./package');
+
+const Base = require('../../lib/model/base');
+
+describe('Base model', () => {
+  describe('create', () => {
+    it('should insert into the appropriate table specified by tableName', () => {
+      const [ TestBase, shim ] = DatabaseShim.shim(Base);
+      const TestSubclass = class extends TestBase {
+        static _tableName() { return 'test_table'; }
+      }
+
+      const model = new TestSubclass();
+      model.create();
+
+      shim.should.be.calledOnceWith('into', 'test_table');
+    });
+
+    it('should save the provided data', () => {
+      const [ TestBase, shim ] = DatabaseShim.shim(Base);
+      const data = { a: 4, b: 8, c: 15, d: 16 };
+
+      const model = new TestBase(data);
+      model.create();
+
+      // check that all k/v pairs of our given data are present.
+      const [ insertedData ] = shim.paramsForCall('insert');
+      for (const key in data)
+        insertedData[key].should.equal(data[key]);
+    });
+
+    it('should inject the creation date', () => {
+      const [ TestBase, shim ] = DatabaseShim.shim(Base);
+
+      const model = new TestBase({ e: 23, f: 42 });
+      model.create();
+
+      const [ insertedData ] = shim.paramsForCall('insert');
+      insertedData.should.have.property('createdAt');
+      insertedData.createdAt.should.be.an.instanceof(Date);
+    });
+
+    it('should return a new instance of the appropriate subclass', () => {
+      const [ TestBase, shim ] = DatabaseShim.shim(Base);
+      const TestSubclass = class extends TestBase {
+        paranoia() { return true; }
+      }
+
+      const model = new TestSubclass();
+      model.create();
+
+      const [ promiseCallback ] = shim.paramsForCall('then');
+      promiseCallback.should.be.a.Function;
+
+      const returnedModel = promiseCallback([]);
+      returnedModel.should.be.an.instanceof(TestSubclass);
+    });
+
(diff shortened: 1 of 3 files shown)


test/model/package.js
```diff
@@ -0,0 +1,75 @@
...
+const shim = (methodName) => function(...args) {
+  this.__calls.push([ methodName, args ])
+  return this;
+};
+for (const method in knex)
+  if (typeof knex[method] === 'function')
+    knex[method] = shim(method);
+
+// Also, insert Promise methods.
+for (const method of [ 'then' ])
```
lognaturel: Wow, `for... of`... that's fun!

test/model/package.js
```diff
@@ -0,0 +1,75 @@
...
+// expressions easier. But it may grow to include more database testing
+// utilities eventually.
+
+const should = require('should');
+
+// Create a standard knex object, then replace all methods we find with a test
+// harness.
+const knex = require('knex')({ client: 'pg' });
+const shim = (methodName) => function(...args) {
+  this.__calls.push([ methodName, args ])
```
lognaturel: noob questions -- how exactly is `this` bound here? Why the double underscore?

clint-tseng:
Ha, okay. Three rules about `this` in JS:

First, `this` is bound directly by caller dot notation. So for instance:

    const foo = { bar: function() { console.log(this.baz); }, baz: 42 };
    foo.bar() // prints 42

but,

    const bar2 = foo.bar;
    bar2(); // prints undefined, because this is bound to window instead of foo

and `window` has no `.baz` property. Though in nodejs, it's `global` instead of `window`.

Second, you can bind manually when invoking a function:

    bar2.call(foo); // prints 42
    bar2.apply(foo); // prints 42

You can look up the difference between `call` and `apply` on MDN; they pertain to passed arguments. Though with es6 you should likely use the `...` spread operator instead.

As a corollary to the first two observations, you should be extremely careful with nested functions:

    class A {
      doSomething() {
        this.someMethod(); // works
        this.on('someEvent', function() {
          this.someMethod(); // dangerous! you don't know how this callback is called!
        });
      }
    }

But this is somewhat easily solved (though in my opinion confusingly so) by the third rule, which is that lambda functions _never_ rebind `this`; they are always bound against the closure lexical scope:

    class A {
      doSomething() {
        this.someMethod(); // works
        this.on('someEvent', () => {
          this.someMethod(); // totally safe
        });
      }
    }

clint-tseng:
Oh. And a clarification on "caller dot notation":

    const foo = { bar: { baz: function() { console.log(this.magicNumber); }, magicNumber: 2 }, magicNumber: 1 };
    foo.bar.baz() // prints 2

test/model/package.js
clint-tseng: It is technically still [20% slower](https://jsperf.com/cxlt-iterations) than a pure `for(;;)` loop buuuuut eh? They're tests.

lognaturel:
Thanks for the education. And forgive me for saying that I find it kind of unpleasant. 😅👵🏼

And the `__`? Does it have meaning beyond a single underscore?

test/model/base.js
```diff
@@ -0,0 +1,211 @@
...
+      const [ TestBase, shim ] = DatabaseShim.shim(Base);
+      const TestSubclass = class extends TestBase {
+        paranoia() { return true; }
+      }
+
+      const model = new TestSubclass();
+      model.create();
+
+      const [ promiseCallback ] = shim.paramsForCall('then');
+      promiseCallback.should.be.a.Function;
```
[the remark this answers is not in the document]
matthew-white: Note: these tests were removed in 7f9dd226fbda4c905e233a404b5afcf60150675f, so we merged this PR without change.

PR #16, opened 2017-10-12
Update merge()

### LINKED ISSUE #11: Update behavior of merge()
My vote would be to have `merge()` in `util.js` match the behavior of `Object.assign()` and spread syntax: I think any differences could be confusing. Right now, `merge()` automatically concatenates values that are arrays (rather than replacing them) and will not replace a value with `null` or `undefined`. In most cases, I think we'd want `merge()` to fully replace values. An implementation could maybe we as simple as:

```javascript
const merge = (...objs) => Object.assign({}, ...objs);
```

Alternatively, perhaps we could have `merge` take an options parameter...

Thoughts?

matthew-white:
The only place `merge()` is used right now is `Base`, and this update doesn't seem to require any changes there.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -63,18 +63,8 @@ const dbErrorHandler = (response) => (error) => {
 // OBJECT OPERATIONS
 
 // returns a /new/ object with a shallow-copy of all keys from given objs overlaid.
-// concatenates arrays.
 // at some point we may replace this with something more robust/deepmerge.
-const merge = (...objs) => {
-  const result = {};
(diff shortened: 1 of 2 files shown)


clint-tseng:
Given it is just a convenience proxy calls, I think we can probably make do with far fewer tests here. "should not merge arrays," for example, only makes sense in context of the removed code.

matthew-white:
Just removed some tests! Let me know if I should remove more.

clint-tseng:
Looks great! Ship it.

PR #20, opened 2017-10-17
Add submissions

### LINKED ISSUE #9: Add Submission model from shootout

matthew-white:
Closes #9.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,17 +4,29 @@
   "rules": {
     "array-bracket-spacing": "off",
     "arrow-parens": "off",
+    "class-methods-use-this": "off",
     "comma-dangle": "off",
+    "consistent-return": "off",
     "curly": "off",
+    "func-names": "off",
+    "guard-for-in": "off",
     "max-len": "off",
+    "no-confusing-arrow": "off",
+    "no-console": "off",
     "no-iterator": "off",
     "no-multiple-empty-lines": "off",
+    "no-param-reassign": [ "error", { "props": false } ],
+    "no-prototype-builtins": "off",
     "no-restricted-syntax": "off",
     "no-underscore-dangle": "off",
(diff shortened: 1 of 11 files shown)


lognaturel:
Now we're in the Javascript paradigm shootout phase! ;)

More seriously, this is more heavily OO than what @cxlt had previously shared and introduces more abstractions. That's not necessarily a bad thing but it's something that I think is important to discuss and weigh the pros and cons of. It's a great conversation to have now as the project is just getting off the ground so we can get some great guidance that keeps the code base coherent.

I love all the comments and I think overall the code is really clean and easy to follow. Perhaps there's a middle ground that maintains those nice properties with just a little bit less heavy infrastructure that needs to be updated and maintained?

It would be good to understand how the different approaches affect testability, flexibility as we learn more over time, ease of understanding and things like that.

matthew-white:
After chatting with @clint-tseng, our two working branches have a fair number of merge conflicts with each other. We're going to close this PR, and Clint's going to add code to his upcoming PR to address the concerns that this PR tries to solve. Specifically:

* We've concluded that functionality like the controllers and the router doesn't need to be OO, as that adds infrastructure. Clint's thinking of introducing a new function named `endpoint()` to reduce boilerplate.
* Clint will introduce a class named `Problem` that is similar to `JubilantError` (but that helpfully does not include Jubilant in its name).
* We may reintroduce chainable queries in the future, but given how few query methods we have right now, that functionality doesn't seem needed yet. If/when we reintroduce them, we should try to minimize uses of `apply()` and rest parameters, as they can reduce performance.
* We're going to remove the `ephemeral` property on the base model class altogether. This will allow developers to call `create()` on a persisted object and `update()` on an ephemeral one, but we don't think that's so likely, and it's also nice to be explicit. It's the developer's responsibility to call the right method.

If Clint's PR doesn't include submissions, I'll add that once it's merged. I'll also file a couple of issues now just so we don't lose track of them.

PR #24, opened 2017-10-24
[WIP] Cxlt/auth

lognaturel:
Making a PR on this so we can leave comments and questions.


lib/util/util.js
```diff
@@ -0,0 +1,29 @@
...
+  return result.join(', ');
+};
+
+const ensureArray = (x) => Array.isArray(x) ? x : [ x ];
+
+const reject = Promise.reject.bind(Promise);
+const resolve = Promise.resolve.bind(Promise);
+
+const Trait = (classFunc) => (superclass) => (context) => classFunc(superclass(context), context);
+const Subclass = (superclass, classFunc) => (context) => classFunc(superclass(context), context);
```
lognaturel: Is this a common pattern? I have a hard time wrapping my head around it and it makes all class definitions look surprising.

clint-tseng: No, it's goofy because of dependency injection. I'm redoing it a bit.

lib/model/grant.js
```diff
@@ -0,0 +1,42 @@
...
+  static revoke(actor, verbs, actee, maybeTrxn) {
+    //
+  }
+
+  static revokeAll(actor, actee, maybeTrxn) {
+    //
+  }
+
+  static getByTriple(actor, verb, actee) {
+    const impliedActors = '"actorId" in (with recursive implied_actors(id) as (select ?::int union all select "parentActorId" as id from implied_actors a, memberships m where a.id = m."childActorId") select id from implied_actors)';
```
lognaturel: This is the big differentiator between making models directly aware of authz vs. using an external dependency, right?

lib/resources/users.js
```diff
@@ -0,0 +1,26 @@
...
+    Actor.getById(40).then((actor) =>
+      actor.grant('whatever', 'user')
+    )
+  ));
+
+  service.get('/permstest/:verb/:formId', endpoint(({ params }) =>
+    Promise.all([
+      Actor.getById(1),
+      Form.getById(params.formId)
+    ]).then(([ actor, form ]) => actor.can(params.verb, form))
```
lognaturel: And this is what most people would need to be able to use, right?

lib/model/migrations/20171010-01-auth.js
```diff
@@ -0,0 +1,79 @@
+
+const up = (knex, Promise) => {
+  const createActees = knex.schema.createTable('actees', (actees) => {
+    actees.string('id', 36).primary();
+    actees.string('species', 36);
```
lognaturel: This is not an enum so that it can also be a reference to a specific model?

lib/model/actor.js
```diff
@@ -0,0 +1,18 @@
+const { Subclass } = require('../util/util');
+const AsActor = require('./trait/actor');
+const AsActee = require('./trait/actee');
+const Base = require('./base');
+
+
+const actorTypes = { system: 'system', user: 'user', group: 'group', proxy: 'proxy' };
```
matthew-white: I think I understand the other types, but what's a proxy?

matthew-white: How are combinations of groups represented? Are they their own user groups? For example, when we implement campaigns such that each form belongs to a campaign and a campaign can have multiple forms, then a single form may be accessible by (1) users with access to the campaign, (2) any individual users given access, and (3) any user groups given access. Would (1) be represented as a user group? And then (1) + (2) + (3) would be combined into its own user group that represents access to the form?

lib/model/migrations/20171010-01-auth.js
```diff
@@ -0,0 +1,79 @@
...
+  });
+  const createActors = knex.schema.createTable('actors', (actors) => {
+    actors.increments('id');
+    actors.enu('type', [ 'system', 'user', 'group', 'proxy' ]).notNull();
+    actors.string('acteeId', 36).notNull();
```
matthew-white: This means that actors can also be actees, right? What's the use case for that?

lib/model/migrations/20171010-01-auth.js
```diff
@@ -0,0 +1,79 @@
...
+    sessions.integer('actorId').notNull();
+    sessions.string('token', 64).notNull();
+    sessions.dateTime('expires').notNull();
+
+    sessions.dateTime('createdAt');
+
+    sessions.foreign('actorId').references('actors.id');
+    sessions.index([ 'actorId', 'expires' ]);
+  });
+  const createMemberships = knex.schema.createTable('memberships', (memberships) => {
```
matthew-white: Rather than a single memberships table, what do you think about a separate table for each membership type? I think that's a pretty common pattern — the formhub logger implements it, for example — though of course that doesn't necessarily make it good. In the example of a campaign accessible to campaign users, other individual users, and other user groups, that would mean we'd have three tables for users, campaign users, and user group users. Would that be more or less performant than a recursive query?

lib/model/actor.js
lognaturel: It's for limited-access revokable tokens on a subset of a user's permissions. To try at some security in an HTTP context.

lib/model/migrations/20171010-01-auth.js
lognaturel: Users. I can potentially change a user's permissions.

lib/model/grant.js
clint-tseng: Hm. Yes, I suppose that's probably true. This query becomes much easier when we are working directly with the objects in question.

lib/resources/users.js
clint-tseng: Correct. `.can()`, `.grant()`, `.grantIf()` are probably the biggest API points. Very simple, and pretty much the same as the `node_acl` API as far as I saw.

lib/model/migrations/20171010-01-auth.js
clint-tseng: So it can be a reference to a species. So for instance the default admin group we'd inject would be granted all permissions on all users, without a grant being created for every single user, as well as to describe the `create` verb for a species.

lib/model/actor.js
clint-tseng: Yes. Optional/v2 type stuff.

clint-tseng: 1 would be a user group indeed, because fundamentally everything that can be given permissions is an `actor`, whether user or group. 1+2+3 would not be combined; each entity would be given its own set of grants. See the following answer for a bit more.

lib/model/migrations/20171010-01-auth.js
clint-tseng: So, this model only thinks in the other direction. Users don't "belong" to a campaign; actors are granted rights on the campaign. The only canonical user group concept is the actor/membership table duo. Remember also that "users" are not a part of the permissioning system here. Users are each tied to an actor, and that actor is what we operate on. So there aren't users, user groups, and campaign users—there are only actors and grants.

clint-tseng:
Closing in favor of a freshly named branch.

PR #29, opened 2017-10-29
Refactor all of the things.

clint-tseng:
* Completely redo model objects to be plain Javascript objects decorated
  with simple data transformation methods, and static convenience
  methods to invoke various queries. See lib/model/instance/instance.js.
  * Reusable behaviour is possible via Traits (which we can easily get
    rid of if they are not useful), which are intended for decorating
    orthogonal behaviour, not for stacking inherited behaviour.
  * In addition, Instance immutability is now enforced via Object.freeze.
* Completely redo query system to be a composable, chainable system that
  automatically manages transactionality and does some minor magic to
  expose a highly fluent interface. See lib/model/query/builder.js.
* Completely redo dependency injection, so that query modules and
  instances can easily interdepend (and even self-depend) without any
  friction at all.
  * This is accomplished via a two-step initialization/injection process
    wherein references and implementation are created separately;
    see lib/model/package.js.
* This commit blows away some previous implementation work; the only
  endpoint that assuredly works is POST /users. Other modules, methods,
  and endpoints have been ported but have not been tested.
* Also, this new stuff isn’t yet tested; one awaits feedback on this
  code as well as suggestions on maximal testability before embarking
  thus.
* The mountains of shaven yak fur should make for a tidy profit if we
  can make some sweaters, here.


lognaturel:
If I understood correctly, @clint-tseng and @matthew-white have discussed this and are feeling good about the interfaces and ready to start building on it. 

@clint-tseng sounds like you want to add some tests here before merge. And you are waiting on some high-level feedback from @ggalmazor as well as some suggestions on how to make things as easily testable as possible, right?

clint-tseng:
Yes; both of those things!

clint-tseng:
I'm sure, for example, Ramda things I do not know of can be leveraged to
accomplish much of what's done here.

lognaturel:
Ok, great! For whatever it's worth, I like the strategy of getting something merged soon and then working through issues and PRs to make any changes. That will leave a better record of what has been tried and make it easier to divide and conquer.

ggalmazor:
I agree. The PR looks fine and we can work on these things by iterating over the merged code.

Anyway, I’ll try to write today a follow-up with testing and ramda suggestions ;)

ggalmazor:
Regarding the use of Ramda, I'd only want to suggest to use `R.merge()`... Not really important right now. I wouldn't change anything else just yet.

Regarding the test strategy, [I've created a gist](https://gist.github.com/ggalmazor/87390af094b28b291178c2107588d593) to illustrate the three options (really four) that come to my mind when facing a web development with database access, generally speaking.

TL;DR: I'd start implementing Option 1 but I'd try to refactor the design to be able to run our tests without having to launch our API and database. I'd fall back to Option 3 if this proves to be too expensive.

**Option 1** is about writing tests without making any assumption about the internal implementation of the app. This let's us refactor everything and the tests will only fail if we break accidentally or purposedly our API. A.K.A.: The tests fail only for the good reasons.

**Option 2** is about doubling the database. This makes sense when there is a big business logic between the web API and the database. I don't think this would be a real option for us because we're already designing queries and commands that hold all the business logic.

**Option 3** assumes that our ExpressJS code is just a thin layer that links routes to queries and commands. If Option 1 is too expensive, I'd choose this strategy.

**Option 4** is not really an option in my opinion: double the web api and the database. I don't like this option because it will require a lot of complex tooling and I we'll hate our lifes one month in, two tops.

My conclusion is that I really believe that we can get away with Option 1 without needing to launch a real web api and database but we need to be open to changing the implementation design. If we focus on writing testable code it will make the project easier to maintain and more able to absorb change in the long run. I also believe that we can achieve that iteratively and without making our code more obscure.

clint-tseng:
It sounds like you are advocating for service-level integration tests primarily, then? (Whether driven by directly calling the service endpoints or by actually sending HTTP tests to a server.)

My primary concern there is the reliance on the database:
1. Slower.
1a. Much more difficult to parallelize tests, for when we eventually have a lot.
2. Requires a test database instance that can be repeatedly nuked (not conceptually difficult of course, just more setup and overhead).
3. Can be finnicky, as you have to carefully test around eg autoincrements, other changing things.

I've had mostly bad experiences writing extensive test libraries that depend on the database. And my general philosophy is to lean heavily into unit tests with the lightest possible layer of integration tests.

But I understand the philosophy behind your advocacy for ensuring the refactorability of the code, and thus not codifying too heavily the internal machinery. I'm just trying to weigh that against the downsides above.

clint-tseng:
I see now I didn't read Option 2 closely enough.

clint-tseng:
If only there were some H2 equivalent to use (as if H2 actually works as advertised in JVM to begin with..)

clint-tseng:
Okay, here is my tentative plan. This branch gets merged, even though it hurts me to just rm all tests and merge a branch. I'll push ahead and work on getting done the following:

* authn/sessions
* authz
* basic Form REST

Then we can look at the code and talk about testing with that structure in mind, or change the structure at that point with actual code to point at to befit the testing.

PR #31, opened 2017-11-03
New Option class to handle values that might not be present 

ggalmazor:
(nulls and undefineds)

### DIFF
--- a/lib/reused/option.js
+++ b/lib/reused/option.js
@@ -1,0 +1,112 @@
+class Option {
+  static of(value) {
+    return value instanceof Option
+      ? value
+      : value === null || value === undefined
+        ? none
+        : new Some(value);
+  }
+
+  static none() {
+    return none;
+  }
+}
+
+class Some extends Option {
+  /* istanbul ignore next */
+  constructor(value) {
+    super();
+    this.value = value;
+  }
+
+  get() {
+    return this.value;
+  }
+
+  map(fn) {
+    return Option.of(fn(this.value));
+  }
+
+  filter(predicate) {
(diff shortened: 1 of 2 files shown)


lib/reused/option.js
```diff
@@ -0,0 +1,112 @@
+class Option {
+  static of(value) {
+    if (value instanceof Option)
+      return value;
+    if (value === null || value === undefined)
+      return none;
```
clint-tseng: I think this will fail the linter but I'll take a look at it.

clint-tseng: (use before declaration)

lib/reused/option.js
```diff
@@ -0,0 +1,112 @@
...
+  get() {
+    return this.value;
+  }
+
+  map(fn) {
+    return Option.of(fn(this.value));
+  }
+
+  filter(predicate) {
+    return predicate(this.value) ? this : none;
```
clint-tseng: Loose truthiness; intentional? (as opposed to `=== true`)

lib/reused/option.js
ggalmazor: I run eslint with the current rules definition and it didn't complain... Maybe I've got the rules wrong...

lib/reused/option.js
ggalmazor: Hmmm, I'm assuming that a predicate will always produce a boolean, but maybe you're right and we would benefit of a type and value check...

lib/reused/option.js
clint-tseng: then i am wrong!

PR #38, opened 2017-11-22
new #34: add api version to application route; expose on request obj.

### LINKED ISSUE #34: Reroute API under /v1 for now

clint-tseng:
* closes #34.
* check request.apiVersion to see the version of the request.
* this could change if eg we’d prefer to actually route based on the
  version, in which case we can simply submount services instead.


matthew-white:
Should we respond with an error if the version isn't 1?

PR #39, opened 2017-11-22
Submissions API

clint-tseng:
WIP.

so far, only POST.


matthew-white:
Looks like we have a minor merge conflict after merging #40.

PR #44, opened 2017-12-26
unit tests for the past month of work

clint-tseng:


clint-tseng:
> > From 6fae574, "tanks the test coverage statistics but oh well"
> You mean because stats then include modules that don't lend themselves so well to unit testing, right?

Yes. But integration tests should ameliorate.

And both of those are indeed copypasta.

clint-tseng:
> I would find it helpful to have make test-coverage suggested in the README in case folks don't think to look in the Makefile.

You mean [like this](https://github.com/nafundi/jubilant-garbanzo/blame/master/README.md#L39)?

lognaturel:
> You mean like this?

Exactly like that. 😖

PR #57, opened 2018-01-18
Field keys

clint-tseng:


clint-tseng:
Still needs: last used, and in extended mode to join createdBy to userdata.

clint-tseng:
Also, the fact that displayName is optional now makes createdBy really awkward to handle..

clint-tseng:
This should be ready.

PR #60, opened 2018-01-31
new: extended metadata for forms on GET /forms and GET /forms/:id.

clint-tseng:
* no tests yet; all forms tests are forthcoming in one batch.

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -1,3 +1,4 @@
+const { rowsToInstances, maybeRowToInstance } = require('../../util/db');
 
 module.exports = {
   create: (form) => ({ actees, simply }) =>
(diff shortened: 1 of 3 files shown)


lib/model/query/forms.js
```diff
@@ -1,10 +1,30 @@
...
 
-  getAll: () => ({ simply, Form }) => simply.getAll('forms', Form)
+  getByXmlFormId: (xmlFormId, extended) => ({ forms, Form }) =>
+    forms._get(extended, { xmlFormId }).then(maybeRowToInstance(Form)),
+
+  getAll: (extended) => ({ forms, Form }) =>
+    forms._get(extended).then(rowsToInstances(Form)),
+
+  _get: (extended = false, condition = []) => ({ db, simply, Form }) => (extended === false)
+    ? simply.getAll('forms', Form)
```
matthew-white: Should this also apply the `condition`, as well as exclude deleted forms?

clint-tseng: Aye. Fixed.

PR #67, opened 2018-02-14
Add hash to URLs

matthew-white:


clint-tseng:
The one awkward thing about this change is that the `?` comes _after_ the `#`, which breaks browser URL parsing.. what does your token extraction code end up looking like?

matthew-white:
I know what you mean... It seems like the Vue router is able to handle it pretty seamlessly though. The query string gets included as part of the `#` fragment, but the router parses the query string (and the rest of the URL) correctly from the fragment.

The URLs that Jubilant has been generating have been working well and have been setting the token correctly in Adventure. Within Adventure, I access `this.$route.query.token`, and it's set correctly.

PR #81, opened 2018-03-28
Form deletion

clint-tseng:
Based off the form state branch/PR because otherwise the merge conflicts would be awful.

See commit messages.


clint-tseng:
For archival reasons, the answer to the question is: submissions won't load without forms; they are subresources.

Issue #37, opened 2017-11-21
Add file-level copyright notices and license information


matthew-white:
We'll do this once we decide on the app's name.

clint-tseng:
@lognaturel still blocked on a name!

PR #91, opened 2018-05-10
code documentation, improved readme, copyright notices

### LINKED ISSUE #35: Add high-level narrative
I didn't see an existing issue/PR for this, so I just wanted to add a +1 to @lognaturel's comment in #32:

> The commit comments are really helpful and I would like to see a little bit of that become comments and/or dev docs. I know that @cxlt you were reluctant to get started with dev docs because understandably you expect things will change but I think just a little bit of high-level narrative now would help. A concrete way to start would be to take `Instance.with(ActeeTrait)(({ Grant, actors, all, grants, simply }) => class` and upack it a little bit. Link out to a good narrative on traits, describe that `package.js` is the dependency injection container, a paragraph on `Option`, things like that. I think a few minutes on that would help a lot with giving someone an idea of where to start. It doesn't have to go in this PR.

clint-tseng:
resolves #35 and #37.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,91 @@
+# Contributing to Jubilant Garbanzo
+
+We want Jubilant Garbanzo to be easy to install and use, yet flexible and powerful for a wide variety of projects and organizations. To do that, we need your help! Even if you don't write code, you are a valuable member of our community, and there are many ways you can help improve the project.
+
+# Getting Involved
+
+One of the easiest ways you can participate is to help us [track down bugs](https://github.com/nafundi/effective-spork/issues), which we do over in the [Effective Spork](https://github.com/nafundi/effective-spork) repository. Did you run into some odd or puzzling behaviour? Did you have trouble installing or updating the server? Is a form upload or data export not working the way you'd expect? We want to know! Please do a quick search of the issues list to see if it's something that's already been submitted. Please do make an effort to figure out if the issue is related to this server, or some other tool you are using (like XLSForm, Build, or Collect). But when in doubt, submit your problem anyway.
+
+## Questions and Discussion
+
+If you are having trouble with making Jubilant Garbanzo work correctly, we encourage you to visit the [community forum](https://forum.opendatakit.org) where a great many of your knowledgeable peers can help you with your problem.
+
+If you're looking for help or discussion on _how_ Jubilant Garbanzo works internally and how to understand or update its code, the ODK [developer Slack](https://opendatakit.slack.com) is for you.
+
+# Contributing Code
+
+## Getting Started
+
+Please see the [project README](https://github.com/nafundi/jubilant-garbanzo#setting-up-a-development-environment) for instructions on how to set up your development environment.
+
+## Guidelines
+
+If you're starting work on an issue ticket, please leave a comment saying so. If you run into trouble or have to stop working on a ticket, please leave a comment as well. As you write code, the usual guidelines apply; please ensure you are following existing conventions:
+
+* Our **code style** is loosely based on the [AirBnB Javascript Style Guide](https://github.com/airbnb/javascript), and you can run `make lint` in the project directory to check your code against it. We do make [some exceptions](https://github.com/nafundi/jubilant-garbanzo/blob/master/.eslintrc.json) to their rules, and we do make exceptions to the linter when we have a reason and consensus to do so.
+* More broadly, we also try to keep our code as maintainable as possible by doing the following:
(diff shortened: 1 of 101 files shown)


lib/model/trait/actee.js
```diff
@@ -1,3 +1,8 @@
+// The Actee Trait simply adds in a couple of methods convenient for dealing
```
lognaturel: As you know, I strongly prefer avoiding words like "just" or "simply"!

CONTRIBUTING.md
```diff
@@ -33,50 +35,57 @@ If you're starting work on an issue ticket, please leave a comment saying so. If
...
+    * `model/migrations/` contains the database migrations for the project. Please follow the date-based naming convention of the existing files. Migrations help us update the database structure itself to support changes and new features.
+    * `model/query/` contains database queries for storing and retrieving `Instance`s. They are composable, and many of them rely on the basic operations available in `simply.js`, which defines generic get/set operations, and `all.js`, which allows multiple queries to be assembled together. All database queries should be put in these files. For each query, the first order arguments are whatever the query needs to work and are how the query will eventually be called, while dependencies are declared and injected as the second order arguments.
```
lognaturel: My hunch is that currying is not familiar to all JS devs. Might be good to have a link to a reading either here or perhaps in a "concepts to be familiar with" section. Ideally it would be framed as something like "if you're not sure what two+ arrow functions means, read this". For "first order arguments" and "second order arguments" here, it might help to have an example (e.g. in `create: (table, record) => ({ db }) =>`, the first order arguments are `table` and `record`, etc.)

CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+* Please format your **commit messages** appropriately:
+    * Ensure that your first commit line is 72 characters or fewer, and if you include further text in your commit message that there is an empty line following the leading line.
+    * Start your commit message with one of the following tags: `new:`, `improve:`, `bug:`, `refactor:`, or `noop:` for new features, improvements to existing functionality or code, fixes, big code refactors, and small low-impact changes (eg style fixes or comments) respectively.
+    * Tag with the GitHub issue number if relevant; eg: `bug #41:`.
+    * Please be as descriptive as possible in the 72 characters allowed. This helps us _a lot_ when writing release notes and tracking down regressions. Here are some examples:
+        * Instead of `Fixing bugs`, consider `bug #1372: negative top/skip values would break odata output`.
+        * Instead of `Updating readme`, consider `improve: making build instructions clearer in README.`
```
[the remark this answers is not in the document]
clint-tseng: [https://github.com/nafundi/jubilant-garbanzo/commits](welllllll)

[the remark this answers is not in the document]
lognaturel: Ohh, nice! Seems a lot of projects don't like periods there but I do.

lib/model/trait/actee.js
clint-tseng: The goal here is to attempt to defuse the sentiment I keep seeing from you, which is fear about what this trait thing is and what it does, by emphasizing that has a very limited use and that _all it does_ is add a couple of methods. Nothing more. So the language attempts to minimize because it is, I think, appropriate to do so. If someone doesn't understand what a trait is, it probably doesn't actually matter. I'm not sure how to do this deëmphasis without resorting to words like "simply" or "just"..

CONTRIBUTING.md
clint-tseng: Technically, we don't curry at all; these are higher order functions. But I can find a place to put it once I find an article about them that isn't confusing or wrong.

clint-tseng:
> I think also pointing to e.g. https://github.com/nafundi/jubilant-garbanzo/blob/d0108f3108e9816085c0329dfd6a0362aa222384/lib/model/instance/instance.js#L74 might help make the injection process a little more concrete.

Hm, it depends on whether your goal is to teach people usage by understanding quirky internals or whether you feel following existing usage patterns is a better direction for beginners. That line is very heavily into quirky internals.

clint-tseng:
> https://github.com/nafundi/jubilant-garbanzo/blob/173124a4f4183922f33198d74b74b3c8565626b3/lib/bin/run-server.js#L34
Feels weird that the port number is hardcoded.

Uh, ha, okay. I'll do that in a different PR though..

clint-tseng:
@lognaturel two threads above awaiting your comments before merge.

lib/model/trait/actee.js
lognaturel: Ok.

PR #95, opened 2018-05-16
bug #93: uniqueness violation problem handles multicolumn constraints.

### LINKED ISSUE #93: Problem details for uniqueness violation
For 4.09b, the user cannot create a form whose combination of form ID and version would conflict with a deleted form. When Jubilant returns the problem to Adventure in that case, the problem's details look like this:

```json
{
  "field": "\"xmlFormId\", version",
  "table": "forms",
  "value": "widgets, 1"
}
```

I see that there are double quotes enclosing `xmlFormId`, I assume that's because column names are case-sensitive in Postgres? Is there a way to strip the quotes before passing the column names to Adventure? Or should I strip the quotes within Adventure?

Also, this may be an unlikely scenario, but I believe the form ID and/or version can theoretically contain commas, perhaps making it challenging to parse `details.value`. Is there a way to parse it robustly?

clint-tseng:
* instead of supplying a problem details of { field, value, table }, it
  now supplies { fields, values, table } in all cases. fields and values
  will always be arrays, even for single-value cases.
* also dequote field names if postgres has quoted them (in particular if
  they are mixed-case).
* the textual error message changed too, but nobody ought to rely on
  those.
* resolves #93.

### DIFF
--- a/lib/util/db.js
+++ b/lib/util/db.js
@@ -88,6 +88,10 @@ const resultCount = (result) => Number(result[0].count);
 ////////////////////////////////////////////////////////////////////////////////
 // ERROR HANDLING
 
+// deals with postgres returning eg '"xmlFormId", version' error text.
+const dequote = (text) => ((text.startsWith('"') && text.endsWith('"')) ? text.slice(1, -1) : text);
+const splitPostgresTuple = (text) => text.split(/,\s*/g);
+
 // Generic database failure handler; attempts to interpret DB exceptions.
 // if it recognizes the problem, it translates it into a predictable
 // error format. either way, it returns a Problem that can be handled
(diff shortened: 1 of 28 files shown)


lib/util/db.js
```diff
@@ -88,6 +88,10 @@ const resultCount = (result) => Number(result[0].count);
 ////////////////////////////////////////////////////////////////////////////////
 // ERROR HANDLING
 
+// deals with postgres returning eg '"xmlFormId", version' error text.
+const dequote = (text) => ((text.startsWith('"') && text.endsWith('"')) ? text.slice(1, -1) : text);
+const splitPostgresTuple = (text) => text.split(/,\s*/g);
```
matthew-white: What happens if an element of the tuple itself contains a comma? (Does Postgres add quotes around such elements?)

matthew-white: Also, if the element of a tuple contains leading white space, will `\s*` remove that? Maybe we could use a single space rather than `\s*`? (Or maybe it doesn't matter if the white space disappears?)

matthew-white: If form IDs can't contain commas, and versions must be numeric, I guess these comments wouldn't affect xmlFormId/version tuples. But maybe they'd affect other Postgres tuples?

clint-tseng: Short of writing a lexer, I can't account for those cases. Postgres will inconsistently quote/notquote things in various cases so I'd have to actually walk the string to make this completely airtight. I think there are no problem cases possible yet, and even if there are, it's just error message output.

matthew-white: Sounds good, and we can always circle back to this if we run into an issue.

PR #97, opened 2018-05-16
bug #92: should return success object rather than string on PATCH forms.

### LINKED ISSUE #92: PATCH /forms/:id returns a string, not a form object
I'm noticing that when I send a request to `PATCH /forms/:id`, the response is the string `true` rather than the updated form object. That isn't a problem for Adventure, which doesn't currently need the updated form object, but I thought I'd mention it in case we want to update the API docs.

clint-tseng:
* see ticket for details.
* should probably eventually return the updated object but we'll worry
  about that later.
* resolves #92.


matthew-white:
Do we also want to update the API docs, or there's no need, because we'll eventually be returning the updated object?

PR #102, opened 2018-06-06
Update notice

matthew-white:


clint-tseng:
:( technically, this PR does not meet the commit naming guidelines listed in its Contributing Guide.

matthew-white:
Oops, my mistake! I'll aim to meet those guidelines in my next PR.

Do you think we'd want to standardize those commit naming guidelines across Backend and Frontend? (@lognaturel, do you have any preferences around commit naming?)

clint-tseng:
(even if we don't care for the prefix scheme, something like `update NOTICE file to fix the copyright attribution` would be much more meaningful than `Update notice`, which is already plainly obvious from the changed-files list)

lognaturel:
I think splitting up and naming commits is a surprisingly challenging art. I find it particularly difficult when making these kinds of small changes that are self-explanatory. I would have gone with something like "Fix the copyright attribution" or "fix: copyright holder is ODK Central Developers" (the only file changed is NOTICE so including that in the message seems redundant) but really even that doesn't add a whole lot of useful context. In conclusion ¯\_(ツ)_/¯.

clint-tseng:
yeah, it's an art for sure. in general i find two things:

1. if you aren't using the whole 70 chars there is probably more you can be
doing
2. sometimes it is clearer to describe what was wrong, sometimes it is
clearer to describe what was changed. often if one is fuzzy the other is
better.

matthew-white:
Catching up on this conversation post-vacation...

> 1. if you aren't using the whole 70 chars there is probably more you can be
doing

I actually often see the recommendation to limit the first line of the commit message to 50 characters or fewer. For example, `git commit --help` reads:

> Though not required, it's a good idea to begin the commit message with
a single short (less than 50 character) line summarizing the change,
followed by a blank line and then a more thorough description. The text
up to the first blank line in a commit message is treated as the commit
title, and that title is used throughout Git. For example, **git-format-
patch**(1) turns a commit into email, and it uses the title on the
Subject line and the rest of the commit in the body.

(Of course, "Update notice" is well short of 50 characters, and I probably should have included at least a little extra description.)

I find it easier to glance over a list of commits when they have shorter commit subjects, so I personally lean toward a 50-character rule-of-thumb. I also haven't found the prefix especially helpful for myself. Especially if there is a 50-character limit, the prefix reduces the amount of space available, so I'd lean toward not requiring it.

That said, I don't feel strongly about those preferences, and I'm happy to adopt the prefix or use commit subjects of >50 characters if either of you have found that to be useful in past projects. However, I do think that it'll be easier for contributors if we have consistent commit guidelines for Backend and Frontend.

clint-tseng:
I think the 50 character thing is vestigial. Most Git clients allow 70 chars and ellipsify afterwards.

I think 50-char commit messages make the most sense to the author, because they remember what they wrote, so the short reminder is all they need. I think usually more context is needed for other contributors.

I think central-frontend@cd5c561 "bug: fix grammar in user admin header blurb." vs central-frontend@fa1fff6 "Update text" is a pretty good example of how a few more words means you can understand the commit much better without having to dig into the diff.

These little details usually matter a lot when eg building a release changelog or running through git blames.

matthew-white:
I could see how that would be quite useful for changelogs... I'll go ahead and start trying to write longer commit messages for Frontend. Is it useful for me to use the prefix as well? Does that end up being most useful for changelogs, or do you find it useful in other cases as well?

PR #106, opened 2018-06-22
Briefcase-compatible .csv.zip output; replaces a use of fast-xml-parser with htmlparser2

clint-tseng:
see commits


clint-tseng:
probably ready for initial poking/prodding/code review

clint-tseng:
i added briefcase's own csv tests into our test suite so we should be pretty sure of this by now. please to review!

PR #112, opened 2018-07-12
Export submission metadata in OData output

clint-tseng:
See commits for more; includes the following structure (and associated schema metadata) in each root level JSON output:

```
__system: {
  submissionDate: '<iso date>',
  submitterId: '<int>',
  submitterName: '<string>'
}
```

### DIFF
--- a/lib/resources/odata.js
+++ b/lib/resources/odata.js
@@ -37,7 +37,7 @@ module.exports = (service, { Form, Submission, env }) => {
     Form.getByXmlFormId(params[0]) // first regexp match
       .then(getOrNotFound)
       .then((form) => auth.canOrReject('read', form)
-        .then(() => Submission.getById(form.id, params[1])
+        .then(() => Submission.getById(form.id, params[1], true)
           .then(getOrNotFound)
           .then((submission) => singleRowToOData(form, submission, domain, decodeURI(originalUrl), query))))));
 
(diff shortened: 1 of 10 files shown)


lib/resources/odata.js
```diff
@@ -37,7 +37,7 @@ module.exports = (service, { Form, Submission, env }) => {
     Form.getByXmlFormId(params[0]) // first regexp match
       .then(getOrNotFound)
       .then((form) => auth.canOrReject('read', form)
-        .then(() => Submission.getById(form.id, params[1])
+        .then(() => Submission.getById(form.id, params[1], true)
```
lognaturel: The `true` here ensures that the join on the actor is always performed so the name can be pulled but it's not possible to get the full actor object in the odata feed, right?

clint-tseng: Correct. The way these records get pulled apart nothing is ever automatically passthrough.

clint-tseng: And besides, actor information is really really boring. It's _user_ information that contains interesting things, like emails and password hashes. Actors are pretty much just id and displayname (and timestamp metadata).

lognaturel: That makes sense. I was getting a little turned around regarding the meaning of the extended parameter vs. the extended header.

clint-tseng: often the latter triggers the former!

lognaturel:
Just realized docs should probably be updated to show this. (https://jubilantgarbanzo.docs.apiary.io/#reference/odata-endpoints/odata-form-service/data-document, etc)

clint-tseng:
oh i suppose the examples could get updated. but the formal docs themselves don't actually even describe `__id`; the odata metadata document documents the available data. i'll update the examples, though.

PR #114, opened 2018-07-24
Completely replace fast-xml-parser

clint-tseng:
commit messages are descriptive but this will probably require interactive review.

### DIFF
--- a/lib/model/instance/form.js
+++ b/lib/model/instance/form.js
@@ -52,7 +52,7 @@ module.exports = Instance.with(ActeeTrait)(({ simply, Form, forms }) => class {
 
   // These two methods call into lib/data/schema.js to provide schema information.
   schema() { return getFormSchema(this); }
-  tables() { return getSchemaTables(this.schema()); }
+  tables() { return this.schema().then(getSchemaTables); }
 
   acceptsSubmissions() { return (this.state === 'open') || (this.state === 'closing'); }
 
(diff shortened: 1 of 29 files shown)


lib/model/instance/form.js
```diff
@@ -65,44 +65,22 @@ module.exports = Instance.with(ActeeTrait)(({ simply, Form, forms }) => class {
...
-    // hash and cache the xml.
-    const hash = createHash('md5').update(xml).digest('hex');
-    return resolve(new Form({ xmlFormId: xmlFormId.get(), xml, name, version, hash }));
+    const dataNode = findOne(node('html'), node('head'), node('model'), node('instance'), node());
+    return traverseXml(xml, [
+      dataNode(attr('id')),
+      dataNode(attr('version')),
+      findOne(node('html'), node('head'), node('title'))(text())
+    ]).then(([ idText, versionText, nameText ]) => {
+      const xmlFormId = idText.map(crushBlankString).orElseGet(() => {
```
clint-tseng: are we sure we didn't mean to crush blank on version??

lib/util/xml.js
```diff
@@ -6,37 +6,286 @@
...
 //
-// Various helper functions for using fast-xml-parser, with useful default options.
-// Also contains some helpers for dealing with XML processing.
+// also in general, they take filters, which are (tagname, attrs) => bool indicating
+// whether the node is a match or not.
+
+// helper for tracking real/imaginary depth traversal; if we are down a matching
+// path, we want to know how many levels deep we are. if we are partway down a
+// matching path then fail a match, we still need to know how deep we are so that
+// we understand when to go back to trying matches.
```
lognaturel: This comment is very good. Maybe just going one step further? "if we are down a matching path, we want to know how many levels deep we are **so we know when to stop**"

"if we are partway down a matching path then fail a match, we **need to know how far down a non-matching path we have gone so we can identify when we see the closing event for the last matching tag we saw**"

Just some ideas.

PR #122, opened 2018-08-13
bug: invalid extant form definitions would fail the FA migration.

clint-tseng:
preflight the expected attachments so we don't attempt to run any invalid queries.


clint-tseng:
on second thought, this only partially solves the problem, and does so backwards. if the same file is referenced multiple times, and it's expected to be the same type, we should happily let that through. so this commit will stay but we should also pre-dedupe at the expected call.

clint-tseng:
closing in favor of #123.

PR #123, opened 2018-08-14
Address duplicate form attachments

clint-tseng:
1. If the same filename is seen multiple times in a form xml def _and_ the expected type each time is the same, only return that filename once from `expectedFormAttachments`. (This will make the database happy with these cases.)
2. If some form still has duplicate attachment violations even given this change, don't fail the initial migration, just bail out on supporting attachments for that form.

### DIFF
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -215,7 +215,22 @@ const expectedFormAttachments = (xml) => {
     // if we have an <input query= at all we expect an itemsets.csv file.
     const itemsets = hasItemsets.orElse(false) ? [{ type: 'file', name: 'itemsets.csv' }] : [];
 
-    return externalInstances.concat(mediaFiles).concat(itemsets);
+    // now we need to deduplicate our expected files. we consider type and name
+    // a composite key, and we only add such pairs once. but duplicate names with
+    // different types will be sent through, as they are different expectations.
+    // n.b. this function returns the /expected/ list of files. so removing a dupe:
+    // 1. fulfills the contract: that one file we expect to see once.
+    // 2. /allows/ the duplicated file as one entry: the database constraint will
+    //    eventually complain about duplicate filenames.
+    const seen = {};
+    const result = [];
+    for (const attachment of externalInstances.concat(mediaFiles).concat(itemsets)) {
+      if ((seen[attachment.name] == null) || (seen[attachment.name] !== attachment.type))
+        result.push(attachment);
+      seen[attachment.name] = attachment.type;
+    }
+
+    return result;
   });
 };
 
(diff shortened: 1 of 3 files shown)


matthew-white:
Just as a background question, do ODK XForms permit two files of the same name but different types? Is there any chance that such forms are actually invalid?

In the form manifest, would the two files have identical `<downloadUrl>` elements? I think right now we don't provide a way to get/post/delete one vs. the other via the API, since those API endpoints are based on name, not name and type.

clint-tseng:
I don't know. My understanding in this area is very loose, the spec is very loose, and the information I have gotten indicates that reality deviates from the spec.

LN has told me that in practice nobody uses (supports??) subpaths. This implies to me that filenames must be unique. What made me add the deduping code was the fact that the standard testing Widgets form indeed references the same file twice.

matthew-white:
The deduping makes sense to me, and I'm wondering whether we can even be stricter about attachments with duplicate names but different types. Do you think it'd be feasible for the time being for us to reject forms with attachments like that, then consider supporting them if/when we encounter a use case for them? If we're going to support the case where a form has two attachments of different types but the same name, I feel like we'll also have to update the form manifest and the API endpoints. For example, in the second test case, it's not clear what `GET /forms/form/attachments/name.file` would return, the image or the audio file.

clint-tseng:
Well, right now we would reject attachments with duplicate names but different types..

matthew-white:
Oh I see, I thought the code allowed that case, but now I realize that there's a database constraint that enforces that attachment names are unique for a given form. 👍

clint-tseng:
Yeah. Removing a duplicate entry actually _allows_ the duplicate. Perhaps I should leave a comment to that effect.

lib/data/schema.js
```diff
@@ -215,7 +215,22 @@ const expectedFormAttachments = (xml) => {
...
+    // a composite key, and we only add such pairs once. but duplicate names with
+    // different types will be sent through, as they are different expectations.
+    // n.b. this function returns the /expected/ list of files. so removing a dupe:
+    // 1. fulfills the contract: that one file we expect to see once.
+    // 2. /allows/ the duplicated file as one entry: the database constraint will
+    //    eventually complain about duplicate filenames.
+    const seen = {};
+    const result = [];
+    for (const attachment of externalInstances.concat(mediaFiles).concat(itemsets)) {
+      if ((seen[attachment.name] == null) || (seen[attachment.name] != attachment.type))
```
matthew-white: I don't think the difference matters in this case, but I suspect the linter will ask for `!==` over `!=`.

clint-tseng: oh, probably

PR #127, opened 2018-09-26
new: perform root table OData paging at the database level for perf.

issa-tseng:
* we still have to get all rows to count them for subtables. but at the
  root level (which most requests will be for) we can make the database
  do it, which is way faster.
* but the plumbing required to allow either/or path gets a little bit
  annoying. it's all explained in the comments.
* likely a followup commit to formalize this database options hash into
  some sort of safer actualized object.

### DIFF
--- a/lib/model/instance/submission.js
+++ b/lib/model/instance/submission.js
@@ -99,6 +99,7 @@ out.Submission = Instance(({ PartialSubmission, SubmissionAttachment, Blob, simp
 
   static getById(formId, instanceId, options) { return submissions.getById(formId, instanceId, options); }
   static getAllByFormId(formId, options) { return submissions.getAllByFormId(formId, options); }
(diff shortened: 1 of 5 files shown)


clint-tseng:
okay i feel better about this PR now with these additional commits. still a ways to go cleaning up the queries but it's some progress.

matthew-white:
I just wanted to leave a few quick questions/thoughts:

1. For the submissions order, how about we order by `id` in addition to `createdAt` so that in the unlikely event of a tie on `createdAt`, the order is still stable? I'm thinking of something like `ORDER BY "createdAt" DESC, id DESC`.
2. How about adding an index on `createdAt` to speed up that order operation? Or we could add the index (or something like it) in the future if we start seeing performance issues?
3. How about rebasing on master so that there aren't CircleCI errors?

PR #130, opened 2018-09-26
security: update npm dependencies.

clint-tseng:


lognaturel:
Verified that this works locally but would be good to get to the bottom of that CI failure.

Issue #111, opened 2018-07-12
Code: better way to define query helpers that need container context


clint-tseng:
Right now the query module system assumes that everything returned by a query function is a `FutureQuery`, or at least a `Promise` of some kind. But this precludes the ability to write a query function that's just a helper, which does not perform its own computation.

This is made doubly problematic by the fact that the function result is `.then()`ed, which forces an execution state for the query when actually you want to get the chain back to do something with (like `.stream()`.

See the submissions query module for an example of working around this, which is to manually pass container state to the helper, which is staged outside the module system. This works for now, but we should come up with a better way to do this before we end up with more helper funcs.

PR #139, opened 2018-10-07
Exclude 'xml' field when X-Extended-Metadata is false

### LINKED ISSUE #103: RESTful form list without XML
`GET /formList` provides an XML list of available forms. `GET /forms` provides a similar json list but with full form XML included.

@tiritea suggests in Slack that a listing without XML through `/forms` would be helpful. Perhaps the form XML could be considered extended metadata.

akshay-ap:
Closes #103 
Modified database select query to exclude `xml` field to produce the following response: 
```
[
    {
        "xmlFormId": "TestForm",
        "version": "",
        "state": "open",
        "hash": "155afd06b2531d726058a5c0ef726817",
        "name": "TestForm",
        "createdAt": "2018-10-07T07:59:22.389Z",
        "updatedAt": null
    }
]
```

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -46,7 +46,7 @@ module.exports = {
   // the modifier allows custom chaining to be inserted into the query, which is used
   // above in getOpen().
   _get: (extended = false, modifier = identity) => ({ db, Form, Actor }) => ((extended === false)
-    ? db.select('*')
+    ? db.select(Form.fields().filter((field)=> field != 'xml'))
       .from('forms')
       .modify(modifier)
       .where({ deletedAt: null })


akshay-ap:
This change breaks OData requests:
```
curl -X GET \
  'http://localhost:8383/v1/forms/TestRatingWidgetnnn.svc/$metadata' \
  -H 'Authorization: Bearer mASpT9Pn6364FIZtWhtACB5IUQ0ki1bA9Vcv8GXDa2I3A29qGF3a9vSIO1!lZExC' \
```
Removing `xml` field from database `select` query is not a right approach. I am trying to find a better way to remove `xml` field from result.

clint-tseng:
Certainly removing the `xml` field unconditionally from all queries will be a problem, yes.

This is a weird case; in the past, the non-extended metadata response has always included all the data local to its own table, and extended metadata has always been a request to get _more_ data by joining with other tables, etc. This is the first time we desire to have the non-extended metadata _not_ include a field that naturally _does_ come back with the database request.

What this means is that we can no longer treat extended metadata as purely a database query formulation concern, because as you see with the OData failure we do expect internally to get all our local fields back when we ask for them.

I see three likely responses:

1. Filter the field out at the HTTP layer, in the `.get()` handler itself. This is the most logical and clean short-term solution, but will become ugly and dangerous if allowed to spread as a pattern.
2. Make instances understand the concept of `forExtendedApi` vs `forApi` (or similar) and have the instance filter out the extraneous fields.
3. Establish three standard database query paths: internal, nonextended, and extended. This would likely be easier handled if we tackled #111 and then standardized how query fragments are put together.

clint-tseng:
hi: since this approach is fundamentally unworkable, i'm going to close this PR for now. we are still looking for a solution here, though, and please do file a new one if you figure out a better way to do this that doesn't break other parts of the system. thanks!

PR #142, opened 2018-10-10
new: add Sentry error handling/logging, incl mock for Sentryless config.

clint-tseng:
* also moved express service generation inside the generation block.
  * shouldn't ever have been a problem but it's "better".
* default configuration is done in the Central docker config.
* also surface exception details in Problem[500]s.
  * it's an open source project, shouldn't be a big deal.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -20,7 +20,8 @@
       "google": {
         "clientId": "660095633112-h7bhsjenhp1agd0c4v3cmqk6bccgkdu0.apps.googleusercontent.com",
         "clientSecret": "lzu2vK1NFqqd6Y5HiN-7ByvE"
-      }
+      },
+      "sentry": {}
     }
   },
   "test": {
--- a/lib/bin/run-server.js
+++ b/lib/bin/run-server.js
@@ -30,12 +30,15 @@ const mail = mailer(merge(config.get('default.email'), { env }));
 const googler = require('../outbound/google');
 const google = googler(config.get('default.external.google'));
 
+// get a sentry client.
+const Sentry = require('../util/sentry').init(config.get('default.external.sentry'));
+
 
 ////////////////////////////////////////////////////////////////////////////////
 // START HTTP SERVICE
 
 // initialize our container, then generate an http service out of it.
-const container = require('../model/package').withDefaults({ db, mail, env, google });
+const container = require('../model/package').withDefaults({ db, mail, env, google, Sentry });
 const service = require('../http/service')(container);
 
 // insert the graceful exit middleware.
--- a/lib/http/service.js
+++ b/lib/http/service.js
@@ -11,13 +11,15 @@
 // defined elsewhere into an actual Express service. The only thing it needs in
 // order to do this is a valid dependency injection context container.
 
-const service = require('express')();
-
 
 module.exports = (container) => {
+  const service = require('express')();
 
   ////////////////////////////////////////////////////////////////////////////////
   // PRERESOURCE HANDLERS
+
+  // apply the Sentry request hook.
+  service.use(container.Sentry.Handlers.requestHandler());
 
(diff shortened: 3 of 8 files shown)


clint-tseng:
@yanokwa still looking for a checkmark here

clint-tseng:
(since @lognaturel mentioned you should look)

yanokwa:
@clint-tseng Couple of quick things that I can see.

1. The tests are failing on CI. Is that expected behavior?
2. I can confirm Sentry shows the errors from your local machine. What would be ideal is if we could deploy this branch to some testing machine (maybe antinod.es) and confirm that the default config does report back to Sentry. Is that possible?
3. Can you explain a more what surface exception details in Problem[500] means and why shouldn't be a big deal. Because I don't know what it means, it sounds like there might be some data leakage about the person deploying Central.

clint-tseng:
1 yeah a PR since has fixed this but it's not part of this tree
2 yeah it'll be done in a few mins
3 i guess their dirtree will leak because the stack will report. do you feel like this is a problem?

clint-tseng:
https://sentry.io/opendatakit/central-backend/issues/722689578/events/34109731460/ this one flagged from antinod.es

yanokwa:
dirtree leakage isn't great. It'll tend to leak username, organization, or project name. If it's easy to remove, we should remove it. If it's a huge lift, then I'm fine keeping it because the data currently goes to trusted maintainers.

And I know we talked about this, but just so future ODKers have this in the public record, the error messages are all combined under one issue. It's not great, but it's work that we don't have the cycles for currently. If it becomes a huge pain, then we can look into fixing it.

clint-tseng:
oh. i should clarify. the goal was to make failures more debuggable _anywhere_ so the error details are surfaced it the 500 http response. because otherwise we have literally nothing to go on; this way at least some things appear. for instance for the one logged earlier it looks like this.

```
{
problemCode: 500.1, 
problemDetails: {
error: {
code: 22P02, 
column: [undefined], 
constraint: [undefined], 
dataType: [undefined], 
detail: [undefined], 
file: numutils.c, 
hint: [undefined], 
internalPosition: [undefined], 
internalQuery: [undefined], 
length: 90, 
line: 62, 
message: [Filtered], 
name: error, 
position: [undefined], 
routine: pg_atoi, 
schema: [undefined], 
severity: ERROR, 
stack: 
error: invalid input syntax for integer: "a"
    at Connection.parseE (/usr/odk/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (/usr/odk/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (/usr/odk/node_modules/pg/lib/connection.js:113:22)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at TCP.onread (net.js:597:20)
, 
table: [undefined], 
where: [undefined]
}
}
}
```

it sounds like you'd rather this information not be available to user responses at least. but if we take this information out of sentry logging there is.. literally nothing left. just "hey a 500 happened!" and that's it. without a stack trace there is nothing to debug.

yanokwa:
Thanks for the clarification, @clint-tseng. I think having the stack trace *is* useful. It's just it'd be nicer if the error reporting was more what Sentry expected so the grouping could be better.  You are the one who is going to be looking at the crashes. And maybe improving that could be a nice contributor friendly issue.

clint-tseng:
since virtually everybody is deploying on `docker`, which isolates identifiable information (eg dirpath is always `/usr/odk/…`, we're punting on anonymization for now.

clint-tseng:
@yanokwa @lognaturel still looking for a checkmark here

Issue #144, opened 2018-10-11
Make email login case-insensitive


clint-tseng:
Background [here](https://forum.opendatakit.org/t/error-with-new-account-creation-odk-central/15599/10). Technically emails are case-sensitive, but this does not seem to be reality as convention, and I just checked bandcamp, patreon, a couple others, and they all seem to be happy with insensitivity.

PR #146, opened 2018-10-11
improve #144: make email addresses case-insensitive everywhere.

clint-tseng:
* the good news is that postgres offers this natively, so that you don't
  have to do 'UPPER('input')=UPPER(email)' all over the place, and so
  that the uniqueness constraint accounts for the case-insensitivity.
* the bad news is that it's offered via the CITEXT extension which is
  bundled but must be manually enabled.
* the worst news is that non-superusers cannot enable extensions, which
  crashes the migration.
* but as far as i know our user /is/ a superuser on docker and on
  circleci (which uses docker).
  * so those cases aren't issues.
  * and i've added a line in the README to manually add the extension at
    database creation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Next, create a database and user in Postgres. Either use the same settings as th
 ```sql
 CREATE USER jubilant WITH PASSWORD 'jubilant';
 CREATE DATABASE jubilant with owner=jubilant encoding=UTF8;
+CREATE EXTENSION IF NOT EXISTS CITEXT;
 ```
 
 Then, go to the repository root in a command line (where this README is) and run `make` with no arguments. This will install all npm dependencies and run all necessary migrations on the database; see the [makefile](Makefile) for details.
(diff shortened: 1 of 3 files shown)


clint-tseng:
now confirmed that production docker and circleci are totally happy with this.. so it's just dev env

matthew-white:
[This forum post](https://forum.opendatakit.org/t/inconsistent-results-running-odk-central-v0-2-beta/15558/7) is making me wonder whether the migration will have trouble if there are already two different users whose email addresses are different case-sensitively but the same case-insensitively. Would using `CITEXT` result in a uniqueness constraint violation in that case? Users should be able to get around this, since we have the edit account page on the way, so they'll be able to update one of the emails so that the emails are different even case-insensitively. But maybe we should include information along these lines in the release notes?

clint-tseng:
Yeah, that would cause a failure. I am not very concerned about it, to be honest.

matthew-white:
Sounds good, it does seem like it should be quite rare.

clint-tseng:
@lognaturel or @matthew-white still looking for a checkmark here

clint-tseng:
yes, that's more or less the recommended approach. and, aye; create database is only possible for superuser. not sure where best to publicize it.

lognaturel:
Thanks! Sharing the update steps on the forum should be enough.

PR #148, opened 2018-10-16
bug #147: fix cases where groups/structures have <bind relevant=> nodes.

### LINKED ISSUE #147: Group <bind>ings for eg relevance break schema structure detection
As found by Felix Mutua [here](https://forum.opendatakit.org/t/missing-data-on-the-exported-csv/15691/2).

The Central system that reads XForms and tries to understand their structure didn't understand that `<bind>` nodes might exist that do not have a `type` attribute and are not for atomic values.

So for instance if you have `<bind nodeset="/facility/bank_contacts" relevant="selected( /facility/feature_type , 'Bank')"/>` the schema detection system will think `/facility/bank_contracts` is a value field, it won't read into the nested structure to figure out that it's a group, and it will record a `type` of `broken`.

Thankfully, this is a one-line fix.

clint-tseng:
* or any <bind> node for that instance.
* just needed to filter out <bind> nodes that don't have type=; we only
  care about the type information.
* resolves #147.

### DIFF
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -81,7 +81,7 @@ const getFormSchema = (form) => {
   const modelNode = findOne(root('html'), node('head'), node('model'));
   return traverseXml(form.xml, [
     modelNode(findOne(root(), node('instance'), node())(tree())),
-    modelNode(findAll(root(), node('bind'))(attr())),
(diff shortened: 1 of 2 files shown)


lognaturel:
TRICKY!

PR #154, opened 2018-11-03
big instance/model refactor

clint-tseng:
still in progress but creating a PR for now for tracking.


clint-tseng:
this is ready but it probably needs an interactive review

lib/model/query/builder.js
```diff
@@ -10,21 +10,25 @@
...
+      const result = module[methodName](...args)(container)
+      if (result.pipe != null) return resolve(result);
+      return result.catch(postgresErrorToProblem);
+    };
   }
 
+  // helper-nested functions differ in that we never assume the result is a run
+  // query, and so we do not coerce catch on it. this is important for streams.
   if (module.helper != null) {
-    result.helper = {};
```
clint-tseng: ??

clint-tseng: okay this is resolved. it worked by accident. having the {} assignment is better, it is now restored.

PR #169, opened 2018-12-20
new #132: include submitter id and name in .csv.zip export.

### LINKED ISSUE #132: Include app user display name in CSV export
A [post in the Forum](https://forum.opendatakit.org/t/odk-central-app-username-on-the-csv/15433) mentions that the app user display name appears in the Frontend submissions table but not the exported CSV file. @clint-tseng, what do you think about adding this to the CSV file? I could see how it could be useful to have that data in the CSV.

Also, I know we include the app user display name in the OData under the `__system` property, but I was wondering whether you would consider that part of the official Central API.

clint-tseng:
* adds them at the very end.
* if we cannot find submitter information we omit the data for the row.
* resolves #132.

### DIFF
--- a/lib/resources/submissions.js
+++ b/lib/resources/submissions.js
@@ -12,6 +12,7 @@ const { always } = require('ramda');
 const sanitize = require('sanitize-filename');
 const { createdMessage } = require('../outbound/openrosa');
 const { getOrNotFound, getOrReject, rejectIf, reject } = require('../util/promise');
+const { QueryOptions } = require('../util/db');
 const { xml } = require('../util/http');
 const Option = require('../util/option');
 const Problem = require('../util/problem');
@@ -108,7 +109,8 @@ module.exports = (service, endpoint) => {
         .then(() => project.getFormByXmlFormId(params.id))
         .then(getOrNotFound)
         .then((form) => all.do([
-          Submission.streamRowsByFormId(form.id),
+          // we always want the extented query; nothing to do w the user request.
+          Submission.streamRowsByFormId(form.id, QueryOptions.extended),
           Submission.streamAttachmentsByFormId(form.id)
         ]).then(([ rows, attachments ]) => {
           const filename = sanitize(form.xmlFormId);
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -267,13 +267,13 @@ describe('api: /forms/:id/submissions', () => {
             zipStreamToFiles(asAlice.get('/v1/projects/1/forms/simple/submissions.csv.zip'), (result) => {
               result.filenames.should.eql([ 'simple.csv' ]);
               const csv = result['simple.csv'].split('\n').map((row) => row.split(','));
-              csv[0].should.eql([ 'SubmissionDate', 'meta-instanceID', 'name', 'age', 'KEY' ]);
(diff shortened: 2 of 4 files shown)


test/integration/api/submissions.js
```diff
@@ -267,13 +267,13 @@ describe('api: /forms/:id/submissions', () => {
             zipStreamToFiles(asAlice.get('/v1/projects/1/forms/simple/submissions.csv.zip'), (result) => {
               result.filenames.should.eql([ 'simple.csv' ]);
               const csv = result['simple.csv'].split('\n').map((row) => row.split(','));
-              csv[0].should.eql([ 'SubmissionDate', 'meta-instanceID', 'name', 'age', 'KEY' ]);
+              csv[0].should.eql([ 'SubmissionDate', 'meta-instanceID', 'name', 'age', 'KEY', 'SubmitterID', 'SubmitterName' ]);
               csv[1].shift().should.be.an.recentIsoDate();
-              csv[1].should.eql([ 'three','Chelsea','38','three' ]);
+              csv[1].should.eql([ 'three','Chelsea','38','three', '5', 'Alice' ]);
```
matthew-white: Am I right in thinking that the CSV includes submitter ID and name only if extended metadata is requested? Or does the CSV include submitter ID and name all the time? This request doesn't seem to request extended metadata, so I was surprised that the resulting CSV includes submitter ID and name.

matthew-white: Also, if extended metadata is used in this way, it might be useful to have one integration test that requests extended metadata and one that doesn't.

test/unit/data/briefcase.js
```diff
@@ -58,7 +60,7 @@ describe('.csv.zip briefcase output @slow', () => {
     callAndParse(form, inStream, (result) => {
       result.filenames.should.eql([ 'mytestform.csv' ]);
       result['mytestform.csv'].should.equal(
-`SubmissionDate,name,age,hometown,KEY
+`SubmissionDate,name,age,hometown,KEY,SubmitterID,SubmitterName
```
matthew-white: What do you think about including column headers for submitter ID and name only if that information was requested? Right now it seems like those headers are included whether or not extended metadata was requested.

test/integration/api/submissions.js
clint-tseng: i think it’s relatively impractical to gate the presence of columns on a header field. how would you ask the browser to download such an object?

matthew-white: Ah I see, I was understanding the addition of `QueryOptions.extended` in the code as meaning that extended metadata was being used to gate. I see now that the CSV endpoint just always passes `QueryOptions.extended` in order to get the data it needs to always add submitter ID and name to the CSV. I agree that not gating is preferable!

clint-tseng:
oh i see; i can leave a clarifying comment on that.

lib/resources/submissions.js
```diff
@@ -108,7 +109,7 @@ module.exports = (service, endpoint) => {
         .then(() => project.getFormByXmlFormId(params.id))
         .then(getOrNotFound)
         .then((form) => all.do([
-          Submission.streamRowsByFormId(form.id),
+          Submission.streamRowsByFormId(form.id, QueryOptions.extended),
```
clint-tseng: //

Issue #151, opened 2018-11-01
Email user when email address or password changes


matthew-white:
As a user, when I update my email address or password, I should receive an email informing me of the change.

In the case of an email address change, I'm not sure whether the email should be sent to the old address, the new one, or both...

alexpatow:
Hi, can I work on this? If so, any what should the text of the emails be?

clint-tseng:
yes, definitely! thanks! my suggestions for text:

password change, email to address on file:

```
Hello!

We are emailing because you have an ODK Central data collection account, and somebody has just changed its password. If this was you, please feel free to ignore this email. Otherwise, please contact your local ODK system administrator immediately.
```

email change, email to old address only:

```
Hello!

We are emailing because you have an ODK Central data collection account, and somebody has just changed the email address associated with the account from this one you are reading right now ($OLDADDRESS) to a new address ($NEWADDRESS). If this was you, please feel free to ignore this email. Otherwise, please contact your local ODK system administrator immediately.
```

alexpatow:
Great, thanks!

PR #172, opened 2018-12-28
improve #151: email notifications for password/email changes

alexpatow:
#151 Adds email notifications when a user updates their email address or password. When a user changes their email address, the email is sent to the email address which was previously on file.

### DIFF
--- a/lib/resources/users.js
+++ b/lib/resources/users.js
@@ -11,6 +11,7 @@ const { success, isTrue } = require('../util/http');
 const { verifyPassword } = require('../util/crypto');
 const Problem = require('../util/problem');
 const { resolve, reject, getOrNotFound } = require('../util/promise');
+const { isPresent } = require('../util/util');
 
 module.exports = (service, endpoint) => {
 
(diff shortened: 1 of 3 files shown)


lib/resources/users.js
```diff
@@ -80,21 +81,29 @@ module.exports = (service, endpoint) => {
...
 
   // TODO: infosec debate around 404 vs 403 if insufficient privs but record DNE.
-  service.patch('/users/:id', endpoint(({ User }, { params, body, auth }) =>
+  service.patch('/users/:id', endpoint(({ User, mail }, { params, body, auth }) =>
     User.getByActorId(params.id)
       .then(getOrNotFound)
       .then((user) => auth.canOrReject('user.update', user.actor)
-        .then(() => user.with(User.fromApi(body)).update()))));
+        .then(() => user.with(User.fromApi(body)).update())
+        .then((response) => {
```
clint-tseng: i'd prefer this to be named `result` or `savedUser` than `response`, since that name in this context implies to me the HTTP/Express response object.

lib/resources/users.js
```diff
@@ -80,21 +81,29 @@ module.exports = (service, endpoint) => {
...
+          if (isPresent(body.email) && body.email !== user.email) {
```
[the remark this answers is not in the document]
clint-tseng: it's not my favorite either but i can't think of a more sensible spot at the moment.

lib/resources/users.js
alexpatow: Sounds good, changed to result in latest commit.

test/integration/api/users.js
```diff
@@ -240,6 +240,31 @@ describe('api: /users', () => {
...
+        asAlice.get('/v1/users/current')
+          .expect(200)
+          .then((before) => asAlice.patch(`/v1/users/${before.body.id}`)
+            .send({ email: 'david123@opendatakit.org' })
+            .expect(200)
+            .then(() => {
+              const email = global.inbox.pop();
+              global.inbox.length.should.equal(0);
+              email.to.should.eql([{ address: 'alice@opendatakit.org', name: '' }]);
+              email.subject.should.equal('ODK Central account email changed');
```
clint-tseng: sorry, one last nit: can you also test that the appropriate values are templated into this body?

alexpatow: Will do!

Issue #150, opened 2018-10-30
Pick up and save deviceIds on form submission


clint-tseng:

alexpatow:
@clint-tseng as mentioned in slack, I'll take a look at this

alexpatow:
@clint-tseng 

Is this the endpoint which needs to grab the deviceId: [https://odkcentral.docs.apiary.io/#reference/openrosa-endpoints/openrosa-form-submission-api/openrosa-form-submission-api](https://odkcentral.docs.apiary.io/#reference/openrosa-endpoints/openrosa-form-submission-api/openrosa-form-submission-api)? 

It looks like the deviceId is part of the query string in the POST request (taken from the Collect app logs):

```
I/HttpClientConnection: Issuing POST request to: https://opendatakit.appspot.com/submission?deviceID=imei%3A358240051111110
```

Please let me know if this approach sounds good:
- new migration to create deviceId column in submissions table
- get deviceId from req.query.deviceID in `resources/submissions`
- extend the `PartialSubmission.complete` method in `model/instance/submission` to accept deviceId as a parameter
- change `.orElseGet(() => partial.complete(form, auth.actor()).create()))` in `resources/submissions` to `.orElseGet(() => partial.complete(form, deviceId, auth.actor()).create()))`
- add deviceID query string to integration tests

Questions: 
- do we need to do any formatting on the deviceID in the query string (looks like it comes in as encoded "imei:<id>")?
- How do these lines work, will I need to add deviceId to them?
```
  all: [ 'id', 'formId', 'instanceId', 'xml', 'submitter', 'createdAt', 'updatedAt', 'deletedAt' ],
  readable: [ 'instanceId', 'submitter', 'createdAt', 'updatedAt' ]
```

Thanks,

Alex

clint-tseng:
hey alex! CIL

> Is this the endpoint which needs to grab the deviceId: https://odkcentral.docs.apiary.io/#reference/openrosa-endpoints/openrosa-form-submission-api/openrosa-form-submission-api?

yeah, that's the one.

> It looks like the deviceId is part of the query string in the POST request (taken from the Collect app logs):
> 
> ```
> I/HttpClientConnection: Issuing POST request to: https://opendatakit.appspot.com/submission?deviceID=imei%3A358240051111110
> ```

great, this was what i sort of remembered from past discussion but i wasn't sure about the details. that seems right to me. i wouldn't count on it always being present, though.

> Please let me know if this approach sounds good:
> 
[...]

PR #175, opened 2018-12-30
improve #150: adds storage of device id on submit

alexpatow:
When a submission is created in the submission endpoint, the device id is now recorded using the deviceID query string property (used by Collect app).

More info: #150

### DIFF
--- a/lib/model/instance/submission.js
+++ b/lib/model/instance/submission.js
@@ -36,28 +36,33 @@ const out = {};
 
 // TODO/CR: should this be in its own file for consistency?
 out.PartialSubmission = Instance()(({ Submission }) => class {
-  complete(form, maybeActor) {
+  complete(form, maybeActor, maybeDeviceId) {
     // TODO it's awkward that this error/message is thrown here; should be closer to edge.
     if (this.version !== form.version)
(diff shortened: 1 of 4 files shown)


lib/model/instance/submission.js
```diff
@@ -36,28 +36,33 @@ const out = {};
 
 // TODO/CR: should this be in its own file for consistency?
 out.PartialSubmission = Instance()(({ Submission }) => class {
-  complete(form, maybeActor) {
+  complete(form, maybeActor, maybeDeviceId) {
```
clint-tseng: nit:

so, here it's `complete(form: Form, maybeActor: Option[Actor], maybeDeviceId: String?)`.

the `maybeActor` must not be null, but it also must be an `Option`, whereas `maybeDeviceId` may be null. but they're both named `maybe` which is confusing.

i would suggest just naming it `deviceId` and either doing `deviceId = null` to indicate it can be nulled, or else leaving a comment explaining what the types are here.

alexpatow: Sounds good! I wasn’t sure how the “maybe” prefix was being used, so I just went with it. I’ll add the “ = null” to indicate it’s nullable.

alexpatow:
Thanks for the comments, I’ll have some time later in the week to update the PR.

clint-tseng:
perfect, thank you!

PR #185, opened 2019-03-05
miscellaneous fixes

clint-tseng:
see commit messages.

roles-api is a pretty significant refactor of some areas and it's easier to rebase this on top of that than to base master and merge.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,5 +31,5 @@ jobs:
             - node_modules
           key: v1a-dependencies-{{ checksum "package.json" }}
 
-      - run: make test
+      - run: make test-full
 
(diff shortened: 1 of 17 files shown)


clint-tseng:
> I'm noticing an error when I run `node lib/bin/cli.js user-create -u [EMAIL] -p [PASSWORD]`:
> 
> ```
> central-backend/lib/model/instance/user.js:64
>     return crypto.hashPassword(cleartext).then((password) => this.with({ password }));
>                   ^
> 
> TypeError: Cannot read property 'hashPassword' of undefined
> ```

bah, good catch again, fixed.

this is fallout from the fact that we have to do our own sort of dirty injection magic to make the database-rollback-after-each-test thing work.

PR #191, opened 2019-04-19
api/improve: use a 409 code rather than 400 on key conflict errors.

clint-tseng:
* ie if a resource already exists.


tiritea:
👍 
409 seems appropriate when POSTing a resource that already exist

ggalmazor:
Just tried this while working on the Briefcase-Central integration and it works like a charm. Thanks!

PR #208, opened 2019-06-06
audit logging changes and API for v0.6

issa-tseng:


lib/resources/audits.js
```diff
@@ -0,0 +1,17 @@
...
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+module.exports = (service, endpoint) => {
+
+  service.get('/audits', endpoint(({ Audit }, { auth, queryOptions }) =>
+    auth.canOrReject('audits.read', Audit.species())
```
matthew-white: Just wanted to check, should this be `'audit.list'` rather than `'audits.read'`?

issa-tseng: yeah i went back and forth on that. ultimately since audits can't be individually addressed and retrieved over the API i decided that there was no need for both, and ultimately read is more descriptive than list.

matthew-white: I think part of the reason I was surprised is that all the other roles use the singular case (for example, `user.list` and `user.read`). I think consistency around the case and the verb could be helpful, but I also don't have strong feelings if you have a preference for `audits.read`.

issa-tseng: oh oops you mean the plural. yeah i'll fix that then.

matthew-white:
I've been using the audits API as part of Frontend development, and so far it's been going well. I do have one request and one suggestion:

- Could we include `projectId` in the actee object when the actee is a form? In Frontend, I want the form to link to the form overview page for the form, but I think I need the project id for that.
- I'm noticing that when `actorId` is `null`, `actor` is not `null`, but is rather an object whose properties have `null` values. That isn't an issue for Frontend, but I personally would lean toward making `actor` `null` in that case.

issa-tseng:
> * Could we include `projectId` in the actee object when the actee is a form? In Frontend, I want the form to link to the form overview page for the form, but I think I need the project id for that.

yes, but only if we _always_ output `projectId` in form responses.

> * I'm noticing that when `actorId` is `null`, `actor` is not `null`, but is rather an object whose properties have `null` values. That isn't an issue for Frontend, but I personally would lean toward making `actor` `null` in that case.

i will try, but this is not easy to do.

matthew-white:
> yes, but only if we _always_ output `projectId` in form responses.

That sounds good to me! :+1:

> i will try, but this is not easy to do.

I don't have strong feelings — just wanted to suggest it!

issa-tseng:
jk i misunderstood the situation and it was trivial

lib/model/trait/child.js
```diff
@@ -37,7 +37,8 @@ const ChildTrait = (ParentName, { // eslint-disable-line object-curly-newline
     }
 
     forApi() {
-      return merge(this[parentName].forApi(), pick(this.constructor.fields.readable, this));
+      const ownReadable = pick(this.constructor.fields.readable, this);
```
issa-tseng: ??

lib/resources/forms.js
```diff
@@ -68,25 +68,33 @@ module.exports = (service, endpoint) => {
...
     Project.getById(params.projectId)
       .then(getOrNotFound)
       .then((project) => auth.canOrReject('form.update', project)
         .then(() => project.getFormByXmlFormId(params.id))
         .then(getOrNotFound)
-        .then((form) => form.with(Form.fromApi(body)).update())
-        // TODO: sucks to have to re-request but this shouldn't be a perf-critical path.
-        .then(() => project.getFormByXmlFormId(params.id))
-        .then(getOrNotFound))));
+        .then((form) => resolve(Form.fromApi(body)) // TODO: lame.
```
issa-tseng: don't need resolve

test/integration/api/audits.js
```diff
@@ -0,0 +1,350 @@
...
+              audits[2].actee.should.eql(plain(project.forApi()));
+              audits[2].details.should.eql({ data: { name: 'audit project' } });
+              audits[2].loggedAt.should.be.a.recentIsoDate();
+            })))));
+
+    it('should not expand actor if there is no actor', testService((service, { Audit }) =>
+      (new Audit({ action: 'backup', details: '{"success":true}' })).create()
+        .then(() => service.login('alice', (asAlice) =>
+          asAlice.get('/v1/audits')
+            .expect(200)
```
issa-tseng: make the ext request dummy

test/integration/api/users.js
```diff
@@ -113,6 +113,28 @@ describe('api: /users', () => {
...
+            Audit.getLatestByAction('user.create').then((o) => o.get())
+          ])
+          .then(([ alice, david, log ]) => {
+            log.actorId.should.equal(alice.actor.id);
+            log.acteeId.should.equal(david.actor.acteeId);
+            log.details.should.eql({
+              data: {
+                email: 'david@opendatakit.org',
+                actor: { displayName: 'david@opendatakit.org', type: 'user' },
+                password: null
```
issa-tseng: ??

test/integration/task/account.js
```diff
@@ -13,6 +13,17 @@ describe('task: accounts', () => {
...
+
+    it('should log an audit entry', testTask(({ Audit, User }) =>
+      createUser('testuser@opendatakit.org', 'aoeu')
+        .then((result) => Promise.all([
+          User.getByEmail('testuser@opendatakit.org').then((o) => o.get()),
+          Audit.getLatestWhere({ action: 'user.create' }).then((o) => o.get())
+        ]))
+        .then(([ user, log ]) => {
+          log.acteeId.should.equal(user.actor.acteeId);
+          log.details.data.email.should.equal(user.email);
```
issa-tseng: odkcmd

test/integration/api/users.js
issa-tseng: it saves the hashed password we good

PR #210, opened 2019-06-10
odata: sanitize field and group/repeat names

issa-tseng:


issa-tseng:
need to also sanitize the .schema.json output

Issue #205, opened 2019-06-03
bug: return a proper 400 error for too-long xmlFormIds


issa-tseng:
as seen here: https://sentry.io/organizations/opendatakit/issues/862023665/?project=1298632

postgres error `22001` indicates that a value is too long for the field. we limit xmlFormId to 64 bytes so we should inform the user if they have selected too long of a name.

akshay-ap:
@opendatakit-bot claim

PR #215, opened 2019-06-24
bug #205: Handle too-long xmlFormIds

akshay-ap:


test/integration/data.js
```diff
@@ -274,6 +274,35 @@ module.exports = {
       two: instance('binaryType', 'two', '<file2>here_is_file2.jpg</file2>'),
       both: instance('binaryType', 'both', '<file1>my_file1.mp4</file1><file2>here_is_file2.jpg</file2>')
     }
-  }
+  },
+  simpleWithLongFormId: `<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa">
```
issa-tseng: rather than create an entire new blob of XML for this case, i would prefer you just replace the string inline with the test. i do something similar here: https://github.com/opendatakit/central-backend/blob/issa/encryption/test/integration/api/submissions.js#L599

lib/util/db.js
```diff
@@ -241,6 +241,11 @@ const postgresErrorToProblem = (error) => {
         reason: 'not a recognized state name'
       }));
     }
+  } else if (error.code === '22001') {
+    const match = /value too long for type character varying\((\d+)\)/.exec(error.message);
+    if (match != null) {
+      return reject(Problem.user.valueTooLong({ reason: match[0] }));
```
issa-tseng: here you provide a reason, which is just the database error message. but in `lib/util/problem.js` you don't do anything with it, that parameter is thrown away. and, people using the web API won't know what a character varying is.

my suggestion is to make the detail object have `{ maxLength: # }`, and put the matched limit in there.

also, if there is any way (i didn't investigate deeply) to extract any kind of information about the actual input from the postgres error (eg the provided length) it would be nice to incorporate it here.

issa-tseng:
thanks, @akshay-ap . i've left a couple of comments. when you have addressed them, please go ahead and just amend your existing commit and force-push on top of your branch rather than creating a second commit.

lib/util/db.js
akshay-ap: Yes, I shall detail with `{maxLength : # }`
But,I don't think there is any way to know for which field he error occurred in this case. As the error message does not have sufficient information.
```
{ error: value too long for type character varying(64)
    at Connection.parseE (<path>/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (<path>/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (<path>/node_modules/pg/lib/connection.js:113:22)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at TCP.onread (net.js:601:20)
  name: 'error',
  length: 98,
  severity: 'ERROR',
  code: '22001',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'varchar.c',
  line: '624',
  routine: 'varchar' }
```

test/integration/data.js
akshay-ap: I thought we will need new blob as it is a form and not a submission instance. In current test data, we don't have a function that can generate form with required `id`. Please let me know if I am missing anything.

issa-tseng: i broke the link, but as you can see here:
https://github.com/opendatakit/central-backend/blob/7aaab0f/test/integration/api/submissions.js#L685

works just fine with forms.

lib/util/db.js
issa-tseng: does that `98` length property match with the attempted input? if it does, we can at least give that back to the user and that will be some hint as to which field went wrong.

akshay-ap: > oes that `98` length property match with the attempted input?

No `98` length property does not match with the attempted input

issa-tseng: that's too bad, okay

issa-tseng:
great, thanks!

PR #217, opened 2019-06-29
submission encryption

issa-tseng:


lib/util/quarantine/oaep.js
```diff
@@ -0,0 +1,203 @@
...
+//
+// and yet, here we are, implementing our own cryptographic primitives. ultimately,
+// the reason this route was chosen was because we can make certain assumptions
+// in our particular use case that radically simplify the algorithm as typically
+// implemented:
+//
+// 1. the digest length will always be 256 bits. this fact is why we are on this
+//    whole stupid journey to begin with.
+// 2. the final cleartext length will always be 32 bytes. usually, once the final
+//    db is obtained, one must hunt for The One (the first 0x01 byte) starting from
```
lognaturel: Pardon my ignorance -- `db` here and in `unpadPkcs1OaepMgf1Sha256` means "data buffer"?

lib/util/quarantine/oaep.js
```diff
@@ -0,0 +1,203 @@
...
+//
+// createHash/update/digest
+// i sure hope these are cryptographically safe.
+//
+// Buffer.allocUnsafe, Buffer#subarray
+// some branching etc occurs internally but all input values are constant.
+//
+// Buffer#writeInt32BE
+// does some branching (checkInt in buffer.js) but the input sequence is fixed:
+// 0 0 1 2 3 4 5 6
```
lognaturel: I had to think through that double 0. So it's only used for the mask generation and the first 0 is for the seed then it's the db sequence.

lib/util/quarantine/pkcs7.js
```diff
@@ -0,0 +1,82 @@
...
+// unpadded result, or else throws.
+const unpadPkcs7 = (buffer) => {
+  const len = buffer.length;
+  const padLen = buffer[len - 1];
+
+  // check a fixed count of 16 bytes (we know we are using aes) for the
+  // padding byte. accumulate errors in good.
+  let good = ge(16, padLen);
+  for (let i = 0; i < 16; i++) {
+    const mask = ge(padLen - 1, i); // TODO: why do libressl and forge BOTH expect paddingbyte + 1?
```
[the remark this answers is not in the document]
issa-tseng: it just doesn't work at all the way they do it. i think maybe they both do some block cipher things that add a byte somewhere maybe; the actual PKCS#7 RFC states that given `05` you knock 5 bytes off. libressl and forge will dutifully knock 6 instead. dunno.

lib/util/quarantine/oaep.js
```diff
@@ -0,0 +1,203 @@
...
+// EXTERNAL FUNCTION ANALYSIS
+// some analysis and justification of external functions used.
+//
+// Math.ceil (v8)
+// is constant time under numbers smaller than 0x80000000 (in which case a cast
+// to uint32 is performed.) additionally, the only values it may be fed are
+// 32 / 32 = 1, and 223 / 32 = 6.96875, neither of which are sensitive.
+//
+// createHash/update/digest
+// i sure hope these are cryptographically safe.
```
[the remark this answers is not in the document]
issa-tseng: well, if we can't trust the actual cryptographic routines provided natively, we're done for.

lib/util/quarantine/oaep.js
issa-tseng: "data block", technically.

lib/util/quarantine/oaep.js
issa-tseng: yes. i think there's some really stripped down version of this where you actually just unmask the seed and then unmask the final 32 bytes of the data block and that saves a lot of work and looping but it's nice for the algorithm to actually look like the algorithm.

lib/util/xml.js
```diff
@@ -158,13 +158,17 @@ const traverseXml = (input, initTraversers) => new Promise((resolve) => {
   if (typeof input.pipe === 'function') {
     // we have a stream if .pipe() exists:
     const decoder = new StringDecoder('utf8');
-    input.on('data', (chunk) => parser.write(decoder.write(chunk)));
```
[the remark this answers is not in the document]
issa-tseng: because then we can't catch errors

CONTRIBUTING.md
```diff
@@ -55,6 +56,7 @@ We have built a lot of custom framework to try to keep ODK Central Backend as fl
 
 * The business logic defining the functionality and behavior of the server for users is the outermost layer, and it is the easiest to understand and update.
 * The model and database logic defining the Things in the server (Users, Forms, etc) and how they are stored and retrieved from the database are the middle layer. They are still easy to understand and update, but you'll need to understand a little bit more about how other code uses these Things and how the underlying framework works in order to make changes.
+    * Whenever possible, we rely on the database to perform data validation rather than implement it in Javascript code. This way, we can be more certain that invalid data hasn't found its way into databases. When the database complains about data validation problems, we translate those problems into user-friendly errors and return them.
```
lognaturel: 👍

lognaturel:
Took another look-see at the tests Just For Fun. They look very complete to me. I really like all of the new contributor guide text.

Issue #211, opened 2019-06-11
sanitize password change requests on sentry


issa-tseng:

NizamLZ:
I would be happy to work on it, can you suggest a starting point? I don't understand much just by the title of the issue.

issa-tseng:
yes, sorry!

when a user tries to [set their password for the first time](https://github.com/opendatakit/central-backend/blob/master/lib/resources/users.js#L57) or [change their password](https://github.com/opendatakit/central-backend/blob/master/lib/resources/users.js#L104) their password input is sent as plaintext as part of the `POST` body.

my understanding from the documentation (and from similar services i've used in the past) was that the input would automatically be sanitized and not sent to sentry if the term password was involved with the request (though now looking at it even if this were true the initial password set _doesn't_ so i messed up either way).

the goal here is for these two requests to never send the payload data to sentry.

NizamLZ:
@opendatakit-bot claim

PR #229, opened 2019-09-04
improve #211: Sanitize pwd change requests sent to sentry

NizamLZ:
This PR will resolve issue #211 

For sensitive endpoints, request payload will be set to `null` before the event is sent to sentry.


issa-tseng:
looks good, thanks! usually for single-statement blocks i don't put curlies but i don't mind them there either.

NizamLZ:
Ohh. Its different from repo to repo in the open source world. Will keep it in mind. Thanks for mentioning and not rejection the PR.

issa-tseng:
:) yeah totally. i think most people don't like to do it this way, without braces.

PR #236, opened 2019-10-16
xlsform import

issa-tseng:


issa-tseng:
if an X-XlsForm-FormId-Fallback is not set we get a 500 right now.

issa-tseng:
also if the connection fails for some reason it's also a 500, which i'm confused by

PR #237, opened 2019-10-16
miscellaneous improvements

issa-tseng:


docs/api.md
```diff
@@ -280,6 +281,10 @@ Currently, there are no paging or filtering options, so listing `User`s will get
 
 Optionally, a `q` querystring parameter may be provided to filter the returned users by any given string. The search is performed via a [trigram similarity index](https://www.postgresql.org/docs/9.6/pgtrgm.html) over both the Email and Display Name fields, and results are ordered by match score, best matches first.
 
+If a `q` parameter is given, and it exactly matches an email address that exists in the system, that user's details will always be returned, even for actors who cannot `user.list`. The request must still authenticate as a valid Actor. This allows non-Administrators to choose a user for an action (eg grant rights) without allowing full search.
+
+Actors who cannot `user.list` will always receive `[]` with a `200 OK` response.
```
lognaturel: This is if there's no `q` parameter given, right? I wonder if it should come before mention of `q`. With the current ordering, it appears to contradict "that user's details will always be returned, even for actors who cannot `user.list`"

issa-tseng: i guess i'm missing the words "at least". i think it'll be fine.....

PR #242, opened 2019-11-19
bug c#114: strip .enc extension suffix when decrypting attachments.

issa-tseng:
* resolves https://github.com/opendatakit/central/issues/114


issa-tseng:
@lognaturel poke. should be quick.

danbjoseph:
thank you @issa-tseng and @lognaturel !

PR #244, opened 2019-12-16
new: add minimal Enketo integration to support form previews

jnm:
See https://github.com/opendatakit/central-frontend/pull/272. This:

* Adds a new `form-viewer` role for read-only access to a single form only, without access to any submissions;
* Adds a `preview_key` actor type whose session tokens are accepted by `fieldKeyHandler`. Actors of this type are reaped once their `expiresAt` time has passed (PR to [central](https://github.com/opendatakit/central) with cron task is forthcoming);
* Adds an Enketo client wrapper (currently with just one function, `preview`) based on the [XLSForm conversion service client](https://github.com/opendatakit/central-backend/blob/master/lib/util/xlsform.js);
* Adds the `/v1/projects/{projectId}/forms/{xmlFormId}/preview` endpoint, which accepts an empty `POST` and then:
    * Creates a `preview_key` that's valid for 15 minutes;
    * Assigns the `preview_key` to the `form-viewer` role for this particular form;
    * `POST`s the OpenRosa `/v1/key/{previewKey}/projects/{projectId}` location and the `xmlFormId` to Enketo's [`/survey/preview`](https://apidocs.enketo.org/v2/#/post-survey-preview) endpoint;
    * returns Enketo's response to the client.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,18 +13,22 @@ run: base
 	node lib/bin/run-server.js
 
 debug: base
-	node --debug --inspect lib/bin/run-server.js
+	./node_modules/nodemon/bin/nodemon.js --debug --inspect=0.0.0.0:9229 lib/bin/run-server.js
+
+# the default test timeout of 2000 ms is too short for some dev. machines.
+# consider increasing it more if you experience sporadic failures.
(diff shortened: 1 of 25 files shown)


issa-tseng:
closed in favor of #269, which includes these commits.

PR #254, opened 2020-03-08
docs: small changes to API documentation.

matthew-white:


matthew-white:
I think there is also a `publishedAt` property on the form response, but I'm not sure how to add that without also including it in the form draft response.

issa-tseng:
not the most useful commit message, but oh well.

PR #261, opened 2020-04-12
bug: ignore extra repeat instance nodes, if they occur locally.

issa-tseng:


test/unit/data/schema.js
```diff
@@ -234,6 +234,141 @@ describe('form schema', () => {
       });
     });
 
+    it('should ignore further repeat instances', () => {
```
lognaturel, line 237: This test form looks perfect.

PR #276, opened 2020-07-09
v0.9 final fixes

issa-tseng:


docs/api.md
```diff
@@ -32,6 +32,14 @@ Finally, **system information and configuration** is available via a set of spec
 
 Here major and breaking changes to the API are listed by version.
 
+### ODK Central v0.9
+
+ODK Central does not change the API except for one minor breaking change.
```
lognaturel, line 37: I think this should start with "ODK Central v0.9" rather than just "ODK Central"?

issa-tseng: oops

PR #283, opened 2020-08-09
bug: account for possible trailing ; in geotrace and geoshape

### LINKED ISSUE #282: OData exports geotrace/geoshape with empty last coordinate (CSV export works fine)
Here is an example form, which captures geopoints/traces/shapes in all ODK Collect widget flavours. 
Submissions of this form are included in R package ruODK (see docs https://docs.ropensci.org/ruODK/reference/index.html#section-example-data > data geo_*) in both WKT and GeoJSON formats, both unparsed (raw) and parsed by ruODK.

Form : https://sandbox.central.getodk.org/#/projects/14/forms/build_Locations_1589344221/submissions
The exported CSV via "Download all records" contains the coordinates as they should be, e.g.  field `path-location_path_gps` (with line breaks added for readability):
```
-31.997176185855277 115.88501214981079 0.0 0.0;
-31.995860539692956 115.88591940701008 0.0 0.0;
-31.995041912207018 115.88509060442449 0.0 0.0;
-31.993920161739386 115.884690284729 0.0 0.0;
```
The geotrace consists of four coordinates.

The same data, exported via OData to ruODK, contains an empty last (fifth) coordinate:
```
data("geo_gj_raw", package="ruODK")
geo_gj_raw[["value"]][[1]][["path"]][["location_path_gps"]]

... (first three coordinates)
$coordinates[[4]]
$coordinates[[4]][[1]]
[1] 115.8847

$coordinates[[4]][[2]]
[1] -31.99392

$coordinates[[4]][[3]]
[1] 0


$coordinates[[5]]
$coordinates[[5]][[1]]
NULL

$coordinates[[5]][[2]]
NULL

```

Same issue with the same data but in WKT format (linebreaks added for readability):
```
data("geo_wkt_raw", package="ruODK")
geo_wkt_raw[["value"]][[1]][["path"]][["location_path_gps"]]

#> "LINESTRING (
  115.88501214981079 -31.997176185855277 0,
  115.88591940701008 -31.995860539692956 0,
  115.88509060442449 -31.995041912207018 0,
  115.884690284729 -31.993920161739386 0,
  undefined NaN)"
```
Note the trailing ",undefined NaN". https://github.com/ropensci/ruODK/issues/88 gets rid of that last ", undefined NaN".

Where does the empty last coordinate come from?
* Since the CSV export does not contain the last empty coordinate, I assume both ODK Collect (have not checked the XML submission) and the raw data in ODK Central (have not checked in DB) are unaffected.
* Since the empty coordinate is contained in the WKT text (unparsed by ruODK, so to ruODK it's just some text), I infer that ruODK does not introduce it.

Could this be a bug in the OData export?

lognaturel:
Closes #282 

Clients may add trailing semicolons to the end of geotrace or geoshape values.

### DIFF
--- a/lib/data/json.js
+++ b/lib/data/json.js
@@ -244,7 +244,7 @@ const submissionToOData = (fields, table, { xml, encHasData, submission, submitt
               dataPtr[name].properties = { accuracy };
           }
         } else if ((type === 'geotrace') || (type === 'geoshape')) {
-          const pointStrs = text.trim().replace(/;$/, '').split(';');
+          const pointStrs = text.split(';');
 
           // we pay the code cost of writing this loop twice so we gain a tighter
           // and more cpu-cache-friendly loop:
@@ -255,6 +255,8 @@ const submissionToOData = (fields, table, { xml, encHasData, submission, submitt
               const [ lat, lon, altitude/*, accuracy*/ ] = str.split(/\s+/g).map(parseFloat);
               // if we have an empty xml node the split won't yield a point:
               if (((lat == null) || (lon == null)) && (pointStrs.length === 1)) return;
+              // if we didn't get a coordinate, skip. this almost certainly only happens on trailing ;
+              if (Number.isNaN(lat) || Number.isNaN(lon)) continue;
 
               let point = `${lon} ${lat}`;
               if ((altitude != null) && !Number.isNaN(altitude)) point += ' ' + altitude;
@@ -271,6 +273,8 @@ const submissionToOData = (fields, table, { xml, encHasData, submission, submitt
               const [ lat, lon, altitude, accuracy ] = str.split(/\s+/g).map(parseFloat);
               // if we have an empty xml node the split won't yield a point:
               if (((lat == null) || (lon == null)) && (pointStrs.length === 1)) return;
+              // if we didn't get a coordinate, skip. this almost certainly only happens on trailing ;
+              if (Number.isNaN(lat) || Number.isNaN(lon)) continue;
 
               const point = [ lon, lat ];
               if ((altitude != null) && !Number.isNaN(altitude)) point.push(altitude);
--- a/lib/data/json.js
+++ b/lib/data/json.js
@@ -244,7 +244,7 @@ const submissionToOData = (fields, table, { xml, encHasData, submission, submitt
               dataPtr[name].properties = { accuracy };
           }
         } else if ((type === 'geotrace') || (type === 'geoshape')) {
-          const pointStrs = text.split(';');
+          const pointStrs = text.trim().replace(/;$/, '').split(';');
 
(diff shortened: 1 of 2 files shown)


lib/data/json.js
```diff
@@ -244,7 +244,7 @@ const submissionToOData = (fields, table, { xml, encHasData, submission, submitt
               dataPtr[name].properties = { accuracy };
           }
         } else if ((type === 'geotrace') || (type === 'geoshape')) {
-          const pointStrs = text.split(';');
+          const pointStrs = text.replace(/;$/, '').split(';');
```
issa-tseng, line 247: is `; ` etc not possible ?

issa-tseng: actually i'm a little confused how the failure case occurs in the first place. if there is a trailing `;` the lat/lon parsing below should fail and it should not output the point, no?

lognaturel: You mean could there be trailing spaces? No client that I know of currently produces them but I'm happy to add a `trim`.

lognaturel: I didn't add a test for the `trim` since it feels like an "extra". If you prefer that I do, do you think it's ok to slip it into one of the existing cases or should it be a dedicated test?

Didn't see the other comment before for some reason. 

> if there is a trailing ; the lat/lon parsing below should fail and it should not output the point, no?

JS is happy to take null and undefined values and make them into floats like `undefined` and `NaN`. I got the same results as in the issue (#282) when running the tests without the `replace`. Did you expect a crash? There's no explicit existence check for lat or lon. In the case of malformed points like only a single float, you'll end up with something like `37 NaN`.

issa-tseng: ah. the NaN bailout logic didn't get copied down.

lognaturel: It feels strange to me to produce a well-formed trace/shape from a bad input instead of including a hint that something went wrong. The ways to get to bad input are a rogue client, a bad default, or a faulty form expression that generates a trace/shape. I suppose they're not super likely. Anyway, fine by me to go with your approach.

issa-tseng: hmm? i'm not sure how my approach is any different in this regard than yours?

issa-tseng: i just don't like text manip. it's really expensive.

lognaturel: Fair about expense. The difference is that previously malformed points within a trace/shape would be identifiable because they'd include a `NaN`. Now they'll just be skipped.

issa-tseng:
oh. huh. i would have instinctively considered that a bug. since it’s not to spec. but i guess that makes sense.

lib/data/json.js
issa-tseng: oh. huh. i would have instinctively considered that a bug. since it’s not to spec. but i guess that makes sense.

PR #285, opened 2020-08-14
rework path token authentication

issa-tseng:
see commits

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -36,10 +36,14 @@ Here major and breaking changes to the API are listed by version.
 
 ODK Central v1.0 adds Public Links to the API, and makes one minor breaking change.
 
+**Added**:
+
+* The new [Public Link](/reference/forms-and-submissions/'-public-access-links) resource lets you create Public Access Links, granting anonymous browser-based access to submit to your Forms using Enketo.
+
 **Changed**:
 
 * The non-extended App User response no longer includes a `createdBy` numeric ID. To retrieve the creator of an App User, request the extended response.
-* We no longer reject the request if multiple authentication schemes are presented, and instead document the priority order of the different schemes [here](TODO).
+* We no longer reject the request if multiple authentication schemes are presented, and instead document the priority order of the different schemes [here](/reference/authentication).
 
 ### ODK Central v0.9
 
@@ -170,7 +174,7 @@ In practice, there are two types of Actors available in the system today:
 
 In a future version of the API, programmatic consumers will be more directly supported as their own Actor type, which can be granted limited permissions and can authenticate over **OAuth 2.0**.
 
-Next, you will find documentation on each of the three authentication methods described above. It is best not to present multiple credentials. If you do, the first _presented_ scheme out of Bearer, Basic, Cookie, then `/key` token will be used for the request.
+Next, you will find documentation on each of the three authentication methods described above. It is best not to present multiple credentials. If you do, the first _presented_ scheme out of `/key` token, Bearer, Basic, then Cookie will be used for the request. If the multiple schemes are sent at once, and the first matching scheme fails, the request will be immediately rejected.
 
 ## Session Authentication [/v1/sessions]
 
@@ -224,7 +228,7 @@ _(There is not really anything at `/v1/example`; this section only demonstrates 
 
 #### Logging out [DELETE /v1/sessions/{token}]
 
-Logging out is not strictly necessary; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token. To do so, issue a `DELETE` request to that token resource.
+Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token. It is also the way Public Link and App User access are revoked. To do so, issue a `DELETE` request to that token resource.
 
 + Parameters
     + token: `lSpAIeksRu1CNZs7!qjAot2T17dPzkrw9B4iTtpj7OoIJBmXvnHM8z8Ka4QPEjR7` (string, required) - The session bearer token, obtained at login time.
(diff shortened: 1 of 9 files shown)


docs/api.md
```diff
@@ -1737,6 +1743,68 @@ Given a `roleId`, which may be a numeric ID or a string role `system` name, and
...
+
+Anybody in possession of a Public Access Link for a Form can use that link to submit data to that Form. Public Links are useful for collecting direct responses from a broad set of respondents, and can be revoked using the administration website or the API at any time.
+
+The API for Public Links is particularly useful, as it can be used to, for example, programmatically create and send individually customized and controlled links for direct distribution.
+
+To revoke the access of any Link, terminate its session `token` by issuing [`DELETE /sessions/:token`](/reference/authentication/session-authentication/logging-out).
+
++ Parameters
+    + projectId: `2` (number, required) - The numeric ID of the Project
+    + xmlFormId: `simple` (string, required) - The friendly name of this form. It is given by the `<title>` in the XForms XML definition.
```
lognaturel, line 1758: This doesn’t seem right it’s the form I’d, not the title?

docs/api.md
```diff
@@ -1737,6 +1743,68 @@ Given a `roleId`, which may be a numeric ID or a string role `system` name, and
...
+    + Attributes (array[Extended Public Link])
+
+### Creating a Link [POST]
+
+To create a new Public Access Link to this Form, you must send at least a `displayName` for the resulting Actor. You may also provide `once: true` if you want to create a link that [can only be filled by each respondent once](https://blog.enketo.org/single-submission-surveys/). This setting is enforced by Enketo using local device tracking; the link is still distributable to multiple recipients, and the enforcement can be defeated by using multiple browsers or devices.
+
++ Request (application/json)
+    + Attributes
+        + displayName: `my public link` (string, required) - The name of the Link, for keeping track of. This name is displayed on the Central administration website but not to survey respondents.
+        + once: `false` (boolean, optional) - If set to `true`, an Enketo [single submission survey](https://blog.enketo.org/single-submission-surveys/) will be created instead of a standard one, limiting respondents to a single submission each.
```
lognaturel, line 1783: The linked blog post actually describes the behavior if one is false. If one is true, it’s that plus browser protection.

lib/http/middleware.js
```diff
@@ -42,21 +42,15 @@ const versionParser = (request, response, next) => {
...
-  const queryKey = Option.of(request.query.st)
-    .map(decodeURIComponent)
-    .filter((k) => /^[a-z0-9!$]{64}$/i.test(k));
+  const queryKey = Option.of(request.query.st).map(decodeURIComponent);
   queryKey.ifDefined((token) => {
     delete request.query.st;
     // we modify the request url to ensure openRosa gives prefixed responses
     // per the requested token. we have to slice off the /v1.
-    request.originalUrl = `/v1/key/${token}${request.originalUrl.slice(3)}`;
+    request.originalUrl = `/v1/key/${token.replace('/', '%2F')}${request.originalUrl.slice(3)}`;
```
matthew-white, line 53: Instead of replacing `/` with `%2F`, would it work to URL-encode the token? Since there's no longer a test that the `st` query parameter matches `/^[a-z0-9!$]{64}$/i`, I think the token could contain other characters that aren't URL-safe in addition to `/`. Or is that case considered user error?

issa-tseng: i originally did this but EncodeURIComponent doesn't escape things it should and EncodeURI overeagerly escapes.

issa-tseng: or possibly the other way around. but it was sufficiently annoying that i changed it to this.

matthew-white: Ah I see, it looks like `encodeURI()` doesn't escape `/`, but `encodeURIComponent()` does escape `$`. Though it looks like `$` is the only token character that `encodeURIComponent()` will encode. If `$` is encoded, I think the URL is valid, and Backend will decode it if it used in a future request. Is the trade-off that the URL is a little longer and less human-readable? Right now I think the URL is mainly used in the form list and manifest, so maybe that's OK. My instinct is that it's better to have an over-encoded URL than a possibly under-encoded one.

docs/api.md
issa-tseng: weird. bad copy/paste.

lib/http/middleware.js
issa-tseng: it just broke like all the tests so i said fuck it.

matthew-white: Haha broken tests are no fun. I guess one question I have is how risky you consider this. If `originalUrl` is set to an invalid URL, could that lead to an issue?

One other thing I'm just now noticing is that only the first occurrence of `/` is replaced here: I think we'd need to use something like `/\//g` to replace all occurrences. I was wondering (and maybe this is part of this that I don't understand), why encode `/` but not other characters? If it's to ensure that the structure of the URL remains intact, I think we'd need to encode `?` and `#` as well.

One option if encoding `$` is breaking tests is that we could encode the token, then decode `$`:

```js
const encodedToken = encodeURIComponent(token).replace(/%24/g, '$');
request.originalUrl = `/v1/key/${encodedToken}${request.originalUrl.slice(3)}`;
```

docs/api.md
lognaturel: This was just not true. 🤦🏻‍♀️ Sorry!

lib/http/middleware.js
issa-tseng: i consider it not remotely risky. if anybody hits this case they have already given invalid input and the likely result is already a `401`. tbh the fact that i even do the replace is massive overkill. again, by the time this branch is hit the user has already given bad input and is already in a failure case.

lognaturel: That seems fine. But maybe replace all occurrences if you're going to do something?

PR #290, opened 2020-09-18
bug: possible transaction override in enketo worker; needed for update.

issa-tseng:
* actually all updates need for update, apparently, but we'll deal with
  that later.

### DIFF
--- a/lib/model/migrations/20200918-01-repair-publishedat-dates.js
+++ b/lib/model/migrations/20200918-01-repair-publishedat-dates.js
@@ -1,0 +1,23 @@
+// Copyright 2020 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/opendatakit/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+
+const up = (db) => db.raw(`
+update form_defs set "publishedAt"="loggedAt"
+  from (select (details->>'newDefId')::int as "publishedId", "loggedAt"
+    from audits
+    where action = 'form.update.publish') as publishes
+  where "publishedId"=form_defs.id
+    and "publishedAt" is null;
+`);
+
+const down = () => {};
+
+module.exports = { up, down };
+
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -65,10 +65,6 @@ select id from form;`,
     ({ forms, FormDef }) =>
       forms._get(FormDef.WithXml, options.withCondition({ projectId, xmlFormId }), version).then(maybeFirst),
 
-  getByActeeId: (acteeId, options = QueryOptions.none, version) =>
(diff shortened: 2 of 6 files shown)


lib/model/migrations/20200918-01-repair-publishedat-dates.js
```diff
@@ -0,0 +1,23 @@
...
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+
+const up = (db) => db.raw(`
+update form_defs set "publishedAt"="loggedAt"
+  from (select (details->'newDefId')::int as "publishedId", "loggedAt"
```
lognaturel, line 13: I don't understand the `int` cast here! That's jsonb and seems we just want it raw and used by the aliases?

lib/model/migrations/20200918-01-repair-publishedat-dates.js
```diff
@@ -0,0 +1,23 @@
...
+    from audits
+    where action = 'form.update.publish') as publishes
```
lognaturel, line 15: Why the publishes alias?

lognaturel: Ah, subqueries need aliases, right.

lib/model/migrations/20200918-01-repair-publishedat-dates.js
lognaturel: Maybe you meant to use `->>` which returns text? Then I guess that would cast both but that's ok because we don't care about `loggedAt`.

lib/model/query/forms.js
```diff
@@ -149,6 +145,21 @@ select id from form;`,
         .orderBy(db.raw('coalesce(forms."name", forms."xmlFormId")'), 'asc')
         .then(map(unjoin))),
 
+  // TODO: temporarily copy+pasta from _get below pending better resolution
+  // of this transaction nightmare. (also, vestigial options)
```
lognaturel, line 149: Does "vestigial options" means that no callers set options at this time? Asking because the old implementation respected options passed in and this one does not.

lib/model/migrations/20200918-01-repair-publishedat-dates.js
issa-tseng: huh? eventually we want to compare that value to `form_defs.id` which is an `int`. so we must cast it. but i can cast it below sure. figured making the data cleaner sooner would be better.

lib/model/query/forms.js
issa-tseng: more that no options are used. but it's sort of the standard call arg structure rn. but also everything to do w queries needs to be really rethought. i'm not sure what better way to do it exists in js.

lib/model/migrations/20200918-01-repair-publishedat-dates.js
issa-tseng: oh i see, it fails in 9.6 this way.

issa-tseng:
pr is ready again

lognaturel (approved):
One note is that draftToken will still be set for affected form defs. I don’t think it matters if there’s a publishedAt but it is unexpected.

PR #291, opened 2020-09-26
bug: could sometimes fail to export all repeat rows. depends on chunks.

issa-tseng:
* this can be a hard bug to reproduce. the number of bytes of csv
  produced per row of data, as well as the size of the rows coming from
  the database.
* whatever the conditions, the issue is that for whatever reason the csv
  stringify stream doesn't always wait to flush out, so it erroneously
  reports to the zip archive that it's done.
* if we wrap it in a PassThrough it works perfectly.
* there is probably a different solution that involves patching
  csv-stringify, but that's a far more expensive operation than we have
  time for rn.

### DIFF
--- a/test/unit/data/briefcase.js
+++ b/test/unit/data/briefcase.js
@@ -395,7 +395,7 @@ Candace,2,three,three/children/child[1]
       </h:html>`;
 
     const inStream = streamTest.fromObjects(
-      (new Array(50)).fill(null).map(_ => instance(uuid(), `<orx:meta><orx:instanceID>${uuid()}</orx:instanceID></orx:meta><name>${uuid()}</name><children><child>${uuid()}</child></children>`)));
+      (new Array(50)).fill(null).map(_ => instance(uuid(), `<orx:meta><orx:instanceID>${uuid()}</orx:instanceID></orx:meta><name>${uuid()}</name><children><child><name>${uuid()}</name></child></children>`)));
 
     callAndParse(inStream, formXml, 'singlerepeat', (result) => {
       result.filenames.should.containDeep([ 'singlerepeat.csv', 'singlerepeat-child.csv' ]);
--- a/test/unit/data/briefcase.js
+++ b/test/unit/data/briefcase.js
@@ -1,5 +1,6 @@
 const appRoot = require('app-root-path');
 const should = require('should');
+const uuid = require('uuid/v4');
 const { construct } = require('ramda');
 const streamTest = require('streamtest').v2;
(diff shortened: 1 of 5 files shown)


issa-tseng:
fwiw the unit test never repro'd the issue. it took the whole service, as well as somewhat exactly the test data you gave me for it to occur.

lognaturel (changes requested):
We’ve talked about this privately but want to make sure the PR status is clear in case anyone else takes a look.

The unit test can be made to fail before code changes by adding 4 child repeat instances per submission. With the code change here, it fails with 12 child repeat instances or 526+ of the original submissions (single repeat instance).

test/unit/data/briefcase.js
```diff
@@ -352,6 +353,59 @@ Candace,2,three,three/children/child[1]
...
+              <input ref="/data/children/child/name">
+                <label>What is the child's name?</label>
+              </input>
+            </repeat>
+          </group>
+        </h:body>
+      </h:html>`;
+
+    const inStream = streamTest.fromObjects(
+      (new Array(50)).fill(null).map(_ => instance(uuid(), `<orx:meta><orx:instanceID>${uuid()}</orx:instanceID></orx:meta><name>${uuid()}</name><children><child><name>${uuid()}</name></child></children>`)));
```
lognaturel, line 398: This test passes without the change. Is that deliberate? It's unlikely that we'd have the exactly same regression so maybe it's ok. I'd tend to do 127 submissions, though, which does fail.

test/unit/data/briefcase.js
```diff
@@ -352,6 +353,59 @@ Candace,2,three,three/children/child[1]
...
+
+    callAndParse(inStream, formXml, 'singlerepeat', (result) => {
+      result.filenames.should.containDeep([ 'singlerepeat.csv', 'singlerepeat-child.csv' ]);
+      result['singlerepeat.csv'].split('\n').length.should.equal(52);
+      result['singlerepeat-child.csv'].split('\n').length.should.equal(52);
+      console.log('aoeuaoeu');
```
lognaturel, line 404: Remove?

test/unit/data/briefcase.js
issa-tseng: yeah i didn't think it mattered but if 127 verifies the change let's do it.

lognaturel (approved):

matthew-white (approved):
I looked over 3889ee3ff2f3c9370907622cee51721cb1894717, which seemed reasonable. :+1:

PR #305, opened 2020-11-02
wip: odata $filter

issa-tseng:


lib/data/odata-filter.js
```diff
@@ -0,0 +1,87 @@
...
+      sql.push('??');
+      bindings.push('actorId'); // TODO: HACK HACK
+    } else {
+      throw Problem.internal.unsupportedODataField({ at: node.position, text: node.raw });
+    }
+  } else if (node.type === 'Literal') {
+    sql.push('?');
+    bindings.push(node.raw); // OR DO WE WANT TO USE .VALUE HERE?
+  } else if (node.type === 'MethodCallExpression') {
+    methodCall(node.value.method, node.value.parameters, sql, bindings);
```
matthew-white, line 50: What happens if the method isn't supported?

lib/data/odata-filter.js
```diff
@@ -0,0 +1,87 @@
...
+  } else if (node.type === 'EqualsExpression') {
+    binaryOp(node.value.left, node.value.right, '=', sql, bindings);
```
matthew-white, line 52: What happens if `node.value.left` and `node.value.right` aren't the same type? For example, something like `__system/submitterId lt 2020-01-01`?

matthew-white (commented):
We'll be reviewing this interactively, but I wanted to note a couple of questions I have going into the review.

matthew-white (approved):
✅ so that issa can merge after making some small changes

lib/data/odata-filter.js
```diff
@@ -0,0 +1,87 @@
...
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const odataParser = require('odata-v4-parser');
+const Problem = require('../util/problem');
+
+////////////////////////////////////////
+// AST NODE TRANSFORMATION
+
+const extractFunctions = [ 'date', 'time', 'year', 'month', 'day', 'hour', 'minute', 'second' ];
```
matthew-white, line 16: Ah I didn't think about this during review, but I'm wondering now whether Postgres `extract()` supports `date` or `time`. (I don't see them in the list here: https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT.)

issa-tseng: okay it looks weird and complicated i'll just take em out.
]

issa-tseng: okay it isn't that complicated but also actually i dunno what good those funcs are rn.

PR #312, opened 2020-12-18
bug: filter by formid when getting OData count

lognaturel:
Additional fix to 9ae6b46fd4f82b435016e98e254488a3aef15ca1


lib/outbound/odata.js
```diff
@@ -345,7 +345,7 @@ const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, t
       const totalCount = (tableCount != null) ? tableCount : counted;
       const nextUrl = nextUrlFor(limit, offset, totalCount, originalUrl);
 
-      // we do toStrring on the totalCount because mustache won't bother rendering
+      // we do toString on the totalCount because mustache won't bother rendering
```
issa-tseng, line 348: i sort of intentionally left this alone when i saw it since i value git blame over a decipherable typo. but it's here already.

issa-tseng (approved):

PR #316, opened 2020-12-22
file extension fix and createdAt sort change

issa-tseng:
i tried for a while to create a database index that would make createdAt not ~30% slower than id sort but it can't be done i think.

### DIFF
--- a/lib/model/query/submission-defs.js
+++ b/lib/model/query/submission-defs.js
@@ -103,7 +103,7 @@ module.exports = {
                 .orWhereIn('form_defs.keyId', keyIds))))
           .modify(applyODataFilter(options.filter))
           .where({ 'submissions.formId': formId, 'submissions.deletedAt': null })
-          .orderBy('submissions.id', 'desc')
+          .orderBy('submissions.createdAt', 'desc')
           .modify(applyPagingOptions(options)),
         unjoin
       ))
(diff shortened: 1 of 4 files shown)


lib/resources/submissions.js
```diff
@@ -226,7 +226,8 @@ module.exports = (service, endpoint) => {
           Key.getDecryptor(passphrases)
         ]).then(([ fields, rows, decryptor ]) => {
           const filename = sanitize(form.xmlFormId);
-          response.append('Content-Disposition', `attachment; filename="${filename}.csv"`);
+          const extension = (rootOnly === true) ? 'csv' : 'csv.zip';
+          response.append('Content-Disposition', `attachment; filename="${filename}.${extension}"`);
           response.append('Content-Type', 'text/csv');
```
matthew-white, line 231: How about we also specify `application/zip` if `rootOnly` is `true`?

lib/model/query/submission-defs.js
```diff
@@ -103,7 +103,7 @@ module.exports = {
                 .orWhereIn('form_defs.keyId', keyIds))))
           .modify(applyODataFilter(options.filter))
           .where({ 'submissions.formId': formId, 'submissions.deletedAt': null })
-          .orderBy('submissions.id', 'desc')
+          .orderBy('submissions.createdAt', 'desc')
```
matthew-white, line 106: Do we want to order by `createdAt` and then also `id` so that if there are ties on `createdAt`, the order is stable? That has to be very unlikely, right?

matthew-white, line 106: This feels less urgent, but for consistency I think we should also update `Submission.getAllByFormId()` so that it orders by `createdAt` rather than `id`. That method is used by the regular/non-OData &hellip;/submissions endpoints.

matthew-white (commented):
Just a few small questions/comments! It also looks like CircleCI failed, though looking at the log, maybe it's a CircleCI issue?

matthew-white (approved):

PR #320, opened 2020-12-25
Export all keepalive data for a longer-running backup

matthew-white:
This code is only run if the end of the first chunk is reached, which will happen for a longer-running backup. In that case, we reset `ptr` so that all data after the first chunk is exported.


issa-tseng:
🤦🏾‍♀️👏

issa-tseng (approved):

PR #324, opened 2021-01-02
Replace opendatakit with getodk

yanokwa:
I've tried to do this safely, but this PR is high-risk because it touches a lot of files and needs careful review! Ideally, this should be merged early in the release cycle to further de-risk it.

I've been careful not to touch:
* XML namespace URLs because those should not change.
* OData namespace because I don't understand the implications. If they can be changed with little impact, they probably should be changed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,7 +83,7 @@ If you override the default `forApi` and `fromApi`, it will be up to you to main
 
 ### Form Encryption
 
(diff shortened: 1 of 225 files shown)


lognaturel (approved):

lognaturel (approved):
I scrolled through every file. It's mostly license headers and email addresses in tests.

issa-tseng:
left it too long sorry. obsoleted by #427.

PR #328, opened 2021-02-01
docs: update Slack link in contributing guidelines.

matthew-white:
Changing the link from sign-in to registration, using the same link as
https://docs.getodk.org/docs-tech-guide/.


matthew-white:
The sign-in page doesn't have a link to registration, so if you aren't registered yet, it can be confusing about what to do next.

lognaturel (approved):

PR #331, opened 2021-02-10
refactor

issa-tseng:


matthew-white (approved):
🎉🎉🎉

lib/model/frames/submission.js
```diff
@@ -7,43 +7,38 @@
...
-    const actorId = Option.of(maybeActor).map((actor) => actor.id).orNull();
-    const formattedDeviceId = blankStringToNull(deviceId);
-    return submissions.create(this, form, actorId, formattedDeviceId);
-  }
-
-  // just a stub for now; we don't need this feature just yet.
-  //createVersion(form, submission, maybeActor) { throw new Error('NYI'); }
+class Submission extends Frame.define(
+  table('submissions'),
+  'id',         readable,               'formId',       readable,
```
matthew-white, line 23: I think we should consider not making these two fields readable: I think `instanceId` is the id that API users should use for the submission, not `id`; we also generally don't expose `formId` (as opposed to `xmlFormId`) over the API.

issa-tseng: yeah that's oops

issa-tseng:
this will be ready to merge as soon as slonik is merged upstream and updated here.

PR #332, opened 2021-02-20
submission versions and review state

issa-tseng:


lib/model/query/submissions.js
```diff
@@ -61,71 +62,94 @@ const createVersion = (partial, deprecated, form) => ({ one, context }) => {
...
+  _getDef(maybeOne, options.withCondition({ 'submission_defs.instanceId': instanceId }), formId, draft);
+
+const getDefsByFormAndLogicalId = (formId, instanceId, draft, options = QueryOptions.none) => ({ all }) =>
+  _getDef(all, options.withCondition({ 'submissions.instanceId': instanceId }), formId, draft);
+
+const getRootForInstanceId = (formId, instanceId, draft) => ({ maybeOne }) => maybeOne(sql`
+select submissions."instanceId" from submissions
+join submission_defs on submission_defs."instanceId"=${instanceId}
+  and submission_defs."submissionId"=submissions.id
+where "formId"=${formId} and draft=${draft}`)
```
matthew-white, line 151: Should deleted submissions be filtered out?

lib/util/http.js
```diff
@@ -68,6 +68,24 @@ const xml = contentType('application/xml');
 const atom = contentType('application/atom+xml');
 const json = contentType('application/json');
 
+const binary = (type, name, content) => (_, response) => {
+  response.set('Content-Type', type);
+  response.set('Content-Disposition', `attachment; filename="${name}"`);
```
matthew-white, line 73: Do we need to sanitize `name`?

matthew-white (commented):
Questions before code review

matthew-white (approved):

matthew-white:
I took a look at `actionCondition()` in `query/audits.js`, and it has some logic around submissions that I didn't remember:

```js
  if (action === 'nonverbose')
    return sql`action not in ('submission.create', 'submission.attachment.update', 'backup')`;
  else if (action === 'user')
    return sql`action in ('user.create', 'user.update', 'user.delete', 'assignment.create', 'assignment.delete')`;
  else if (action === 'project')
    return sql`action in ('project.create', 'project.update', 'project.delete')`;
  else if (action === 'form')
    return sql`action in ('form.create', 'form.update', 'form.delete', 'form.attachment.update', 'form.update.draft.set', 'form.update.draft.delete', 'form.update.publish')`;
    return sql`action in ('submission.create', 'submission.attachment.update')`;
  return sql`action=${action}`;
};
```

Now that there's `submission.update` and `submission.update.version`, I think those should be included for `action === 'submission'`.

`action === 'nonverbose'` excludes the existing submission actions. Should it also exclude the new submission actions? My instinct is that it should, because if submission edits are part of your workflow, you probably don't want to see those on the System audits page.

matthew-white:
I'm noticing that at least some of the `PATCH` endpoints don't seem to be calling `forApi()`. For example, the submission `PATCH` endpoint is returning an object with a `deletedAt` property even though the `Submission` frame doesn't include `deletedAt`. I think I'm seeing something similar for the project `PATCH` endpoint.

If they don't already, maybe it'd be useful for tests of `PATCH` endpoints to make assertions about the response body?

lib/util/http.js
matthew-white: It looks like this Sentry error is related to `Content-Disposition` containing an invalid character: https://sentry.io/organizations/getodk/issues/2251111354/

lib/resources/submissions.js
```diff
@@ -281,47 +335,52 @@ module.exports = (service, endpoint) => {
...
+        Submissions.getByIds(params.projectId, params.formId, params.instanceId, draft, queryOptions)
+      ])
+        .then(([ form, maybeSub ]) => Promise.all([
+          auth.canOrReject('submission.read', form),
+          maybeSub.map(resolve).orElseGet(() =>
+            Submissions.getRootForInstanceId(form.id, params.instanceId, draft)
+              .then(getOrNotFound)
+              .then((rootId) => redirect(originalUrl.replace(
+                `/submissions/${params.instanceId}`,
+                `/submissions/${rootId}/versions/${params.instanceId}`
```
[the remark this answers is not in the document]
issa-tseng: i encoded these but it's probably a problem in more than this place. i think in my head i still assume they're uuids even though i shouldn't.

matthew-white:
@issa-tseng, just wanted to make sure that you saw my comments above about `actionCondition()` and about PATCH endpoints.

Also, I think `X-Action-Notes` needs to be percent-decoded.

lib/resources/submissions.js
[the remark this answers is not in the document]
matthew-white: It looks like c1e41ba encodes the second URL, but I think the first still needs to be encoded.

issa-tseng:
i don’t think it should be? it’s comparing against a decoded string no?

matthew-white:
I thiiink `originalUrl` is encoded, but `params.instanceId` is decoded. If that's right, I think `params.instanceId` need to be encoded in order to be found in `originalUrl`.

Issue #333, opened 2021-02-25
Include "instance ID" first column to the combined client audit CSV export


lognaturel:
Collect can record an audit log of data collectors’ actions if it is requested by a form ([documentation](https://docs.getodk.org/form-audit-log/)). For forms that request audits, a Central ZIP export includes a summary CSV that combines logs from all submissions. That export does not connect audit events to the submissions they belong to [as reported on the forum](https://forum.getodk.org/t/31887).

The implementation will require careful changes to the database schema so that the export remains performant.

matthew-white:
Closed by #335. 🎉

PR #335, opened 2021-03-04
include instanceId in client audit csv export

ktuite:
This addresses issue #333 and forum comment https://forum.getodk.org/t/update-the-audit-log-structure/31887. Now each row of client audit events contains an extra `uuid` column referring to the unique ID of the corresponding submission.

The new `uuid` column id named that because of the forum post, but it should possibly be named `KEY` or `instanceID` to be more consistent with other exported files.

Turns out, although there's a complicated query in `model/query/client-audits.js` spanning many tables, the `submissions` table with `instanceId` column is already one of the tables active in that big join so this doesn't impact the database performance.

### DIFF
--- a/lib/data/client-audits.js
+++ b/lib/data/client-audits.js
@@ -59,8 +59,10 @@ const parseClientAudits = (buffer) => {
 };
 
 // helper for streamClientAudits below.
-const formatRow = (row) => {
+const formatRow = (row, instanceId) => {
   const out = [];
+  // prepend instance id (not part of ClientAudit headers, but extracted nearby)
+  out.push(instanceId);
   for (const header of headers) out.push(row[header]);
   return out;
 };
@@ -73,6 +75,9 @@ const streamClientAudits = (inStream, form) => {
   const csvifier = new Transform({
     objectMode: true,
     transform(x, _, done) {
+      // data here contains ClientAudit attchement info as well as associated
+      // submission instanceId fetched through query in
+      // model/query/client-audits.js
       const data = x.row;
 
       // TODO: we do not currently try/catch this block because it feels low risk.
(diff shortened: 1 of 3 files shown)


lognaturel:
Yay for your first PR! 🎉

Oops, I wasn't very explicit in the issue. The new column should be the first one and it should be named `instance ID`. This is to exactly match what Briefcase exports for the same dataset so folks can go between the two and use the same analysis pipeline. Hopefully we can eventually deprecate Briefcase but for now there are a few things people need from it that are not in Central such as splitting select multiples and stripping group names ([docs](https://docs.getodk.org/briefcase-using/#export-forms-to-csv)).

You and @issa-tseng will no doubt discuss the details. I have a couple high-level/process suggestions. Github has a nifty "draft" feature you might consider using for PRs that are not ready to merge. You can also convert a PR once it's up. [The contribution docs](https://github.com/getodk/central-backend/blob/master/CONTRIBUTING.md) have some guidance on writing commit messages.

ktuite:
I've moved the `instanceID` field to the first column! And updated my commit message to better match style guidelines.

I've also tried it on more realistic data (submissions collected through Collect app with auditing enabled).

**Central export screenshot**
Screen Shot 2021-03-05 at 12 32 52 PM

I've noticed a couple other things:
1. Briefcase audit export has different columns/headers. It doesn't explicitly say "old value"/"new value" but it does have an extra unnamed change reason column. It also doesn't include the GPS audit info available from Central export.

**Briefcase export screenshots**
*different headers*
Screen Shot 2021-03-05 at 12 37 16 PM

*change reason present*
Screen Shot 2021-03-05 at 12 43 12 PM

2. Because of how the Collect app appears to handle client audits as attachments all named `audit.csv`, the media export folder includes a single `audit.csv` file. This might be the source of the forum commenter (https://forum.getodk.org/t/update-the-audit-log-structure/31887) saying:

> I have seen that the audit CSV file included in the *.zip now merges the audit data for all submissions, which is great (while it seems to me in previous versions it was only available for the last submission).

Screen Shot 2021-03-05 at 12 45 08 PM

lognaturel:
Thanks for actually trying the Briefcase export to compare! I think you're mostly running into https://github.com/getodk/briefcase/issues/740. Briefcase just concatenates all the CSVs and uses a fixed header which definitely should be addressed at some point. I would definitely expect the location columns if https://docs.getodk.org/form-audit-log/#location-tracking is enabled on the Briefcase side but it's not clear to me whether maybe they're after all the change columns and not visible because they have no values and no header? I use something like [this app](https://play.google.com/store/apps/details?id=com.lexa.fakegps&hl=en_US&gl=US) so Collect gets locations that aren't my home when testing.

> the media export folder includes a single audit.csv file

Yes, the files are technically apparently in the zip but must zip clients won't make them accessible.  The combined log is definitely higher priority but at some point it might be nice to do something like put audits in their own folder and add a submissionID prefix to the filenames.

I see a bunch of test failures on CI. Are those expected and also failing locally or a CI special?

ktuite:
Tests have been updated now!

issa-tseng (approved):

PR #344, opened 2021-03-24
Keeping Form name/title consistent with XForm title

ktuite:
This resolves an issue brought up in [getodk/central#29](https://github.com/getodk/central/issues/29#issuecomment-800617039) and described in the [Central block 12 release criteria](https://docs.google.com/document/d/10uUU1nlIFRPJmtTr-zukKDXFY0flFAw_kuvDXZqLgxg).

There are two main changes:

**1. Update form name based on latest XForm title when new draft is uploaded**

Previously, Central's form display name was read once from the initial form's title. Now, the form display name will update if a user uploads a new version of a form with a different title. More specifically, any time a user uploads a new draft, the form name/title will update. 

Changing the name here (vs. on form publication or some other place) seemed like the best option, especially when working with forms only in draft mode. Drafts seem to be the only channel through which an existing form can change.

Theoretically, there could be an open, published form with an old title, and a new draft with a new title, and Central would show the new (draft) title while Collect would show the old title from the old (published) form. Hopefully this isn't that common, but having a title that is too new seems better than having a title that never gets updated.

**2. Remove ability to update form name via PATCH**

Previously, central-backend supported updating the form's name via the PATCH request, but it was not exposed through central-frontend. Since we overwrite form names with any new draft, we have removed the ability to change a form's name through the API. 

This PR doesn't address the renaming "form name" --> "form title" in the early discussion of [getodk/central#29](https://github.com/getodk/central/issues/29). The Form field that refers to the XForm title is still called "name".

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -807,7 +807,6 @@ You can inspect the Request format for this endpoint to see the exact nested dat
         + forms: (array, optional) - If given, the Form metadata to update.
             + (object)
                 + xmlFormId: `simple` (string, required) - The `id` of this form as given in its XForms XML definition.
-                + name: `Simple Form` (string, optional) - The name to display for this form
                 + state: (Form State, required) - The present lifecycle status of this form.
                 + assignments: (array, optional) - If given, the Assignments to apply to this Form. And if given, any existing Assignments that are not specified here will be revoked.
                     + (object)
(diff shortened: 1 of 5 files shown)


matthew-white:
Ah I forgot that Frontend sends the `name` property back to Backend as part of the `PUT` request to `/projects/:id`:

https://github.com/getodk/central-frontend/blob/fa97e326475d3ab4f9021fff0230e400e5e0d2c1/src/components/project/form-access.vue#L145-L152

Not a big deal, because the `name` property will just be ignored once this PR is merged, but would you mind removing that in Frontend whenever you have a chance? 🙏

issa-tseng (approved):

issa-tseng:
💯

Issue #274, opened 2020-07-02
Generate additional audit log entries?


matthew-white:
I'm noticing that when a user's password is set using /users/:id/password, a user.update audit log entry is created. However, /users/reset/initiate and /users/reset/verify do not similarly generate audit log entries. Should those endpoints also generate audit log entries?

matthew-white:
One other thought, how about we have Central generate an audit log entry after backups are configured or terminated?

florianm:
Another idea: audit changes to submitted data (once we can use Enketo to edit existing submissions)

issa-tseng:
> Another idea: audit changes to submitted data (once we can use Enketo to edit existing submissions)

this is not in question.

florianm:
The audit or updating existing submissions on ODK Central?

issa-tseng:
whenever we do submission edit there's no way we don't audit the crap out of it

florianm:
So you mean that submission edit is on the table? (Audit would be awesome but my project's holy grail is submission edit)

matthew-white:
Submission edit is definitely on the table! It's listed under "Features we’ll likely add after June 2020" in a forum post [here](https://forum.getodk.org/t/whats-coming-in-central-over-the-next-few-years/19677).

parthibd:
@matthew-white Hi! Is there any updates on the edit and delete submissions feature ?

matthew-white:
Hi @parthibd! The upcoming release v1.2 will introduce related functionality. See getodk/central#198 for more information about that release.

matthew-white:
About audit log entries, now that #331 is merged, I believe that more actions are being logged, which is great.

/v1/users/reset/verify now seems to generate two audit log entries: a `user.update` entry for the user and a `user.delete` entry for the single-use actor.

For the `user.update` entry, the actor is the single-use actor, which makes sense. The actor display name has the format "Reset token for [user id]". Frontend links that display name to a user page that 404s, so I think I should update Frontend to only render a link for the actor if the actor is a user.

For the `user.delete` entry, my instinct is that it'd be OK not to log an entry when a single-use actor is consumed.

Relatedly, a `user.delete` entry is now logged after an app user or public link is deleted (only possible right now over the API). Would it be possible to have the action that's logged depend on the actor type? For example, logging `field_key.delete` or `public_link.delete` rather than `user.delete`. Either way, I think I should update Frontend to display these new entries nicely.

[...]

PR #348, opened 2021-03-26
block 12 misc bugfixes and cleanup.

issa-tseng:
see commits/tickets.


lib/formats/odata.js
```diff
@@ -226,9 +226,12 @@ const edmxTemplater = template(`<?xml version="1.0" encoding="UTF-8"?>
 const typeMap = {
   int: 'Edm.Int64',
   decimal: 'Edm.Decimal',
+  boolean: 'Edm.Boolean',
   geopoint: 'Edm.GeographyPoint',
   geotrace: 'Edm.GeographyLineString',
   geoshape: 'Edm.GeographyPolygon',
+  date: 'Edm.Date',
+  time: 'Edm.TimeOfDay',
```
matthew-white, line 234: While using Power BI, @lognaturel encountered the following error, which seems related:

> OLE DB or ODBC error: [DataSource.Error] OData: Cannot convert the literal ‘09:04:00.000Z’ to the expected type ‘Edm.TimeOfDay’.

The form used was: https://test.getodk.cloud/#/projects/149/forms/all-widgets/submissions

issa-tseng: oh huh. why is there a tz?

lognaturel: https://www.w3.org/TR/xmlschema-2/#time

issa-tseng: interesting. i didn't know anyone did time w tz. last i understood times w tz are meaningless bc you need the date to actually determine what that time means? either way odata doesn't like it so i guess we'll go back to doing timeofday as string.

lognaturel: Agreed it's not a great idea and I try to steer clear of this data type myself. We've previously discussed introducing a new type that can't have a timezone but it hasn't been a priority. Using string for that one seems great.

issa-tseng:
i don't know why ci isn't working anymore.

lib/model/query/audits.js
```diff
@@ -31,15 +33,15 @@ values (${actorId}, ${action}, ${acteeId}, ${(details == null) ? null : JSON.str
...
   else if (action === 'project')
     return sql`action in ('project.create', 'project.update', 'project.delete')`;
   else if (action === 'form')
     return sql`action in ('form.create', 'form.update', 'form.delete', 'form.attachment.update', 'form.update.draft.set', 'form.update.draft.delete', 'form.update.publish')`;
   else if (action === 'submission')
-    return sql`action in ('submission.create', 'submission.attachment.update')`;
+    return sql`action in ('submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update')`;
 
   return sql`action=${action}`;
 };
```
matthew-white, line 47: Related to #274, could we also add `if`s for `action === 'field_key'` and `action === 'public_link'`?

Also, I'm noticing that the action filter above the audit log will filter if `session.end` (Delete) is selected, but not if `session` (Session Actions) is selected. What do you think about adding `action === 'session'` as well?

https://user-images.githubusercontent.com/5970131/113957871-14171600-97ee-11eb-9434-7a4a2e52d7b0.mov

matthew-white (commented):
Just a couple of random comments, haven't taken a look at things yet.

README.md
```diff
@@ -8,6 +8,8 @@ ODK Central Backend is a minimal [Open Data Kit](https://opendatakit.org/) serve
 
 This repository contains only the code for the backend API server: [Central Frontend](https://github.com/opendatakit/central-frontend) holds frontend code, and [Central](https://github.com/opendatakit/central) contains both the Docker-based production deployment infrastructure for the combined frontend/backend, as well as project management and issue tickets.
 
+> **The `master` branch of this repository reflects ongoing development for the next version of ODK Central.** It may or may not be in sync with the `master` branch of the `central-backend` repository. For the latest stable version, see the [release tags](https://github.com/getodk/central-backend/releases).
```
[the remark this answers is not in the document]
issa-tseng: oh, i sort of liked the `>` there?

lib/model/query/audits.js
matthew-white: I just noticed one other thing, which is that an `assignment.create` action is logged for a public link when the public link is created. That makes sense to me; however, right now Frontend will label the action "User > Assign Role" even though the actee isn't a user.

I think the easiest way to account for this case would be for Frontend to label `assignment.create` differently: something like "Role > Assign" or "Assignment > Create". (`assignment.delete` would become something like "Role > Revoke".) One thing I like about this approach is that it mirrors how Backend organizes these actions, as being about assignments rather than strictly users.

If `assignment.create` and `assignment.delete` aren't just about users, then I think they can be removed from the `action === 'user'` case above. Then we'd have a situation where the actions in each list have the same prefix: filtering on `'project'` filters for a list of actions that all begin with `project.`, filtering on `'user'` filters for a list of actions that all begin with `user.`, and so on. Maybe that would simplify `actionCondition()`?

issa-tseng: i made a new commit. does it look good to you?

matthew-white: Yes, looks great! I think it'd be great to add `field_key.delete` and `public_link.delete` as well, but I'll mention that in a separate comment below.

lib/model/query/audits.js
```diff
@@ -31,15 +33,19 @@ values (${actorId}, ${action}, ${acteeId}, ${(details == null) ? null : JSON.str
...
   if (action === 'nonverbose')
-    return sql`action not in ('submission.create', 'submission.attachment.update', 'backup')`;
+    return sql`action not in ('submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup')`;
   else if (action === 'user')
-    return sql`action in ('user.create', 'user.update', 'user.delete', 'assignment.create', 'assignment.delete')`;
+    return sql`action in ('user.create', 'user.update', 'user.delete', 'user.assignment.create', 'user.assignment.delete')`;
+  else if (action === 'field_key')
+    return sql`action in ('field_key.create', 'field_key.assignment.create', 'field_key.assignment.delete', 'field_key.session.end')`;
+  else if (action === 'public_link')
+    return sql`action in ('public_link.create', 'public_link.assignment.create', 'public_link.assignment.delete', 'public_link.session.end')`;
```
matthew-white, line 42: I think it'd be great to add `field_key.delete` and `public_link.delete` as well, since those are now logged (though they're only possible over the API).

lib/model/query/audits.js
matthew-white: I've tried to summarize the audits-related changes in the release criteria. Let me know if I missed anything! I've also specified new text for Frontend there, so let me know also if anything there should change.

lib/resources/comments.js
```diff
@@ -0,0 +1,40 @@
...
+// except according to the terms contained in the LICENSE file.
+
+const { Comment } = require('../model/frames');
+const { getOrNotFound } = require('../util/promise');
+
+module.exports = (service, endpoint) => {
+  service.get('/projects/:projectId/forms/:formId/submissions/:id/comments', endpoint(({ Comments, Forms, Submissions }, { auth, params, queryOptions }) =>
+    Submissions.getByIds(params.projectId, params.formId, params.id, false)
+      .then(getOrNotFound)
+      .then((submission) => Promise.all([
```
matthew-white, line 17: Could `Submissions.getByIds()` go into the `Promise.all()`?

lib/model/query/sessions.js
```diff
@@ -42,7 +42,9 @@ const terminate = (session) => ({ run }) =>
 terminate.audit = (session) => (log) => {
   // don't audit user logouts, since they're just normal user actions.
   if ((session.actor != null) && (session.actor.type === 'user')) return;
-  return log('session.end', session.actor);
+  const prefix = ((session.actor != null) && (session.actor.type != null))
+    ? `${session.actor.type}.` : '';
+  return log(`${prefix}session.end`, session.actor);
```
matthew-white, line 47: `session.actorId` is not nullable in the database, so maybe it's possible to remove the null checks for `session.actor`?

lib/data/odata.js
```diff
@@ -134,7 +134,9 @@ const submissionToOData = (fields, table, submission, options = {}) => new Promi
         attachmentsPresent: attachment.present || 0,
         attachmentsExpected: attachment.expected || 0,
         status: encrypted ? (encHasData ? 'notDecrypted' : 'missingEncryptedFormData') : null,
-        reviewState: submission.reviewState || null
+        reviewState: submission.reviewState || null,
+        deviceId: submission.deviceId || null,
+        edits: submission.aux.edit.count
```
matthew-white, line 139: Can this be 0, or will it be `null`? I thought it would be `null` and made a related commit in getodk/central-frontend#430, but I think I might have misunderstood the comment about the Edits column of the CSV.

lib/model/query/submissions.js
```diff
@@ -190,6 +193,10 @@ ${encrypted ? sql`
 left outer join (select id, "keyId" as "encKeyId" from form_defs) as form_defs on form_defs.id=submission_defs."formDefId"
 left outer join (select id, content as "encData" from blobs) as blobs on blobs.id=submission_attachments."blobId"`
     : sql`join (select null as "encKeyId", null as "encData") as enc on true`}
+left outer join
```
matthew-white, line 196: Could an inner join be used? I think there will always be at least one submission def?

matthew-white (commented):
Just making note of some questions to ask during interactive review.

test/unit/formats/odata.js
```diff
@@ -129,8 +129,37 @@ describe('odata message composition', () => {
       }));
 
     it('should express repeats as entity types behind navigation properties', () =>
```
issa-tseng, line 137: title??

test/integration/api/public-links.js
```diff
@@ -176,6 +176,15 @@ describe('api: /projects/:id/forms/:id/public-links', () => {
           .expect(200)
           .then(({ body }) => asAlice.delete(`/v1/projects/1/forms/simple/public-links/${body.id}`)
             .expect(404)))));
+
+    it('should delete the token', testService((service) =>
```
issa-tseng: title

matthew-white (approved):

PR #351, opened 2021-04-01
bug: decode/encode URLs for OData .../Submissions(':id') endpoints.

matthew-white:
The base branch of this PR is `issa/1.2-misc`. However, the parent commit is 79f7863835eb1cb60ada0691c727b0dafe8a8bba on `master`, so this PR could also be merged into `master`.

The goal of this PR is to update the OData .../Submissions(':instanceId') endpoint so that the `uuid:` prefix of an instance ID may be encoded. The encoding will be preserved in the OData URLs that the endpoint returns: both the form ID and the instance ID will remain encoded. Not all instance IDs have a `uuid:` prefix, but some do. This PR intersects with getodk/central-frontend#429, which adds a submission detail page that uses this endpoint.

With these changes, I think the endpoint may also be pretty close to being able to accept an instance ID that is not a UUID.

CircleCI is failing, but I think it's also failing on `master`. I was able to successfully run test/integration/api/odata.js locally.

See the commit message for additional notes. I'll also add some notes here about specific lines.

### DIFF
--- a/lib/resources/odata.js
+++ b/lib/resources/odata.js
@@ -39,7 +39,7 @@ module.exports = (service, endpoint) => {
 
     // serves filtered single-row data.
     //const singleRowRegex = /^\/projects\/(\d+)\/forms\/([a-z0-9-_]+).svc\/Submissions\((?:'|%27)((?:uuid:)?[a-z0-9-]+)(?:'|%27)\)(\/.*)*$/i;
-    const uuidRegex = /^(?:'|%27)((?:uuid:)?[a-z0-9-]+)(?:'|%27)$/i;
+    const uuidRegex = /^'((?:uuid:)?[a-z0-9-]+)'$/i;
     const getUuid = (str) => {
       const matches = uuidRegex.exec(str);
       if (matches == null) throw Problem.user.notFound();
@@ -51,7 +51,7 @@ module.exports = (service, endpoint) => {
           Forms.getFields(form.def.id),
           Submissions.getForExport(form.id, getUuid(params.uuid), draft).then(getOrNotFound)
         ])
-          .then(([ fields, row ]) => singleRowToOData(fields, row, env.domain, decodeURI(originalUrl), query))));
+          .then(([ fields, row ]) => singleRowToOData(fields, row, env.domain, originalUrl, query))));
 
     // TODO: because of the way express compiles the *, we have to register this twice.
     service.get(`${base}/Submissions\\((:uuid)\\)`, singleRecord);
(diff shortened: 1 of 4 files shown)


lib/resources/odata.js
```diff
@@ -39,7 +39,7 @@ module.exports = (service, endpoint) => {
 
     // serves filtered single-row data.
     //const singleRowRegex = /^\/projects\/(\d+)\/forms\/([a-z0-9-_]+).svc\/Submissions\((?:'|%27)((?:uuid:)?[a-z0-9-]+)(?:'|%27)\)(\/.*)*$/i;
-    const uuidRegex = /^(?:'|%27)((?:uuid:)?[a-z0-9-]+)(?:'|%27)$/i;
+    const uuidRegex = /^'((?:uuid:)?[a-z0-9-]+)'$/i;
```
matthew-white, line 42: When I first started working on this PR, I thought I had to add `%3A` to this regular expression for `:`, but the change needed turned out to be elsewhere. After taking a look at this endpoint, I'm pretty sure that we don't need `%27` in the regular expression. `uuidRegex` is only used in `getUuid()`, which is passed `params.uuid`, but Express already ensures that `params.uuid` is decoded. My guess is that this regular expression used to be part of the route itself, not used for the param, and the single quote needed to be encoded there.

test/integration/api/odata.js
```diff
@@ -242,7 +258,7 @@ describe('api: /forms/:id.svc', () => {
           .then(({ body }) => {
             body.should.eql({
               '@odata.context': 'http://localhost:8989/v1/projects/1/forms/doubleRepeat.svc/$metadata#Submissions.children.child',
-              '@odata.nextLink': "http://localhost:8989/v1/projects/1/forms/doubleRepeat.svc/Submissions('double')/children/child?%24skip=2",
+              '@odata.nextLink': "http://localhost:8989/v1/projects/1/forms/doubleRepeat.svc/Submissions(%27double%27)/children/child?%24skip=2",
```
matthew-white, line 261: I encode the instance ID above, but I don't encode `'`, so it's very interesting to me that `'` becomes `%27` here. I think what's happening is that `%27` was part of the `originalUrl`, because supertest is encoding the `'` (though I don't think it needs to).

matthew-white:
@issa-tseng, this should be ready for you to take a look!

lib/resources/odata.js
issa-tseng: yes, that is the exact history there.

test/integration/api/odata.js
issa-tseng: yeah this is probably what led me to do the unencoding at all.

issa-tseng (approved):

Issue #355, opened 2021-04-07
When there is a published form version, don't update form name from new drafts


lognaturel:
https://github.com/getodk/central-backend/pull/344 updates the form name based on new form definitions. It does this update every time a new draft is created. This becomes confusing if the title change is anything other than a typo fix and especially confusing if a draft with a title change is created and then abandoned.

The expected behavior is that the title of the most recent draft is used until there's a published version at which point the title of the most recent published version is always used whether or not there's an active draft.

PR #357, opened 2021-04-16
Fix for issue #355: Don't update form name from drafts if published version exists

ktuite:
This PR addresses issue #355 and updating form titles at the right times in the work flow.

Expected behavior, for which there are now tests: 
* If no published version, update form title with draft title
    * needs a conditional check
* If published version, update title on next publish
    * needs to get title from somewhere (reading xml stream of from form def)
        * How to add something to form def?

This PR has gone through a couple phases:

**Phase 1: Parse form name from FormDef XML on publish**

First, I set up some checks in the draft upload part of the code so the new title will only get applied from a draft to the overall form if there is no published versions.

For transferring the title from the form from a draft on publish, I first tried an approach of reading the title directly from the form def xml. It had to wait for that xml to get loaded and parsed, so I don't know if it's a great approach (but also it's ok if publish is a little slow). It also had some clunky async/await code that seemed a little fragile, so I moved on to the next phase.

**Phase 2: Add name to FormDef** 

Since each form XML definition has a title (or null title) that can change with each version, it makes a lot of sense to store that name here. (In the XML, the form name is called *title*, but everywhere in the code, including the database column name, this field is called *name* for consistency.)

I made a migration that adds the name column and backfills it with title information read from each FormDef xml. 

I then use that FormDef.name to update the Form.name at the right places in the code.

**Phase 3: Remove Form.name and just use the appropriate FormDef.name (not reflected in this PR yet)**

Copying over the form name is confusing and redundant and in an effort to normalize the data, I'm working on removing Form.name and using the appropriate FormDef.name, since the form_def table is joined in any query about forms already.

### DIFF
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -112,7 +112,7 @@ class Form extends Frame.define(
       const version = versionText.orElse('');
       const name = nameText.orNull();
       const key = pubKey.map((k) => new Key({ public: k }));
-      return new Form.Partial({ xmlFormId, name }, { def: new Form.Def({ version, name: name }), key });
+      return new Form.Partial({ xmlFormId, name }, { def: new Form.Def({ version, title: name }), key });
     });
 
     return (typeof xml === 'string')
@@ -165,7 +165,7 @@ Form.Def = Frame.define(
   'sha256',      readable,              'draftToken',   readable,
   'enketoId',    readable,              'createdAt',
   'publishedAt', readable,              'xlsBlobId',
-  'name',        readable
+  'title',        readable
 );
 Form.Xml = Frame.define(into('xml'), 'xml');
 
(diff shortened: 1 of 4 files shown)


ktuite:
Current response for `/v1/projects/<proj_id>/forms/<form_id>` when the form only exists as a draft looks like this: 

**name** is filled in (from Form.name) but **title** is null (from FormDef.title but all form def information is withheld because the form is only a draft)

      {
          "createdAt": "2021-04-16T00:03:03.593Z",
          "draftToken": null,
          "enketoId": null,
          "enketoOnceId": null,
          "hash": null,
          "keyId": null,
          "name": "Titled Form",
          "projectId": 3,
          "publishedAt": null,
          "sha": null,
          "sha256": null,
          "state": "open",
          "title": null,
          "updatedAt": "2021-04-21T18:17:32.709Z",
          "version": null,
          "xmlFormId": "titled-form"
      }

ktuite:
Closing this and going with PR #359 to fully remove `name` from Form and use name stored on FormDef instead.

PR #359, opened 2021-04-23
Update form name by storing name on FormDef instead of Form

### LINKED ISSUE #355: When there is a published form version, don't update form name from new drafts

ktuite:
Closes #355 by moving a form's `name` (or <title> as defined in the form XML) from Form to FormDef. Form will always reference the appropriate FormDef and get the name from there.

This PR includes a migration that adds `name` to the `form_defs` table and backfills it with form title information read from form XML. It also includes a migration to drop the `name` field from the `forms` table. (Er, I was nervous about dropping that field so I just changed the column name but will change the migration to drop the column before I finalize this PR.)

One thing that changes is that a little bit more information about an unpublished form comes through the API (like its `version` and `sha`) when that information used to be `null`. It is still possible to determine if a form is unpublished by looking at its `publishedAt` field, though. We should probably add a note about this to the docs.

Endpoints that only show published forms (like the OpenRosa `formList`) are unaffected and work the same as before.

### DIFF
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -51,10 +51,10 @@ class Form extends Frame.define(
   table('forms'),
   'id',                                 'projectId',    readable,
   'xmlFormId',    readable,             'state',        readable, writable,
-  'currentDefId',                       'draftDefId',
-  'enketoOnceId', readable,             'enketoId',     readable,
+  'currentDefId',
+  'draftDefId',                         'enketoId',     readable,
+  'enketoOnceId', readable,             'acteeId',
   'createdAt',    readable,             'updatedAt',    readable,
-  'acteeId',
   embedded('createdBy'), embedded('publishedBy')
 ) {
   get def() { return this.aux.def; }
@@ -149,7 +149,6 @@ Form.Extended = class extends Frame.define(
       submissions: this.submissions || 0,
       lastSubmission: this.lastSubmission,
       excelContentType: this.excelContentType,
-      name: this.name,
     };
   }
 };
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -113,7 +113,7 @@ class Form extends Frame.define(
       const version = versionText.orElse('');
       const name = nameText.orNull();
       const key = pubKey.map((k) => new Key({ public: k }));
-      return new Form.Partial({ xmlFormId, name }, { def: new Form.Def({ version, name: name }), key });
+      return new Form.Partial({ xmlFormId, name }, { def: new Form.Def({ version, name }), key });
     });
 
     return (typeof xml === 'string')
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -40,6 +40,7 @@ const Problem = require('../../util/problem');
 
 // sentinel values used below (oh how i long for case classes..)
 const DraftVersion = Symbol('draft version');
+const PublishedVersion = Symbol('published version');
 const AllVersions = Symbol('all versions');
 
 
@@ -50,10 +51,10 @@ class Form extends Frame.define(
   table('forms'),
   'id',                                 'projectId',    readable,
   'xmlFormId',    readable,             'state',        readable, writable,
-  'name',         readable,             'currentDefId',
-  'draftDefId',                         'enketoId',     readable,
-  'enketoOnceId', readable,             'acteeId',
+  'currentDefId',                       'draftDefId',
+  'enketoOnceId', readable,             'enketoId',     readable,
   'createdAt',    readable,             'updatedAt',    readable,
+  'acteeId',
   embedded('createdBy'), embedded('publishedBy')
 ) {
   get def() { return this.aux.def; }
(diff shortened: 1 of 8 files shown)


lib/model/migrations/20210423-02-drop-form-name.js
```diff
@@ -0,0 +1,22 @@
...
+// except according to the terms contained in the LICENSE file.
+
+const { Form } = require('../frames');
+
+const up = async (db) => {
+  await db.raw('ALTER TABLE forms RENAME COLUMN name to nonactive_name');
+};
+
+const down = async (db) => {
+  await db.raw('ALTER TABLE forms RENAME COLUMN nonactive_name to name');
```
ktuite, line 17: I need to change this migration to just drop the column

ktuite:
The frontend COULD show the historical form names, now, and not just the version...

Screen Shot 2021-04-23 at 12 31 07 PM

lib/model/frames/form.js
```diff
@@ -112,7 +113,7 @@ class Form extends Frame.define(
       const version = versionText.orElse('');
       const name = nameText.orNull();
       const key = pubKey.map((k) => new Key({ public: k }));
-      return new Form.Partial({ xmlFormId, name }, { def: new Form.Def({ version }), key });
+      return new Form.Partial({ xmlFormId, name }, { def: new Form.Def({ version, name: name }), key });
```
issa-tseng, line 116: probably if you `make lint` you'll get a complaint about this line.

lib/model/frames/form.js
```diff
@@ -50,10 +51,10 @@ class Form extends Frame.define(
   table('forms'),
   'id',                                 'projectId',    readable,
   'xmlFormId',    readable,             'state',        readable, writable,
-  'name',         readable,             'currentDefId',
```
issa-tseng, line 53: i'd probably rather just name get taken out here and the rest left alone to preserve the line history.

lib/model/query/forms.js
```diff
@@ -241,8 +242,9 @@ order by coalesce(forms.name, forms."xmlFormId") asc`)
 /* eslint-disable indent */
 const versionJoinCondition = (version) => (
   (version === '___') ? versionJoinCondition('') :
-  (version == null) ? sql`form_defs.id=forms."currentDefId"` :
+  (version == null) ? sql`form_defs.id=coalesce(forms."currentDefId", forms."draftDefId")` :
```
issa-tseng, line 245: ah brilliant

lib/model/frames/form.js
```diff
@@ -140,13 +142,14 @@ Form.Partial = class extends Form {};
...
-  'excelContentType',   readable
+  'excelContentType',   readable,       'name',           readable
 ) {
   forApi() {
     return {
       submissions: this.submissions || 0,
       lastSubmission: this.lastSubmission,
-      excelContentType: this.excelContentType
+      excelContentType: this.excelContentType,
+      name: this.name,
```
issa-tseng, line 152: why is this necessary? does it not come through the def in some cases?

lib/model/query/forms.js
```diff
@@ -106,9 +106,10 @@ const createVersion = (partial, form, publish = false) => ({ run, one, FormAttac
         ? def.with({ publishedAt: new Date(), xml: partial.xml })
         : def.with({ draftToken: _getDraftToken(form), xml: partial.xml })))
       .then(compose(one, insert))
-      .then(ignoringResult((savedDef) => ((publish === true)
-        ? Forms._update(form, { currentDefId: savedDef.id, name: partial.name })
-        : Forms._update(form, { draftDefId: savedDef.id, name: partial.name })))),
+      .then(ignoringResult((savedDef) =>
+        ((publish === true)
+          ? Forms._update(form, { currentDefId: savedDef.id })
+          : Forms._update(form, { draftDefId: savedDef.id })))),
```
issa-tseng, line 112: love this diff

issa-tseng: well, i'm not sure why the whitespace change but i love the rm bits

issa-tseng (changes requested):
just a couple questions then i'll do a squash merge

lib/model/frames/form.js
ktuite: done!

lib/model/query/forms.js
ktuite: oops, all fixed up now

lib/model/frames/form.js
ktuite: you're right, this isn't necessary... something i tried along the way, i think

issa-tseng:
merged via https://github.com/getodk/central-backend/commit/ee7ba91488e6b3d83e2d0d69eed09e6143902e90

PR #360, opened 2021-04-23
docs: update api documentation for v1.2.

issa-tseng:


docs/api.md
```diff
@@ -1848,6 +1881,157 @@ This endpoint supports retrieving extended metadata; provide a header `X-Extende
 + Response 403 (application/json)
     + Attributes (Error 403)
 
+### Getting Submission metadata [GET /v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}]
+
+Like how `Form`s are addressed by their XML `formId`, individual `Submission`s are addressed in the URL by their `instanceId`.
```
lognaturel, line 1886: It also feels like it would be helpful to call out here that the instance id is the id of the original submission and that only the latest version is returned. Maybe the section introduction could have that but it seems important enough to me to repeat in each section so they truly stand alone.

docs/api.md
```diff
@@ -1848,6 +1881,157 @@ This endpoint supports retrieving extended metadata; provide a header `X-Extende
...
++ Response 400 (application/json)
+    + Attributes (Error 400)
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
++ Response 409 (application/json)
+    + Attributes (Error 409)
+
+### Updating Submission Data [PUT /v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}]
```
lognaturel, line 1971: Maybe link to https://github.com/getodk/central-backend/pull/360/files#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620abR2329 somewhere in here and/or in the submission access sections? Or do you prefer to limit cross references?

docs/api.md
```diff
@@ -2950,6 +3263,29 @@ As the vast majority of clients only support the JSON OData format, that is the
 + Response 501 (application/json)
     + Attributes (Error 501)
 
+### Data Download Path [GET /dl/projects{projectId}/forms/{xmlFormId}/submissions/{instanceId}/attachments/{filename}]
+
+_(introduced: version 1.2)_
+
+This route is a web browser oriented endpoint intended for user-interactive usage only. It's not part of the Central API, but is documented here as it can be useful.
```
lognaturel, line 3270: 👍

lognaturel:
> maybe we should really be fixing this by fixing it (#369)

This has caused me some pain as well though I don't have a great fix in mind. Agreed lowering the threshold might solve most issues.

matthew-white (commented):
I haven't taken a close look — just wanted to leave a couple of random comments!

docs/api.md
```diff
@@ -3253,6 +3254,29 @@ As the vast majority of clients only support the JSON OData format, that is the
...
+
+If you are writing or configuring an OData client and have submission media files to deal with, you can run into authentication problems directly fetching or linking the media file URLs that are provided in the OData feed. This can be due to several reasons: if the user is not logged into the Central administration site (and thus has no valid cookie), if the request comes from a foreign origin (and thus cookies are not sent by the browser), and more.
+
+To help manage this, the frontend provides a `/dl` path that allows file download. Just take a normal attachment download path and replace the `/v1` near the beginning of the path with `/dl`, and the user will be taken to a page managed by the Central administration website that will ensure the user is logged in, and offer the file as a download.
```
matthew-white, line 3274: The path prefix is actually the very beautiful `/#/dl`.

matthew-white (commented):
Just took a look at 816ed856762a522f654c57d391e62a7cf8bff0fd about `/dl`. That looks great — just had one small comment.

docs/api.md
```diff
@@ -1043,9 +1070,9 @@ This endpoint supports retrieving extended metadata; provide a header `X-Extende
...
 
-For XLSForm upload, either `.xls` or `.xlsx` are accepted. You must provide the `Content-Type` request header corresponding to the file type: `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` for `.xlsx` files, and `application/vnd.ms-excel` for `.xls` files. You must also provide an `X-XlsForm-FormId-Fallback` request header with the `formId` you want the resulting form to have, if the spreadsheet does not already specify.
+For XLSForm upload, either `.xls` or `.xlsx` are accepted. You must provide the `Content-Type` request header corresponding to the file type: `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` for `.xlsx` files, and `application/vnd.ms-excel` for `.xls` files. You must also provide an `X-XlsForm-FormId-Fallback` request header with the `formId` you want the resulting form to have, if the spreadsheet does not already specify. This header field accepts percent-encoded values.
```
ktuite, line 1075: What's an example of a percent-encoded value you'd want to put in a form id fallback request header here?

docs/api.md
```diff
@@ -1819,6 +1846,12 @@ You can fully delete a link by issuing `DELETE` to its resource. This will remov
 + Response 403 (application/json)
     + Attributes (Error 403)
 
+# Group Submissions
```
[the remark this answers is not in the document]
ktuite: I think this is just for Apiary

docs/api.md
```diff
@@ -1819,6 +1846,12 @@ You can fully delete a link by issuing `DELETE` to its resource. This will remov
...
+
+`Submission`s are filled-out forms (also called `Instance`s in some other ODK documentation). Each is associated with a particular Form (and in many cases with a particular _version_ of a Form), and is also created out of a standard XML format based on the Form itself. Submissions can be sent with many accompanying multimedia attachments, such as photos taken in the course of the survey. Once created, the Submissions themselves as well as their attachments can be retrieved through this API.
+
+These subsections cover only the modern RESTful API resources involving Submissions. For documentation on the OpenRosa submission endpoint (which can be used to submit Submissions), or the OData endpoint (which can be used to retrieve and query submission data), see those sections below.
+
 ## Submissions [/v1/projects/{projectId}/forms/{xmlFormId}/submissions]
 
 `Submission`s are available as a subresource under `Form`s. So, for instance, `/v1/projects/1/forms/myForm/submissions` refers only to the Submissions that have been submitted to the Form `myForm`.
```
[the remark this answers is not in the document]
ktuite: I think this should be explicit.

docs/api.md
lognaturel: 😭 I was pleased reading the docs to learn that just `/dl` was supported. Oops!

docs/api.md
```diff
@@ -1848,6 +1881,157 @@ This endpoint supports retrieving extended metadata; provide a header `X-Extende
...
+
+### Updating Submission Data [PUT /v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}]
+
+_(introduced: version 1.2)_
+
+You can use this endpoint to submit _updates_ to an existing submission.
+
+The `instanceId` that is submitted with the initial version of the submission is used permanently to reference that submission logically, which is to say the initial submission and all its subsequent versions. Each subsequent version will also provide its own `instanceId`. This `instanceId` becomes that particular version's identifier.
+
+To perform an update, you need to provide in the submission XML an additional [`deprecatedID` metadata node](https://getodk.github.io/xforms-spec/#metadata) with the `instanceID` of the particular submission version you are replacing. If the `deprecatedID` you give is anything other than the identifier of the current version of the submission at the time the server receives it, you will get a `409 Conflict` back.
```
ktuite, line 1979: I'm confused about this. Does "current version of the submission" mean the initial instance id? I don't think so, but if not, how do you *get* the latest ID?

Like, if there are 3 updates to a submission:
`A`
`B`
`C`

`A` will be how you refer to that submission even when it updates... but do you have to set the deprecated ID to `B` to make the `C` update?

docs/api.md
```diff
@@ -1848,6 +1881,157 @@ This endpoint supports retrieving extended metadata; provide a header `X-Extende
...
++ Response 409 (application/json)
+    + Attributes (Error 409)
+
+### Getting an Enketo Edit URL [GET /v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}/edit]
+
+_(introduced: version 1.2)_
+
+This endpoint redirects the user to an Enketo-powered page that allows the user to interactively edit th e submission. Once the user is satisfied, they can perform the update submission directly through the Enketo interface.
+
+The Enketo instance is already hosted inside of ODK Central. There is no reason to create or use a separate Enketo installation.
```
ktuite, line 2019: (For production setups / the full Central setup...)

> There is no reason to create or use a separate Enketo installation.

You might want to set up a separate enketo installation for dev work.

docs/api.md
ktuite: I see there's more info about this in the openrosa part of the API... would you want to link from here to there?

For my above comment, I think a clarification like "current version - possibly not the initial instance ID if already deprecated" would help.

docs/api.md
```diff
@@ -3492,8 +3816,10 @@ These are in alphabetic order, with the exception that the `Extended` versions o
 
 ## Submission (object)
 + instanceId: `uuid:85cb9aff-005e-4edd-9739-dc9c1a829c44` (string, required) - The `instanceId` of the `Submission`, given by the Submission XML.
-+ submitterId: `23` (number, required) - The ID of the `Actor` (`App User` or `User`) that submitted this `Submission`.
++ instanceName: `village third house` (string, optional) - The `instanceName`, if any, given by the Submission XML in the metadata section.
++ submitterId: `23` (number, required) - The ID of the `Actor` (`App User`, `User`, or `Public Link`) that submitted this `Submission`.
 + deviceId: `imei:123456` (string, optional) - The self-identified `deviceId` of the device that collected the data, sent by it upon submission to the server.
++ reviewState: `received` (Submission Review State, optional) - The current review state of the submission.
```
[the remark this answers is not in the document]
ktuite: i agree - it's never going to be `received`

ktuite (commented):
Lots of good stuff in the docs here!

matthew-white (commented):
I'm currently taking a look at changes to the form response and whether any small changes are needed for Frontend. I thought I'd leave a couple of comments here related to the form response.

docs/api.md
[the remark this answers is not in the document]
issa-tseng: this route doesn't return anything that really changes meaningfully across versions. the xml itself isn't in this response. the result didn't change when we added versions so i didn't touch it.

docs/api.md
lognaturel: Maybe it would be helpful to explicitly describe requesting the latest instance XML (which may or may not have the same instanceId as the API uses to reference the submission) and then as part of that edit changing the instanceId from the XML to be the deprecatedId and populating a new instanceId. That is, whatever client performs the edit will use the latest instance XML so it doesn't care what id Central uses the reference the submission by.

docs/api.md
issa-tseng: i try to keep these sections short, because of the apiary layout. the main bar is really meant to be more of an index. a lot of content and the reading gets difficult and cramped. the parameter descriptions themselves (i think..) always describe exactly what is happening in each case. but i'll add it to the section introduction as well.

docs/api.md
```diff
@@ -3194,7 +3194,11 @@ The data documents are the straightforward JSON representation of each table of
...
+
+Note that the `submissionDate` has a time component. This means that any comparisons you make need to account for the full time of the submission. It might seem like `$filter=__system/submissionDate le 2020-01-31` would return all results on or before 31 Jan 2020, but in fact only submissions made before midnight of that day would be accepted. To include all of the month of January, you need to filter by either `$filter=__system/submissionDate le 2020-01-31T23:59:59.999` or `$filter=__system/submissionDate lt 2020-02-01`. Remember also that you can [query by a specific timezone](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC).
```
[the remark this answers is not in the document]
issa-tseng: Z

docs/api.md
issa-tseng: mmmmmm usually this is to support unicode values which i dunno will example super good.. but i'll mention the unicode.

docs/api.md
lognaturel: I /think/ the primary confusion is "how do you get the latest ID?". I believe the answer is that you don't explicitly, you just request the latest XML and do your edits including the id switching from there.

issa-tseng: yeah, i understood that part.

ktuite (approved):

PR #361, opened 2021-04-24
improve: return xml odata endpoint errors in xml.

issa-tseng:
* so far as i can tell this should only ever be 403/404 but it should
  work okay no matter what errors.

### DIFF
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -103,6 +103,12 @@ const template = (body) => {
 ////////////////////////////////////////////////////////////////////////////////
 // MESSAGES
 
+const odataXmlError = template(`<?xml version="1.0" encoding="UTF-8"?>
+<error code="{{problemCode}}">
+  <message>{{message}}</message>
+  <details>{{problemDetails}}</details>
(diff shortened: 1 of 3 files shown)


lib/formats/odata.js
```diff
@@ -103,6 +103,12 @@ const template = (body) => {
 ////////////////////////////////////////////////////////////////////////////////
 // MESSAGES
 
+const odataXmlError = template(`<?xml version="1.0" encoding="UTF-8"?>
+<error code="{{problemCode}}">
+  <message>{{message}}</message>
+  <details>{{problemDetails}}</details>
```
matthew-white, line 109: Will `{{problemDetails}}` become `[object Object]`, or is it automatically JSON-stringified?

matthew-white (approved):

lib/formats/odata.js
issa-tseng: uh yeah they'll probably end up object object. maybe i'll just take the details out for now, i don't know of any errors this could trip that would have meaningful details.

PR #362, opened 2021-04-24
deps: update dependencies for security.

issa-tseng:


issa-tseng:
oh i think i need to take the knex upgrade out of this.

matthew-white (approved):

PR #366, opened 2021-04-27
Add issue template config

matthew-white:
Using the same config as the `central` repository. See getodk/central#216.


issa-tseng (approved):
hate adding  another root folder/file but i guess what choice is there

matthew-white:
Just pushed the latest `config.yml` from `central` (see getodk/central#218).

PR #375, opened 2021-05-12
Implement additional filtering for Sentry

### LINKED ISSUE #347: Implement additional filtering for Sentry
We currently filter many Sentry events. This issue is to implement additional filtering. Here are some ideas for that filtering:

1. Filter out the request body for additional endpoints
    - /sessions (POST)
    - /users (POST)
    - /users/:id (PATCH)
    - /users/reset/initiate
    - /projects/:id/key
    - /config/backups/initiate
    - /backup (at least the POST version)
    - Any others?
2. Filter the request for the submission download endpoints. Filtering out the request body would work for the GET endpoints, but the POST endpoints use query parameters, not the request body.
3. Filter out the `q` query parameter from `/users`
4. Filter out tokens from URLs (@lognaturel mentioned that this would be a good idea)
    - App users
    - Form drafts
    - Public links (the `st` query parameter)
5. Filter out the CSRF token from the request body
6. Filter out the `Authorization` header. I think Sentry filters it out server-side, but recommends not sending it in the first place. I think it would be easy to remove?

ktuite:
Closes #347. This issue requested filtering out the request body for a number of endpoints. These are endpoints that might contain emails and other sensitive information. (Note that extra sensitive information, such as passwords, was already filtered by Sentry, but with this change, that information is erased before it is even sent to Sentry.) The issue also requested filtering out the query parameters of certain endpoints.

* /sessions
* /users
* /users/:id (covered by endpoint regex `/users`)
* /users/reset/initiate
* /projects/:id/key
* /config/backups/initiate
* /backup (at least the POST version)
* any more...?

Changes:
1. I've expanded the list of sensitive endpoints, and any matching endpoint, regardless of what kind of GET/POST/PUT request it is, will have both it's request body filtered out, and have all query parameters removed from the request and the URL that is sent to Sentry.

2. Cookies, including CSRF tokens, are more thoroughly removed. Previously, cookies were summarized in the request header, but getting passed through in the request object and still visible in Sentry.

3. The `Authorization` header is removed when it is present.

4. Authentication tokens that are part of a URL, such as those used with [App User Authentication](https://odkcentral.docs.apiary.io/#reference/authentication/app-user-authentication/using-app-user-authentication) and [Draft Testing](https://odkcentral.docs.apiary.io/#reference/openrosa-endpoints/draft-testing-endpoints/openrosa-form-submission-api) are removed.

5. [Help?] I'm not sure what the `st` query parameter of a public link refers to. I see one that is part of an Enketo URL, which is outside of `central-backend`'s + Sentry's realm.

How to test this is going to be interesting! I have been trying it out with a dev sentry account, breaking the code, making various API requests, and checking that the information is getting removed... Not sure if there's a more repeatable way to test.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -1,7 +1,93 @@
 const { inspect } = require('util');
 const { isBlank } = require('../util/util');
 
-const sensitiveEndpoints = ['/users/reset/verify', '/users/(.*?)/password'];
+
+// Endpoints where query strings are sensitive and should be removed
+const sensitiveEndpoints = [
+  ['GET', '/users[^/]'], // ?q=<part of name or email>
+  ['GET', '/projects/([^/]*)/forms/([^/]*)/submissions.csv'], // ?key=passphrase for managed encryption
+  ['GET', '/projects/([^/]*)/forms/([^/]*)/draft/submissions.csv'], // ?key=passphrase for managed encryption
+  ['GET', '/projects/([^/]*)/formList'], // ?st=token for enketo public link access (showing form)
+  ['HEAD', '/projects/([^/]*)/formList'], // ?st=token for enketo public link access (showing form)
+  ['POST', '/projects/([^/]*)/submission'], // ?st=token for enketo public link access (submitting)
+];
+
+const isSensitiveEndpoint = (request) => {
+  for (const [method, endpoint] of sensitiveEndpoints) {
+    if (request.method === method && request.url.match(endpoint)) {
+      return true;
+    }
+  }
+  return false;
+};
+
+const filterTokenFromUrl = (url) => {
+  // remove user token from any URL
+  if (url.match('/v1/key/(.*?)')) {
+    return url.replace(/v1\/key\/([^/]*)/, 'v1/key/[FILTERED]');
+  }
+
+  // remove draft token from any URL
+  if (url.match('/v1/test/(.*?)')) {
+    return url.replace(/v1\/test\/([^/]*)/, 'v1/test/[FILTERED]');
+  }
+
+  // return original URL if there are no tokens
+  return url;
+};
+
+const sanitizeEventRequest = (event) => {
+  /* eslint-disable no-param-reassign */
+  if (event.request) {
+
+    // clear out all cookie values
+    if (event.request.headers && event.request.headers.cookie) {
+      const cookies = event.request.headers.cookie.split('; ')
+        .map((cookie) => cookie.slice(0, cookie.indexOf('=')))
+        .join(', ');
+
+      event.request.headers.cookie = cookies;
+    }
+
+    // clear out cookie values not stored in request header
+    if (event.request.cookies)
+      event.request.cookies = null;
+
+    // clear out Authorization header
+    if (event.request.headers && event.request.headers.authorization)
+      event.request.headers.authorization = null;
+
+    // clear out referer header (from enketo submission, can have ?st=token in url)
+    if (event.request.headers && event.request.headers.referer)
+      event.request.headers.referer = null;
+
+    // clear out x-action-notes
+    if (event.request.headers && event.request.headers['x-action-notes'])
+      event.request.headers['x-action-notes'] = null;
+
+    // clear out request body data for all endpoints
+    if (event.request.data)
+      event.request.data = null;
+
+    // remove sensitive info from URL via query string or path
+    if (event.request.url) {
+      // handle endpoints that might expose sensitive data through query params
+      if (isSensitiveEndpoint(event.request) && event.request.query_string) {
+        // remove query string from request data and from URL
+        event.request.query_string = null;
+        // remove it from URL, too, otherwise Sentry will find it
+        event.request.url = event.request.url.split('?')[0]; // eslint-disable-line prefer-destructuring
+      }
+
+      // filter out access tokens embedded in URLs
+      event.request.url = filterTokenFromUrl(event.request.url);
+    }
+  }
+
+  /* eslint-enable no-param-reassign */
+
+  return event;
+};
 
 const init = (config) => {
   if ((config == null) || isBlank(config.key) || isBlank(config.project)) {
@@ -24,35 +110,17 @@ const init = (config) => {
   Sentry.init({
     dsn: `https://${config.key}@sentry.io/${config.project}`,
     beforeSend(event) {
-      if (event.request) {
-        // clear out all cookie values
-        if (event.request.headers && event.request.headers.cookie) {
-          const cookies = event.request.headers.cookie.split('; ')
-            .map((cookie) => cookie.slice(0, cookie.indexOf('=')))
-            .join(', ');
-
-          event.request.headers.cookie = cookies; // eslint-disable-line no-param-reassign
-        }
-
-        // clear out all request body data for sensitive endpoints (logging in, etc)
-        if (event.request.url) {
-          for (const endpoint of sensitiveEndpoints) {
-            if (event.request.url.match(endpoint)) {
-              event.request.data = null; // eslint-disable-line no-param-reassign
-            }
-          }
-        }
-      }
+      const sanitizedEvent = sanitizeEventRequest(event);
 
       // only file the event if it is a bare exception or it is a true 500.x Problem.
-      const error = event.extra.Error;
-      if (error == null) return event; // we aren't sure why there isn't an exception; pass through.
-      if ((error.isProblem !== true) || (error.httpCode === 500)) return event; // throw exceptions.
+      const error = sanitizedEvent.extra.Error;
+      if (error == null) return sanitizedEvent; // we aren't sure why there isn't an exception; pass through.
+      if ((error.isProblem !== true) || (error.httpCode === 500)) return sanitizedEvent; // throw exceptions.
       return null; // we have a user-space problem.
     }
   });
   return Sentry;
 };
 
-module.exports = { init };
+module.exports = { init, sanitizeEventRequest, isSensitiveEndpoint, filterTokenFromUrl };
 
--- a/test/unit/external/sanitize-sentry.js
+++ b/test/unit/external/sanitize-sentry.js
@@ -1,0 +1,570 @@
+require('should');
+const appRoot = require('app-root-path');
+const { sanitizeEventRequest, isSensitiveEndpoint, filterTokenFromUrl } = require(appRoot + '/lib/external/sentry');
+
+// These cases are based on real requests!
+const cases = [
+  // Request body with sensitive data
+  [
+    {
+      cookies: {},
+      data: '{"email":"test-email","password":"test-pass"}',
+      headers: {
+        host: 'localhost:8383',
+        'user-agent': 'HTTPie/2.4.0',
+        'accept-encoding': 'gzip, deflate',
+        accept: 'application/json, */*;q=0.5',
+        connection: 'keep-alive',
+        'content-type': 'application/json',
+        'content-length': '48'
+      },
+      method: 'POST',
+      query_string: null,
+      url: 'http://localhost/v1/sessions'
+    },
+    {
+      cookies: null,
+      data: null,
+      headers: {
+        host: 'localhost:8383',
+        'user-agent': 'HTTPie/2.4.0',
+        'accept-encoding': 'gzip, deflate',
+        accept: 'application/json, */*;q=0.5',
+        connection: 'keep-alive',
+        'content-type': 'application/json',
+        'content-length': '48'
+      },
+      method: 'POST',
+      query_string: null,
+      url: 'http://localhost/v1/sessions'
+    }
+  ],
+  // Authorization header
+  [
+   {
+      cookies: {},
+      data: '{}',
+      headers: {
+        host: 'localhost:8383',
+        'user-agent': 'HTTPie/2.4.0',
+        'accept-encoding': 'gzip, deflate',
+        accept: '*/*',
+        connection: 'keep-alive',
+        authorization: 'Bearer sdff.....sdQ'
+      },
+      method: 'GET',
+      query_string: null,
+      url: 'http://localhost/v1/projects/3/forms'
+    },
+    {
+      cookies: null,
+      data: null,
+      headers: {
+        host: 'localhost:8383',
+        'user-agent': 'HTTPie/2.4.0',
+        'accept-encoding': 'gzip, deflate',
+        accept: '*/*',
+        connection: 'keep-alive',
+        authorization: null
+      },
+      method: 'GET',
+      query_string: null,
+      url: 'http://localhost/v1/projects/3/forms'
+    }
+  ],
+  // Cookies and query strings from the frontend
+  [
+    {
+      cookies: {
+        csrftoken: 'j0j...U5',
+        __enketo_meta_deviceid: 's:localhost:zgC...hAE',
+        __csrf: 'kGa...w1d',
+        session: 'Aj...nR'
+      },
+      data: '{}',
+      headers: {
+        'x-forwarded-proto': 'https',
+        host: 'localhost:8383',
+        connection: 'close',
+        'sec-ch-ua': '" Not A;Brand";v="99", "Chromium";v="90", "Google Chrome";v="90"',
+        'sec-ch-ua-mobile': '?0',
+        'upgrade-insecure-requests': '1',
+        'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.93 Safari/537.36',
+        accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9',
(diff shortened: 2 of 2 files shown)


ktuite:
Some more thoughts:

This might filter more aggressively than we want... (other endpoints matching the regexes, maybe you DO want to keep the body/query string for a certain URL in a POST but not a GET, or the other way around?)

It could also potentially be configurable. Maybe you're running your own server and you want more of this sensitive information to come through?

In many cases, I've set certain values to null, but they could also be set to something that indicates we actively/intentionally filtered them out.

lib/external/sentry.js
```diff
@@ -1,7 +1,19 @@
 const { inspect } = require('util');
 const { isBlank } = require('../util/util');
 
-const sensitiveEndpoints = ['/users/reset/verify', '/users/(.*?)/password'];
+const sensitiveEndpoints = [
+  '/users/reset/verify',
+  '/users/(.*?)/password',
+  '/users',
```
matthew-white, line 7: Maybe be worth adding a comment mentioning that this is also intended to match /users/:id?

matthew-white (commented):
I'll let @issa-tseng take a closer look, but I thought I'd leave some thoughts!

> any matching endpoint, regardless of what kind of GET/POST/PUT request it is, will have both it's request body filtered out, and have all query parameters removed from the request and the URL that is sent to Sentry.

> maybe you DO want to keep the body/query string for a certain URL in a POST but not a GET, or the other way around?

I think it's OK to always remove the request body. I feel less sure about always removing all query parameters, but maybe it's OK for now? In the list of endpoints, the only endpoints that stand out to me as having query parameters that we wouldn't want to filter are the submission download endpoints, for example the `attachments` and `$filter` query parameters. I think it'd be ideal to preserve those, but it also might not be worth the extra logic.

> This might filter more aggressively than we want... (other endpoints matching the regexes,

Ah yes, I just made a comment along those lines. If it's easy to make those regular expressions more specific, that seems nice. But if the worst case is just extra scrubbing, it might not matter. Maybe it'd be worth adding `/v1` to the regular expressions so that it's more likely that they're matching from the start of the path?

> Previously, cookies were summarized in the request header, but getting passed through in the request object and still visible in Sentry.

Ah, good catch!

> I'm not sure what the `st` query parameter of a public link refers to. I see one that is part of an Enketo URL, which is outside of `central-backend`'s + Sentry's realm.

Exactly, it's used as part of public links. I think in the context of public links, Enketo will specify the `st` query parameter when sending requests to Backend.

> How to test this is going to be interesting! I have been trying it out with a dev sentry account, breaking the code, making various API requests, and checking that the information is getting removed... Not sure if there's a more repeatable way to test.

Good idea to do this with a dev Sentry account! I think the dream would be to have tests around the logic here, but I'm not sure how to achieve that. Just a thought, maybe you could export the `beforeSend()` function and unit-test that on its own? You'd probably need to mock the `event` object.

> It could also potentially be configurable. Maybe you're running your own server and you want more of this sensitive information to come through?

Very interesting! My instinct is that we could wait for this sort of configuration to be requested — it'd be interesting to hear about the use case and how much configuration is needed.

> In many cases, I've set certain values to null, but they could also be set to something that indicates we actively/intentionally filtered them out.

No strong preference!

For the CSRF token, that can be passed through a cookie and can also be passed in the request body. So I think we should always remove the `__csrf` property from the request body, similar to how we always remove the `Authorization` header. This would only be if the request body is a simple object (not a string or something else).

lib/external/sentry.js
```diff
@@ -1,7 +1,19 @@
...
+  '/sessions',
+  '/projects/(.*?)/key',
+  '/projects/([^/]*)/forms/([^/]*)/submissions.csv.zip',
+  '/projects/([^/]*)/forms/([^/]*)/submissions.csv',
+  '/projects/([^/]*)/forms/([^/]*)/draft/submissions.csv',
+  '/projects/([^/]*)/forms/([^/]*)/draft/submissions.csv.zip',
+  '/config/backups/initiate',
+  '/backup',
+];
```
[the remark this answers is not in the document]
matthew-white: Ah I'm realizing that /users will also cover that one?

matthew-white (commented):
> any more...?

I took a quick look at the API docs to see if I could think of any others:

- POST /v1/projects/{projectId}/forms and POST /v1/projects/{projectId}/forms/{xmlFormId}/draft
  - The XForm can be sensitive.
- That's probably also true of the endpoints to create a submission or to update its XML?
- POST /v1/projects/{projectId}/forms/{xmlFormId}/draft/attachments/{filename}
  - Is the request body sent to Sentry in this case?
- Maybe also the endpoints to send a submission attachment?
- Maybe submission comments and the `X-Action-Notes` header?

lib/external/sentry.js
```diff
@@ -34,16 +46,40 @@ const init = (config) => {
...
+          event.request.headers.authorization = null; // eslint-disable-line no-param-reassign
+
+        // clear out all request body data and query params for sensitive endpoints (logging in, etc)
         if (event.request.url) {
           for (const endpoint of sensitiveEndpoints) {
             if (event.request.url.match(endpoint)) {
               event.request.data = null; // eslint-disable-line no-param-reassign
+
+              // remove query string from request data and from URL
+              event.request.query_string = null; // eslint-disable-line no-param-reassign
```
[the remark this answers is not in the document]
ktuite: yep, there are now some examples in the unit test i added

ktuite:
This isn't done yet, but a lot has been said/changed and I wanted to consolidate comments:

Great suggestion @matthew-white to unit-test what's in that `lib/external/sentry.js`! There are now tests for stripping out information that are based on real requests, and there are tests for the regex-heavy parts of the code (identifying sensitive endpoints and filtering out tokens). This will make it easier to think about which endpoints are really sensitive or not, and make sure the filters work as expected.

TODO:
- Possibly be more specific with sensitive endpoint regular expressions
- Always remove request body?
- Don't always remove all query parameters?
- What's an example of a non-sensitive URL that we want to preserve data for, and what kind of data?
- Look into CSRF token getting passed in from request body? It's in a field called `cookie` and the previous code removed it
- Check how Enketo talks to to backend server (I've done curl and frontend, but not Enketo)
- Look at how XML data gets sent, probably filter it out since XForm data itself is sensitive
	* POST /v1/projects/{projectId}/forms and POST /v1/projects/{projectId}/forms/{xmlFormId}/draft
- Look at how submission attachment data gets sent and probably filter it out, too
	* POST /v1/projects/{projectId}/forms/{xmlFormId}/draft/attachments/{filename}
- Look at submission comments and the X-Action-Notes header?

ktuite:
I've decided to remove request body data from all requests instead of trying to identify specific requests where sensitive data may come through.

Looking into submissions from Enketo: Enketo sends the `csrf` token in the request body (among other places) for some reason, but that now gets cleared out.

I checked for sensitive data within forms or submissions themselves  (XML form or submission files, attachments, etc.), but didn't see any raw request data coming through to Sentry.

I thought a lot about query parameters. I thought about filtering them all out, or filtering them out *except* in certain cases, but I ultimately decided there are many useful and non-sensitive query parameters (`?invalidate`, `?publish`, `?odata`), so I ultimately identified the few that _are_ potentially sensitive and have those removed:

* `/users?q=<something personal in the name or email>`
* When the key id and password is used in a `GET` request for exporting encrypted submissions: `projects/<proj id>/forms/<form id>/submissions.csv?<keyId>=<password>&<other possibly useful filters>`
	* Because I don't want to carefully remove JUST encryption keys, I just remove the whole query string here.
	* The `POST` version of the request preserves the query string filters because it expects the sensitive information to come through the body.

I've updated the tests to reflect these changes!

matthew-white (commented):
The tests look great! Nice that you were able to grab real request objects and use them in tests. I also think the code is structured well: I think it makes things clear to split the logic into separate functions.

lib/external/sentry.js
```diff
@@ -1,7 +1,80 @@
...
+  }
+
+  // remove draft token from any URL
+  if (url.match('/v1/test/(.*?)')) {
+    return url.replace(/v1\/test\/([^/]*)/, 'v1/test/[FILTERED]');
+  }
+
+  // return original URL if there are no tokens
+  return url;
+};
```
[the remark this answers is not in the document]
ktuite: Ah, I hadn't seen this before because I hadn't tried using enketo with public links. I've now had a chance to look into it, and into the middleware that transforms it into the `key` url. There is now new code and tests (including ones I saw when I manually broke the middleware) to reflect this case.

I can't easily remove just this specific query parameter, but I added these URLs that enketo sends `st` with to the list of sensitive endpoints that will have all their query parameters stripped.

I also saw this `st` param getting captured in the request header `referer` field, so I stripped that out of all requests, too.

lib/external/sentry.js
```diff
@@ -1,7 +1,80 @@
...
+
+    // clear out request body data for all endpoints
+    if (event.request.data)
+      event.request.data = null;
+
+    // handle endpoints that might expose sensitive data through query params
+    if (event.request.url) {
+      if (isSensitiveEndpoint(event.request) && event.request.query_string) {
+        // remove query string from request data and from URL
+        event.request.query_string = null;
```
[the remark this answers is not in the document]
ktuite: In my experiments, I've only seen enketo device ids mentioned in a cookie: ` __enketo_meta_deviceid`. But if it gets in there as a query parameter, I believe it will get filtered out along with the st query parameter.

lib/external/sentry.js
```diff
@@ -1,7 +1,80 @@
...
+      event.request.headers.cookie = cookies;
+    }
+
+    // clear out cookie values not stored in request header
+    if (event.request.cookies)
+      event.request.cookies = null;
+
+    // clear out Authorization header
+    if (event.request.headers && event.request.headers.authorization)
+      event.request.headers.authorization = null;
```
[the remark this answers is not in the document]
ktuite: done!

lib/external/sentry.js
```diff
@@ -1,7 +1,92 @@
...
 const { isBlank } = require('../util/util');
 
-const sensitiveEndpoints = ['/users/reset/verify', '/users/(.*?)/password'];
+
+// Endpoints where query strings are sensitive and should be removed
+const sensitiveEndpoints = [
+  ['GET', '/users[^/]'], // ?q=<part of name or email>
+  ['GET', '/projects/([^/]*)/forms/([^/]*)/submissions.csv'], // ?key=passphrase for managed encryption
+  ['GET', '/projects/([^/]*)/forms/([^/]*)/draft/submissions.csv'], // ?key=passphrase for managed encryption
+  ['HEAD', '/projects/([^/]*)/formList'], // ?st=token for enketo public link access (showing form)
```
matthew-white, line 10: Also GET in addition to HEAD I think.

lib/external/sentry.js
```diff
@@ -1,7 +1,92 @@
 const { inspect } = require('util');
 const { isBlank } = require('../util/util');
 
-const sensitiveEndpoints = ['/users/reset/verify', '/users/(.*?)/password'];
+
+// Endpoints where query strings are sensitive and should be removed
+const sensitiveEndpoints = [
+  ['GET', '/users[^/]'], // ?q=<part of name or email>
```
matthew-white, line 7: Does `[^/]` need to be in there? I think it'll just match the `?` of the query string?

matthew-white (commented):
Sorry for the delay in review! This is looking good from my perspective. I've been reviewing things with an eye to the underlying logic/behavior (how the filtering should work). I didn't think about code style or anything along those lines. Would be great for @issa-tseng to take a look as well! (I'll let @issa-tseng leave the final ✅.)

lib/external/sentry.js
matthew-white: Oh I see, it's there in order not to match endpoints like `/v1/users/reset/initiate`. :+1:

lib/external/sentry.js
```diff
@@ -1,7 +1,93 @@
...
+      const cookies = event.request.headers.cookie.split('; ')
+        .map((cookie) => cookie.slice(0, cookie.indexOf('=')))
+        .join(', ');
+
+      event.request.headers.cookie = cookies;
+    }
+
+    // clear out cookie values not stored in request header
+    if (event.request.cookies)
+      event.request.cookies = null;
```
issa-tseng, line 54: for all of these null-out assignments, i wonder if it would be a better idea to do something like eg `event.request.cookies.replace(/./g, '*')` or even just `'***'` so we know there _was_ something there?

lib/external/sentry.js
```diff
@@ -1,7 +1,93 @@
...
+    if (event.request.headers && event.request.headers.authorization)
+      event.request.headers.authorization = null;
+
+    // clear out referer header (from enketo submission, can have ?st=token in url)
+    if (event.request.headers && event.request.headers.referer)
+      event.request.headers.referer = null;
+
+    // clear out x-action-notes
+    if (event.request.headers && event.request.headers['x-action-notes'])
+      event.request.headers['x-action-notes'] = null;
```
issa-tseng, line 66: nah let this through i think

lib/external/sentry.js
```diff
@@ -1,7 +1,93 @@
...
+
+    // clear out request body data for all endpoints
+    if (event.request.data)
+      event.request.data = null;
```
issa-tseng, line 70: oh i didn't know we were going this far w it. i would have let this be. i know there's the POST passphrases we'd have to clear though.

lib/external/sentry.js
```diff
@@ -24,25 +110,7 @@ const init = (config) => {
...
-        // clear out all request body data for sensitive endpoints (logging in, etc)
-        if (event.request.url) {
-          for (const endpoint of sensitiveEndpoints) {
-            if (event.request.url.match(endpoint)) {
-              event.request.data = null; // eslint-disable-line no-param-reassign
-            }
-          }
-        }
-      }
+      event = sanitizeEventRequest(event); // eslint-disable-line no-param-reassign
```
issa-tseng, line 113: could just `const sanitized =` rather than disabling eslint. if it's as simple as that to make the linter happy i usually do.

issa-tseng:
code looks fine. did we decide to basically filter any and all user data period?

issa-tseng:
it feels like at the point we're taking _this_ much stuff out sentry isn't going to be useful for more than "there is an error somewhere." if so maybe it would be easier to just pick out the four or five values that have any kind of meaning anymore and ship em.. url, verb, host, user agent, error (message/stack trace), whatever.

than implement _all_ this filtering i mean.

matthew-white:
> it feels like at the point we're taking _this_ much stuff out sentry isn't going to be useful for more than "there is an error somewhere." if so maybe it would be easier to just pick out the four or five values that have any kind of meaning anymore and ship em.. url, verb, host, user agent, error (message/stack trace), whatever.
> 
> than implement _all_ this filtering i mean.

I can see this both ways! I agree that the code has a lot more logic now. We'll also need to update it whenever there's a new endpoint with a sensitive query parameter.

On the other hand, I think we'll need to do some filtering no matter what. For example, we'll need to filter a token from the URL. I also think that with the exception of request data, the current filtering is pretty targeted: we only remove the query string for select endpoints, and we only filter out select headers.

We could try to be more targeted, by filtering out the request data for only select endpoints. I don't quite remember, but I think there would be maybe 10 of those (`master` does that filtering for a couple of endpoints today, and #347 lists several more). As with filtering out query parameters, that would require updating the list whenever there's a new endpoint with potentially sensitive request data.

I feel like we have two goals that are in tension: maximizing how much useful data is sent to Sentry while minimizing code complexity. The current solution seems somewhere in the middle. We could make the code simpler by always removing query parameters and removing all headers, but then we wouldn't see that information in Sentry. Alternatively, we could do more targeted filtering of the request data, but that'd result in more code.

I can definitely see it both ways and don't have strong feelings. The current code is well covered by unit tests, which makes me feel better about the amount of code. If we wanted to, I think we could add targeted filtering of the request data without too much extra cost (maybe we could have the `sensitiveEndpoints` constant cover both the query string and request data, rather than having that be two separate pieces of logic). On the other hand, I think it's hard to remember to update the `sensitiveEndpoints` constant, and maybe it'd be best to go with simple code that we'd almost never have to change. 🤔

lib/external/sentry.js
ktuite: I like the idea of clearly showing when something was filtered out by us vs. just not set for some reason.

lib/external/sentry.js
ktuite: I think it got brought up as something to filter out because _might_ write personal info in these audit logs?

lib/external/sentry.js
ktuite: i'm having a hard time detecting empty objects and not overwriting them with '***' though (because of some weird nuance of javascript)

lib/external/sentry.js
issa-tseng: people _will_ put personal info through the whole system.. if that's the rubric we shouldn't collect any data at all. example: the form instanceID could have sensitive text in it, so the url could be no good. just my opinion idk what to say or recommend here.

matthew-white (approved):
@lognaturel mentioned on Slack that she'd like to try what's here, so I'll go ahead and merge. I'm sure this won't be the last time we touch Sentry, but I think this PR provides a great basis for future work, in large part because there are now a number of tests.

PR #376, opened 2021-05-13
final 2

issa-tseng:
2 more bugfixes maybe

### DIFF
--- a/lib/data/client-audits.js
+++ b/lib/data/client-audits.js
@@ -16,7 +16,7 @@ const { zipPart } = require('../util/zip');
 const headers = [ 'event', 'node', 'start', 'end', 'latitude', 'longitude', 'accuracy', 'old-value', 'new-value' ];
 
 // used by parseClientAudits below.
-const parseOptions = { bom: true, trim: true, skipEmptyLines: true };
+const parseOptions = { bom: true, trim: true, skipEmptyLines: true, relax: true };
 const headerLookup = {};
 for (const header of headers) headerLookup[header] = true;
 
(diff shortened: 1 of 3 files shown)


lib/resources/backup.js
```diff
@@ -115,12 +116,12 @@ const backup = (keys, response) => {
...
   // we don't bother catching any errors here since we can't return any kind of
   // useful message to the user anyway, and there is nothing to clean up.
-  const dump = (tmpdir) => pgdump(tmpdir.path).then(() => {
+  tmp.withDir((tmpdir) => pgdump(tmpdir.path).then(() => {
     // halt the keepalive.
     release();
 
     // TODO: mostly copypasta for now from lib/task/fs:
-    promisify(readdir)(tmpdir.path).then((files) => {
+    return promisify(readdir)(tmpdir.path).then((files) => {
```
ktuite, line 124: Did this line make a big difference? The rest of it makes sense. This change seems like a good idea but now I don't get what it was doing before.

lib/data/client-audits.js
```diff
@@ -16,7 +16,7 @@ const { zipPart } = require('../util/zip');
 const headers = [ 'event', 'node', 'start', 'end', 'latitude', 'longitude', 'accuracy', 'old-value', 'new-value' ];
 
 // used by parseClientAudits below.
-const parseOptions = { bom: true, trim: true, skipEmptyLines: true };
+const parseOptions = { bom: true, trim: true, skipEmptyLines: true, relax: true };
```
ktuite, line 19: Yasss, one little form that has been broken for me since I reproduced this issue (https://github.com/getodk/collect/issues/4473) now exports properly.

ktuite (approved):

lib/resources/backup.js
issa-tseng: i think it was not working before lol

PR #379, opened 2021-05-28
Move formId extension check from Form.fromXml to Form ingestion resource (migration patch)

ktuite:
There was an issue with database migrations where the migration to backfill form def names used `Form.fromXml`, which was checking for `.xml` and `.xslx` extensions inside formIDs and throwing an error. This was causing the whole migration to fail, which was no good! The original purpose of this formID check was to prevent forms with this formID pattern from even being uploaded.

This extension check has now been moved from `fromXml` to a `getPartial()` method used in the form ingestion resource.

### DIFF
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -67,8 +67,14 @@ module.exports = (service, endpoint) => {
       Form.fromXml(input))
     )
     /* eslint-enable indent */
+      .then(rejectIf((partial) => (/\.(xlsx?|xml)$/.test(partial.xmlFormId)),
+        (partial) => Problem.user.unexpectedValue({
+          field: 'formId',
+          value: partial.xmlFormId,
+          reason: 'The Form ID cannot end in .xls, .xlsx, or .xml. Please either specify an allowed ID in the form itself, or rename the file (eg change form.xls.xls to form.xls).'
+        })))
       // if we don't have managed encryption, or the form carries its own key,
-      // we can use the form xml as-is. otherwise we must inject things.
+      // we can use the form xml as-is. otherwise we must inject things
       .then((partial) => (((project.keyId == null) || partial.aux.key.isDefined())
         ? partial
         : Keys.getById(project.keyId)
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -103,13 +103,6 @@ class Form extends Frame.define(
       const xmlFormId = idText.map(blankStringToNull)
         .orThrow(Problem.user.missingParameter({ field: 'formId' }));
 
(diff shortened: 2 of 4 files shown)


lib/resources/forms.js
```diff
@@ -69,6 +69,13 @@ module.exports = (service, endpoint) => {
     /* eslint-enable indent */
       // if we don't have managed encryption, or the form carries its own key,
       // we can use the form xml as-is. otherwise we must inject things.
+      .then((partial) => (/\.(xlsx?|xml)$/.test(partial.xmlFormId)) // eslint-disable-line no-confusing-arrow
```
issa-tseng, line 72: i’ve always appeased this rule actually

lib/resources/forms.js
```diff
@@ -69,6 +69,13 @@ module.exports = (service, endpoint) => {
     /* eslint-enable indent */
       // if we don't have managed encryption, or the form carries its own key,
       // we can use the form xml as-is. otherwise we must inject things.
+      .then((partial) => ((/\.(xlsx?|xml)$/.test(partial.xmlFormId))
```
issa-tseng, line 72: the only suggestion i'd make is to consider using `rejectIf` if you feel like it. it does exist and it's used a couple places but there's nothing wrong w this really.

lib/resources/forms.js
```diff
@@ -69,6 +69,13 @@ module.exports = (service, endpoint) => {
     /* eslint-enable indent */
       // if we don't have managed encryption, or the form carries its own key,
       // we can use the form xml as-is. otherwise we must inject things.
```
lognaturel, line 71: I think this comment should now move down to the next `.then`?

ktuite: ah yes, moved down!

lib/resources/forms.js
ktuite: ok, i tried using rejectIf

lib/resources/forms.js
```diff
@@ -67,15 +67,14 @@ module.exports = (service, endpoint) => {
       Form.fromXml(input))
     )
     /* eslint-enable indent */
-      // if we don't have managed encryption, or the form carries its own key,
-      // we can use the form xml as-is. otherwise we must inject things.
-      .then((partial) => ((/\.(xlsx?|xml)$/.test(partial.xmlFormId))
-        ? reject(Problem.user.unexpectedValue({
+      .then(rejectIf((partial) => (/\.(xlsx?|xml)$/.test(partial.xmlFormId)),
+        (partial) => (Problem.user.unexpectedValue({
```
issa-tseng, line 71: now you don't need this `(` before `Problem` but _(shrug!)_

PR #382, opened 2021-07-06
new: add basic user access logging.

matthew-white:


lib/resources/sessions.js
```diff
@@ -24,17 +24,27 @@ module.exports = (service, endpoint) => {
     return Users.getByEmail(email)
       .then(getOrReject(Problem.user.authenticationFailed()))
       .then((user) => bcrypt.verify(password, user.password)
-        .then((verified) => ((verified !== true)
-          ? Problem.user.authenticationFailed()
-          : Sessions.create(user.actor)
```
matthew-white, line 29: I thought about using `Promise.all()` within the ternary, but that seemed less clear to me. Instead, I started using `rejectIf()`.

test/integration/api/app-users.js
```diff
@@ -220,7 +220,7 @@ describe('api: /projects/:id/app-users', () => {
       service.login('alice', (asAlice) =>
         asAlice.post('/v1/projects/1/app-users').send({ displayName: 'condemned' }).expect(200)
           .then(({ body }) => asAlice.delete('/v1/projects/1/app-users/' + body.id).expect(200))
-          .then(() => asAlice.get('/v1/audits')
+          .then(() => asAlice.get('/v1/audits?action=field_key')
```
matthew-white, line 223: After the addition of `user.session.create`, a number of tests related to the audit log started failing. Where possible, I specify the `action` filter in the test, which seemed like the easiest way to update the test.

issa-tseng (approved):

issa-tseng:
probably in one or many of code comments/api docs/user docs we should note which actions are nondestructive.

matthew-white:
> probably in one or many of code comments/api docs/user docs we should note which actions are nondestructive.

I've added the new actions to the API docs, but I haven't noted which ones are nondestructive. `user.session.create` also seems a little hard to categorize, since it's not entirely a read-only operation. 🤔

PR #383, opened 2021-07-13
Consider updatedAt for lastSubmission date in "extended" project forms list

punkch:
Hello,

It will be great if the `lastSubmission` timestamp in the response from `projects/:projectId/forms` also indicate if a submission was recently edited.

I am using this endpoint to figure which forms have new submissions and hence import them in another application. Now, with submission editing in place (:champagne:), I can't find an efficient method to catch the "edits".

Thank you!

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -750,7 +750,7 @@ The Projects listing endpoint is somewhat unique in that it is freely accessible
 
 Currently, there are no paging or filtering options, so listing `Project`s will get you every Project you have access to.
 
-This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to additionally retrieve the `appUsers` count of App Users and `forms` count of Forms within the Project, as well as the `lastSubmission` timestamp of the latest submission to any for in the project, if any.
+This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to additionally retrieve the `appUsers` count of App Users and `forms` count of Forms within the Project, as well as the `lastSubmission` timestamp of the latest submission to any form in the project, if any. The response will also include the `lastSubmissionUpdate` timestamp, indicating when a submission to any form in the project was last modified.
 
 + Response 200 (application/json)
     This is the standard response, if Extended Metadata is not requested:
@@ -784,7 +784,7 @@ To create a Project, the only information you must supply (via POST body) is the
 
 To get just the details of a single Project, `GET` its single resource route by its numeric ID.
 
-This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to additionally retrieve the `appUsers` count of App Users and `forms` count of forms within the Project, as well as the `lastSubmission` timestamp of the latest submission to any for in the project, if any.
+This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to additionally retrieve the `appUsers` count of App Users and `forms` count of forms within the Project, as well as the `lastSubmission` timestamp of the latest submission to any form in the project, if any. The response will also include the `lastSubmissionUpdate` timestamp, indicating when a submission to any form in the project was last modified.
 
 In addition, the extended metadata version of this endpoint (but not the overall Project listing) returns an array of the `verbs` the authenticated Actor is able to perform on/within the Project.
 
@@ -1070,7 +1070,7 @@ Currently, there are no paging or filtering options, so listing `Form`s will get
 
 As of version 1.2, Forms that are unpublished (that only carry a draft and have never been published) will appear with full metadata detail. Previously, certain details like `name` were omitted. You can determine that a Form is unpublished by checking the `publishedAt` value: it will be `null` for unpublished forms.
 
-This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to additionally retrieve the `submissions` count of the number of `Submission`s that each Form has and the `lastSubmission` most recent submission timestamp, as well as the Actor the Form was `createdBy`.
+This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to additionally retrieve the `submissions` count of the number of `Submission`s that each Form has and the `lastSubmission` most recent submission timestamp, as well as the Actor the Form was `createdBy`. The response will also include the `lastSubmissionUpdate` timestamp, indicating when a submission was last modified.
 
 + Response 200 (application/json)
     This is the standard response, if Extended Metadata is not requested:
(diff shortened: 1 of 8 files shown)


issa-tseng:
thanks for the pr!
i think this would be a great _option_ for the api, but i think we would need to update the ui to communicate this behavior to the user if it's going to be the default sort.
you can look into usages of the queryOptions args methods for receiving and applying  options.

alternatively, you can produce a `lastSubmissionUpdate` (or similar i didn't think hard about the name) field _alongside_ `lastSubmission`. this is probably the most useful for the most people.

matthew-white:
@punkch, one thing to consider is that while `updatedAt` is updated when there is a new submission version, it is also updated when a submission is reviewed. It sounds like you want something more like "time of last version." I think the best way to access that information is `submission_defs."createdAt"`. Another difference between that and `submission."updatedAt"` is that (I think) `updatedAt` will be `null` for a submission that has never been edited or reviewed.

We've had the thought that it could be useful to add a field like that to the OData response. That query already pulls from `submission_defs` and calculates the edit count, so in some ways, that might be the easiest place to make this change:

https://github.com/getodk/central-backend/blob/9f4add7c7432e97d3b7e506b1b8b6dc31f4384b7/lib/model/query/submissions.js#L173-L207

That said, if we added that to the OData response, we'd probably also want to filter on it. But that might be challenging, because in some cases, we also count the number of submissions, and that query doesn't join in `submission_defs`:

https://github.com/getodk/central-backend/blob/9f4add7c7432e97d3b7e506b1b8b6dc31f4384b7/lib/model/query/submissions.js#L105-L107

@issa-tseng, how important is the `countByFormId` query? Could Central always do the counting instead of Postgres? Alternatively, what do you think about adding `submission_defs` to the `countByFormId()` query? Or adding a `lastSubmittedAt` column to the `submissions` table that is modified only when a new version is created?

punkch:
Hi @issa-tseng  and @matthew-white 

Thank you very much for looking into this and the deliberate explanations!

It is even better that `updatedAt` changes when there is a review of the submission, so in the future I'll also be able to syncronize these as well. I've changed back the logic for `lastSubmission` to what it was and added the `lastSubmissionActivity` field alongside.

Unfortunately, this is the best I can do, as the queryOptions and the OData API are way beyond my understanding of the codebase so far.

matthew-white:
Thanks, @punkch!

> It is even better that `updatedAt` changes when there is a review of the submission, so in the future I'll also be able to syncronize these as well.

Just to make sure I understand your use case, are you thinking that even if there isn't a new submission version, it'd be useful to import the latest review state into the other application? So `updatedAt` is actually better than "time of last version" in order to keep things synchronized?

punkch:
Hi @matthew-white 

Well, I basically import all the project forms list on a scheduled basis. If there is a change in the lastSubmission date for some form, I'll import the submissions that I am missing from the OData API. Now with editing, I will have to import the submissions that are updated, and in the future, will also be able to keep the comments in sync.

matthew-white:
Thanks, that's helpful! One thing to consider is that while `submissions."updatedAt"` is updated when the submission's review state changes, it isn't updated when there is a new comment on the submission (at least right now).

matthew-white:
@punkch, I wanted to let you know that we've merged a related PR, #386, which adds `submissions."updatedAt"` to the OData response (under `__system/updatedAt`). This will be `null` when a submission is first created, then updated when the submission is edited or reviewed. It sounds like you're already using the OData API, so you may find it helpful to use this field as part of the `$filter` query parameter. You could specify something like the following for the value of `$filter`:

```
__system/submissionDate gt 2021-08-01T01:23:45.678Z or __system/updatedAt gt 2021-08-01T01:23:45.678Z
```

We'd also be happy to include `lastSubmissionActivity` on the extended form response if you'd find that helpful. It'd be great if you could look into adding that property to tests (primarily `test/integration/api/forms.js` and the `ExtendedForm` assertion in `test/assertions.js`), as well as updating the API docs.

punkch:
Hi @matthew-white,
Thanks for letting me know. I definitely want to finish up on this PR, just bear with me for couple of weeks more, as I am currently on a small vacation.

punkch:
Hi @matthew-white 

The mentioned tests and the documentation are now updated to include `lastSubmissionActivity`. Let me know if something.

issa-tseng:
hey @punkch, i am so sorry it's been such a long delay on this.

i really wasn't sure what i wanted to do here. i think i'd like to suggest that we change the feature back to what i suggested, which is `lastSubmissionUpdate`.

as it is, we are providing `max(createdAt)` as `lastSubmission` and `max(coalesce(updatedAt, createdAt))` and i think it would be cleaner and more straightforward if we didn't do the coalesce.

ultimately, i think your use case would be far better served by maybe an across-projects form listing than any change to the projects api. but as far as this suggestion is concerned, i think that's the direction i'd rather go. thank you, and sorry again.

punkch:
Hi @issa-tseng 

So should I change it to have both, or just the `max("updatedAt") as lastSubmissionUpdate` ?
```sql
select 
       "formId",
        count(id)::integer as "submissions", 
        max("createdAt") as "lastSubmission",
        max("updatedAt") as "lastSubmissionUpdate",
        max(coalesce("updatedAt","createdAt")) as "lastSubmissionActivity"
from submissions
```
There is also something similar going on with the [projects query](https://github.com/getodk/central-backend/blob/4da6d37bb30dd44716ef68f279669672d279189a/lib/model/query/projects.js#L78). Is it ok to add these new fields (or field) there as well?

punkch:
Hi @issa-tseng 

Sorry for taking so long, I just couldn't get to this.
I've now added the lastSubmissionUpdate to both, projects and forms when ExtendedMetadata is true, and rebased a little. Let me know if I am missing on something, or if I can improve the change somehow.

I am not sure about the cross project form list, but here is my plan on how to use these fields:
 - on a scheduled basis, I'll get the project list and compare the lastSubmission and lastSubmissionUpdate with those from the previous run. Technically, I save these in a firestore collection of projects metadata, and if there is a change for specific form, it will trigger the next step.
 - whenever there is a change in a forms lastSubmission, or lastSubmissionUpdate, I will get the new and changed submissions as JSON via the OData APIs with expanded=* and proper filters for createdAt and updatedAt.
 - Then I save each submission's json (as a Firestore document), keeping the data for the latest submission version.
 - Whenever there is a change to a submission (create, or update) I action it, depending on the form, project, etc. At the moment, I only stream these changes to a Google BigQuery table per form.

I could probably do this much better, if there was a possiblity in central to somehow subscribe external systems to the events going on there.

issa-tseng:
i'm wondering if you can maybe use the audit table and the audit api to read update events and use that instead to know when to pull different things?

i'm sorry i keep dragging my feet on this. i keep thinking we'll have to support this field and calculation forever even though maybe there's a better way to do this that doesn't create any new fields and helps the synchronization case better.

if you are still interested in pursuing the pr:
i think my last complaint about the code that i didn't notice before (sorry) is that while you've fixed the existing tests to not break with the new code, i don't really see anything in here that actually asserts that the value is something we expect at a moment we expect it (eg does an update and looks to see that the value is there).

punkch:
Hi @issa-tseng 

Actually, I can make my integration much more efficient if I base it around the audits api. Thank you very much for the pointer!
Thus said, I don't want to persue this PR anymore.

matthew-white:
@punkch, I'm glad it's worked to base your integration around the audits API! I'll go ahead and close this PR, but thank you for the useful discussion!

PR #384, opened 2021-07-16
new: add endpoints for configuring analytics.

matthew-white:
This PR is very much a work in progress! But I have a few questions, so I thought I'd push what I have so that we could discuss with a little context. I'll add a few specific questions below.

### DIFF
--- a/lib/model/frames.js
+++ b/lib/model/frames.js
@@ -75,7 +75,12 @@ const Comment = Frame.define(
   embedded('actor')
 );
 
-const Config = Frame.define(table('configs'), 'key', 'value', 'setAt', species('config'));
+const Config = Frame.define(
+  table('configs'),
+  'key',        readable,               'value',        readable, writable,
+  'setAt',      readable,
+  species('config')
+);
 
 class FieldKey extends Frame.define(
   table('field_keys'), aux(Actor),
--- a/lib/model/frames.js
+++ b/lib/model/frames.js
@@ -7,7 +7,6 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-const { identity, pick } = require('ramda');
 const { sql } = require('slonik');
 const { Frame, table, readable, writable, aux, species, embedded } = require('./frame');
 const { isBlank } = require('../util/util');
@@ -76,56 +75,7 @@ const Comment = Frame.define(
   embedded('actor')
 );
 
-// configKeys holds options for each config. It determines whether the config
-// can be directly set over the API, and if so, how. It also determines
-// whether/how the config value is returned over the API and saved in the audit
-// log. (Using Object.create(null) to prevent access to Object.prototype
-// properties.)
-const configKeys = Object.assign(Object.create(null), {
-  analytics: {
-    fromApi: (value) => {
-      const result = { enabled: value.enabled === true };
-      if (result.enabled) {
-        if ((typeof value.email === 'string') && (value.email.length !== 0))
-          result.email = value.email;
-        if ((typeof value.organization === 'string') && (value.organization.length !== 0))
-          result.organization = value.organization;
-      }
-      return result;
-    },
-    forApi: identity
-  },
-  'backups.main': {
-    forApi: pick(['type'])
-  },
-  'backups.google': {
-    forApi: () => 'New or refreshed Google API credentials'
-  }
-});
-
-class Config extends Frame.define(
-  table('configs'),
-  'key',        readable,               'value',
-  'setAt',      readable,
-  species('config')
-) {
-  // Returns `true` if the config for `key` can be directly set over the API and
-  // `false` if not.
-  static settableFromApi(key) {
-    return (configKeys[key] != null) && (configKeys[key].fromApi != null);
-  }
-
-  static valueFromApi(key, value) {
-    return this.settableFromApi(key) ? configKeys[key].fromApi(value) : null;
-  }
-
-  forApi() {
-    const data = { key: this.key, setAt: this.setAt };
-    const forApi = configKeys[this.key] != null ? configKeys[this.key].forApi : null;
-    if (forApi != null) data.value = forApi(this.value);
-    return data;
-  }
-}
+const { Config } = require('./frames/config');
 
 class FieldKey extends Frame.define(
   table('field_keys'), aux(Actor),
--- a/lib/model/frames.js
+++ b/lib/model/frames.js
@@ -7,6 +7,7 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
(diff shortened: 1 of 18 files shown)


lib/resources/config.js
```diff
@@ -67,11 +69,46 @@ module.exports = (service, endpoint) => {
...
+  };
+
+  service.get('/config/:key', endpoint(({ Configs }, { auth, params }) =>
+    auth.canOrReject('config.read', Config.species)
+      .then(checkKey(params.key))
+      .then(() => Configs.get(params.key))
+      .then(getOrNotFound)));
+
+  service.delete('/config/:key', endpoint(({ Configs }, { auth, params }) =>
+    auth.canOrReject('config.set', Config.species)
```
matthew-white, line 107: Right now, I'm using a single verb for both POST and DELETE: `config.set`. However, maybe I should use two verbs, `config.set` and `config.unset`? That would seem to nicely parallel the query methods `Configs.set()` and `Configs.unset()`. (Note that I haven't added any new verbs to the `admin` role yet.)

lib/model/migrations/20210716-01-config-value-jsonb.js
```diff
@@ -0,0 +1,16 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/opendatakit/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) =>
+  db.raw('alter table config alter value type jsonb using value::jsonb');
```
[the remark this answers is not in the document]
issa-tseng: strong agree.

lib/resources/config.js
```diff
@@ -67,11 +69,46 @@ module.exports = (service, endpoint) => {
...
+        field: 'key',
+        value: key,
+        reason: 'the key is reserved for internal use.'
+      });
+  };
+
+  service.get('/config/:key', endpoint(({ Configs }, { auth, params }) =>
+    auth.canOrReject('config.read', Config.species)
+      .then(checkKey(params.key))
+      .then(() => Configs.get(params.key))
```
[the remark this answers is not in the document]
issa-tseng: i'd rather move the backup api in the direction of the `:key` sample you show above than the other way around. i don't think we should do any magic merging.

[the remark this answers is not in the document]
issa-tseng: but i think if it were me i wouldn't worry about what backup is doing right now.

lib/resources/config.js
```diff
@@ -67,11 +69,46 @@ module.exports = (service, endpoint) => {
...
+    auth.canOrReject('config.set', Config.species)
+      .then(() => {
+        const value = { enabled: !!body.enabled };
+        if (value.enabled) {
+          if (isPresent(body.email)) value.email = body.email.toString();
+          if (isPresent(body.organization))
+            value.organization = body.organization.toString();
+        }
+        return Configs.set('analytics', value);
+      })));
```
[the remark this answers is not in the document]
issa-tseng: this is sort of why i abandoned the `/config/:key` code i think. i ended up figuring that a pile of small bespoke endpoints might be better so each can take just what it expects and do what it wants w it. if there are eg config changes that imply tasks get loaded into the queue etc you're going to end up w a huge headache routing all those custom extra things into the generic stuff.

lib/resources/config.js
issa-tseng: i don't feel like there will ever be a meaningful distinction between set and unset at the site config level.
maybe if  we were talking about eg change vs clear some like project level flag that has a fallback from a higher level or something.

lib/model/query/audits.js
```diff
@@ -33,7 +33,7 @@ values (${actorId}, ${action}, ${acteeId}, ${(details == null) ? null : JSON.str
 // TODO: some sort of central table for all these things, not here.
 const actionCondition = (action) => {
   if (action === 'nonverbose')
-    return sql`action not in ('submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'config.set')`;
+    return sql`action not in ('submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup')`;
```
matthew-white, line 36: Now that there are more types of config and more details are logged for `config.set`, I think it makes sense to include `config.set` in the Frontend audit log.

matthew-white (commented):
I've pushed a new commit, and I think the PR is now ready for review. I'm thinking that an interactive review might work best.

One significant change from the previous approach is that the new approach will log details about the config value when a config is set. That will happen for the analytics config and also the backups config. To set that up, I've overridden `forApi()` for the `Config` frame so that it returns only select properties of the config value.

After overriding `forApi()`, it felt like the right move to move the validation of the analytics config from the endpoint to the frame. The frame now has a "from API" concept that will determine whether a config can be directly set over the API, and if so, how it is validated. Making that change also removes the concept of `reservedKeys` from the first commit.

It wasn't a lot of code to add this "for API" / "from API" functionality, yet the approach feels flexible and allows there to be generic config endpoints that are pretty small. I think in some cases, there will continue to be bespoke config endpoints. However, I'm guessing that many future configs will simply involve getting/setting an object, similar to the analytics config.

lib/resources/config.js
matthew-white: The new commit continues to use a single verb: `config.set`. I've added logging for the existing query method `Configs.unset()`, logging a `config.set` action.

lib/model/query/audits.js
issa-tseng: yes

lib/model/frames.js
```diff
@@ -75,7 +76,56 @@ const Comment = Frame.define(
   embedded('actor')
 );
 
-const Config = Frame.define(table('configs'), 'key', 'value', 'setAt', species('config'));
+// configKeys holds options for each config. It determines whether the config
+// can be directly set over the API, and if so, how. It also determines
+// whether/how the config value is returned over the API and saved in the audit
+// log. (Using Object.create(null) to prevent access to Object.prototype
+// properties.)
+const configKeys = Object.assign(Object.create(null), {
```
issa-tseng, line 84: i think this is a fine approach. my two thoughts are
1 maybe this is heavy enough to move to its own `frames/` file. any time these things sprout too much logic i feel like this file doesn't want it.
2 i wonder if there is a way to encode these different capsules as separate frame types or subtypes rather than with sort of this internal guts way? feels like it would be more confluent w the other declarations and probably keep these overrides from becoming messy

issa-tseng (approved):
very very fine as is, but i left some suggestions

PR #385, opened 2021-07-21
improve the diff implementation with schema input and myers algo

issa-tseng:


lib/data/submission.js
```diff
@@ -59,136 +61,239 @@ const submissionXmlToFieldStream = (fields, xml) => {
...
+/* eslint-disable no-param-reassign */
+const _decorated = (obj, array = false) => {
+  obj[subhash] = createHash('sha1');
+  if (array) obj[subhashes] = [];
+  else obj[keys] = [];
+  return obj;
+};
+
+// we can't just count diff lines to determine how "much" change a delta represents
+// because we have to count nested keys.
```
issa-tseng, line 99: probably comment more here why this is

issa-tseng: also comment which util takes which and  goes in what order

ktuite (approved):
just did an interactive review with issa!

also checked what these diffs (w/ insertions and deletions in the middle of lists) look like when hooked into the frontend and they look great!

PR #386, opened 2021-07-21
new: add ability to filter submissions on updatedAt.

matthew-white:


lognaturel:
Define what `updatedAt` actually means since it's surprising that it includes both new submissions and review state changes.

issa-tseng (approved):
yup this looks exactly like what this easy painful terrible tedious exercise should lol

PR #390, opened 2021-07-26
bug: fix reference to query module in reapSessions task.

matthew-white:
Making a small change to the `reapSessions` task. (I ran the `reapSessions` task as part of debugging something unrelated and noticed that the task threw an error.)


issa-tseng:
ah oops

issa-tseng (approved):

PR #394, opened 2021-08-09
improve #377: support more ways of connecting to the database.

### LINKED ISSUE #377: Postgres connection string
Filing an issue in connection with this forum post: https://forum.getodk.org/t/connection-timeout-after-upgrading-to-central-v1-2/33690

In v1.1, we passed the `default.database` config to Knex. In v1.2, we construct a connection string using the config, then pass it to Slonik:

https://github.com/getodk/central-backend/blob/24d20de20fdf2e03b55e687542fbc0cae889a9ff/lib/external/slonik.js#L15-L18

However, I think there are differences in some cases:

- I think we need to percent-encode special characters in the password.
- It used to be possible to specify a `port` property, though I don't think we documented that. I think the user can just specify the port as part of the `host` property, right?
- It used to be possible to specify the named parameter `?ssl=true` (by specifying `"ssl": true` in the config), though again I don't think we documented that. It seems like that might something that'd be useful to officially support?

One idea is that we could allow `default.database` to be either an object or a string. That way, if you're using the typical properties, you can use the object, but if you want to do something else, you can specify a string.

matthew-white:
Closes #377.

### DIFF
--- a/lib/model/migrate.js
+++ b/lib/model/migrate.js
@@ -11,13 +11,14 @@
 // top-level operations with a database, like migrations.
 
 const knex = require('knex');
+const { connectionObject } = require('../util/db');
 
 // Connects to the postgres database specified in configuration and returns it.
-const connect = (connectionSettings) => knex({ client: 'pg', connection: connectionSettings });
+const connect = (config) => knex({ client: 'pg', connection: connectionObject(config) });
 
 // Connects to a database, passes it to a function for operations, then ensures its closure.
-const withDatabase = (connectionSettings) => (mutator) => {
-  const db = connect(connectionSettings);
+const withDatabase = (config) => (mutator) => {
+  const db = connect(config);
   return mutator(db).finally(() => db.destroy());
 };
 
--- a/lib/util/db.js
+++ b/lib/util/db.js
@@ -24,8 +24,8 @@ const { construct } = require('../util/util');
 const validateConfig = (config) => {
   const { host, port, database, user, password, ssl, ...unsupported } = config;
 
-  if (!(ssl == null || ssl === true || ssl === false))
-    return Problem.internal.invalidDatabaseConfig({ reason: 'ssl can only be true or false.' });
+  if (ssl != null && ssl !== true)
+    return Problem.internal.invalidDatabaseConfig({ reason: 'If ssl is specified, its value can only be true.' });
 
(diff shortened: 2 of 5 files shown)


lib/util/db.js
```diff
@@ -18,6 +18,45 @@ const { PartialPipe, mapStream } = require('../util/stream');
...
 
 
+////////////////////////////////////////////////////////////////////////////////
+// DATABASE CONFIG
+
+const validateConfig = (config) => {
+  const { host, port, database, user, password, ssl, ...unsupported } = config;
+
+  if (!(ssl == null || ssl === true || ssl === false))
+    throw Problem.internal.invalidDatabaseConfig({ reason: 'ssl can only be true or false.' });
```
matthew-white, line 28: Should I use a Problem even though it's not an HTTP response?

lib/util/db.js
```diff
@@ -18,6 +18,45 @@ const { PartialPipe, mapStream } = require('../util/stream');
...
+  const unsupportedKeys = Object.keys(unsupported);
+  if (unsupportedKeys.length !== 0)
+    throw Problem.internal.invalidDatabaseConfig({
+      reason: `'${unsupportedKeys[0]}' is not supported.`
+    });
+};
+
+// Returns a connection string that will be passed to Slonik.
+const connectionString = (config) => {
+  validateConfig(config);
```
matthew-white, line 39: Return a Problem or null? Or an Option of a Problem? If there is a Problem here, throw it.

matthew-white (commented):
Notes from interactive review

issa-tseng (approved):

lib/model/migrate.js
```diff
@@ -11,13 +11,14 @@
 // top-level operations with a database, like migrations.
 
 const knex = require('knex');
+const { connectionObject } = require('../util/db');
```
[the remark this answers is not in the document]
matthew-white: There shouldn't be side effects here. In general, side effects when requiring are rare, and any issue that that causes should be immediately clear.

lib/util/db.js
matthew-white: It's good to use a Problem for anything user-facing, even if it's not an HTTP response.

florianm:
Testing on my UAT server indicates this works well with external DBs with SSL enabled. Ideally, we'd like to provide a certificate, but the implementation here works well for us.

PR #396, opened 2021-08-20
Improve setup instructions / provide a make rule for starting a Postgres instance

x3ro:
Hey folks,

I've tried setting up a development instance for the first time and ran into some problems, which turned out to be mainly Node.js version related. ODK Central doesn't seem to currently play well with Node.js >= 15, where most of the integration tests fail with

````
TypeError: Cannot read property 'authorization' of undefined
````

I also tried to provide an easier way of starting a local postgres instance, as close as possible to what currently happens in CircleCI. Let me know if there's anything that can be improved.

Thanks for building ODK Central and have a great weekend 🎉

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,3 +37,11 @@ run-multi: base
 stop-multi:
 	node node_modules/naught/lib/main.js stop
 
+run-docker-postgres: stop-docker-postgres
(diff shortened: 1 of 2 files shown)


issa-tseng (approved):

issa-tseng:
thank you!

PR #397, opened 2021-08-21
Fixing a bug that resulted in Collect not being able to download forms from Central instances using custom ports

j-graham98:
Hi!

My first time contributing here, so my apologies if I've failed to follow appropriate procedures. Long story short, it turns out that when you install Central using non-standard ports, the server appears to work normally until you try to download forms to Collect. Collect is able to list the forms on the server, but errors out when trying to download them. A user on the forums did some detective work and noticed that when the formList endpoint is queried, Central currently returns a downloadUrl that omits the custom port number defined in the overall central .env file, resulting in Collect defaulting to attempting to download the forms on port 443 and failing.

In this PR I edited forms.js and openrosa.js to specify the custom port, resulting in a correct downloadUrl being returned. I'm reading the port a 'port' entry in the config. I am happy to go into the main central repo and make changes to config.json.template, start-odk.sh, and the docker-compose file so that the HTTPS_PORT variable from the .env file is added into local.json as 'port' during install, but I wanted to make sure that this solution was acceptable in general before I went and did that as well.

Thanks!

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -19,8 +19,9 @@
     },
     "enketo": {},
     "env": {
-      "domain": "http://localhost:8989",
-      "sysadminAccount": "no-reply@getodk.org"
+      "domain": "http://localhost",
+      "sysadminAccount": "no-reply@getodk.org",
+      "port": "8989"
     },
     "external": {
       "google": {
--- a/lib/formats/openrosa.js
+++ b/lib/formats/openrosa.js
@@ -46,7 +46,7 @@ const formListTemplate = template(200, `<?xml version="1.0" encoding="UTF-8"?>
       <name>{{def.name}}{{^def.name}}{{xmlFormId}}{{/def.name}}</name>
       <version>{{def.version}}</version>
       <hash>md5:{{def.hash}}</hash>
-      <downloadUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}.xml</downloadUrl>
+      <downloadUrl>{{{domain}}}:{{{port}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}.xml</downloadUrl>
       {{#aux.openRosa.hasAttachments}}
       <manifestUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}/manifest</manifestUrl>
       {{/aux.openRosa.hasAttachments}}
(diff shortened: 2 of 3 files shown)


j-graham98:
Test failure seems to be due to the fact that I cut the port number out of localhost in the default config file as I figured in production the domain is set to the value listed in the .env file and that value is not supposed to have port numbers in it. Not sure the best way to go about fixing this for testing purposes.

issa-tseng:
hey @j-graham98, thanks for the pr. it's really appreciated.

i think my only issue with the approach is that we can't simply just change the configuration like that.. people will have written all sorts of things in that field already and asking them to go edit the configuration to do this precise port splitting thing but only if they even have a custom port to begin with.. it's just a lot to ask people to understand and perform. especially just to upgrade their server to a new version.

is it possible to leave the port in the url and appropriately parse or incorporate it where it should go? is the port that's currently in the configuration getting dropped in the output, is that the problem?

config/default.json
```diff
@@ -19,8 +19,9 @@
     },
     "enketo": {},
     "env": {
-      "domain": "http://localhost:8989",
-      "sysadminAccount": "no-reply@getodk.org"
+      "domain": "http://localhost",
+      "sysadminAccount": "no-reply@getodk.org",
+      "port": "8989"
```
issa-tseng, line 24: i guess my question is, why this split? can we leave the `8989` where it already was?

j-graham98:
Hey @issa-tseng!

In production, the domain variable in local.json is set programatically through start-odk.sh (which is run by docker-compose up when you start central). start-odk.sh reads the value it passes for the domain variable from the docker .env file in the main central directory, which according to odk's [spec](https://docs.getodk.org/central-install-digital-ocean/#changing-server-settings), separates domain name and port specification. So a user who follows ODK's documentation will specify their domain in their .env file, and then separately in the .env file, specify their ports (and indeed I'm pretty sure nginx won't configure correctly if you specify port together with domain in the .env file). I figured I'd keep that paradigm and separate the two.

That said, there's no reason we couldn't concatenate domain and port when it's read out to local.json. I'm not familiar with how env.domain is used more generally through the server codebase, so I was cautious of adding port directly in that way (as again, in production, it would never be specified that way in local.json for the reason noted above), but if there is no issue with just directly specifying domain+port under domain in config from the rest of the server codebase's end we could totally do it that way!

issa-tseng:
thank you for taking a look!

yanokwa:
@j-graham98 Just a heads up that we'd like to get a Central release out this month. If you'd like your PR to be included, you have until close of business on Sept 17th and earlier is preferred!

j-graham98:
Thanks Yaw - should be doable, will try to get it in by then!

j-graham98:
@yanokwa Unfortunately, I've been tied up with my job - won't make this release, but will aim for soonish and you guys can integrate it whenever.

PR #408, opened 2021-09-30
bug: account for .enc filename suffix when decrypting client audit logs.

matthew-white:
This PR follows up on the work from 8188047f74cb8a0d8b3dbeba4e485dc1f99161bb. On the QA server, a client audit log wasn't decrypted for a form in an encrypted project. I took a look and found that the associated submission attachment was named `audit.csv.enc`, not `audit.csv`, which meant that it was not marked as a client audit log. I've changed the logic to look for `audit.csv.enc`, not `audit.csv`.


lib/model/migrations/20210903-01-backfill-encrypted-client-audits.js
```diff
@@ -11,15 +11,15 @@ const up = (db) => db.raw(`
 update submission_attachments sa set "isClientAudit"=true
   from submission_defs sd
   where
-    sa.name = 'audit.csv' and
+    sa.name = 'audit.csv.enc' and
```
matthew-white, line 14: I thought about writing a new migration, but instead opted to modify the existing one. Mostly that seems cleaner to me, but I may need to manually rerun this migration on the QA server.

lib/model/query/submission-attachments.js
```diff
@@ -58,7 +58,7 @@ const _extractEncryptedAttachments = (ensure, submission, binaryFields, fileLook
       for (; i < names.length; i += 1) {
         if (names[i].isDefined()) {
           const name = names[i].get().trim();
-          const isClientAudit = (name === 'audit.csv');
+          const isClientAudit = (name === 'audit.csv.enc');
```
matthew-white, line 61: Encrypted files always have an `.enc` suffix, right? Or should we be looking for `audit.csv` _or_ `audit.csv.enc`?

lognaturel: All files with an encrypted submission including the submission xml are supposed to have a .enc suffix. So no need to look for audit.csv without suffix if the suffix isn’t explicitly stripped at that point.

issa-tseng (approved):

PR #409, opened 2021-09-30
bug: fix csv parsing options to allow ragged line lengths.

issa-tseng:
* also fix the name of this other option that i guess is now or always
  was (or maybe it likes both?) underscores and not camels.

### DIFF
--- a/lib/data/client-audits.js
+++ b/lib/data/client-audits.js
@@ -16,7 +16,7 @@ const { zipPart } = require('../util/zip');
 const headers = [ 'event', 'node', 'start', 'end', 'latitude', 'longitude', 'accuracy', 'old-value', 'new-value' ];
 
 // used by parseClientAudits below.
-const parseOptions = { bom: true, trim: true, skipEmptyLines: true, relax: true };
+const parseOptions = { bom: true, trim: true, skip_empty_lines: true, relax_column_count: true };
 const headerLookup = {};
 for (const header of headers) headerLookup[header] = true;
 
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

matthew-white:
Thinking more about it, I'm thinking that it probably makes sense to also update the submission attachment worker, since the worker will call `parseClientAudits()` on the content of any attachment marked `isClientAudit`. If a client audit log is encrypted, it can't be parsed, so it seems best not to call `parseClientAudits()` at all in that case. That seems especially true now that `parseClientAudits()` is more relaxed about input: is it possible that `parseClientAudits()` would now not throw when given an encrypted client audit log and would manage to create a `ClientAudit`?

As background, we saw a Sentry error related to `parseClientAudits()`: https://sentry.io/organizations/getodk/issues/2683905482/?project=5724763. There are a few things that make it seem like the error's coming from the worker, not from downloading submissions:

- I used the server audit log to see which submissions were downloaded around the time of the error. I did the download myself and saw that there was no new Sentry error.
- The Sentry error is missing a lot of context, for example, there's no URL.
- There are exactly 5 Sentry errors, each separated by about 10 minutes. That matches the behavior of a Backend worker.

matthew-white:
The submission attachment worker will be updated as part of #416, so I'm going to go ahead and merge this PR.

And just to note for posterity, we had a different Sentry error that also seemed to be from a worker: https://sentry.io/organizations/getodk/issues/2703129106/?project=5724763. The stack trace is cut off before we can see which worker threw the error (and to address that, we've increased the stack trace limit in #413). However, I think it's possible that the error is the submission attachment worker attempting to save the contents of an encrypted client audit log. A quick search indicates that \u0000 is a common culprit behind the error, and if an encrypted client audit log got through `parseClientAudits()` and then was attempted to be saved, its `remainder` `jsonb` column could contain \u0000. But again, that case should be addressed in #416.

PR #413, opened 2021-10-09
improve: increase stack trace limit.

matthew-white:
Making this change because we've encountered errors in Sentry whose stack trace ends before it indicates what line in Central triggered the error. Example: https://sentry.io/organizations/getodk/issues/2703129106/?project=5724763

### DIFF
--- a/lib/bin/run-server.js
+++ b/lib/bin/run-server.js
@@ -32,8 +32,11 @@ const mail = mailer(merge(config.get('default.email'), { env }));
 const googler = require('../external/google');
 const google = googler(config.get('default.external.google'));
 
-// get a sentry and xlsform client, and a password module.
+// get a sentry and configure errors.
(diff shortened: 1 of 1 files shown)


lib/bin/run-server.js
```diff
@@ -32,8 +32,11 @@ const mail = mailer(merge(config.get('default.email'), { env }));
 const googler = require('../external/google');
 const google = googler(config.get('default.external.google'));
 
-// get a sentry and xlsform client, and a password module.
+// get a sentry and configure errors.
 const Sentry = require('../external/sentry').init(config.get('default.external.sentry'));
+Error.stackTrackLimit = 20;
```
matthew-white, line 37: The default `stackTraceLimit` is 10, so I bumped it to 20. See https://nodejs.org/docs/latest-v14.x/api/errors.html#errors_error_stacktracelimit

PR #415, opened 2021-10-13
Update API docs about v1.3

matthew-white:
The main addition of this PR is an entry in the changelog for v1.3. I also added `analytics` to the list of audit log actions.

In both cases, it felt easiest to use the language of usage reporting. For consistency, I replaced most existing references to "analytics" with "usage reporting". I don't think the API paths related to usage reporting should ever change, but I think it helps to otherwise use the language of usage reporting.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -3747,6 +3747,7 @@ Server Audit Logs entries are created for the following `action`s:
 * `submission.attachment.update` when a Submission Attachment binary is set or cleared, but _only via the REST API_. Attachments created alongside the submission over the OpenRosa `/submission` API (including submissions from Collect) do not generate audit log entries.
 * `config.set` when a system configuration is set.
 * `backup` when a backup operation is attempted.
+* `analytics` when a Usage Report is attempted.
 
 ### Getting Audit Log Entries [GET /v1/audits{?action,start,end,limit,offset}]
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -26,12 +26,41 @@ Forms and their submissions are also accessible through two **open standards spe
...
+  * A `form.submissions.export` action will be logged when a User exports Form Submissions to CSV.
+* The Submission update timestamp is now included in OData (as `__system/updatedAt`). Resources that accept the `$filter` query parameter can be filered on `__system/updatedAt`.
+* All groups are now included in OData, even if they are not relevant. For more information, see [this post](https://forum.getodk.org/t/include-non-relevant-groups-and-fields-in-odk-central-api-responses/33536) in the ODK Forum.
+
+**Fixed**:
+
+* An XLSForm is now available in more cases:
+  * Previously, if an XLSForm was uploaded as a Draft Form, then the version string was changed as the Draft was published, the XLSForm would not be available.
+  * Also, if you created a Draft Form without providing a definition (thereby copying the published definition), the XLSForm would not be available on the Draft.
+  * The XLSForm is now available in both cases.
```
issa-tseng, line 58: i think omit. we don't talk about bugfixes here much.

docs/api.md
```diff
@@ -26,12 +26,41 @@ Forms and their submissions are also accessible through two **open standards spe
...
+* The `Content-Disposition` header now specifies the `filename*` parameter, allowing filenames to contain Unicode.
+* When exporting Form Submissions to CSV under Project Managed Encryption, a decrypted combined client audit log will now be exported.
+* Previously, ODK Central could hang if an incorrect passphrase was provided for decryption. This has been fixed.
+* If a Form definition contains an invalid `search()` appearance, an informative 400 error will be returned rather than a 500 error.
+
```
issa-tseng, line 63: probably also omit

issa-tseng (approved):

issa-tseng:
you'll see that when i've used the "fixed:" heading in the past it was actually fixing bugs _in_ the documentation, not fixing problems w the api.

if we made an api change i call it a change. if we fixed something it just should have worked.

yanokwa:
Yes, I'd like to change the API paths too, but is that reasonable to do at this stage? I suppose better now than later?

issa-tseng:
if the docs are wrong anyway ?

PR #416, opened 2021-10-15
v1.3 patches

issa-tseng:


matthew-white (commented):
The commits so far look 💯!

Submission CSV files will now be square. @yanokwa also mentioned seeing a combined client audit log CSV file that was ragged, which wouldn't change with this PR. @yanokwa, do you have an example of that handy?

matthew-white (approved):

PR #418, opened 2021-10-19
bug: return to supporting quote inside unquoted field of client audit.

matthew-white:
We used to support that case, but it was unintentionally removed in 62df8a770b05a87e8d20e686d86d28db540f0d68.


issa-tseng:
oops oops i knew there was something else i was supposed to do

PR #421, opened 2021-10-21
bug: don't log password hashes on cli user create. (+backfill)

issa-tseng:


matthew-white (commented):
I tested this PR on the QA server. I can see that:

- `odk-cmd user-create` does not log the password hash.
- Past log entries no longer have a `password` property.

However, I think there's still one case to consider, which is that POST /v1/users accepts an optional `password` field. If `password` is provided, then `Users.withPassword()` is called before `Users.create()`, which I think means that the password hash will be logged. Actually, I think that that's the last place where `Users.withPassword()` is used. Maybe it could be replaced with what's done in this PR: `Users.create()` followed by `Users.updatePassword()`? Or alternatively, `Users.create.audit` could omit the `password` property?

I don't know how many API users are using that optional `password` field. Since we don't use it in Frontend, I'd completely forgotten that it was even possible. Part of me thinks that we should handle that case later, but that'd mean having to run this migration again later.

Side note: I'm noticing that `Users.create.audit` only logs fields from the `users` table and not from the `actors` table. That means that the audit log doesn't log the user's initial display name. I don't think that matters much, but thought I'd mention it.

issa-tseng:
@matthew-white i think this can be rereviewed

issa-tseng:
@matthew-white still i think ready for re-review.

lib/resources/users.js
```diff
@@ -33,13 +33,13 @@ module.exports = (service, endpoint) => {
 
   service.post('/users', endpoint(({ Users, mail }, { body, auth }) =>
     auth.canOrReject('user.create', User.species)
-      // here we manually reincorporate the given password as it's allowed to
-      // be set upon user creation (but not update).
-      .then(() => Users.withPassword(User.fromApi(body), body.password))
-      .then((user) => user.forV1OnlyCopyEmailToDisplayName())
+      .then((user) => User.fromApi(body).forV1OnlyCopyEmailToDisplayName())
```
matthew-white, line 36: I think `user` is no longer used (and possibly doesn't exist).

lib/model/migrations/20211021-remove-hashes-from-audits.js
```diff
@@ -0,0 +1,18 @@
+// Copyright 2021 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/opendatakit/central-backend/blob/master/NOTICE.
```
matthew-white, line 3: One last opendatakit→getodk I think

matthew-white (approved):
Very sorry for the long-delayed review! This is looking good. 🚀

I still think that it'd be useful to test that POST /v1/users?password doesn't log the password hash: this PR tests that the task doesn't log the password hash, and I'm thinking that it'd be good to test that for the endpoint as well.

I also wonder whether it'd be helpful to remove `withPassword()`, which is no longer used anywhere. It's a little hard to imagine how `withPassword()` could be used without the password hash getting logged. My instinct is that it'd be safest to remove `withPassword()` or to add a comment above it that warns about that possibility.

PR #423, opened 2021-10-24
Update dependencies

x3ro:
Hey folks,

updating Sentry sort of escalated and led me to checking which other dependencies I could painlessly update without breaking the tests. I ran `make test-full` locally and it passed, but wasn't sure whether there's more automated testing I could do.

If this is somehow undesirable feel free to let me know. I can also cut this into smaller chunks.

Cheers
Lucas

PS: Would it potentially make sense to enable something like [dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates) for this repo?


issa-tseng:
hey @x3ro , thanks for this pr too. usually we do this just before we cut a release because node security breaks so much it's a waste of time to do it now, we'll just have to do it again later.

that said, if you'd still like it merged here are my notes:

as noted in the other one, i'd prefer sticking to the convention `~` that already exists in this project, rather than change some of them to `^` and leave the rest `~`. it's already hard to reason about the one operator.

the if statements you added don't fit the code style of the project. we leave a space after `if`.

the commit message does not meet [our guidelines](https://github.com/getodk/central-backend/blob/master/CONTRIBUTING.md#guidelines). it would also be nice if the additional code changes that aren't updating dependencies are noted.

issa-tseng:
closing to inactivity. please feel free to reopen if you have a new patch.

PR #425, opened 2021-11-02
deps: upgrade prompt to remove warning (thanks @x3ro for identifying).

issa-tseng:
* also upgrade some test deps for security issues.


matthew-white:
Rebased so that we can release off the commit here.

matthew-white (approved):

PR #426, opened 2021-11-05
Delete API that fetches list of trashed forms and lets trashed forms be restored

ktuite:
This exposes two new endpoints:

1. `/projects/:projectId/deletedForms` that fetches list of deleted forms for display in the frontend (in a trash section below the rest of a project's forms). All the same info about a form is there as well as an additional `deletedAt` field.

2. `/projects/:projectId/forms/:id/restore` that restores a soft-deleted form with a given (numeric) form id.

There are two quirks of this `restore` endpoint:
1. While you can (soft-)delete a form and then make a new one with the same `xmlFormId`, you can't restore the deleted form while an active form of the same `xmlFormId` exists. This will be prevented and explained in both the backend API and the frontend.

~2. It's possible to have multiple deleted forms with the same `xmlFormId`, but different integer form ids in the database, in which case the above restore endpoint is ambiguous and wont work.~ We resolved this by making the restore endpoint take the integer id of the form, which is now exposed through the API. 

Screen Shot 2021-11-09 at 9 54 25 AM

### DIFF
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -55,6 +55,7 @@ class Form extends Frame.define(
   'draftDefId',                         'enketoId',     readable,
   'enketoOnceId', readable,             'acteeId',
   'createdAt',    readable,             'updatedAt',    readable,
+  'deletedAt',
   embedded('createdBy'), embedded('publishedBy')
 ) {
   get def() { return this.aux.def; }
@@ -83,7 +84,10 @@ class Form extends Frame.define(
     const enketoOnceId = (this.def.id === this.currentDefId) ? this.enketoOnceId : null;
     /* eslint-enable indent */
 
-    return Object.assign(Frame.prototype.forApi.call(this), { enketoId, enketoOnceId });
+    // Include deletedAt in response only if it is not null (most likely on resources about soft-deleted forms)
+    const deletedAtField = this.deletedAt ? { deletedAt: this.deletedAt } : null;
+
+    return Object.assign(Frame.prototype.forApi.call(this), { enketoId, enketoOnceId }, deletedAtField);
   }
 
   // Given an XML string, returns Promise[Object]. If the Promise rejects, the XML
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -49,7 +49,7 @@ const AllVersions = Symbol('all versions');
 
 class Form extends Frame.define(
   table('forms'),
-  'id',           readable,             'projectId',    readable,
+  'id',                                 'projectId',    readable,
   'xmlFormId',    readable,             'state',        readable, writable,
   'currentDefId',
   'draftDefId',                         'enketoId',     readable,
@@ -85,9 +85,10 @@ class Form extends Frame.define(
     /* eslint-enable indent */
 
     // Include deletedAt in response only if it is not null (most likely on resources about soft-deleted forms)
-    const deletedAtField = this.deletedAt ? { deletedAt: this.deletedAt } : null;
+    // and also include the numeric form id (used to restore)
+    const deletedAtIdFields = this.deletedAt ? { deletedAt: this.deletedAt, id: this.id } : null;
 
-    return Object.assign(Frame.prototype.forApi.call(this), { enketoId, enketoOnceId }, deletedAtField);
+    return Object.assign(Frame.prototype.forApi.call(this), { enketoId, enketoOnceId }, deletedAtIdFields);
   }
 
   // Given an XML string, returns Promise[Object]. If the Promise rejects, the XML
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -49,7 +49,7 @@ const AllVersions = Symbol('all versions');
 
 class Form extends Frame.define(
   table('forms'),
-  'id',                                 'projectId',    readable,
+  'id',           readable,             'projectId',    readable,
   'xmlFormId',    readable,             'state',        readable, writable,
   'currentDefId',
   'draftDefId',                         'enketoId',     readable,
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -316,7 +316,8 @@ const getByProjectId = (projectId, xml, version, options = QueryOptions.none, de
   _get(all, options.withCondition({ projectId }), xml, version, deleted);
 const getByProjectAndXmlFormId = (projectId, xmlFormId, xml, version, options = QueryOptions.none, deleted = false) => ({ maybeOne }) =>
   _get(maybeOne, options.withCondition({ projectId, xmlFormId }), xml, version, deleted);
-
+const getByProjectAndNumericId = (projectId, id, xml, version, options = QueryOptions.none, deleted = false) => ({ maybeOne }) =>
+  _get(maybeOne, options.withCondition({ projectId, 'forms.id': id }), xml, version, deleted);
 
 ////////////////////////////////////////////////////////////////////////////////
 // SCHEMA
@@ -387,7 +388,8 @@ module.exports = {
   _update, update, _updateDef, del, restore,
   setManagedKey,
   getByAuthForOpenRosa,
-  getVersions, getByActeeIdForUpdate, getByActeeId, getByProjectId, getByProjectAndXmlFormId,
+  getVersions, getByActeeIdForUpdate, getByActeeId,
+  getByProjectId, getByProjectAndXmlFormId, getByProjectAndNumericId,
   getFields, getBinaryFields, getStructuralFields, getMergedFields,
   lockDefs, getAllSubmitters
 };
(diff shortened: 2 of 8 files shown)


lib/model/frames/form.js
```diff
@@ -83,7 +84,10 @@ class Form extends Frame.define(
     const enketoOnceId = (this.def.id === this.currentDefId) ? this.enketoOnceId : null;
     /* eslint-enable indent */
 
-    return Object.assign(Frame.prototype.forApi.call(this), { enketoId, enketoOnceId });
+    // Include deletedAt in response only if it is not null (most likely on resources about soft-deleted forms)
+    const deletedAtField = this.deletedAt ? { deletedAt: this.deletedAt } : null;
```
ktuite, line 88: This feels weird but I wasn't sure how else to do this (expose a certain field only in some cases - in particular if it is not null). I also tried it the frame extender way but I ran into ambiguous database column situations and i don't understand the frames stuff well enough to get around it. Thoughts, @issa-tseng?

lib/model/query/forms.js
```diff
@@ -172,6 +172,14 @@ const del = (form) => ({ run, Assignments }) =>
   Promise.all([ run(markDeleted(form)), Assignments.revokeByActeeId(form.acteeId) ]);
 del.audit = (form) => (log) => log('form.delete', form);
 
+const restore = (form) => ({ run, Forms }) =>
+  Forms.getByProjectAndXmlFormId(form.projectId, form.xmlFormId)
+    .then((activeForm) => (activeForm.isDefined()
+      ? reject(Problem.user.uniquenessViolation({ fields: 'xmlFormId', values: form.xmlFormId }))
+      : run(markUndeleted(form))));
+restore.audit = (form) => (log) => log('form.restore', form);
```
matthew-white, line 180: I think it makes sense for there to be a new audit action. For it to show up in Frontend, it'll also need to be added to the nonverbose list in `lib/model/query/audits.js`. (Frontend will also need to be modified to display the action nicely in the audit log and to show it in the filter.)

lib/model/query/forms.js
```diff
@@ -294,16 +296,16 @@ ${extend|| sql`
   left outer join actors on audits."actorId"=actors.id
   left outer join (select id, "contentType" as "excelContentType" from blobs) as xls
     on form_defs."xlsBlobId"=xls.id`}
-where ${equals(options.condition)} and forms."deletedAt" is null
+where ${equals(options.condition)} and ${filterDeleted(deleted)}
```
ktuite, line 307: inline this function and make it shorter

lib/resources/forms.js
```diff
@@ -40,6 +40,12 @@ module.exports = (service, endpoint) => {
       .then((project) => auth.canOrReject('form.list', project))
       .then((project) => Forms.getByProjectId(project.id, false, undefined, queryOptions))));
 
+  service.get('/projects/:projectId/deletedForms', endpoint(({ Forms, Projects }, { auth, params, queryOptions }) =>
```
ktuite, line 43: consider making a query parameter on /forms above

lib/model/query/forms.js
```diff
@@ -172,6 +172,14 @@ const del = (form) => ({ run, Assignments }) =>
   Promise.all([ run(markDeleted(form)), Assignments.revokeByActeeId(form.acteeId) ]);
 del.audit = (form) => (log) => log('form.delete', form);
 
+const restore = (form) => ({ run, Forms }) =>
+  Forms.getByProjectAndXmlFormId(form.projectId, form.xmlFormId)
+    .then((activeForm) => (activeForm.isDefined()
+      ? reject(Problem.user.uniquenessViolation({ fields: 'xmlFormId', values: form.xmlFormId }))
```
ktuite, line 178: DB should catch this error itself so catch it there and use problem/translate thing to make useful error

lib/model/query/forms.js
ktuite: yes, will add this verb for admins and project managers

lib/resources/forms.js
```diff
@@ -268,6 +268,20 @@ module.exports = (service, endpoint) => {
       .then((form) => Forms.getVersions(form.id, queryOptions))));
 
 
+  ////////////////////////////////////////
+  // RESTORE / UNDELETE
+
+  service.post('/projects/:projectId/forms/:id/restore', endpoint(({ Forms }, { auth, params }) =>
+    Forms.getByProjectAndXmlFormId(params.projectId, params.id, false, null, QueryOptions.none, true)
```
ktuite: do a permissions check

test/integration/api/forms.js
```diff
@@ -1428,6 +1380,116 @@ describe('api: /projects/:id/forms', () => {
...
+            body.forEach((form) => form.hasOwnProperty('deletedAt').should.be.false());
+          })
+          .then(() => asAlice.get('/v1/projects/1/forms')
+            .expect(200)
+            .then(({ body }) => {
+              body.forEach((form) => form.hasOwnProperty('deletedAt').should.be.false());
+            })))));
+  });
+
+  describe('/:id/restore (undeleting trashed forms)', () => {
```
ktuite, line 1438: note to future selves @ktuite and @issa-tseng to break these painfully long test files into more files

lib/resources/forms.js
```diff
@@ -268,6 +268,20 @@ module.exports = (service, endpoint) => {
...
+      .then(getOrNotFound)
+      .catch((err) => ((err.name === 'DataIntegrityError')
+        ? Promise.reject(Problem.internal.notImplemented({ feature: 'restoring ambiguous xmlFormId' }))
+        : Promise.reject(err)))
+      .then((form) => auth.canOrReject('form.delete', form))
```
ktuite, line 281: move this permission check up: fetch it and immediately check auth 

form.delete --> form.restore

test/integration/api/forms.js
```diff
@@ -1428,6 +1380,116 @@ describe('api: /projects/:id/forms', () => {
...
+          .then(() => asAlice.post('/v1/projects/1/forms/')
+            .send(testData.forms.simple.replace('id="simple"', 'id="simple" version="two"'))
+            .set('Content-Type', 'application/xml')
+            .expect(200))
+          .then(() => asAlice.delete('/v1/projects/1/forms/simple')
+            .expect(200))
+          .then(() => asAlice.post('/v1/projects/1/forms/simple/restore')
+            .expect(501)
+            .then(({ body }) => {
+              body.code.should.equal(501.1);
```
ktuite: usually this would be a 409 conflict instead

lib/resources/forms.js
```diff
@@ -271,8 +271,9 @@ module.exports = (service, endpoint) => {
   ////////////////////////////////////////
   // RESTORE / UNDELETE
 
+  // Instead of the xmlFormId, this endpoint uses the numeric form id in the database
   service.post('/projects/:projectId/forms/:id/restore', endpoint(({ Forms }, { auth, params }) =>
-    Forms.getByProjectAndXmlFormId(params.projectId, params.id, false, null, QueryOptions.none, true)
+    Forms.getByProjectAndNumericId(params.projectId, params.id, false, null, QueryOptions.none, true)
```
ktuite, line 276: Broader suggestion: make all form resources able to take either numeric form id or xml form id

Maybe for now: just return form int id with deleted forms so it can be used here

lib/resources/forms.js
ktuite: remove this problem

ktuite (commented):
interactive review with @issa-tseng

issa-tseng (approved):

lib/model/query/forms.js
```diff
@@ -172,6 +172,11 @@ const del = (form) => ({ run, Assignments }) =>
   Promise.all([ run(markDeleted(form)), Assignments.revokeByActeeId(form.acteeId) ]);
 del.audit = (form) => (log) => log('form.delete', form);
 
+const restore = (form) => ({ run, Forms }) =>
+  run(markUndeleted(form));
+restore.audit = (form) => (log) => log('form.restore', form);
```
ktuite, line 177: add form.restore to https://github.com/getodk/central-backend/blob/master/lib/model/query/audits.js#L45-L46 to show up on the frontend

test/integration/api/forms.js
```diff
@@ -1380,6 +1380,137 @@ describe('api: /projects/:id/forms', () => {
...
+          .expect(200)
+          .then(() => service.login('chelsea', (asChelsea) =>
+            asChelsea.post('/v1/projects/1/forms/1/restore').expect(403))))));
+
+    it('should reject restoring if referenced by the xml form id', testService((service) =>
+      service.login('alice', (asAlice) =>
+        asAlice.delete('/v1/projects/1/forms/simple')
+          .expect(200)
+          .then(() => asAlice.post('/v1/projects/1/forms/simple/restore')
+            .expect(400))))); // bad request
```
ktuite, line 1451: clarify that this is 400.11 `message: "Invalid input data type: expected 'simple' to be (type integer)",`

ktuite (commented):
notes from review with @matthew-white

matthew-white (approved):

PR #428, opened 2021-11-11
improve: add knexfile.js for Knex migration CLI.

matthew-white:
We want to be able to manage Knex migrations from the command line: that will facilitate troubleshooting when a migration fails in production and will also aid local development. In particular, we want to be able to list completed and pending migrations, and we want to be able to run one migration at a time.

We thought about writing our own CLI, but I think we can actually benefit from the built-in Knex [migration CLI](https://knexjs.org/#Migrations-CLI). The Knex CLI is not easy to set up, but once it's set up, it seems to work nicely. The output is clear, and in contrast to the JavaScript API, the CLI automatically outputs the stack trace if there is a migration failure. There is also a `--debug` flag that seems potentially useful. The CLI also allows you to run a specific migration.

The best way to set up the Knex CLI seems to be to use a [`knexfile.js`](https://knexjs.org/#knexfile), which this PR adds.

I've tested this out locally and on a testing server.

### DIFF
--- a/lib/model/knexfile.js
+++ b/lib/model/knexfile.js
@@ -1,0 +1,33 @@
+// Copyright 2021 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+/*
+This file is for use with the Knex migration CLI. Run `npx knex` from the
+repository root, specifying the path of this file to --knexfile. Because the
+Knex CLI will change the working directory to the directory of this file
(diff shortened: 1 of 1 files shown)


lib/model/knexfile.js
```diff
@@ -0,0 +1,33 @@
...
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+/*
+This file is for use with the Knex migration CLI. Run `npx knex` from the
+repository root, specifying the path of this file to --knexfile. Because the
+Knex CLI will change the working directory to the directory of this file
+(lib/model), you will also need to specify the NODE_CONFIG_DIR environment
```
matthew-white, line 14: This environment variable is documented here: https://github.com/lorenwest/node-config/wiki/Environment-Variables#node_config_dir

matthew-white: As an alternative to setting `$NODE_CONFIG_DIR`, I also tried using the `--cwd` option of the Knex CLI. However, that didn't work: the working directory was still changed to `lib/model` before `config` was `require`d.

issa-tseng (approved):

PR #429, opened 2021-11-12
bug: set highWaterMark to prevent migration from being killed.

matthew-white:


issa-tseng (approved):

matthew-white:
Unfortunately, this approach does not seem to work after all. When I try this PR on the problem server, the migration fails at some form def, sometimes after the 100th form def and sometimes before. (I've tried a few different `highWaterMark`s, and that value affects which def it is at which the migration fails.) I thought this approach was working, but I think I probably also commented out `work.push()` at some point:

```js
//work.push(db.update(data).into('form_defs').where({ id: def.id }));
```

I think I must have commented it out, because when I try the code here now, it succeeds if `work.push()` is commented out and fails if it is not. More specifically:

- If you run the migration without this PR, the migration fails after exactly the 100th form def.
- If you simply comment out `work.push()`, the migration still fails after exactly the 100th form def.
- If you comment out `work.push()` _and_ set `highWaterMark` to a sufficiently low number, then the migration completes.
- However, if you simply set `highWaterMark`, as in this PR, then you get the behavior described above: the migration fails.

It's very surprising to me that `work.push()` has any effect here. I tried a few variations on that line and observed the following:

```js
// The original line doesn't work:
const data = { name: partial.def.name };
work.push(db.update(data).into('form_defs').where({ id: def.id }));

// It also doesn't work if you remove the reference to db:
work.push([def.id, partial.def.name]);

// Or even if you only push the name:
work.push(partial.def.name);

// Making work a Map doesn't help:
work.set(def.id, partial.def.name);

// Destructuring beforehand doesn't work, even though that removes all object
// references:
for await (const { id, xml } of db.select('*').from('form_defs').stream({ highWaterMark: 10 })) {
  const { def: name } = await Form.fromXml(xml);
  if (name == null) continue;
  work.push([id, name]);
}

// The loop does complete if you only push literals onto work:
work.push(0);
// or
work.push('Some name');
// So it has to be something more than the reference to work that's the issue.
```

I double-checked, and `typeof def.id` and `typeof partial.def.name` are `number` and `string`, respectively. I also checked the size of `partial.def.name` in case there was somehow a very lengthy title, and all titles were short.

Again, the loop does complete if the reference to `work` is commented out, and it makes progress at a steady pace. That makes me think that this can't be explained by an issue with `Form.fromXml()` or `stream()`.

Setting `highWaterMark` to 1 was not successful. (It may even have made things worse?) I also tried using `readable` events, setting `highWaterMark` to 5, but that was likewise not successful.

This feels like a garbage collection issue to me: somehow the XML strings (at an average of 1.5 MB apiece) are not being cleaned up. I read some about memory leaks in Node, including [this](https://blog.meteor.com/an-interesting-kind-of-javascript-memory-leak-8b47d2e7f156) interesting article. However, most discussion did not seem relevant. The following Node issue does seem similar: nodejs/node/issues/30298. Interestingly, that issue mentions both `for await` and streams. I also tried to produce a minimal reproducible example to try to understand what's happening here and maybe report it, but I wasn't able to easily reproduce outside the migration.

I'm considering an alternative approach using the Postgres XML data type that would avoid the `for await` and the use of `work` entirely. At this point, looking into that alternative approach feels like a better use of time than continuing to investigate why it is that this implementation isn't working.

matthew-white:
Closing in favor of #430, which was successful on the problem server.

PR #430, opened 2021-11-14
bug: prevent migration from being killed by using Postgres xml type.

matthew-white:
A migration is being killed in rare cases. I tried to patch the migration's JavaScript in #429, but ran into issues. This PR tries to fix the issue by taking a different approach, replacing the JavaScript with a single Postgres statement. That Postgres statement makes use of the Postgres `xml` data type. I tried this PR on the problem server, and with it, the migration was able to complete (after about 30 seconds). 🎉 It seems that Postgres is better able to manage the resources around parsing the relatively large XForms on the server.

I'll also add comments about specific lines below.

References:

https://www.postgresql.org/docs/9.6/datatype-xml.html
https://www.postgresql.org/docs/9.6/functions-xml.html

### DIFF
--- a/lib/model/migrations/20210423-01-add-name-to-form-def.js
+++ b/lib/model/migrations/20210423-01-add-name-to-form-def.js
@@ -19,17 +19,12 @@ const up = async (db) => {
   });
 
   await db.raw('ALTER TABLE form_defs DISABLE TRIGGER check_managed_key');
-
-  const work = [];
-  for await (const def of db.select('*').from('form_defs').stream()) {
-    const partial = await Form.fromXml(def.xml);
-    if (partial.def.name == null) continue;
-
-    const data = { name: partial.def.name };
-    work.push(db.update(data).into('form_defs').where({ id: def.id }));
-  }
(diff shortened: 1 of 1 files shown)


lib/model/migrations/20210423-01-add-name-to-form-def.js
```diff
@@ -19,17 +19,12 @@ const up = async (db) => {
...
-    const data = { name: partial.def.name };
-    work.push(db.update(data).into('form_defs').where({ id: def.id }));
-  }
-  await Promise.all(work);
-
+  await db.raw(`update form_defs
+set name = (xpath(
+  '/*[local-name() = ''html'']/*[local-name() = ''head'']/*[local-name() = ''title'']/text()',
+  xml::xml
+))[1]::text
```
matthew-white, line 26: `xpath()` returns an array of `xml` values. For an ODK XForm, the resulting array should either have one element or (if the form has no title) be empty. `[1]` will return `null` if the array is empty.

issa-tseng (approved):

PR #432, opened 2021-11-16
track and return more of: formVersion, deviceId, userAgent

issa-tseng:
the goal here is to do the following:

1. return an extra `FormVersion`/`formVersion` field on csv/odata export. this field reflects only the original "root" version of a submission, it does not update if the submission is edited.
2. return an extra `formVersion` field on extended submission version request. ditto it's the root version.
3. track and return `userAgent` for each submission version.
4. track and return `deviceId` for each submission version.

this pr does _not_ deprecate the existing `deviceId` on the _logical_ submission, which will also always reflect the initial submission. it only adds an _additional_ `deviceId` to version requests.

this pr does _not_ add `userAgent` to submission logical requests. maybe it should.

### DIFF
--- a/test/integration/other/encryption.js
+++ b/test/integration/other/encryption.js
@@ -215,6 +215,7 @@ describe('managed encryption', () => {
           .then((keyId) => new Promise((done) =>
             zipStreamToFiles(asAlice.get(`/v1/projects/1/forms/simple/submissions.csv.zip?${keyId}=supersecret`), (result) => {
               result.filenames.should.eql([ 'simple.csv' ]);
+              console.log(result['simple.csv']);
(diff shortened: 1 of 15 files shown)


test/integration/other/encryption.js
```diff
@@ -215,6 +215,7 @@ describe('managed encryption', () => {
           .then((keyId) => new Promise((done) =>
             zipStreamToFiles(asAlice.get(`/v1/projects/1/forms/simple/submissions.csv.zip?${keyId}=supersecret`), (result) => {
               result.filenames.should.eql([ 'simple.csv' ]);
+              console.log(result['simple.csv']);
```
issa-tseng: rm

ktuite (approved):

Issue #388, opened 2021-07-22
Form preview tasks unexpectedly stop running


yanokwa:
I've seen this issue on two servers now. Users report that the preview buttons are not enabled. 

An investigation of the service logs show that `start/finish processing event` are not present in the logs. A `docker-compose stop && docker-compose up -d` fixes it and the logs then immediately show the correct events.

Both servers are running v1.2.x, but I don't know if this is a regression. I haven't looked closely, but this doesn't seem associated with a big spike in CPU/RAM or anything like that.

matthew-white:
@yanokwa, can you tell whether the fix for v1.4 resolved this issue? If it's unclear, maybe we close for now and reopen if something comes up?

yanokwa:
I haven't experienced the problem recently, so I think it's safe to close for now.

PR #436, opened 2021-12-02
improve: worker resiliency, hopefully.

issa-tseng:
* one try/catch we didn't have before, around checker invocation.
* one .catch() we didn't have before, around checker promise result.
* watchdog timer which kicks the loop if it looks stale.


issa-tseng:
resolves #388 hopefully.

ktuite (approved):

yanokwa:
Restored because I need to try this branch on a production server.

issa-tseng:
@yanokwa i just pushed an update to the branch now, adding some additional sentry logging.

PR #437, opened 2021-12-02
Fix to populating project viewer and edited submission counts

ktuite:
issa noticed that the counts for number of viewers and number of submissions in edited state was mysteriously always zero. Turns out I wasn't handing the data returned from the query correctly.

### DIFF
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -17,7 +17,7 @@ const _cutoffDate = sql`current_date - cast(${DAY_RANGE} as int)`;
 // Gets a pointer to the project (repeat) in the metrics report for a specific
 // project, creating it first from the project metrics template if necessary
 const _getProject = (projects, id) => {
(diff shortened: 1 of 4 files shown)


lib/model/query/analytics.js
```diff
@@ -17,7 +17,7 @@ const _cutoffDate = sql`current_date - cast(${DAY_RANGE} as int)`;
 // Gets a pointer to the project (repeat) in the metrics report for a specific
 // project, creating it first from the project metrics template if necessary
 const _getProject = (projects, id) => {
-  if (!Object.hasOwnProperty.call(projects, id)) {
+  if (!projects[id]) {
```
issa-tseng, line 20: we make all our checks explicit in -backend (and i think frontend too), so this needs `!= null` rather than just the negation. js has a lot of bad behavior around implied falsiness.

issa-tseng (approved):

PR #451, opened 2022-02-05
bug: need to flag all upgrade-form.purged audits as processed.

issa-tseng:
* otherwise the workers try to pick it up.
* maybe the workers should be smarter.

### DIFF
--- a/lib/model/migrations/20220121-02-purge-deleted-forms.js
+++ b/lib/model/migrations/20220121-02-purge-deleted-forms.js
@@ -30,8 +30,8 @@ with redacted_audits as (
     where audits."acteeId" = forms."acteeId"
     and forms."deletedAt" is not null
   ), purge_audits as (
-    insert into audits ("action", "acteeId", "loggedAt")
-    select 'form.purge', "acteeId",  clock_timestamp()
+    insert into audits ("action", "acteeId", "loggedAt", "processed")
+    select 'form.purge', "acteeId", clock_timestamp(), clock_timestamp()
     from forms
     where forms."deletedAt" is not null
   ), update_actees as (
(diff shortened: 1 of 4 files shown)


lib/model/migrations/20220121-02-purge-deleted-forms.js
```diff
@@ -30,8 +30,8 @@ with redacted_audits as (
     where audits."acteeId" = forms."acteeId"
     and forms."deletedAt" is not null
   ), purge_audits as (
-    insert into audits ("action", "acteeId", "loggedAt")
-    select 'form.purge', "acteeId",  clock_timestamp()
+    insert into audits ("action", "acteeId", "loggedAt", "processed")
+    select 'form.purge', "acteeId", clock_timestamp(), clock_timestamp()
```
ktuite, line 34: if this processed flag is getting set here, it also probably needs to get set in Forms.purge()

ktuite: yeah, this `insert into audits` is in that file, too. it's not the normal auditing functionality because it updates the audit log for a bunch of deleted forms at once.

matthew-white:
We were considering updating Slonik after noticing this warning in a log:

```
(node:51) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 notice listeners added to [Client]
. Use emitter.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
```

This seems related to gajus/slonik#167. (When I Googled the part of the warning about `[Client]`, that issue showed up right away.) That issue was resolved in Slonik 23.6.3; we're on 23.6.0.

However, @issa-tseng noticed some issues after updating Slonik. From Slack:

> i tried upgrading only slonik and it's also timing a ton of queries out

> okay. if i don't upgrade slonik the tests pass perfectly. so somewhere between 23.6.0 and 27.1 is a performance regression. it's probably to do with streams, since the tests fail a little too consistently right at the start of the odata tests

Because of that, and because it's not clear that the warning is actually causing any problems, we've decided not to update Slonik in v1.4.

matthew-white (approved):

PR #452, opened 2022-02-07
Addressing xlsBlobId issue in form and blob purging

ktuite:
This error came up while applying the form purging migration on the QA server on Feb 2, 2022, and it turns out to be an issue in the regular blob purging SQL as well. That code checks for blobs with no references and tries to delete them, but it wasn't considering the reference to `xlsBlobId` in the `form_defs` table, so it was trying (and failing) to delete blobs that shouldn't have been deleted. 

      service_1             | migration file "20220121-02-purge-deleted-forms.js" failed
      service_1             | migration failed with error:
      service_1             | delete from blobs
      service_1             |   using blobs as b
      service_1             |   left join client_audits as ca on ca."blobId" = b.id
      service_1             |   left join submission_attachments as sa on sa."blobId" = b.id
      service_1             |   left join form_attachments as fa on fa."blobId" = b.id
      service_1             | where (blobs.id = b.id and
      service_1             |   ca."blobId" is null and
      service_1             |   sa."blobId" is null and
      service_1             |   fa."blobId" is null) - update or delete on table "blobs" violates foreign key constraint "form_defs_xlsblobid_foreign" on table "form_defs"

### DIFF
--- a/lib/model/migrations/20220121-02-purge-deleted-forms.js
+++ b/lib/model/migrations/20220121-02-purge-deleted-forms.js
@@ -13,7 +13,15 @@
 // forms, but since there was no way to access forms deleted prior to this release, we
 // are removing old deleted forms.
 
-// Purging steps
+// NOTE: copypasta alert!
+// The following purge queries are roughly the same as in the form and blob
+// query modules, with the first roughly matching Forms.purge() (except where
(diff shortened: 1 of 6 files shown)


issa-tseng (approved):

issa-tseng:
only optional nits

PR #453, opened 2022-02-11
Purge unneeded form draft definitions

ktuite:
If a user does something that makes a particular Draft Form unneeded, it should be immediately hard-deleted from the database.

There are three scenarios for this:
1. Uploading a new Draft while there is currently a draft, which replaces the existing one
2. Enabling Project Managed Encryption, which replaces all existing drafts
3. Deleting a draft directly (only when there is a published version already on the form)

### DIFF
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -109,7 +109,8 @@ module.exports = (service, endpoint) => {
         .then((partial) => Promise.all([
           Forms.createVersion(partial, form, false),
           Submissions.clearDraftSubmissions(form.id)
-        ])))
+        ]))
+        .then(() => Forms.purgeUnneededDrafts(form))) // remove drafts made obsolete by new draft
       .then(success)));
 
   service.post('/projects/:projectId/forms/:id/draft/publish', endpoint(({ Forms, Submissions }, { params, auth, query }) =>
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -110,7 +110,7 @@ module.exports = (service, endpoint) => {
           Forms.createVersion(partial, form, false),
           Submissions.clearDraftSubmissions(form.id)
         ]))
-        .then(() => Forms.purgeUnneededDrafts(form))) // remove drafts made obsolete by new draft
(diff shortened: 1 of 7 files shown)


matthew-white:
I haven't taken a close look at the code, but I think it would be good to call `Forms.purgeUnneededDrafts()` when a form draft is deleted using the DELETE /v1/projects/:projectId/forms/:id/draft endpoint. That endpoint won't be used when a form doesn't have a published version: instead, the frontend will delete the form (DELETE /v1/projects/:projectId/forms/:id). However, if the draft is deleted for a form with a published version, I think it'd be good to then purge that draft. Looking at that endpoint, I see a TODO along those lines:

https://github.com/getodk/central-backend/blob/b23bad1029e43a4e51d6739b576b27058ba75adb/lib/resources/forms.js#L144-L149

lib/resources/forms.js
```diff
@@ -109,7 +109,8 @@ module.exports = (service, endpoint) => {
         .then((partial) => Promise.all([
           Forms.createVersion(partial, form, false),
           Submissions.clearDraftSubmissions(form.id)
-        ])))
+        ]))
+        .then(() => Forms.purgeUnneededDrafts(form))) // remove drafts made obsolete by new draft
```
ktuite, line 113: naming: maybe call this clear or hardDelete or something other than the other kind of purging on a cycle that we do

lib/resources/forms.js
```diff
@@ -142,8 +143,7 @@ module.exports = (service, endpoint) => {
       .then(rejectIf(((form) => form.currentDefId == null), noargs(Problem.user.noPublishedVersion)))
       .then(rejectIf(((form) => form.draftDefId == null), noargs(Problem.user.notFound)))
       .then((form) => Promise.all([
-        // TODO: for now we just cast away the draft to nothing. eventually probably sweep+delete.
-        Forms.clearDraft(form),
+        Forms.clearDraft(form).then(Forms.purgeUnneededDrafts(form)),
```
ktuite, line 146: () => Forms.purgeUnneededDrafts(forms)

ktuite (commented):
interactive review with issa

issa-tseng (approved):

PR #454, opened 2022-02-16
Disallow passwords that are too short

ktuite:
Complain if a user tries to set a password that is too short (<10 chars) on several API endpoints and in the central CLI. Includes creating and updating a user password and specifying a passphrase when enabling managed encryption on a project.


matthew-white (approved):

matthew-white:
Going to go ahead and merge so that we can get this out to QA.

PR #455, opened 2022-02-17
fix: use groupPaths query parameter in .../submissions.csv endpoints.

matthew-white:


matthew-white:
@ktuite, I think this can be reviewed async!

matthew-white:
Just changed the commit message prefix from "fix" to "bug".

ktuite (approved):

PR #457, opened 2022-02-24
New metrics: archived projects, unique roles

ktuite:
Adds 5 new metrics:
1. number of total archived projects (no recent vs. total, just how many there are at the time of computing analytics)
2. number of unique project managers (if a user is a manager on multiple projects, this is only counted once. if a user has two different roles on different projects, each type of role is counted.)
3. same as above but for project viewers
4. same as above but for data collectors
5. number of reused form ids: will report how many active forms use an (`xmlFormid`) ID that was ever deleted or purged in the past.
    * this will give us a sense of whether people are reusing form IDs at all (which will they have to use form deletion to do).

### DIFF
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -235,7 +235,7 @@ group by fs."projectId"`);
 
 // This query checks the audit log for 'form.delete' actions
 // and joins with two sources: purged actees and soft-deleted forms
-// to produce a list of all deleted formId/xmlFormId/projectId entries.
+// to produce a list of all deleted xmlFormId/projectId entries.
 // The list of active forms is then compared against this
 // deleted forms list to see how many active forms reuse an xmlFormId.
 const countReusedFormIds = () => ({ all }) => all(sql`
@@ -244,8 +244,7 @@ from forms
 inner join (
   select
   coalesce(a."details"->>'xmlFormId', f."xmlFormId") as xml_form_id,
-  coalesce((a."details"->>'projectId')::int, f."projectId") as proj_id,
-  coalesce((a."details"->>'formId')::int, f."id") as form_id
+  coalesce((a."details"->>'projectId')::int, f."projectId") as proj_id
   from audits as au
   join actees as a on au."acteeId" = a."id"
   left join forms as f on au."acteeId" = f."acteeId"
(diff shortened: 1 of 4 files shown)


test/integration/other/analytics-queries.js
```diff
@@ -359,6 +403,51 @@ describe('analytics task queries', () => {
...
+      await service.login('alice', (asAlice) =>
+        asAlice.delete('/v1/projects/1/forms/simple')
+          .expect(200));
+
+      // no deleted forms reused yet
+      const emptyRes = await container.Analytics.countReusedFormIds();
+      emptyRes.should.eql([]);
+
+      // one purged form reused
+      await container.Forms.purge(force=true);
```
ktuite, line 417: don't need force

test/integration/other/analytics-queries.js
```diff
@@ -359,6 +403,51 @@ describe('analytics task queries', () => {
...
+          .set('Content-Type', 'application/xml')
+          .then(() => asAlice.delete('/v1/projects/1/forms/simple2'))
+          .then(() => asAlice.post('/v1/projects/1/forms')
+            .send(testData.forms.simple2)
+            .set('Content-Type', 'application/xml')));
+
+      // delete multiple times (only count 1 active form with reused id)
+      const proj2 = await createTestProject(service, container, 'New Proj');
+      await service.login('alice', (asAlice) =>
+        asAlice.post('/v1/projects/1/forms')
```
ktuite, line 433: posting into the old project??

lib/model/query/analytics.js
```diff
@@ -212,6 +233,35 @@ from (
...
+  case
+    when f.id is not null then f."deletedAt" is not null
+    when a."purgedAt" is not null then true
+    else false
+  end
+  group by (xml_form_id, form_id, proj_id)
+) as deleted_form_ids
+on forms."xmlFormId" = deleted_form_ids."xml_form_id" and
+   forms."projectId" = deleted_form_ids."proj_id"
+where forms.id <> deleted_form_ids."form_id"
```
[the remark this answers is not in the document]
ktuite: ok cool, we have fixed the query and have a plan for this just to return `1`, the number of active forms with an xmlformid that matches any previously deleted xmlformid in that project.

ktuite (commented):
discussions with Matt

matthew-white (approved):

PR #458, opened 2022-02-25
bug: fix splitSelectMultiples for draft submissions.

issa-tseng:
* we now just get the fields for the form def in question rather than
  try to merge them all. we did the latter for paranoia+compatibility
  with deletedFields = true.
* but theoretically nobody can be filling out data their form version
  doesn't ask for, so it should cover all the values in the instance.

### DIFF
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -2235,6 +2235,18 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff
                 done();
               })))))));
 
+    it('should not log the action in the audit log', testService((service) =>
+      service.login('alice', (asAlice) =>
+         asAlice.post('/v1/projects/1/forms/simple/draft')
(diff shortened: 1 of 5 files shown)


test/integration/api/submissions.js
```diff
@@ -2235,17 +2235,34 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff
                 done();
               })))))));
 
-    it('should not log the action in the audit log', testService((service) =>
```
matthew-white, line 2238: Ah I think I added this test, in #382. The test description is not great, but I think it's still a useful test. It's trying to test that when you download draft submissions, a `form.submission.export` audit log entry is not created.

matthew-white (approved):

test/integration/api/submissions.js
issa-tseng: oh i see

Issue #461, opened 2022-03-07
Server should not come up if migrations fail


yanokwa:
I upgraded to Central v1.4. I was able to login, but the the server audit logs page did not load. Looking at the logs, I saw this...

It would be ideal if the server does not load if migrations fail in some way (or in this case, run out of memory).

```
root@example:~/central# docker logs central_service_1
wait-for-it.sh: waiting 15 seconds for example.foo.bar:5432
wait-for-it.sh: example.foo.bar:5432 is available after 0 seconds
generating local service configuration..
running migrations..

<--- Last few GCs --->

[30:0x4a26ef0]    65185 ms: Mark-sweep (reduce) 968.3 (972.4) -> 968.2 (973.4) MB, 299.2 / 0.0 ms  (average mu = 0.552, current mu = 0.000) last resort GC in old space requested
[30:0x4a26ef0]    65457 ms: Mark-sweep (reduce) 968.2 (972.4) -> 968.2 (973.4) MB, 272.2 / 0.0 ms  (average mu = 0.365, current mu = 0.000) last resort GC in old space requested


<--- JS stacktrace --->

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: 0xa25510 node::Abort() [node]
 2: 0x9664d3 node::FatalError(char const*, char const*) [node]
 3: 0xb9a8be v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xb9ac37 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xd56ca5  [node]
 6: 0xd692c1 v8::internal::Heap::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 7: 0xd2e9fd v8::internal::Factory::AllocateRaw(int, v8::internal::AllocationType, v8::internal::AllocationAlignment) [node]
 8: 0xd28884 v8::internal::FactoryBase<v8::internal::Factory>::AllocateRawWithImmortalMap(int, v8::internal::AllocationType, v8::internal::Map, v8::internal::AllocationAlignment) [node]
 9: 0xd2abd1 v8::internal::FactoryBase<v8::internal::Factory>::NewRawTwoByteString(int, v8::internal::AllocationType) [node]
[...]

PR #462, opened 2022-03-07
Check for pending migrations before running server

ktuite:
Adds a helper function for checking for pending migrations that returns an error exit code (1) if there are any pending migrations. This is used to prevent running the server after automatically applying migrations if there happened to be a problem with the migrations. See issues #461 and https://github.com/getodk/central/issues/258. 

This check has been added to the Makefile so the dev server won't run if there is a problem with the migrations, but it also needs to be used in `start-odk.sh` of the main central repo.

Here's an example non-make script (similar to `start-odk.sh`) that will exit before starting the server if the migration check fails:
  
    #!/bin/sh
    
    echo "pretending to work like start-odk.sh in the main central repo"
    
    node -e 'const { withDatabase, migrate } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(migrate);'
    
    node -e 'const { withDatabase, checkMigrations } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(checkMigrations);'
    
    if [ $? -eq 1 ]; then
      echo "migration error: there are still pending migrations... central will not run until fixed"
      exit 1
    fi
    
    node lib/bin/run-server.js

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,12 @@ node_modules:
 migrations: node_modules
 	node -e 'const { withDatabase, migrate } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(migrate);'
 
+check-migrations: node_modules
+	node -e 'const { withDatabase, checkMigrations } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(checkMigrations);'
+
 base: node_modules migrations
 
-run: base
+run: base check-migrations
 	node lib/bin/run-server.js
 
 debug: base
--- a/Makefile
+++ b/Makefile
@@ -10,9 +10,9 @@ migrations: node_modules
 check-migrations: node_modules
 	node -e 'const { withDatabase, checkMigrations } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(checkMigrations);'
 
-base: node_modules migrations
+base: node_modules migrations check-migrations
 
-run: base check-migrations
+run: base
 	node lib/bin/run-server.js
 
 debug: base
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,9 +7,12 @@ node_modules:
 migrations: node_modules
 	node -e 'const { withDatabase, migrate } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(migrate);'
 
+check-migrations: node_modules
+	node -e 'const { withDatabase, checkMigrations } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(checkMigrations);'
+
 base: node_modules migrations
 
-run: base
+run: base check-migrations
```
matthew-white, line 15: What do you think about putting `check-migrations` under `base` (after `migrations`) rather than under `run`?

lib/model/migrate.js
```diff
@@ -25,5 +25,12 @@ const withDatabase = (config) => (mutator) => {
 // Given a database, initiates migrations on it.
 const migrate = (db) => db.migrate.latest({ directory: `${__dirname}/migrations` });
 
-module.exports = { connect, withDatabase, migrate };
+// Checks for pending migrations and returns an exit code of 1 if any are
+// still pending/unapplied (e.g. automatically running migrations just failed).
+const checkMigrations = (db) => db.migrate.list({ directory: `${__dirname}/migrations` })
+  .then((res) => (res[1].length === 0
+    ? process.exit(0)
+    : process.exit(1)));
```
matthew-white, line 33: This area is new to me, but I'm reading [here](https://nodejs.dev/learn/how-to-exit-from-a-nodejs-program) that `process.exit()` "ungracefully" terminates the program. `withDatabase()` will want to call `db.destroy()` after running `checkMigrations`, so if there's another way to terminate, that seems preferable. (That link seems to suggest setting `process.exitCode`.)

matthew-white (approved):

Issue #466, opened 2022-03-14
Revisit testing database migrations


ktuite:
In PR #464, we added a new column `description` to Projects, which broke tests of previous database migrations ([here](https://github.com/getodk/central-backend/blob/15419b41cf84bb5a49465d3d422b4a146c9229ec/test/integration/other/migrations.js#L29-L34)).

The problem with the migration tests is that the code (`frames.js`) reflects the latest structure of the database (with `projects.description` column), but the migration test starts by applying migrations only up to an earlier database state (without that `description` column). So the code to populate the database with the initial project doesn't work because it's expecting a database column that doesn't exist. The way these database migration tests work needs to fundamentally change to not have this code state / database state inconsistency.

Some thoughts:
1) maybe the migrations served their purpose at the time but they don’t need to be re-tested going forward?
2) we need to see how other people test migrations because surely they run into this problem.
3) the data needs to be put in the database in a different way (e.g. raw sql or loading a pgdump or something, not created by the server) that is more consistent with how it will look at the time of the migration in question.
4) there has been talk of more extensive database testing that looks at performance, and that could potentially include migration performance.

For now, we will just skip the migration tests until we come up with a better solution.

matthew-white:
@alxndrsn is working on a new approach to testing migrations here: #1252

alxndrsn:
Closed by https://github.com/getodk/central-backend/pull/1252

PR #467, opened 2022-03-14
Do not automatically create a Default Project

ktuite:
I wanted to give this a shot... of changing the migrations that create default projects. Default projects get created in an old migration `20181206-01-add-projects.js` and removing that behavior from the migration in the first place seemed cleaner than trying to undo that behavior later by trying to remove the default project or something else more risky.

It turned out to be a bit more complicated than that, though.
1) default project creation should still happen in the case that forms exist (without projects) at the time of running the migration.
2) a later roles migration `20181212-01-add-roles.js` assumes that a project exists.
3) the text fixture data assumes that a default project exists.

I think I addressed these three issues and tested them (with migration tests that don't use any server code and should continue to work going forward... see issue #466). 

Here's what a fresh database (with all migrations applied) looks like through the latest frontend, with and without admin privileges:
Screen Shot 2022-03-14 at 2 22 01 PM
Screen Shot 2022-03-14 at 2 22 21 PM

Leaving PR as a draft for now in case a different approach comes to mind.

### DIFF
--- a/lib/model/migrations/20181206-01-add-projects.js
+++ b/lib/model/migrations/20181206-01-add-projects.js
@@ -9,59 +9,48 @@
 
 const uuid = require('uuid/v4');
 
-const up = (db) =>
+const up = async (db) => {
   // first create the projects table itself.
-  db.schema.createTable('projects', (projects) => {
+  await db.schema.createTable('projects', (projects) => {
     projects.increments('id');
     projects.text('name').notNull();
     projects.string('acteeId', 36).notNull().unique();
     projects.dateTime('createdAt');
     projects.dateTime('updatedAt');
     projects.dateTime('deletedAt');
-  })
+  });
+
+  // first add projectId column to forms
+  await db.schema.table('forms', (forms) => {
+    forms.integer('projectId');
+  });
 
   // and create the project actee species.
-    .then(() => db.insert({ id: 'project', species: 'species' }).into('actees'))
+  await db.insert({ id: 'project', species: 'species' }).into('actees');
 
   // create a default project if forms already exist
-    .then(() => db.count('*').from('forms').where({ deletedAt: null }).limit(1)
-      .then(([{ count }]) => {
-        if (Number(count) > 0) {
-          const name = 'Forms you made before projects existed';
-          return db.insert({ id: uuid(), species: 'project' }).into('actees').returning('id')
-            .then(([ id ]) =>
-              db.insert({ name, acteeId: id, createdAt: new Date() }).into('projects').returning('*'));
-        }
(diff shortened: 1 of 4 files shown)


lib/model/migrations/20181206-01-add-projects.js
```diff
@@ -23,27 +23,31 @@ const up = (db) =>
...
-      db.count('*').from('forms').where({ deletedAt: null }).limit(1)
-        .then(([{ count }]) => count),
-      db.insert({ id: uuid(), species: 'project' }).into('actees').returning('id')
-        .then(([ id ]) => id)
-    ])
-      .then(([ count, acteeId ]) => {
-        const name = (Number(count) > 0) ? 'Forms you made before projects existed' : 'Default Project';
-        return db.insert({ name, acteeId, createdAt: new Date() }).into('projects').returning('*');
+  // create a default project if forms already exist
+    .then(() => db.count('*').from('forms').where({ deletedAt: null }).limit(1)
```
ktuite, line 27: get rid of `deletedAt` in this query

ktuite: also what would happen if there were some forms but they were all deleted? would this all break? maybe

lib/model/migrations/20181206-01-add-projects.js
```diff
@@ -23,27 +23,31 @@ const up = (db) =>
...
+              db.insert({ name, acteeId: id, createdAt: new Date() }).into('projects').returning('*'));
+        }
       }))
 
   // now create a project_id column in forms.
-    .then(([ project ]) =>
+    .then((projects) =>
+      // first add projectId column to forms
       db.schema.table('forms', (forms) => {
         forms.integer('projectId');
```
ktuite, line 41: i want this changing the form table schema to happen earlier in this migration?

test/integration/other/migrations.js
```diff
@@ -189,3 +190,40 @@ describe.skip('database migrations', function() {
...
+
+    const proj = projects[0];
+    proj.name.should.equal('Forms you made before projects existed');
+
+    const formCount = await container.oneFirst(sql`select count(*) from forms where "projectId"=${proj.id}`);
+    formCount.should.equal(1);
+  }));
+
+  it('should not make a default project if no forms', testServiceFullTrx(async (service, container) => {
+    // before 20181206-01-add-projects.js
```
ktuite, line 222: remove this comment

ktuite (commented):

matthew-white (approved):
This approach is making sense to me!

matthew-white (approved):

PR #470, opened 2022-03-17
In-depth project + form queries

ktuite:
This PR is to be used by the frontend for the new Project list page (the Central homepage) in Block 15, where forms and key details about each form are shown on that page.

This backend PR makes two key changes:
1. It extends what an "Extended" Form is to also include counts of Submissions in specific review states, such as 'edited', 'hasIssues', and 'received' (the default state for new Submissions)
2. It adds a query parameter `?forms=true` to the Project listing API call to get a list of Forms (extended with the fields described above) associated with each project that are visible to the authenticated Actor. 
    - Actors that are admins or managers of Projects will be able to see all the Forms of a Project including drafts, while Actors who are data collectors will only be able to see the published Forms.

### DIFF
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -140,11 +140,19 @@ Form.Partial = class extends Form {};
 
 Form.Extended = class extends Frame.define(
   'submissions',        readable,       'lastSubmission', readable,
-  'excelContentType',   readable
+  'excelContentType',   readable,
+  // counts of submissions in various review states
+  'receivedCount',                      'hasIssuesCount',
+  'editedCount'
 ) {
   forApi() {
     return {
       submissions: this.submissions || 0,
+      reviewStates: {
+        received: this.receivedCount || 0,
+        hasIssues: this.hasIssuesCount || 0,
+        edited: this.editedCount || 0
+      },
       lastSubmission: this.lastSubmission,
       excelContentType: this.excelContentType
     };
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -379,7 +379,7 @@ where "acteeId"=${acteeId} and "deletedAt" is null`)
   .then(map(_unjoiner));
 
 // there are many combinations of required fields here so we compose our own extender variant.
-const _getSql = ((fields, extend, options, version, deleted = false) => sql`
+const _getSql = ((fields, extend, options, version, deleted = false, actorId) => sql`
 select ${fields} from forms
 left outer join form_defs on ${versionJoinCondition(version)}
 ${extend|| sql`
@@ -397,13 +397,30 @@ ${extend|| sql`
   left outer join actors on audits."actorId"=actors.id
   left outer join (select id, "contentType" as "excelContentType" from blobs) as xls
     on form_defs."xlsBlobId"=xls.id`}
+${(actorId == null) ? sql`` : sql`
+inner join
+  (select id, max(assignment."showDraft") as "showDraft" from projects
+    inner join
+      (select "acteeId", 0 as "showDraft" from assignments
+        inner join (select id from roles where verbs ? 'form.read') as role
+          on role.id=assignments."roleId"
+        where "actorId"=${actorId}
+      union all
+      select "acteeId", 1 as "showDraft" from assignments
+        inner join (select id from roles where verbs ? 'form.update') as role
+          on role.id=assignments."roleId"
+        where "actorId"=${actorId}) as assignment
+      on assignment."acteeId" in ('*', 'project', projects."acteeId")
+    group by id) as filtered
+  on filtered.id=forms."projectId"`}
 where ${equals(options.condition)} and forms."deletedAt" is ${deleted ? sql`not` : sql``} null
+${(actorId == null) ? sql`` : sql`and (form_defs."publishedAt" is not null or filtered."showDraft" = 1)`}
 order by coalesce(form_defs.name, "xmlFormId") asc`);
 
 const _getWithoutXml = extender(Form, Form.Def)(Form.Extended, Actor.into('createdBy'))(_getSql);
 const _getWithXml = extender(Form, Form.Def, Form.Xml)(Form.Extended, Actor.into('createdBy'))(_getSql);
-const _get = (exec, options, xml, version, deleted) =>
-  ((xml === true) ? _getWithXml : _getWithoutXml)(exec, options, version, deleted);
+const _get = (exec, options, xml, version, deleted, actorId) =>
+  ((xml === true) ? _getWithXml : _getWithoutXml)(exec, options, version, deleted, actorId);
 
 const getByProjectId = (projectId, xml, version, options = QueryOptions.none, deleted = false) => ({ all }) =>
   _get(all, options.withCondition({ projectId }), xml, version, deleted);
@@ -411,6 +428,8 @@ const getByProjectAndXmlFormId = (projectId, xmlFormId, xml, version, options = 
   _get(maybeOne, options.withCondition({ projectId, xmlFormId }), xml, version, deleted);
 const getByProjectAndNumericId = (projectId, id, xml, version, options = QueryOptions.none, deleted = false) => ({ maybeOne }) =>
   _get(maybeOne, options.withCondition({ projectId, 'forms.id': id }), xml, version, deleted);
+const getAllByAuth = (auth, options = QueryOptions.none) => ({ all }) =>
+  _get(all, options, null, null, false, auth.actor.map((actor) => actor.id).orElse(-1));
 
 ////////////////////////////////////////////////////////////////////////////////
 // SCHEMA
@@ -486,6 +505,7 @@ module.exports = {
   getByAuthForOpenRosa,
   getVersions, getByActeeIdForUpdate, getByActeeId,
   getByProjectId, getByProjectAndXmlFormId, getByProjectAndNumericId,
+  getAllByAuth,
   getFields, getBinaryFields, getStructuralFields, getMergedFields,
   lockDefs, getAllSubmitters
 };
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -384,7 +384,11 @@ select ${fields} from forms
 left outer join form_defs on ${versionJoinCondition(version)}
 ${extend|| sql`
   left outer join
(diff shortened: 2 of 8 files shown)


lib/model/frames/form.js
```diff
@@ -140,11 +140,16 @@ Form.Partial = class extends Form {};
 
 Form.Extended = class extends Frame.define(
   'submissions',        readable,       'lastSubmission', readable,
+  'received',           readable,       'has_issues',       readable,
```
matthew-white, line 143: Just a random comment, I think it'd be great to name the field `hasIssues` rather than `has_issues`, since other fields are camelCase.

lib/model/frames/form.js
```diff
@@ -140,11 +140,16 @@ Form.Partial = class extends Form {};
...
+  'edited',             readable,
```
matthew-white, line 144: Could it be confusing that the `edited` field isn't about the form itself and whether it's been edited, but is instead about the form's submissions? Would `editedSubmissions` be better? (Though I guess that pattern doesn't really work for the `hasIssues` review state...) Another alternative would be to have a single field (named `reviewStates`?) that is an object with a property for each of these review states. Quite possibly no change needed, mostly just wanted to flag a potential for confusion.

ktuite: I changed it a bit so the states on the `Form` frame are still flat, but they get put into a `reviewState`(s - ahh maybe it should be pluralized) object for the API response.

matthew-white: Ah nice! I do think it's helpful to group them together like that. In that case, I think these fields shouldn't be `readable`, since they won't be available as top-level fields on the API response.

Another strategy that might be possible would be to define a frame for the `reviewState` object, then use `extender()` to insert these counts into that frame. (I'm not quite sure about the details, but I think there are similar examples in the codebase.) Possibly it doesn't add much here though, since you already have to have a custom `forApi()` in order for `submissions` to default to 0.

test/integration/api/projects.js
```diff
@@ -1180,3 +1180,29 @@ describe('api: /projects', () => {
   });
 });
 
+describe('api: /projects-forms', () => {
+  describe('GET', () => {
+    it('should return projects with nested extended forms', testService((service) =>
+      service.login('alice', (asAlice) => asAlice.get('/v1/projects-forms')
+        // todo: revisit this...
+        // extended=true to get form details but not really needed for projects
```
matthew-white, line 1188: I also think that it'd be nice for performance for the projects query not to hit the `submissions` table. Maybe this endpoint shouldn't require the extended metadata header and should default to extended metadata for forms?

lib/model/frames/form.js
```diff
@@ -140,11 +140,18 @@ Form.Partial = class extends Form {};
...
   'submissions',        readable,       'lastSubmission', readable,
+  'received',           readable,       'hasIssues',       readable,
+  'edited',             readable,
   'excelContentType',   readable
 ) {
   forApi() {
     return {
       submissions: this.submissions || 0,
+      reviewState: {
+        received: this.received || 0,
```
matthew-white, line 151: Everywhere else in the API, the received review state is represented as `null` rather than as `'received'`. I'd love to somehow use `null` here as well, but I can't think of a good way to do so. Probably no change needed, just wanted to throw out this thought.

lib/model/query/projects.js
```diff
@@ -112,9 +112,24 @@ const getAllByAuth = (auth, options = QueryOptions.none) => ({ all }) =>
...
+const getAllByAuthWithForms = (auth, queryOptions = QueryOptions.none) => ({ Projects, Forms }) =>
+  Promise.all([
+    Projects.getAllByAuth(auth, queryOptions),
+    Forms.getAllByAuth(auth, QueryOptions.extended)
+  ])
+    .then(([ projects, forms ]) => {
+      const projectsWithForms = projects.reduce((acc, proj) => ({ ...acc, [proj.id]: { ...proj.forApi(), forms: [] } }), {});
+      for (const form of forms) {
+        if (form.projectId in projectsWithForms)
+          projectsWithForms[form.projectId].forms.push(form.forApi());
```
ktuite, line 124: `forms` used to contain a count of forms in the extended Project so maybe this should be called something else so that still works.

lib/resources/projects.js
```diff
@@ -9,13 +9,16 @@
...
 const { getOrNotFound, getOrReject, reject } = require('../util/promise');
-const { success } = require('../util/http');
+const { isTrue, success } = require('../util/http');
 const { QueryOptions } = require('../util/db');
 const Problem = require('../util/problem');
 
 module.exports = (service, endpoint) => {
-  service.get('/projects', endpoint(({ Projects }, { auth, queryOptions }) =>
-    Projects.getAllByAuth(auth, queryOptions)));
+  // specify ?forms=true to get forms nested within projects
```
ktuite, line 17: could be a different query parameter? also had a version where the projects with nested forms was a completely different api endpoint. not sure which way to go except adding it here to an existing endopint via `/v1/projects?someflag=true` means fewer documentation changes.

matthew-white: The idea of having a single endpoint makes a lot of sense to me! The response will still be an array of projects, so I think it's a good fit.

lib/model/query/forms.js
```diff
@@ -393,14 +397,33 @@ ${extend|| sql`
   left outer join actors on audits."actorId"=actors.id
   left outer join (select id, "contentType" as "excelContentType" from blobs) as xls
     on form_defs."xlsBlobId"=xls.id`}
+${(actorId == null) ? sql`` : sql`
+inner join
+  (select id, max(assignment."showDraft") as "showDraft" from projects
```
matthew-white, line 402: I think you could also make `showDraft` a boolean, then use `bool_or(assignment."showDraft")` instead of `max()`.

lib/model/query/projects.js
```diff
@@ -112,9 +112,24 @@ const getAllByAuth = (auth, options = QueryOptions.none) => ({ all }) =>
 const getById = (id, options = QueryOptions.none) => ({ maybeOne }) =>
   _get(maybeOne, options.withCondition({ id }));
 
+const getAllByAuthWithForms = (auth, queryOptions = QueryOptions.none) => ({ Projects, Forms }) =>
+  Promise.all([
+    Projects.getAllByAuth(auth, queryOptions),
+    Forms.getAllByAuth(auth, QueryOptions.extended)
+  ])
+    .then(([ projects, forms ]) => {
+      const projectsWithForms = projects.reduce((acc, proj) => ({ ...acc, [proj.id]: { ...proj.forApi(), nestedForms: [] } }), {});
```
matthew-white, line 121: Maybe break this onto multiple lines?

test/integration/api/projects.js
```diff
@@ -1180,3 +1180,51 @@ describe('api: /projects', () => {
...
+describe('api: /projects?forms=true', () => {
+  describe('GET', () => {
+    it('should return projects with nested extended forms', testService((service) =>
+      service.login('alice', (asAlice) => asAlice.get('/v1/projects?forms=true')
+        .expect(200)
+        .then(({ body }) => {
+          body.length.should.equal(1);
+          body[0].should.be.a.Project();
+          const form = body[0].nestedForms[0];
+          form.should.be.a.Form();
```
matthew-white, line 1192: There's also an `ExtendedForm()` assertion. Actually, probably a good idea to update that assertion to incorporate the new `reviewState` property.

test/integration/api/projects.js
```diff
@@ -1180,3 +1180,51 @@ describe('api: /projects', () => {
   });
 });
 
+describe('api: /projects?forms=true', () => {
```
matthew-white, line 1183: How about a test where there are multiple projects, each with at least one form, to verify that each form is put on the right project?

test/integration/api/forms/list.js
```diff
@@ -45,6 +45,9 @@ describe('api: /projects/:id/forms (listing forms)', () => {
               body.forEach((form) => form.should.be.an.ExtendedForm());
```
matthew-white, line 45: How about updating the `ExtendedForm()` assertion to incorporate the new `reviewState` property?

matthew-white (commented):
Small questions/comments before interactive review

lib/model/frames/form.js
```diff
@@ -140,11 +140,19 @@ Form.Partial = class extends Form {};
...
-  'excelContentType',   readable
+  'excelContentType',   readable,
+  // counts of submissions in various review states
+  'receivedCount',                      'hasIssuesCount',
+  'editedCount'
 ) {
   forApi() {
     return {
       submissions: this.submissions || 0,
+      reviewStates: {
```
ktuite, line 151: Add in the API docs that extended forms have this new piece of data.

test/assertions.js
```diff
@@ -177,8 +177,10 @@ should.Assertion.add('ExtendedForm', function() {
   this.params = { operator: 'to be a ExtendedForm' };
 
   this.obj.should.be.a.Form();
-  Object.keys(this.obj).should.containDeep([ 'submissions', 'lastSubmission' ]);
+  Object.keys(this.obj).should.containDeep([ 'submissions', 'lastSubmission', 'reviewStates' ]);
   if (this.obj.submissions != null) this.obj.submissions.should.be.a.Number();
+  if (this.obj.reviewStates != null)
```
ktuite, line 182: Maybe null checks not needed for this or for `submissions`

test/integration/api/projects.js
```diff
@@ -1180,3 +1180,78 @@ describe('api: /projects', () => {
...
+            .expect(200))
+          .then(() => Users.getByEmail('bob@getodk.org').then((o) => o.get()))
+          .then((bob) => asAlice.post(`/v1/projects/${body.id}/assignments/formfill/${bob.actorId}`)))
+        .then(() => service.login('bob', (asBob) => asBob.get('/v1/projects?forms=true')
+          .expect(200)
+          .then(({ body }) => {
+            body.length.should.equal(2);
+            body[0].formList.length.should.equal(2);
+            body[1].formList.length.should.equal(1);
+            body[1].formList[0].name.should.equal('Simple');
```
ktuite, line 1254: forms in each project are named `Simple`, maybe I should name it something more distinctive

ktuite (commented):
Notes from review with @matthew-white

matthew-white (approved):

Issue #468, opened 2022-03-03
Central sends emails even if account does not exist


yanokwa:

lognaturel:
Thanks for taking this on, @sadiqkhoja. #472 also adds a message in the UI saying that the email address has no associated account. I don't think we should disclose that. @yanokwa, @matthew-white what did you have in mind? I was imagining  the generic message becomes something like "If there is an account with that email address, it has been sent a password reset message."

yanokwa:
Agreed with the generic message.

matthew-white:
> #472 also adds a message in the UI saying that the email address has no associated account. I don't think we should disclose that.

That sounds right to me. That means that the API should return the same successful response whether or not there is a user with the specified email.

Right now Frontend shows the following message if it receives a successful response (in the [`AccountResetPassword`](https://github.com/getodk/central-frontend/blob/master/src/components/account/reset-password.vue) component):

> An email has been sent to {email} with further instructions.

I'm finding it a little hard to think of alternative wording that's not a little clunky. Here's one possibility that mentions the "if" at the end of the sentence:

> An email will be sent to {email} with further instructions if that account exists in the system.

If that feels too clunky, we might also be fine leaving the message as-is.

lognaturel:
> we might also be fine leaving the message as-is.

Sounds fine to me.

sadiqkhoja:
I have updated the PR, keeping the message as-is

PR #472, opened 2022-04-15
improve #468: no send email if no account exists

sadiqkhoja:
- Updated `/users/reset/initiate` API such that it does not send any email if no account is found and returns 404
- Updated integration test to validate the change
- Performed manual testing to see how error appears on UI

image

### DIFF
--- a/lib/resources/users.js
+++ b/lib/resources/users.js
@@ -55,7 +55,7 @@ module.exports = (service, endpoint) => {
         .orElseGet(() => Users.emailEverExisted(body.email)
           .then((existed) => ((existed === true)
             ? mail(body.email, 'accountResetDeleted')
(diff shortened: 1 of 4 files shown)


lib/resources/users.js
```diff
@@ -55,7 +55,7 @@ module.exports = (service, endpoint) => {
         .orElseGet(() => Users.emailEverExisted(body.email)
           .then((existed) => ((existed === true)
             ? mail(body.email, 'accountResetDeleted')
-            : mail(body.email, 'accountResetFailure'))))
```
matthew-white, line 58: With this gone, I think we can go ahead and remove this message in `lib/formats/mail.js`.

matthew-white (approved):
Looks great! 💯 Thanks, @sadiqkhoja!

PR #474, opened 2022-04-20
bug #473: always check auth if invalidate is true

### LINKED ISSUE #473: Password reset: always check auth if ?invalidate=true
Normally, it is not possible to use /v1/users/reset/initiate to learn whether a particular email exists on a Central server: if the specified email exists, the endpoint returns a successful response, and if it doesn't exist, the endpoint returns the same successful response. That is also true if the query parameter `?invalidate=true` is specified, as long as valid auth is provided. However, if valid auth is not provided, `?invalidate=true` can be used to check the existence of a particular email: if the specified email exists, the endpoint will return a 403, but if it doesn't exist, the endpoint will return a 200. That's because the endpoint only checks auth if it will take an action that requires auth, which only happens if the email exists. Instead, I think that if `?invalidate=true` is specified, the endpoint should always check auth. That would mean that even if the specified email doesn't exist, the endpoint will return a 403 if valid auth is not provided. With that change, it would no longer be possible to use the endpoint to check the existence of a particular email.

sadiqkhoja:
Fixes #473:

- Return 403 if invalidate is true and account doesn't exist
- Added integration test to verify it


lib/resources/users.js
```diff
@@ -55,7 +55,7 @@ module.exports = (service, endpoint) => {
         .orElseGet(() => Users.emailEverExisted(body.email)
           .then((existed) => ((existed === true)
             ? mail(body.email, 'accountResetDeleted')
-            : resolve())))
+            : (isTrue(query.invalidate) ? reject(Problem.user.insufficientRights()) : resolve()))))
```
matthew-white, line 58: I'm wondering whether the `isTrue(query.invalidate)` check should come before the `Users.emailEverExisted()` query. If the check came right after `orElseGet()`, that'd be similar to how it's positioned after `map()` above. Right now, since `isTrue(query.invalidate)` isn't checked if `existed` is `true`, it'd be still possible for an unauthed user to use this endpoint to determine whether an email was deleted (if the email was deleted, it'd be a 200 instead of a 403). That's probably not a big deal, but it'd be nice to have this endpoint not provide any information at all to a user who doesn't have sufficient auth (for example, Chelsea in testing).

I'm also wondering what the right behavior is if the authed user is a sidewide administrator (like Alice in testing). I'm thinking that Alice shouldn't receive a 403 here, since the issue isn't really that Alice has insufficient rights. I think it'd be fine for Alice to continue to receive a 200; we just want a user without sufficient auth to receive a 403 (so that they can't gain information). Instead of returning `reject(Problem.user.insufficientRights())`, maybe we could do an auth check in `orElseGet()` that's similar to what's done above in `map()`.

The auth check in `map()` is `auth.canOrReject('user.password.invalidate', user.actor)`, but of course there's no `user` in `orElseGet()`. But I'm thinking it might work to specify `User.species` as the actee when you complete the auth check in `orElseGet()`.

sadiqkhoja: Incorporated your feedback. Can you please rerun the build, I think it failed because two builds where running simultaneously

test/integration/api/users.js
```diff
@@ -287,6 +287,22 @@ describe('api: /users', () => {
               .expect(401);
           }))));
 
+    it('should return 403 if user doesn\'t have rights to invalidate', testService((service) =>
```
matthew-white, line 290: How about we update this test description to make it clear that it's different from the similar test above (the test that "should fail the request if invalidation is requested but not allowed"). Maybe something like "should return 403 if user doesn't have rights to invalidate and email doesn't exist" or "should fail the request if invalidation is not allowed and email doesn't exist".

test/integration/api/users.js
```diff
@@ -287,6 +287,22 @@ describe('api: /users', () => {
...
+      service.login('chelsea', (asChelsea) =>
+        asChelsea.post('/v1/users/reset/initiate?invalidate=true')
+          .send({ email: 'winnifred@getodk.org' })
+          .expect(403)
+          )));
```
matthew-white, line 295: We generally keep closing parentheses on the same line, so it'd be great to combine these two lines.

matthew-white (commented):
This is looking good! I left just a couple of small comments about testing.

matthew-white (approved):
This looks great, thanks again, @sadiqkhoja! 🎉 I'll go ahead and merge now.

By the way, I'd love to learn more about what sort of issues you're interested in. Are you also interested in the frontend (Vue/HTML/CSS), or are you most interested in the backend (Node/database/API)? Do you also have an interest in ops or things related to Docker? Just wanted to hear more about what's interesting to you!

sadiqkhoja:
Thanks @matthew-white  I am open to anything, if you have top 5 priority issues, please share I'll take them on.

Issue #417, opened 2021-10-18
PUT .../submissions/:instanceId does not return expected fields


matthew-white:
When I send a PUT request to edit a submission using the REST API, I receive the following response:

```json
{
  "instanceId": "4bfcb0f9-7bb7-4bda-a165-ab731fbcaffc",
  "submitterId": 16,
  "createdAt": "2021-10-18T18:39:10.064Z",
  "instanceName": "White-Matt-9435bee7-c929-497e-874a-61a2f47389e4",
  "current": true
}
```

However, the schema of this JSON differs from what's normally returned for submissions:

- `current` usually isn't returned.
- Other fields usually are returned, including `deviceId` and `reviewState`.

ktuite:
Closing this issue because we have determined that this PUT response is consistent with what is returned when fetching a specific version of a submission. It does not have all the information of a full Submission, but it does have the info of what the code internally calls a `Submission.Def`, essentially a snapshot of a submission for a particular version. Opening a new issue #481 to fix the docs. 

```javascript
"PUT /v1/projects/1/forms/simple/submissions/one HTTP/1.1"
{
  instanceId: 'two',
  submitterId: 6,
  createdAt: '2022-05-06T17:17:46.990Z',
  instanceName: null,
  current: true,
  deviceId: null,
  userAgent: 'node-superagent/3.8.3'
}
```

```javascript
"GET /v1/projects/1/forms/simple/submissions/one/versions/two HTTP/1.1" 200 162
{
  submitterId: 6,
  createdAt: '2022-05-06T17:00:49.002Z',
  instanceName: null,
  instanceId: 'two',
  current: true,
  deviceId: null,
  userAgent: 'node-superagent/3.8.3'
}
```

PR #475, opened 2022-04-23
improve 417: correct the response body for PUT submission

sadiqkhoja:
I have changed the sql so that it returns reviewState and updatedAt. Also removed the `readable` attribute from `current` field. 

In `lib/model/frames/submission.js` file, I see several HACK/TODO tags around constructor of submission. @issa-tseng can you please share some details around that, what should be the ideal way to create submission object?

### DIFF
--- a/lib/model/frames/submission.js
+++ b/lib/model/frames/submission.js
@@ -90,7 +90,7 @@ Submission.Def = Frame.define(
   'localKey',                           'encDataAttachmentName',
   'signature',                          'createdAt',    readable,
   'instanceName', readable,             'instanceId',   readable,
-  'current',      readable,             'xml',
+  'current',                            'xml',
   'root',                               'deviceId',     readable,
   'userAgent',    readable,
   embedded('submitter')
--- a/lib/model/query/submissions.js
+++ b/lib/model/query/submissions.js
@@ -65,13 +65,14 @@ const createVersion = (partial, deprecated, form, deviceIdIn = null, userAgentIn
 
   // we already do transactions but it just feels nice to have the cte do it all at once.
   return one(sql`
-with logical as (update submissions set "reviewState"='edited', "updatedAt"=clock_timestamp()
-  where id=${deprecated.submissionId})
+with logical as (update submissions set "reviewState"='edited', "updatedAt"=clock_timestamp() 
+  where id=${deprecated.submissionId} returning "reviewState","updatedAt")
 , upd as (update submission_defs set current=false where "submissionId"=${deprecated.submissionId})
-${_defInsert(deprecated.submissionId, partial, form.def.id, actorId, null, deviceId, userAgent)}`)
+, def as (${_defInsert(deprecated.submissionId, partial, form.def.id, actorId, null, deviceId, userAgent)})
+select def.*, logical.* from def, logical`)
     // TODO/HACK: lame that we are reconstructing this this way.
-    .then((saved) => new Submission({ instanceId: partial.instanceId },
-      { def: new Submission.Def(saved), xml: new Submission.Xml({ xml: partial.xml }) }));
+    .then(({reviewState, updatedAt, ...savedDef}) =>  new Submission({ instanceId: partial.instanceId, reviewState, updatedAt },
+      { def: new Submission.Def(savedDef), xml: new Submission.Xml({ xml: partial.xml }) }));
 };
 
 createVersion.audit = (partial, deprecated, form) => (log) =>


issa-tseng:
hmm, i’ve reread the file several times and did not find a TODO comment in it? am i looking at the wrong version?

sadiqkhoja:
@issa-tseng  my bad, it's `lib/model/query/submissions.js`

issa-tseng:
hey @sadiqkhoja the comment doesn't have to do with the constructor so much as we usually have the database return us its complete current data for the object at the end of the operation and return that as the result, but here we are instead copying in memory values that we _believe_ the database has inserted.

matthew-white:
Hi @sadiqkhoja, we haven't had a chance to review this PR yet, but we're hoping to take a look soon. Thanks for looking into this issue!

ktuite:
Hi @sadiqkhoja! I wanted to jump in and introduce myself -- I'm Kathleen, part of the core Central team! I wasn't able to look into this particular PR and issue and wrap my head around it until now, but here's what I'm thinking:

I've been looking at the API responses for submission versions (when the contents of a submission is updated and you get access to the previous versions). This PUT request is how a submission gets updated/edited, and the response actually matches what is returned by `../submissions/<instanceId>/versions` and `../submissions/<instanceId>/versions/<instanceId>`. It's not a full Submission, it's like a snapshot of a Submission, what internally in the code is called a `SubmissionDef`.

Anyway, I think (and have discussed with @matthew-white) that the original issue #417 is not the problem after all, and the real issue is that the documentation is wrong and describes a full Submission response in places where we don't actually return that.

I wrote this test to just print what the different requests look like. If you'll notice, the first response and the last match!

I have some specific comments on your code changes. I think your approach was absolutely in the right direction for what the issue said, and there is still a change you could make to improve the clarity of the code. I hope it is still a good learning experience for you -- it is for me! 

```javascript
it.only('should return PUT response in correct format', testService((service) =>
    asAlice.post('/v1/projects/1/forms/simple/submissions')
      .send(testData.instances.simple.one)
      .expect(200)
      .then(() => service.login('bob', (asBob) =>
        asBob.put('/v1/projects/1/forms/simple/submissions/one')
          .send(withSimpleIds('one', 'two'))
          .expect(200)
          .then(({ body }) => {
            console.log("put response in question");
            console.log(body);
            //body.should.be.a.Submission(); // assertion fails
            body.should.be.a.SubmissionDef();
          })))
      .then(() => asAlice.get('/v1/projects/1/forms/simple/submissions/one')
        .expect(200)
        .then(({ body }) => {
          console.log("what it a submission looks like");
          console.log(body);
        }))
      .then(() => asAlice.get('/v1/projects/1/forms/simple/submissions/one/versions')
        .expect(200)
        .then(({ body }) => {
          console.log("what list of versions looks like");
          console.log(body);
        }))
      .then(() => asAlice.get('/v1/projects/1/forms/simple/submissions/one/versions/two')
        .expect(200)
        .then(({ body }) => {
          console.log("what it specific version of a submission looks like");
          console.log(body);
        })))));
```

```javascript
  api: /forms/:id/submissions
    /:instanceId PUT
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "POST /v1/sessions HTTP/1.1" 200 228
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "POST /v1/projects/1/forms/simple/submissions HTTP/1.1" 200 127
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "POST /v1/sessions HTTP/1.1" 200 228
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "PUT /v1/projects/1/forms/simple/submissions/one HTTP/1.1" 200 162
put response in question
{
  instanceId: 'two',
  submitterId: 6,
  createdAt: '2022-05-06T16:39:48.511Z',
  instanceName: null,
  current: true,
  deviceId: null,
  userAgent: 'node-superagent/3.8.3'
}
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "GET /v1/projects/1/forms/simple/submissions/one HTTP/1.1" 200 153
what it a submission looks like
{
  instanceId: 'one',
  submitterId: 5,
  deviceId: null,
  createdAt: '2022-05-06T16:39:48.489Z',
  updatedAt: '2022-05-06T16:39:48.511Z',
  reviewState: 'edited'
}
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "GET /v1/projects/1/forms/simple/submissions/one/versions HTTP/1.1" 200 328
what list of versions looks like
[
  {
    submitterId: 6,
    createdAt: '2022-05-06T16:39:48.511Z',
    instanceName: null,
    instanceId: 'two',
    current: true,
    deviceId: null,
    userAgent: 'node-superagent/3.8.3'
  },
  {
    submitterId: 5,
    createdAt: '2022-05-06T16:39:48.489Z',
    instanceName: null,
    instanceId: 'one',
    current: false,
    deviceId: null,
    userAgent: 'node-superagent/3.8.3'
  }
]
::ffff:127.0.0.1 - - [06/May/2022:16:39:48 +0000] "GET /v1/projects/1/forms/simple/submissions/one/versions/two HTTP/1.1" 200 162
what it specific version of a submission looks like
{
  submitterId: 6,
  createdAt: '2022-05-06T16:39:48.511Z',
  instanceName: null,
  instanceId: 'two',
  current: true,
  deviceId: null,
  userAgent: 'node-superagent/3.8.3'
}
      ✔ should return PUT response in correct format (342ms)
```

lib/model/frames/submission.js
```diff
@@ -90,7 +90,7 @@ Submission.Def = Frame.define(
   'localKey',                           'encDataAttachmentName',
   'signature',                          'createdAt',    readable,
   'instanceName', readable,             'instanceId',   readable,
-  'current',      readable,             'xml',
+  'current',                            'xml',
```
ktuite, line 93: Because this `current` field is important to knowing which version of a submission is the current one in `GET ...submissions/<instanceId>/versions`, I don't think we should make it un-readable here (which, as you've figured out, controls whether or not a field is returned in an API response).

lib/model/query/submissions.js
```diff
@@ -65,13 +65,14 @@ const createVersion = (partial, deprecated, form, deviceIdIn = null, userAgentIn
...
+with logical as (update submissions set "reviewState"='edited', "updatedAt"=clock_timestamp() 
+  where id=${deprecated.submissionId} returning "reviewState","updatedAt")
 , upd as (update submission_defs set current=false where "submissionId"=${deprecated.submissionId})
-${_defInsert(deprecated.submissionId, partial, form.def.id, actorId, null, deviceId, userAgent)}`)
+, def as (${_defInsert(deprecated.submissionId, partial, form.def.id, actorId, null, deviceId, userAgent)})
+select def.*, logical.* from def, logical`)
     // TODO/HACK: lame that we are reconstructing this this way.
-    .then((saved) => new Submission({ instanceId: partial.instanceId },
-      { def: new Submission.Def(saved), xml: new Submission.Xml({ xml: partial.xml }) }));
+    .then(({reviewState, updatedAt, ...savedDef}) =>  new Submission({ instanceId: partial.instanceId, reviewState, updatedAt },
```
ktuite, line 74: Overall, I think this was the right approach for what we thought the problem was -- getting the right info from the database and passing it along to the hacky `Submission` constructor.

But maybe it shouldn't be constructing a `Submission` here, just a `Submission.Def`? Because that is more consistent with what we expect to be returned? 

... Hmm, briefly tried that myself and it makes the step that follows https://github.com/getodk/central-backend/blob/master/lib/resources/submissions.js#L204 fail because it is expecting to work with a `Submission`. But maybe line 205 right after could return `submission.def` instead? And/or we could just add more explanatory comments about what we've learned along this journey...

matthew-white: Just to follow up on this comment, L204 fails because `SubmissionAttachments.create()` expect XML, not just a `Submission.Def`. However, like @ktuite said, if we change L205 to `submission.def` instead of `submission`, I think that would make the intention of that endpoint clearer. If there are places where additional comments would make things clearer, those would be great as well.

matthew-white:
@sadiqkhoja and @ktuite, could you two take another look at this PR as part of v2023.1? Making small code changes and/or adding comments if that still seems useful, or otherwise closing out the PR.

matthew-white:
@sadiqkhoja, now that you've made the changes in #783, do you think this PR can be closed? Is there anything else from this PR that we should capture?

PR #476, opened 2022-05-05
improve: increase maximum pool size.

matthew-white:
By default, Slonik [limits](https://github.com/gajus/slonik#api) the database connection pool to 10 connections. However, 10 seems to be on the low side. Right now if a request is unable to acquire a connection, there is usually an unhandled Slonik `ConnectionError` and a 500 response.

We have also observed that in some cases, trying to create a connection when the pool is full results in a connection leak. We're going to try to solve that issue more systematically, but for now, increasing the pool size should make it less likely that that issue comes up.

There has been some discussion about what number to choose instead of 10. In one Slack discussion, @issa-tseng wrote:

> 10 is lowish. but for a 1GB server it's probably about right. 20-25 is probably a reasonable default. maybe even 30.

@yanokwa wrote:

> AWS max_db_connection is 83 on DO […] its 100.

We decided on 32 partly with the thought that it's the JDBC default.

I have tried this change on the QA server. When `maximumPoolSize` was 10, I observed that there could be up to 9 concurrent requests for a large OData feed. After changing the max to 32, I observed that there could be up to 31 concurrent requests. So it looks like Central is able to utilize the new max.

This commit branches off of v1.4.2 in case we want to try to release it as a patch.

### DIFF
--- a/lib/external/slonik.js
+++ b/lib/external/slonik.js
@@ -15,6 +15,7 @@ const timestamptzTypeParser = { name: 'timestamptz', parse: (x) => new Date(x) }
 
 const slonikPool = (config) => createPool(connectionString(config), {
   captureStackTrace: false,
+  maximumPoolSize: 32,
   typeParsers: [
     createDateTypeParser(),
     createBigintTypeParser(),


lognaturel:
I did a little bit more quick reading and recollection on pool sizes and I'm feeling less excited about this change. I believe that this would increase the likelihood of Central/the db running out of memory, right? A host with 1Gb of RAM is unlikely to be able to have more than 10 connections. What's the failure mode when there's an attempt to add another connection but there's no available memory?

> on DO […] its 100.

Might be good to get a reference there. I see:

> Clusters provide 25 connections per 1 GB of RAM. 3 connections per cluster are reserved for maintenance, and all remaining connections can be allocated to connection pools.

https://docs.digitalocean.com/products/databases/postgresql/how-to/manage-connection-pools/

Also:

> Additionally, each active connection uses about 10 MB of RAM.

https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing has some good thoughts on tuning.

matthew-white:
> I believe that this would increase the likelihood of Central/the db running out of memory, right?

I don't have a deep understanding of this, but that sounds plausible to me. I feel a little reassured that the QA server (with 2 GB of RAM I believe) was able to support some heavy lifting: 31 concurrent requests for a large OData feed.

> What's the failure mode when there's an attempt to add another connection but there's no available memory?

I'm not sure. Hopefully the connection attempt would be rejected, but maybe things would grind to a halt and/or be killed.

> > Clusters provide 25 connections per 1 GB of RAM.

If that's what Digital Ocean has landed on, would it be reasonable for us to make our max 25? That'd still be a sizable increase over the current 10.

I also still think it wouldn't be hard to make this configurable, if we want to keep the default on the lower side, but then have the ability to increase it for more robust servers.

> I want to be explicit that I generally had some sort of nonsense mental model when we were talking about this.

A lot of this is new to me as well! I've been excited to learn more about this area while working on this issue.

matthew-white:
I'm going to go ahead and close this PR in favor of a new one that allows `maximumPoolSize` to be configured.

PR #480, opened 2022-05-06
improve: add ability to configure maximumPoolSize.

matthew-white:
This PR adds the ability to configure `maximumPoolSize` for Slonik. See #476 for more on why it could be useful to change the pool size. We originally planned to set `maximumPoolSize` to a constant, but ultimately decided to make it configurable. With this PR, the user can specify `maximumPoolSize` in their database config; it defaults to 10. We use the database config to connect to both Slonik and Knex, but we only need to use `maximumPoolSize` with Slonik: Knex has its own approach to pooling. That means that part of the work of this PR is forwarding `maximumPoolSize` on to Slonik but not Knex.

### DIFF
--- a/lib/external/slonik.js
+++ b/lib/external/slonik.js
@@ -15,6 +15,7 @@ const timestamptzTypeParser = { name: 'timestamptz', parse: (x) => new Date(x) }
 
 const slonikPool = (config) => createPool(connectionString(config), {
   captureStackTrace: false,
+  maximumPoolSize: config.maximumPoolSize ?? 10,
   typeParsers: [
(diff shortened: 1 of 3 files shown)


ktuite (approved):

lognaturel:
For posterity: we decided to keep the default at 10 for now because we bumped it to 25 on a server with 4gb of RAM and started seeing many more premature close errors. It's not entirely clear that the errors were related to the pool size change and it's particularly weird that they continued once the value was changed back to 10. But we didn't have more time to spend on this immediately.

PR #491, opened 2022-06-03
Updated changelog for block 15 and added missing tests

ktuite:
Added 1.5 change log to API docs...

But as I was writing about the new description field, realized that while the `PUT` endpoint is tested, the more basic Project endpoints didn't seem to have description-related tests.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -32,6 +32,18 @@ Finally, **system information and configuration** is available via a set of spec
 
 Here major and breaking changes to the API are listed by version.
 
+### ODK Central v1.5
+
+ODK Central v1.5 adds editable Project descriptions as well as more detailed information about Forms and Submissions when listing Projects (via the API and shown in new Project and Form lists on Frontend).
+
+**Added**:
+
+* New `description` field returned for each [Project](/reference/project-management/projects) that can be set or updated through `POST`/`PATCH`/`PUT` on `/projects/…`
+    * Note that for the `PUT` request, the Project's description must be included in the request. [Read more](/reference/project-management/projects/deep-updating-project-and-form-details).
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -32,6 +32,18 @@ Finally, **system information and configuration** is available via a set of spec
 
 Here major and breaking changes to the API are listed by version.
 
+### ODK Central v1.5
+
+ODK Central v1.5 adds editable Project descriptions as well as more detailed information about Forms and Submissions when listing Projects (via the API and shown in new Project and Form lists on Frontend).
```
matthew-white, line 37: "Frontend" is more of an internal name. In the API docs, we use a few different terms for Frontend, for example, "the administration website". But since these are the API docs, it might not be necessary to mention Frontend here.

docs/api.md
```diff
@@ -32,6 +32,18 @@ Finally, **system information and configuration** is available via a set of spec
...
+
+**Added**:
+
+* New `description` field returned for each [Project](/reference/project-management/projects) that can be set or updated through `POST`/`PATCH`/`PUT` on `/projects/…`
+    * Note that for the `PUT` request, the Project's description must be included in the request. [Read more](/reference/project-management/projects/deep-updating-project-and-form-details).
+* [Form extended metadata](/reference/forms/individual-form/getting-form-details) now includes a `reviewStates` object of counts of Submissions with specific review states.
+    * e.g. `{ received: 12, edited: 5, hasIssues: 3 }`
```
matthew-white, line 44: I think we should add quotes around the property names, since that's needed in JSON.

matthew-white (approved):
Looks good! Just a couple of small comments.

PR #492, opened 2022-06-03
Update multer to 1.4.5-lts.1

matthew-white:


lognaturel (approved):
Read through the `multer` thread about versions and the vulnerability we want to address. Did a quick sanity check submitting a form with multiple binary attachments from Enketo and Collect.

PR #494, opened 2022-06-12
improve 493: add  support in odata (Draft)

sadiqkhoja:
Added `$select` support in odata endpoint.

### DIFF
--- a/lib/data/odata.js
+++ b/lib/data/odata.js
@@ -104,7 +104,8 @@ const navigationLink = (schemaStack, instanceId, slicer = identity) => {
 };
 
 // manually extracts fields from a row into a js obj given a schema fieldlist.
-// NOTE: expects extended submission metadata, for exporting submitter metadata!
+// pass list of system properties in `options.metadata` that you want in return object
+// returns all system properties if `options.metadata` is falsy
 const submissionToOData = (fields, table, submission, options = {}) => new Promise((resolve) => {
   const { submitter, attachment } = submission.aux;
   const encHasData = (submission.aux.encryption != null) ? submission.aux.encryption.encHasData : false;
@@ -129,30 +130,45 @@ const submissionToOData = (fields, table, submission, options = {}) => new Promi
   // for now some odata clients will report the missing records as Errors because
   // a null field isn't provided. that's not the worst thing ever, really.
   const base = encrypted ? {} : generateDataFrame(schemaStack);
-  const root = Object.assign({ __id: submission.instanceId }, base);
+  const root = Object.assign({ }, base);
 
+  // The instanceId and submission metadata we decorate onto every root row based
+  // on option.metadata even though they are not part of the form's own schema.
+  // So rather than try to inject them into the xml transformation below, we just
+  // formulate them herein advance:
+  if (!options.metadata || options.metadata.includes('__id')) {
+    root.__id = submission.instanceId;
+  }
   if (table === 'Submissions') {
-    // the instanceId and submission metadata we decorate onto every root row even
-    // though they are not part of the form's own schema. so rather than try to
-    // inject them into the xml transformation below, we just formulate them here
-    // in advance:
-    const submitterId = (submitter.id == null) ? null : submitter.id.toString(); // => string for future-proofing
-    Object.assign(root, {
-      __system: {
-        submissionDate: submission.createdAt,
-        updatedAt: submission.updatedAt,
-        submitterId,
-        submitterName: submitter.displayName || null,
-        attachmentsPresent: attachment.present || 0,
-        attachmentsExpected: attachment.expected || 0,
-        status: encrypted ? (encHasData ? 'notDecrypted' : 'missingEncryptedFormData') : null,
-        reviewState: submission.reviewState || null,
-        deviceId: submission.deviceId || null,
-        // because of how we compute this value we don't need to default it to 0:
-        edits: submission.aux.edit.count,
-        formVersion: submission.aux.exports.formVersion
+    const systemObj = {
+      submissiondate: { key: 'submissionDate', value: submission.createdAt },
+      updatedat: { key: 'updatedAt', value: submission.updatedAt },
+      submitterid: { key: 'submitterId', value: (submitter.id == null) ? null : submitter.id.toString() }, // => string for future-proofing
+      submittername: { key: 'submitterName', value: submitter.displayName || null },
+      attachmentspresent: { key: 'attachmentsPresent', value: attachment.present || 0 },
+      attachmentsexpected: { key: 'attachmentsExpected', value: attachment.expected || 0 },
+      status: { key: 'status', value: encrypted ? (encHasData ? 'notDecrypted' : 'missingEncryptedFormData') : null },
+      reviewstate: { key: 'reviewState', value: submission.reviewState || null },
+      deviceid: { key: 'deviceId', value: submission.deviceId || null },
+      edits: { key: 'edits', value: submission.aux.edit.count },
+      formversion: { key: 'formVersion', value: submission.aux.exports.formVersion },
+    };
+
+    if (options.metadata) {
+      const systemProp = options.metadata.filter(p => p.startsWith('__system')).map(p => p.substr(9));
+      if (systemProp.length > 0) {
+        root.__system = {};
+        for (const property of systemProp) {
+          root.__system[systemObj[property].key] = systemObj[property].value;
+        }
       }
-    });
+    } else {
+      root.__system = {};
+      for (const property of Object.getOwnPropertyNames(systemObj)) {
+        root.__system[systemObj[property].key] = systemObj[property].value;
+      }
+    }
+
 
     // we always return an array result, but if we want to return the root record
     // we won't have a repeat step-in to seed the one record we'll return. so if
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -22,7 +22,7 @@ const { PartialPipe } = require('../util/stream');
 const { sanitizeOdataIdentifier } = require('../util/util');
 const { submissionToOData } = require('../data/odata');
 const { SchemaStack } = require('../data/schema');
-
+const { resolve } = require('../util/promise');
 
 ////////////////////////////////////////////////////////////////////////////////
 // UTIL
@@ -70,8 +70,14 @@ const nextUrlFor = (limit, offset, count, originalUrl) =>
     : urlWithQueryParams(originalUrl, { $skip: (offset + limit), $top: null }));
 
 // Given a querystring object, returns an object of relevant OData options. Right
-// now that is only { wkt: Bool, expand: String }
-const extractOptions = (query) => ({ wkt: isTrue(query.$wkt), expand: query.$expand });
+// now that is only { wkt: Bool, expand: String, metadata: Array }
+const extractOptions = (query) => ({
+  wkt: isTrue(query.$wkt),
+  expand: query.$expand,
+  metadata: query.$select?.split(',')
+    .map(p => p.trim().toLowerCase())
+    .filter(p => p.startsWith('__'))
+});
 
 // Given a tableParts: [ String ] and a schemaLookup: Object given by schemaAsLookup(),
 // ensures that the table implied by tableParts actually exists in schemaLookup.
@@ -374,6 +380,12 @@ const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, t
   return PartialPipe.of(inStream, parserStream);
 };
 
+const getTableFromOriginalUrl = (originalUrl) => {
+  const serviceRoot = getServiceRoot(originalUrl);
+  const targetContext = extractPathContext(urlPathname(originalUrl).slice(serviceRoot.length)); // only the service subpath
+  const tableParts = targetContext.map((pair) => pair[0]);
+  return tableParts.join('.');
+};
 
 // Given a single Submission object, returns a Stream[String] of the JSON OData response
 // for that single record at the depth specified in the originalUrl. Parameters:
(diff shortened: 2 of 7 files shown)


lib/formats/odata.js
```diff
@@ -418,6 +430,48 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+      const propSegments = property.split('/');
+      let path = '';
+      for (let i = 0; i < propSegments.length; i+=1) {
+        path += `/${propSegments[i]}`;
+        const field = fieldMap[path];
+        if (!field) throw Problem.user.propertyNotFound({ field: property });
+        // it's ok to include field with repeat type so that user can have navigation link
+        // but child field of a repeat field is not supported
+        if (field.type === 'repeat' && i < propSegments.length - 1) throw Problem.user.unsupportedODataSelectField({ field: property });
+        filteredFields.add(field);
```
matthew-white, line 467: I'm not totally sure, but I think that if the user `$select`s a field in a group, but doesn't `$select` the group, we should also add the group to `filteredFields`. For example: `$select=meta/instanceID` instead of `$select=meta,meta/instanceID`. Otherwise I think `generateDateFrame()` won't work as expected.

lib/formats/odata.js
```diff
@@ -418,6 +430,48 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+  '__system/reviewstate',
+  '__system/deviceid',
+  '__system/edits',
+  '__system/formversion',
+]);
+
+// Validates $select query parameter including metadata properties and returns list of FormFields
+const filterFields = (formFields, select, table) => {
+  const filteredFields = new Set();
+  const fieldMap = formFields.reduce((map, field) => ({ ...map, [field.path.toLowerCase()]: field }), {});
```
matthew-white, line 450: The OData name of a field can be different from the name of the field in the form: see `sanitizeOdataIdentifier()`. I'm thinking that `fieldMap` should use those OData names to map.

lib/data/odata.js
```diff
@@ -129,30 +130,45 @@ const submissionToOData = (fields, table, submission, options = {}) => new Promi
...
+      formversion: { key: 'formVersion', value: submission.aux.exports.formVersion },
+    };
+
+    if (options.metadata) {
+      const systemProp = options.metadata.filter(p => p.startsWith('__system')).map(p => p.substr(9));
+      if (systemProp.length > 0) {
+        root.__system = {};
+        for (const property of systemProp) {
+          root.__system[systemObj[property].key] = systemObj[property].value;
+        }
```
matthew-white, line 163: I think the user could specify `$select=__system` to get all the data in `__system`: in OData, `__system` is itself a property that is a `ComplexType`.

matthew-white (commented):
Thanks, @sadiqkhoja! I left a few comments below, but this is definitely looking like it's going in the right direction. (By the way, sorry for getting these comments to you late: I was busy during the day, so decided to flip things around and work in the evening.)

I think your approach of implementing `$select` for `__id` and `__system` one way (using `options.metadata`) and form fields another way (using `selectFields()`) makes a lot of sense. I like your approach of filtering the fields within the endpoint, before passing the fields to `submissionToOData()`. I was thinking that `submissionToOData()` was going to have to have more logic around which form fields were `$select`ed (I thought `generateDataFrame()` might have to change, for example). But it'd be great if that's not needed! (`submissionToOData()` is already pretty complex without that logic.) I'm pretty sure that `submissionToOData()` can handle data in the submission XML that doesn't correspond to one of the `fields` passed to `submissionToOData()`.

It looks like you've implemented `$select` so that it's case-insensitive (and I can see that that's making things more complicated in a couple of places). Is it part of the OData spec that `$select` is case-insensitive? (If so, I wonder if that's the case for `$filter` as well. At the moment, our implementation of `$filter` is case-sensitive.)

sadiqkhoja:
Thanks @matthew-white I'll incorporate your comments soon

lib/formats/odata.js
```diff
@@ -70,8 +70,14 @@ const nextUrlFor = (limit, offset, count, originalUrl) =>
...
-// now that is only { wkt: Bool, expand: String }
-const extractOptions = (query) => ({ wkt: isTrue(query.$wkt), expand: query.$expand });
+// now that is only { wkt: Bool, expand: String, metadata: Array }
+const extractOptions = (query) => ({
+  wkt: isTrue(query.$wkt),
+  expand: query.$expand,
+  metadata: query.$select?.split(',')
+    .map(p => p.trim().toLowerCase())
+    .filter(p => p.startsWith('__'))
+});
```
issa-tseng, line 80: it seems to me most of your use of the `metadata` object is as a lookup, not as a list. can you use a `{}` instead of an array here?

lib/formats/odata.js
```diff
@@ -418,6 +430,48 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+  '__system/submitterid',
+  '__system/submittername',
+  '__system/attachmentspresent',
+  '__system/attachmentsexpected',
+  '__system/status',
+  '__system/reviewstate',
+  '__system/deviceid',
+  '__system/edits',
+  '__system/formversion',
+]);
```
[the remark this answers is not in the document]
sadiqkhoja: done

lib/formats/odata.js
```diff
@@ -418,6 +430,48 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+        // but child field of a repeat field is not supported
+        if (field.type === 'repeat' && i < propSegments.length - 1) throw Problem.user.unsupportedODataSelectField({ field: property });
+        filteredFields.add(field);
+      }
+    }
+  }
+  return Array.from(filteredFields);
+};
+
+const selectFields = (query, table) => (fields) => (query.$select ? resolve(filterFields(fields, query.$select, table)) : resolve(fields));
```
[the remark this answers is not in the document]
sadiqkhoja: done

lib/formats/odata.js
```diff
@@ -418,6 +430,48 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+      if (!(property === '__id' || (table === 'Submissions' && systemFields.has(property))))
+        throw Problem.user.propertyNotFound({ field: property });
+    } else {
+      // we have to include parents fields in the result to handle grouped fields
+      const propSegments = property.split('/');
+      let path = '';
+      for (let i = 0; i < propSegments.length; i+=1) {
+        path += `/${propSegments[i]}`;
+        const field = fieldMap[path];
+        if (!field) throw Problem.user.propertyNotFound({ field: property });
```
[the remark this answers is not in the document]
sadiqkhoja: Thanks for highlighting this. I have added code to handle property selection for subtables

lib/formats/odata.js
```diff
@@ -418,6 +430,48 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+        // it's ok to include field with repeat type so that user can have navigation link
+        // but child field of a repeat field is not supported
+        if (field.type === 'repeat' && i < propSegments.length - 1) throw Problem.user.unsupportedODataSelectField({ field: property });
```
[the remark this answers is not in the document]
sadiqkhoja: Yes, you are right if $expand is used with $select then empty repeat groups are returned. 
Since only `$expand=*` is allowed, I can throw unsupported feature problem if user provides both `$select` and `$expand` or we can simply ignore `$select` option if `$expand` is also present, what do you recommend?

lib/formats/odata.js
sadiqkhoja: This is handled by adding group fields automatically to `filteredFields`

lib/formats/odata.js
sadiqkhoja: Fixed it by sanitizing the path

lib/data/odata.js
sadiqkhoja: Added this feature

lib/formats/odata.js
sadiqkhoja: Thanks, I have refactored the code

sadiqkhoja:
@matthew-white I have made following changes:

- corrected the handling of $select for subtable
- made $select case-sensitive
- added ability to provide just __system in $select
- added integration tests
- incorporated other feedback

Regarding case-sensitivity of query parameters like `$select', `$filter`, etc, the latest version (4.01) of OData specs require them to be case insensitive. In fact `$` sign is optional too. However, I couldn't find any requirement related to property names. In any case, I have removed `toLowerCase` for simplicity's sake. Maybe we can create separate issue to make OData options case-insensitivity across the board.

matthew-white (commented):
Thanks, @sadiqkhoja! This is looking good. The tests provide good coverage and look well organized.

You may have noticed this already, but it looks like one test is failing on CircleCI.

It's exciting that there's now support for `$select`ing properties in repeat group tables; I didn't think we'd be able to do that yet. I also like that we support `$select=*`. I think it was helpful for code clarity to make `$select` case-sensitive: there's already plenty going on in the code.

I also wanted to let you know that I noticed a separate issue while thinking about the code here: #496. I don't think that issue will affect your work here, but I thought I'd point it out.

lib/formats/odata.js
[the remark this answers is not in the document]
sadiqkhoja: Added validation in `endpoint.js`

lib/formats/odata.js
sadiqkhoja: Tested in `should return a single row result with selected properties` integration test

lib/data/odata.js
sadiqkhoja: will add the behavior for other complexType in a separate PR. It's quite complex

lib/formats/odata.js
```diff
@@ -418,6 +430,44 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
   });
 };
 
+// Validates $select query parameter including metadata properties and returns list of FormFields
+const filterFields = (formFields, select, table) => {
+  const filteredFields = new Set();
+  const fieldMap = formFields.reduce((map, field) => ({ ...map, [`${field.path.replace(/[^/]*$/, '')}${sanitizeOdataIdentifier(field.name)}`]: field }), {});
```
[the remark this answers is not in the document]
sadiqkhoja: done

lib/formats/odata.js
```diff
@@ -418,6 +430,44 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+  for (const property of select.split(',').map(p => p.trim())) {
+    // validate metadata properties. __system/.. properties are only valid for Submission table
+    if (property.startsWith('__')) {
+      if (!(property === '__id' || (table !== 'Submissions' && property === `__${table.replace(/^(.*)\.[^.]+\.[^.]+$/, '$1').replace(/\./g, '-')}-id`) || (table === 'Submissions' && systemFields.has(property))))
```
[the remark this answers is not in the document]
sadiqkhoja: I've simplified the condition. If user requests `__id` then I return all the IDs (parent and child)

lib/formats/odata.js
```diff
@@ -418,6 +430,44 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
...
+        throw Problem.user.propertyNotFound({ field: property });
+    } else {
+
+      let path = '';
+
+      // For subtables we have to include parents fields
+      if (table !== 'Submissions') {
+        for (const tableSegment of table.replace(/Submissions\./, '').split('.')) {
+          path += `/${tableSegment}`;
+          filteredFields.add(fieldMap[path]); // fieldMap is guaranteed to have structure field
```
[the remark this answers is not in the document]
sadiqkhoja: fixed

matthew-white:
@sadiqkhoja, I'm still taking a look at your changes (I should be done before we meet tomorrow). In the meantime, I'm noticing that there are a few small conflicts. Would you mind taking a look at those? Also, now that #545 is merged, it'd be awesome if you could verify that linting passes for the new tests.

lib/formats/odata.js
```diff
@@ -464,7 +475,16 @@ const filterFields = (formFields, select, table) => {
       }
     }
   }
-  return Array.from(filteredFields);
+
+  const filteredFieldsArr = Array.from(filteredFields);
+  // Sort the fields by `order` as Set() doesn't gurantee an order
```
matthew-white, line 480: I think `Set`s and `Map`s iterate according to insertion order, in case that helps!

lib/formats/odata.js
```diff
@@ -350,10 +350,17 @@ const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, t
       // per row, we do our asynchronous parsing, jam the result onto the
       // text resultstream, and call done to indicate that the row is processed.
       submissionToOData(fields, table, row, options).then((data) => {
+
+        const parentIdProperty = Object.keys(data[0] ?? true).filter(p => /^__.*-id$/.test(p))[0];
+
         for (const field of data) {
+
+          // if ID (of subtable) is not requested then remove it
+          const fieldRefined = (options.metadata && !options.metadata.__id) ? (({ [parentIdProperty]: __, ...item }) => item)(field) : field;
```
matthew-white, line 359: There's a utility function in lib/util/util.js named `without()` that I think would be helpful here.

lib/formats/odata.js
```diff
@@ -344,10 +350,17 @@ const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, t
       // per row, we do our asynchronous parsing, jam the result onto the
       // text resultstream, and call done to indicate that the row is processed.
       submissionToOData(fields, table, row, options).then((data) => {
+
+        const parentIdProperty = Object.keys(data[0] ?? true).filter(p => /^__.*-id$/.test(p))[0];
```
matthew-white, line 354: I had no idea that `Object.keys()` could be called with `true`! Is `true` automatically converted to an object? I feel like a ternary might be a little clearer here, something like:

```js
const parentIdProperty = data.length !== 0
  ? Object.keys(data[0]).find(p => /^__.*-id$/.test(p))
  : null;
```

(Also using `find()`, which I think is a little more performant than `.filter()[0]`.)

lib/formats/odata.js
```diff
@@ -464,7 +475,16 @@ const filterFields = (formFields, select, table) => {
       }
     }
   }
-  return Array.from(filteredFields);
+
+  const filteredFieldsArr = Array.from(filteredFields);
+  // Sort the fields by `order` as Set() doesn't gurantee an order
+  filteredFieldsArr.sort((a, b) => a.order - b.order);
+  // Resequence the order because downstream function(s) are dependent on it e.g schema.children
```
matthew-white, line 482: Ah very interesting, good catch!

lib/formats/odata.js
```diff
@@ -70,8 +69,15 @@ const nextUrlFor = (limit, offset, count, originalUrl) =>
...
-// now that is only { wkt: Bool, expand: String }
-const extractOptions = (query) => ({ wkt: isTrue(query.$wkt), expand: query.$expand });
+// now that is only { wkt: Bool, expand: String, metadata: Array }
+const extractOptions = (query) => ({
+  wkt: isTrue(query.$wkt),
+  expand: query.$expand,
+  metadata: query.$select !== '*' && query.$select?.split(',')
+    .map(p => p.trim())
+    .filter(p => p.startsWith('__'))
+    .reduce((map, field) => ({ ...map, [field]: true }), {})
```
matthew-white, line 79: I think it'd be a little better to initialize this as `Object.create(null)` instead of `{}`, since the keys are user-provided. Someone might have the bright idea of specifying `$select=__proto__`, in which case reasoning about this starts getting confusing. 😆 Or I think a `Set` would work well here!

lib/formats/odata.js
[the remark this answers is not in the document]
matthew-white: Unless the OData spec calls for that behavior, I think my preference would be just to return `__id`. I can see how it'd be nice to provide the user a way to access all IDs, but we're already providing a lot of useful `$select` functionality with this PR. I think we could come back to other IDs in a future PR.

matthew-white (commented):
These changes are looking good! I've added a few questions/comments below. Could you also document our support of `$select` in the API docs?

lib/formats/odata.js
sadiqkhoja: insertion order is based on the value of `$select`, which is defined by the user. So I have to order it here

lib/formats/odata.js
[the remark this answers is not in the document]
sadiqkhoja: done

matthew-white (approved):
Looks great! ✨ I left a couple of very minor comments. Feel free to merge after taking a look at those!

lib/data/odata.js
sadiqkhoja: added this in https://github.com/getodk/central-backend/pull/717

PR #501, opened 2022-06-24
README: add extra step to db setup instructions

alxndrsn:
Without this, I see the following error when running `make test`:

```bash
$ make test
env BCRYPT=no node node_modules/mocha/bin/mocha --recursive --exit


migration file "20181011-make-email-case-insensitive.js" failed
migration failed with error: create extension if not exists CITEXT - permission denied to create extension "citext"
  1) "before all" hook: initialize in "{root}"

  0 passing (548ms)
  1 failing

  1) "before all" hook: initialize in "{root}":
     error: create extension if not exists CITEXT - permission denied to create extension "citext"
      at Parser.parseErrorMessage (node_modules/pg-protocol/dist/parser.js:278:15)
      at Parser.handlePacket (node_modules/pg-protocol/dist/parser.js:126:29)
      at Parser.parse (node_modules/pg-protocol/dist/parser.js:39:38)
      at Socket.<anonymous> (node_modules/pg-protocol/dist/index.js:10:42)
      at Socket.emit (domain.js:475:12)
      at addChunk (internal/streams/readable.js:293:12)
      at readableAddChunk (internal/streams/readable.js:267:9)
      at Socket.Readable.push (internal/streams/readable.js:206:10)
      at TCP.onStreamRead (internal/stream_base_commons.js:188:23)



make: *** [Makefile:22: test] Error 1
$
```

To recreate, try:

```sql
ALTER USER jubilant WITH nosuperuser;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ First, create a database and user in Postgres. Either use the same settings as t
 ```sql
 CREATE USER jubilant WITH PASSWORD 'jubilant';
 CREATE DATABASE jubilant_test with owner=jubilant encoding=UTF8;
+\c jubilant_test;
+CREATE EXTENSION IF NOT EXISTS CITEXT;
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
 CREATE DATABASE jubilant with owner=jubilant encoding=UTF8;
 \c jubilant;
 CREATE EXTENSION IF NOT EXISTS CITEXT;


ktuite:
Hmmm, interesting! There are some heavy duty things happening in some of those migrations. I guess this one from 2018 (`20181011-make-email-case-insensitive.js`) is responsible for enabling the `citext` case-insensitive extension. Another migration `20190225-01-add-actor-trigram-indices.js` also adds one: `pg_trgm`. 

My guess is these failed in `make test` but not `make run` (that's what happened, right?) because the setup instructions include adding those extensions for the main `jubilant` database but not `jubilant_test`. Maybe an alternative solution is add instructions in the README for adding the extensions for `jubilant_test`, too. 

```
\c jubilant_test;
CREATE EXTENSION IF NOT EXISTS CITEXT;
CREATE EXTENSION IF NOT EXISTS pg_trgm;
```

OR if we stick with the superuser approach, we don't need the extension setup lines in the README. 

I will also add I have the dreaded "🤷‍♀️ works on my machine" experience. I wonder why?? On my own dev setup, I don't seem to need to have the user be a superuser to add new extensions. I verified the extensions (psql command `\dx`) weren't there before the migrations ran and were there afterwards. I'm on a mac, postgres looks like version 13.2.

alxndrsn:
> My guess is these failed in `make test` but not `make run` (that's what happened, right?) because the setup instructions include adding those extensions for the main `jubilant` database but not `jubilant_test`. 

Correct!  I think your alternative suggestion is much cleaner than requiring `superuser` - I'll update the PR.

ktuite (approved):

PR #502, opened 2022-06-27
Makefile: add non-file-based targets to .PHONY list

alxndrsn:
As per https://github.com/getodk/central-backend/pull/497#discussion_r906957505, I think all the `make` targets except for `node_modules` should be listed as phony.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,53 +1,68 @@
-.PHONY: test
 default: base
 
 node_modules: package.json
 	npm install
 
+.PHONY: node_version
 node_version: node_modules
 	node lib/bin/enforce-node-version.js
 
+.PHONY: migrations
 migrations: node_version
 	node lib/bin/run-migrations.js
 
+.PHONY: check-migrations
 check-migrations: node_version
 	node lib/bin/check-migrations.js
 
+.PHONY: base
 base: node_version migrations check-migrations
 
+.PHONY: run
 run: base
 	node lib/bin/run-server.js
 
+.PHONY: debug
 debug: base
 	node --debug --inspect lib/bin/run-server.js
 
+.PHONY: test
 test: node_version
 	env BCRYPT=no node node_modules/mocha/bin/mocha --recursive --exit
+.PHONY: test-full
 test-full: node_version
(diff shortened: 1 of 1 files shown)


matthew-white (commented):
The StackOverflow [question](https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile) that you linked to in [#497](https://github.com/getodk/central-backend/pull/497#discussion_r906957505) was quite helpful. From reading that, I think the reason why `test` needs to be `.PHONY` is that there's a directory named `test` in the same directory as the makefile. `lint` (for example) doesn't need to be `.PHONY`, because there's no directory or file named `lint`. From a quick look at the list here, it looks like `test` is the only command that needs to be `.PHONY`. If that's the case, I think I'd lean toward including only `test` in this list. That way, there won't be a need to update `.PHONY` each time there's a new command. However, I do think a comment here would be quite useful. What do you think?

alxndrsn:
> the reason why `test` needs to be `.PHONY` is that there's a directory named `test` in the same directory

I can see how it might seem like clutter, but personally I'd err on the side of caution, and mark any target which is phony.

> That way, there won't be a need to update `.PHONY` each time there's a new command.

Maybe a preferrable pattern would be to include `.PHONY` above each phony target?  e.g.:

```
.PHONY: base
base: node_modules migrations check-migrations

.PHONY: run
run: base
	node lib/bin/run-server.js

.PHONY: debug
debug: base
	node --debug --inspect lib/bin/run-server.js
```

matthew-white:
Sounds good! I guess even though it's extra code, it can actually help you understand how `make` is operating.

> Maybe a preferrable pattern would be to include .PHONY above each phony target?

That sounds nice! I think that'd help establish a clear pattern for whenever the next command is added.

alxndrsn:
PR updated to include `.PHONY` directly above each phony target definition

matthew-white:
Looks like there's a conflict (with #497 I think).

alxndrsn:
Conflicts resolved :+1:

matthew-white (approved):

PR #506, opened 2022-06-28
Makefile: make node_modules depend on package.json

alxndrsn:
If I've understood correctly, this means that a change in `package.json` should trigger a re-run of `make node_modules` on the next invocation of `make`.

This should be helpful e.g. when switching to a branch with different dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: test
 default: base
 
-node_modules:
+node_modules: package.json
 	npm install
 
 migrations: node_modules


alxndrsn:
Build failing due to unrelated test timeout:

```
  1 failing

  1) analytics task queries
       combined analytics
         should fill in all project.submissions queries:
     Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/repo/test/integration/other/analytics-queries.js)
      at listOnTimeout (internal/timers.js:557:17)
      at processTimers (internal/timers.js:500:7)



Makefile:24: recipe for target 'test-full' failed
make: *** [test-full] Error 1

Exited with code exit status 2
```

matthew-white (approved):
> This should be helpful e.g. when switching to a branch with different dependencies.

I've certainly had the experience of forgetting to reinstall dependencies after switching branches or after pulling, so this seems helpful. Part of me wonders whether there are cases where you wouldn't want to reinstall dependencies in that case. But I think usually this is the behavior you'd want (but might forget to do), so I think we should merge.

matthew-white:
> Build failing due to unrelated test timeout

I reran CircleCI and tests are passing now. I think I've also seen that test fail at times. That test suite calls `this.timeout()`, allowing tests to run for longer than the default:

https://github.com/getodk/central-backend/blob/80d60d783a399abc24fc2f8e652bee1f42f322d5/test/integration/other/analytics-queries.js#L737

If we end up seeing this failure often enough, we could consider further increasing that timeout.

PR #507, opened 2022-06-28
projects: return combined verbs for users with multiple roles

### LINKED ISSUE #486: Return combination of verbs if user is assigned multiple roles on the same project
This is for the unusual case (not even supported by frontend) that the same user is assigned multiple roles on the same project.

The endpoint `/v1/projects/<id>` (extended version) returns a list of verbs the user can do on that project that is computed with this `auth.verbsOn` query:https://github.com/getodk/central-backend/blob/56757cae67e0c29a3fd7b5d855c461a7d80b7b56/lib/model/query/auth.js#L54-L65 and assigned to the Project here: https://github.com/getodk/central-backend/blob/56757cae67e0c29a3fd7b5d855c461a7d80b7b56/lib/resources/projects.js#L32

The endpoint `/v1/projects?forms=true` also returns verbs for each project (added in Central v1.5), but is computed with the `Projects.getAllByAuthWithForms` query.  Currently. this query chooses the verbs of the role with the lowest ID. There are some assumptions baked in here about the lower the ID, the more permissive the role, which is true for the current set of roles, but could certainly change.

https://github.com/getodk/central-backend/blob/56757cae67e0c29a3fd7b5d855c461a7d80b7b56/lib/model/query/projects.js#L96-L105

Instead of this approach, we might want to change the query to return the combined list of verbs for both roles. A different approach to try is in this comment of the PR adding verbs to the project + form list: https://github.com/getodk/central-backend/pull/484#discussion_r873091682

From @matthew-white:
> Just to recap our conversation from yesterday, things get tricky when a user has multiple roles (for example, if a user is both an administrator and a project manager for some reason). Right now in that case, this query returns the verbs from a single role. It selects the role with the smallest id, with the thought that in Central right now, more powerful roles have had smaller ids. I think it'd be awesome if we could somehow aggregate the verbs from all the roles. That way, /v1/projects and /v1/projects/:id would return the same set of verbs for any given project. I took a look at the [list](https://www.postgresql.org/docs/9.6/functions-json.html) of JSON functions, but I didn't see a function that would do that aggregation specifically. However, one strategy seems to be to convert each element of each array to a row, then aggregate the rows. Based on a Stack Overflow [answer](https://stackoverflow.com/a/46835477/2420970) that looks promising, here's a rough attempt at such a subquery:

```sql
inner join (
  select id, jsonb_agg(distinct value) as verbs
  from (
    select projects.id, roles.verbs
    from projects
    join assignments on
      assignments."actorId" = ${actorId} and
      assignments."acteeId" in ('*', 'project', projects."acteeId")
    join roles on roles.id = assignments."roleId"
    group by projects.id, roles.verbs
  ) as project_verbs, jsonb_array_elements_text(verbs)
  group by id
  having verbs ? 'project.read'
) as filtered on filtered.id = projects.id
```

alxndrsn:
Closes #486

### DIFF
--- a/lib/model/query/projects.js
+++ b/lib/model/query/projects.js
@@ -94,16 +94,16 @@ ${extend|| sql`
     on field_keys."projectId"=projects.id`}
 ${(actorId == null) ? sql`` : sql`
 inner join
-  (select id, min("roleId") as "bestRole" from projects
+  (select id, array_agg(distinct verb) as verbs from projects
     inner join
-      (select "acteeId", role.id as "roleId", role.verbs from assignments
+      (select "acteeId", jsonb_array_elements_text(role.verbs) as verb from assignments
         inner join (select id, verbs from roles where verbs ?& array['project.read', 'form.list']) as role
           on role.id=assignments."roleId"
         where "actorId"=${actorId}) as assignment
       on assignment."acteeId" in ('*', 'project', projects."acteeId")
     group by id) as filtered
 on filtered.id=projects.id
-inner join (select id, verbs from roles) as roles_no_name on roles_no_name.id = "bestRole" `}
+`}
 where "deletedAt" is null and ${equals(options.condition)}
 order by coalesce(archived, false) asc, name asc`);
 
--- a/lib/model/query/projects.js
+++ b/lib/model/query/projects.js
@@ -97,11 +97,13 @@ inner join
   (select id, array_agg(distinct verb) as verbs from projects
     inner join
       (select "acteeId", jsonb_array_elements_text(role.verbs) as verb from assignments
-        inner join (select id, verbs from roles where verbs ?& array['project.read', 'form.list']) as role
+        inner join roles as role
           on role.id=assignments."roleId"
         where "actorId"=${actorId}) as assignment
       on assignment."acteeId" in ('*', 'project', projects."acteeId")
-    group by id) as filtered
+    group by id
+    having array_agg(distinct verb) @> array['project.read', 'form.list']
+  ) as filtered
 on filtered.id=projects.id
 `}
 where "deletedAt" is null and ${equals(options.condition)}
--- a/test/assertions.js
+++ b/test/assertions.js
@@ -280,6 +280,8 @@ should.Assertion.add('EncryptedSimpleCsv', function() {
 });
 
 should.Assertion.add('eqlInAnyOrder', function(expectedUnsorted) {
+  this.params = { operator:'to be equal in any order' };
+
   const actualSorted = [ ...this.obj ].sort();
   const expectedSorted = [ ...expectedUnsorted ].sort();
   actualSorted.should.eql(expectedSorted);
--- a/test/assertions.js
+++ b/test/assertions.js
@@ -279,7 +279,7 @@ should.Assertion.add('EncryptedSimpleCsv', function() {
   csv[4].should.eql([ '' ]);
 });
 
-should.Assertion.add('deepEqualInAnyOrder', function(expectedUnsorted) {
+should.Assertion.add('eqlInAnyOrder', function(expectedUnsorted) {
   const actualSorted = [ ...this.obj ].sort();
   const expectedSorted = [ ...expectedUnsorted ].sort();
   actualSorted.should.eql(expectedSorted);
--- a/test/assertions.js
+++ b/test/assertions.js
@@ -279,3 +279,8 @@ should.Assertion.add('EncryptedSimpleCsv', function() {
   csv[4].should.eql([ '' ]);
 });
 
+should.Assertion.add('deepEqualInAnyOrder', function(expectedUnsorted) {
+  const actualSorted = [ ...this.obj ].sort();
+  const expectedSorted = [ ...expectedUnsorted ].sort();
+  actualSorted.should.eql(expectedSorted);
+});
--- a/test/integration/api/projects.js
+++ b/test/integration/api/projects.js
@@ -346,21 +346,28 @@ describe('api: /projects', () => {
             body.verbs.should.not.containDeep([ 'project.create' ]);
           }))));
 
-    it('should return verb information with extended metadata (dave)', testService((service) =>
-      service.login('dave', (asDave) =>
-        asDave.get('/v1/projects/1')
-          .set('X-Extended-Metadata', 'true')
-          .expect(200)
-          .then(({ body }) => {
-            body.verbs.should.eqlInAnyOrder([
-              'project.read',      // from role(s): formfill, viewer
-              'form.list',         // from role(s): formfill, viewer
-              'form.read',         // from role(s): formfill, viewer
-              'submission.read',   // from role(s): viewer
-              'submission.list',   // from role(s): viewer
-              'submission.create', // from role(s): formfill
-            ]);
-          }))));
+    it('should return verb information with extended metadata (chelsea)', testService((service) =>
+      service.login('alice', (asAlice) =>
+        service.login('chelsea', (asChelsea) =>
+          asChelsea.get('/v1/users/current').expect(200)
+            .then(({ body }) => body.id)
+            .then((chelseaId) => Promise.all([
+               asAlice.post(`/v1/projects/1/assignments/viewer/${chelseaId}`).expect(200),
+               asAlice.post(`/v1/projects/1/assignments/formfill/${chelseaId}`).expect(200),
+            ]))
+            .then(() => asChelsea.get('/v1/projects/1')
+              .set('X-Extended-Metadata', 'true')
+              .expect(200)
+              .then(({ body }) => {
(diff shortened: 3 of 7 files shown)


alxndrsn:
Test failing because of array ordering.

alxndrsn:
Tests now passing - ready for review.

test/integration/api/projects.js
```diff
@@ -1219,8 +1235,7 @@ describe('api: /projects?forms=true', () => {
           body.length.should.equal(1);
           body[0].should.be.a.Project();
           const { formList, verbs } = body[0];
-          // verbs seems to have 1 more verb than it should - one is duplicated
-          verbs.length.should.be.greaterThan(40);
+          verbs.length.should.equal(40); // REVIEWER: master currently returns duplicate of 'submission.update'
```
matthew-white, line 1238: It seems like a nice effect of `array_agg(distinct)` that the duplicated verb is no longer returned over the API. (Maybe at some point we should also remove it from the database. 🤔) Can the REVIEWER comment be removed now?

test/integration/fixtures/01-users.js
```diff
@@ -8,19 +8,22 @@ module.exports = ({ Assignments, Users, Projects, bcrypt }) => {
...
   return Projects.create(proj)
     .then(() => withPasswords)
     .then((users) => mapSequential(users, Users.create))
-    .then(([ alice, bob, chelsea ]) => Promise.all([
+    .then(([ alice, bob, chelsea, dave ]) => Promise.all([
       Assignments.grantSystem(alice.actor, 'admin', '*'),
-      Projects.getById(1).then((project) => Assignments.grantSystem(bob.actor, 'manager', project.get()))
+      Projects.getById(1).then((project) => Assignments.grantSystem(bob.actor, 'manager', project.get())),
+      Projects.getById(1).then((project) => Assignments.grantSystem(dave.actor, 'formfill', project.get())),
+      Projects.getById(1).then((project) => Assignments.grantSystem(dave.actor, 'viewer', project.get())),
```
matthew-white, line 26: What do you think about calling `Projects.getById()` just once?

test/assertions.js
```diff
@@ -279,3 +279,8 @@ should.Assertion.add('EncryptedSimpleCsv', function() {
   csv[4].should.eql([ '' ]);
 });
 
+should.Assertion.add('deepEqualInAnyOrder', function(expectedUnsorted) {
```
matthew-white, line 282: I'm not totally sure whether Should.js requires this, but the other custom assertions here specify `this.params`.

lib/model/query/projects.js
```diff
@@ -94,16 +94,16 @@ ${extend|| sql`
     on field_keys."projectId"=projects.id`}
 ${(actorId == null) ? sql`` : sql`
 inner join
-  (select id, min("roleId") as "bestRole" from projects
+  (select id, array_agg(distinct verb) as verbs from projects
     inner join
-      (select "acteeId", role.id as "roleId", role.verbs from assignments
+      (select "acteeId", jsonb_array_elements_text(role.verbs) as verb from assignments
         inner join (select id, verbs from roles where verbs ?& array['project.read', 'form.list']) as role
```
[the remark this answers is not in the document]
alxndrsn: :+1: it would be helpful to have a test case reflecting this...

test/integration/fixtures/01-users.js
alxndrsn: E.g.

```
    .then(([ alice, bob, chelsea, dave ]) => Promise.all([
      Assignments.grantSystem(alice.actor, 'admin', '*'),
      Projects.getById(1).then((project) => Promise.all([
        Assignments.grantSystem(bob.actor, 'manager', project.get()),
        Assignments.grantSystem(dave.actor, 'formfill', project.get()),
        Assignments.grantSystem(dave.actor, 'viewer', project.get()),
      ])),
    ]));
```

?  Maybe more efficient?  Is it more readable?

test/assertions.js
alxndrsn: Maybe required?  https://shouldjs.github.io/#assertion-assert

I've added something.

lib/model/query/projects.js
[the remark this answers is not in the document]
matthew-white: Yeah, what I'm suggesting isn't really relevant for our current roles. 😅 You could write a test where you create a new role along these lines, then assign it to Dave. But my instinct is also that it's OK if just the current tests continue to pass.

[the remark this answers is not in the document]
matthew-white: I think I have a test getting at this! (Though possibly I should have split it into two&hellip;)

```js
it('should return the correct projects with the correct verbs', testService((service) =>
    service.login('chelsea', (asChelsea) => Promise.all([
      asChelsea.get('/v1/users/current')
        .expect(200)
        .then(({ body }) => body.id),
      asAlice.post('/v1/projects')
        .send({ name: 'Another Project' })
        .expect(200)
        .then(({ body }) => body.id)
    ])
      .then(([chelseaId, projectId]) => Promise.all([
        asAlice.post(`/v1/projects/1/assignments/viewer/${chelseaId}`)
          .expect(200),
        asAlice.post(`/v1/projects/1/assignments/app-user/${chelseaId}`)
          .expect(200),
        asAlice.post(`/v1/projects/${projectId}/assignments/app-user/${chelseaId}`)
          .expect(200)
      ]))
      .then(() => asChelsea.get('/v1/projects?forms=true')
        .expect(200)
        .then(({ body }) => {
          body.length.should.equal(1);
          const project = body[0];
          project.id.should.equal(1);
          project.verbs.should.eqlInAnyOrder([
            'project.read',     // from role(s): viewer
            'form.list',        // from role(s): viewer
            'form.read',        // from role(s): viewer, app-user
            'submission.read',  // from role(s): viewer
            'submission.list',  // from role(s): viewer
            'submission.create' // from role(s): app-user
          ]);
        }))))));
```

- This test fails on `master`, because the query there only returns verbs from a single role.
- It also fails for the latest commit of this PR (345be343478c5d5df8e3e2f3b9662e3d1357d966) because the query there only returns verbs from roles that have the verbs `project.read` and `form.list`, which the `app-user` role does not.
- It also fails for the latest commit of #535 (290d1b86df46b6400c2e1bda3794ebecc638cfb1) because the query there doesn't filter on verbs at all. /v1/projects?forms=true should only return projects for which the user can `project.read` and `form.list`, which means that it should only return a single project in this test.

I threw out this idea when we chatted today, but I think the answer is probably to remove the `WHERE` clause, as 290d1b86df46b6400c2e1bda3794ebecc638cfb1 does today, but then to also add a `HAVING` clause that does that filtering at a more outer stage, something like:

```sql
(select id, array_agg(distinct verb) as verbs from projects
  inner join -- ...
  group by id
  having array_agg(distinct verb) @> array['project.read', 'form.list']) as filtered
```

test/integration/fixtures/01-users.js
```diff
@@ -8,19 +8,22 @@ module.exports = ({ Assignments, Users, Projects, bcrypt }) => {
   const users = [
     new User({ email: 'alice@getodk.org', password: 'alice' }, { actor: new Actor({ type: 'user', displayName: 'Alice' }) }),
     new User({ email: 'bob@getodk.org', password: 'bob' }, { actor: new Actor({ type: 'user', displayName: 'Bob' }) }),
-    new User({ email: 'chelsea@getodk.org', password: 'chelsea' }, { actor: new Actor({ type: 'user', displayName: 'Chelsea' }) })
+    new User({ email: 'chelsea@getodk.org', password: 'chelsea' }, { actor: new Actor({ type: 'user', displayName: 'Chelsea' }) }),
+    new User({ email: 'dave@getodk.org', password: 'dave' }, { actor: new Actor({ type: 'user', displayName: 'Dave' }) }),
```
matthew-white, line 12: After writing the test above, I'm having second thoughts about adding Dave, since Dave is only needed for a couple of specific tests. What do you think about having Alice assign Chelsea the `formfill` and `viewer` roles on project 1, similar to how the test above is structured?

issa-tseng: i should have put more thought into the fixtures system. as it is, too many tests depend on the exact fixture values today, and it's impractical to ever edit them. probably some system where a test can elect to use a known alternate fixture would be better. there are a lot of eg form/submissions tests that every single one of them add the same test data over and over.

alxndrsn: I've dispatched with Dave.

lognaturel:
Looks like one test plus related changes to go (https://github.com/getodk/central-backend/pull/507#discussion_r927056860) and this should be complete?

lib/model/query/projects.js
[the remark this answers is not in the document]
alxndrsn: :+1: I've added the test case and corresponding fix to the PR

matthew-white (approved):

PR #508, opened 2022-06-28
Circleci node version

alxndrsn:
Ref https://github.com/getodk/central/blob/master/service.dockerfile#L1

Should tests be run against the same version used in production?


matthew-white (approved):

matthew-white:
Good catch, thanks!

PR #511, opened 2022-06-28
Add project flag to purge-forms.js

### LINKED ISSUE #487: Add project flag to purge-forms.js
I sometimes like to force a form purge by manually running `purge-forms.js` but it'd be great if I could bind it to a project space.

alxndrsn:
Closes #487?


lib/bin/purge-forms.js
```diff
@@ -16,7 +16,8 @@ const { purgeForms } = require('../task/purge');
 const cli = require('cli');
 const cliArgs = {
   force: [ 'f', 'Force any soft-deleted form to be purged right away.', 'bool', false ],
-  formId: [ 'i', 'Purge a specific form based on its id.', 'int' ]
+  formId: [ 'i', 'Purge a specific form based on its id.', 'int' ],
+  projectId: [ 'p', 'Restrict purging to a specific project.', 'int' ],
```
ktuite, line 20: `options.projectId` needs to be passed into `purgeForms` a few lines down in this same file!

ktuite (commented):
Tried running the command line tool `node lib/bin/purge-forms.js -p <project id> -f` and it was deleting more projects than it should have, but I found the problem! Everything else looks good to me.

lib/bin/purge-forms.js
alxndrsn: ooops, thanks!

ktuite (approved):
All good!

Issue #500, opened 2022-06-24
Get insights into postgres performance


alxndrsn:
I think this relates to #204 (adding `pgbadger` reports).

Postgres has a couple of handy built-ins for getting visibility into real-life performance:

* [`log_min_duration_statement`](https://www.postgresql.org/docs/9.6/runtime-config-logging.html#GUC-LOG-MIN-DURATION-STATEMENT)
* [`pg_stat_statements`](https://www.postgresql.org/docs/9.6/pgstatstatements.html)

These are very low cost, and I'm guessing could be enabled via a db migration.  For a simple intro, try https://www.crunchydata.com/blog/five-tips-for-a-healthier-postgres-database-in-the-new-year

alxndrsn:
PR for enabling `pg_stat_statements` at https://github.com/getodk/central-backend/pull/512

matthew-white:
Changing `log_min_duration_statement` via a migration does sound quite nice. If you're seeing "permission denied", is there a way that we could increase the permissions of the database user running the migrations? (Is that a bad idea?) You're just trying to alter the database, not the entire Postgres system, so it seems pretty surprising (at least to me) that the user doesn't have the permissions to do so.

matthew-white:
I took a look at the links about `pg_stat_statements` in the issue description, and it sounds good to me to use that in Central. 👍 I'll also try to read up on `log_min_duration_statement` soon.

alxndrsn:
Various approaches now in place.

PR #512, opened 2022-06-28
postgres: enable statement statistics gathering

alxndrsn:
Partial fix for #500.

### DIFF
--- a/lib/model/migrations/20220628-01-enable-pg-stat-statements.js
+++ b/lib/model/migrations/20220628-01-enable-pg-stat-statements.js
@@ -1,0 +1,14 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw('create extension if not exists pg_stat_statements');
+
+const down = () => {};
+
+module.exports = { up, down };
--- a/README.md
+++ b/README.md
@@ -49,10 +49,12 @@ CREATE DATABASE jubilant_test with owner=jubilant encoding=UTF8;
 \c jubilant_test;
 CREATE EXTENSION IF NOT EXISTS CITEXT;
 CREATE EXTENSION IF NOT EXISTS pg_trgm;
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
 CREATE DATABASE jubilant with owner=jubilant encoding=UTF8;
 \c jubilant;
 CREATE EXTENSION IF NOT EXISTS CITEXT;
 CREATE EXTENSION IF NOT EXISTS pg_trgm;
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
 ```
 
 ### Sending email


lib/model/migrations/20220628-01-enable-pg-stat-statements.js
```diff
@@ -0,0 +1,16 @@
...
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const { sql } = require('slonik');
+
+const up = (db) => { db.raw('create extension if not exists pg_stat_statements');
```
matthew-white, line 12: I think this needs a closing brace. (That might explain the CircleCI failure.)

lib/model/migrations/20220628-01-enable-pg-stat-statements.js
```diff
@@ -0,0 +1,16 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const { sql } = require('slonik');
```
matthew-white, line 10: We actually use Knex rather than Slonik for migrations! We've mostly moved away from Knex to Slonik, but migrations still use Knex.

matthew-white (commented):
I still need to read up on `pg_stat_statements`, but I thought I'd comment on a couple of small things I noticed.

lib/model/migrations/20220628-01-enable-pg-stat-statements.js
alxndrsn: :facepalm:

lib/model/migrations/20220628-01-enable-pg-stat-statements.js
alxndrsn: ...and also this isn't even used!  :facepalm: :facepalm:

alxndrsn:
@matthew-white sorry this (tiny) PR was such a mess - I've fixed the issues you mentioned and updated the PR

alxndrsn:
This will also need `shared_preload_libraries` to include `pg_stat_statements` if it does not already.

This needs to be confirmed in `central`'s docker/docker-compose setup; if it's not already set then this setting will face a similar challenge to those outlined in https://github.com/getodk/central-backend/issues/500#issuecomment-1168656710 relating to setting postgres conf for `log_min_duration_statement`.

lib/model/migrations/20220628-01-enable-pg-stat-statements.js
```diff
@@ -0,0 +1,14 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw('create extension if not exists pg_stat_statements');
```
matthew-white, line 10: Similar migrations seem to call `db.schema.raw()`:

https://github.com/getodk/central-backend/blob/4f9d19c1e5a604385a36c5a3901e1d8d0f3c4bda/lib/model/migrations/20181011-make-email-case-insensitive.js#L10-L11
https://github.com/getodk/central-backend/blob/4f9d19c1e5a604385a36c5a3901e1d8d0f3c4bda/lib/model/migrations/20190225-01-add-actor-trigram-indices.js#L10-L11


Is that needed here? Is there a difference here between `db.raw()` and `db.schema.raw()`?

matthew-white (commented):
I'm seeing this message in CircleCI:

```
migration file "20220628-01-enable-pg-stat-statements.js" failed
migration failed with error: create extension if not exists pg_stat_statements - permission denied to create extension "pg_stat_statements"
  1) "before all" hook: initialize in "{root}"

  0 passing (414ms)
  1 failing

  1) "before all" hook: initialize in "{root}":
     error: create extension if not exists pg_stat_statements - permission denied to create extension "pg_stat_statements"
      at Parser.parseErrorMessage (node_modules/pg-protocol/dist/parser.js:278:15)
      at Parser.handlePacket (node_modules/pg-protocol/dist/parser.js:126:29)
      at Parser.parse (node_modules/pg-protocol/dist/parser.js:39:38)
      at Socket.<anonymous> (node_modules/pg-protocol/dist/index.js:10:42)
      at Socket.emit (domain.js:470:12)
      at addChunk (internal/streams/readable.js:290:12)
      at readableAddChunk (internal/streams/readable.js:265:9)
      at Socket.Readable.push (internal/streams/readable.js:204:10)
      at TCP.onStreamRead (internal/stream_base_commons.js:188:23)
```

I'm actually not sure that anything needs to change in the migration though. I think you maybe just need to update .circleci/initdb.js. (The commit 963751365451c3794ec326e7ab76a6c6d743baae seems related to that.)

matthew-white:
> This needs to be confirmed in `central`'s docker/docker-compose setup

You have access to a Central server now, right? Could you check things out there? In case you end up needing to run this migration and roll it back, I also wanted to point you to lib/model/knexfile.js, which might help with that.

Issue #509, opened 2022-06-28
make migrations & make check-migrations exit cleanly when migration directory is corrupt


alxndrsn:
Commit: 0a1a59bf339cb6a498c350c6f9043e82a163794f
Node version: `v14.19.3`

I'd expect a target called `check-migrations` should be exiting with a non-zero code if it fails, but in this instance that doesn't seem to be the case.

## Steps to recreate

```
rm lib/model/migrations/20170920-01-initial.js
make check-migrations
```

## Example
```
$ make check-migrations; echo "Exited with status: $?"
node -e 'const { withDatabase, checkMigrations } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(checkMigrations);'
(node:769870) UnhandledPromiseRejectionWarning: Error: The migration directory is corrupt, the following files are missing: 20220627-01-add-random-counter.js
    at validateMigrationList (/x/odk/central-backend/node_modules/knex/lib/migrate/Migrator.js:575:11)
    at Migrator.list (/x/odk/central-backend/node_modules/knex/lib/migrate/Migrator.js:290:7)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:769870) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:769870) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Exited with status: 0
```

# `make migrations`

This target also fails but exits with `0` if there are migrations missing:

```
$ make migrations
node -e 'const { withDatabase, migrate } = require("./lib/model/migrate"); withDatabase(require("config").get("default.database"))(migrate);'
(node:71825) UnhandledPromiseRejectionWarning: Error: The migration directory is corrupt, the following files are missing: 20170920-01-initial.js, 20220627-01-add-random-counter.js
    at validateMigrationList (/x/odk/backend/node_modules/knex/lib/migrate/Migrator.js:575:11)
[...]

PR #513, opened 2022-06-30
migrations: move scripts to lib/bin

alxndrsn:
Precursor to fixing https://github.com/getodk/central-backend/issues/509

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,10 @@ node_version: node_modules
 	node lib/bin/enforce-node-version.js
 
 migrations: node_version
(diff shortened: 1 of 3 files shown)


matthew-white:
Looks like there's a conflict, I think with #497, which I just merged.

matthew-white:
I do think it helps the makefile to have moved the JavaScript to separate files in lib/bin. I'll take a closer look tomorrow and will probably merge then unless @ktuite thinks this isn't the right approach.

matthew-white (approved):

PR #515, opened 2022-07-04
worker: prevent blocking on audit table

alxndrsn:
It looks like this code should be skipping locked rows, as in its current form it looks like it falls prey to this:

> If two are started at exactly the same moment, the subSELECTs for each will be processed first. (It’s not that simple, but we can pretend for this purpose. Don’t rely on subqueries executing in any particular order for correctness). Each one scans for a row with [supplied `WHERE` query]. Both find the same row and attempt to lock it. One succeeds, one waits on the other one’s lock.

https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/

### DIFF
--- a/lib/worker/worker.js
+++ b/lib/worker/worker.js
@@ -69,7 +69,7 @@ with q as
       and ("lastFailure" is null or "lastFailure" < (now() - interval '10 minutes'))
     order by "loggedAt" asc
     limit 1
-    for update)
+    for update skip locked)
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Very interesting, I did not know about this! I would guess that we don't have enough concurrency right now for this to be a live issue. But that might change, and in general this seems like a good change to make.

PR #517, opened 2022-07-04
First pass of parsing entity xml

ktuite:
Key issues to work out:

1. **Unified parsing of form defs**

Ideally, entity parsing happens at the same time a form def is uploaded and parsed for the existing reasons (extracting form fields and their types). The form def data comes in from the request stream and is processed as as stream.

2. **Namespaces and namespace prefixes**

We're trying to support a new entities namespace: `xmlns:entities="http://www.opendatakit.org/xforms/entities"`. The prefix could be anything, not just `entities:` But given the stream reading issues above, it's tricky (but probably not impossible) to read and check for namespace mappings at the beginning of the stream and use them later. We might need to get into the core _Iteratee_ code (`lib/util/xml.js`) to do this. This might look like tracking the namespace mappings once we see them and being able to use them in node processing, i.e. looking for `entity/create` nodes of any namespace but then confirming the namespace on that node. Or we could take a different approach to this whole issue, like with a convention prefix. 

3. **Double-checking the meaning of `ref`/`saveTo` attributes on `bind`**

Related to namespace issues above, we were discussing what to call the attribute in a `bind` statement that specifies the entity prop that a form field maps to, and how critical having the `entities` namespace is. The latest notes in the spec now have this being called `saveTo`.

### DIFF
--- a/lib/data/entities.js
+++ b/lib/data/entities.js
@@ -1,0 +1,54 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const { traverseXml, findOne, findAll, and, root, node, hasAttr, attr } = require('../util/xml');
+const { submissionXmlToFieldStream } = require('./submission');
+const { Form } = require('../model/frames');
+const { construct } = require('../util/util');
+
+// turn Option[Array[Option[x]]] into Array[x]; used just below in getFormFields
+const getList = (x) => x.map((xs) => xs.map((y) => y.orElse({}))).orElse([]);
+
+const getEntityDef = (xml) => {
+  const modelNode = findOne(root('html'), node('head'), node('model'));
+  return traverseXml(xml, [
+    modelNode(findOne(node('meta'), and(node('entity'), hasAttr('dataset')))(attr('dataset'))),
+    modelNode(findAll(root(), and(node('bind'), hasAttr('ref')))(attr()))
+  ]).then(([entity, bindings]) => {
+    if (entity.isEmpty()) return null;
+    const dataset = entity.get();
+    const mapping = getList(bindings)
+      .filter((bind) => ('entities:ref' in bind))
+      .map((bind) => ({
+        path: bind.nodeset.replace(/^(\/data)/, ''),
+        entity_prop: bind['entities:ref']
+      }));
+    return { dataset, mapping };
+  });
+};
+
+const getEntityFromSub = (fields, xml) => new Promise((resolve, reject) => {
+  const values = {};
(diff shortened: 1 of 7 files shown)


lib/data/entities.js
```diff
@@ -0,0 +1,54 @@
...
+  return traverseXml(xml, [
+    modelNode(findOne(node('meta'), and(node('entity'), hasAttr('dataset')))(attr('dataset'))),
+    modelNode(findAll(root(), and(node('bind'), hasAttr('ref')))(attr()))
+  ]).then(([entity, bindings]) => {
+    if (entity.isEmpty()) return null;
+    const dataset = entity.get();
+    const mapping = getList(bindings)
+      .filter((bind) => ('entities:ref' in bind))
+      .map((bind) => ({
+        path: bind.nodeset.replace(/^(\/data)/, ''),
```
matthew-white, line 29: Based on the XForm [spec](https://getodk.github.io/xforms-spec/#primary-instance), I believe the root node won't necessarily be `<data>`: it could be anything. So here I think we can remove whatever that first element is. How about we validate it before removing it though, checking that the first element corresponds to the child of the primary instance?

lib/data/entities.js
```diff
@@ -0,0 +1,54 @@
...
+const getEntityDef = (xml) => {
+  const modelNode = findOne(root('html'), node('head'), node('model'));
+  return traverseXml(xml, [
+    modelNode(findOne(node('meta'), and(node('entity'), hasAttr('dataset')))(attr('dataset'))),
+    modelNode(findAll(root(), and(node('bind'), hasAttr('ref')))(attr()))
+  ]).then(([entity, bindings]) => {
+    if (entity.isEmpty()) return null;
+    const dataset = entity.get();
+    const mapping = getList(bindings)
+      .filter((bind) => ('entities:ref' in bind))
```
matthew-white, line 27: I don't think it should happen here, but I think we should validate this attribute somewhere (maybe when the form def is uploaded).

lib/data/entities.js
```diff
@@ -0,0 +1,54 @@
...
+const { Form } = require('../model/frames');
+const { construct } = require('../util/util');
+
+// turn Option[Array[Option[x]]] into Array[x]; used just below in getFormFields
+const getList = (x) => x.map((xs) => xs.map((y) => y.orElse({}))).orElse([]);
+
+const getEntityDef = (xml) => {
+  const modelNode = findOne(root('html'), node('head'), node('model'));
+  return traverseXml(xml, [
+    modelNode(findOne(node('meta'), and(node('entity'), hasAttr('dataset')))(attr('dataset'))),
```
matthew-white, line 21: Not sure, but should we search for this specifically within the primary instance rather than anywhere in `<model>`?

lib/data/entities.js
```diff
@@ -0,0 +1,54 @@
...
+      .map((bind) => ({
+        path: bind.nodeset.replace(/^(\/data)/, ''),
+        entity_prop: bind['entities:ref']
+      }));
+    return { dataset, mapping };
+  });
+};
+
+const getEntityFromSub = (fields, xml) => new Promise((resolve, reject) => {
+  const values = {};
```
matthew-white, line 37: Since the keys of the objects are user-defined, I think it's probably a little safer to use `Object.create(null)`.

ktuite:
FYI: This branch is getting really experimental trying to find the possible/best places to entity extraction code.

ktuite:
replaced by newer and better PRs

PR #518, opened 2022-07-12
Add default enketo config suitable for development

alxndrsn:
As per settings in https://github.com/getodk/central-frontend/blob/master/docs/enketo.md.

I think this change would be helpful for developers because:

* this brings backend, frontend & enketo-express 1 step closer to working out-of-the-box for developers
* it means backend repo is not constantly dirty when working with enketo
* it removes the risk of accidentally committing these changes
* it removes the risk of accidentally overriding these changes.  This can be particularly frustrating, as the error message(s) relating to missing enketo config can be cryptic and hard to resolve if unfamiliar.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -17,7 +17,10 @@
       "host": "localhost",
       "port": 5000
     },
-    "enketo": {},
+    "enketo": {
+      "url": "http://localhost:8005/-",
+      "apiKey": "enketorules"
+    },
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Thanks, @alxndrsn! Would you mind also removing this step from the instructions in Frontend or otherwise updating those instructions?

PR #520, opened 2022-07-12
Update Node to 14.19.3 to match production

### LINKED ISSUE #341: Use next-gen CircleCI images
I just had to update the Frontend CircleCI config to use a next-gen CircleCI image. I thought I'd add some notes for Backend as well.

CircleCI is moving to a set of new images: see https://circleci.com/docs/2.0/circleci-images/ for more information. There are new images for Node, but I don't believe there's a next-gen service image for Postgres yet.

Based on [this](https://circleci.com/docs/2.0/migrating-from-1-2/) migration guide, I also think `machine:` can be removed from the Backend CircleCI config:

https://github.com/getodk/central-backend/blob/8d1133acfa7b4c4036af096856a23d131c01a7fb/.circleci/config.yml#L2-L4

matthew-white:
Also closes #341.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,9 @@
-version: 2
-machine:
-  node:
-    version: stable
+version: 2.1
 jobs:
   build:
     docker:
-      - image: circleci/node:14.17.6
-      - image: circleci/postgres:9.6-alpine-ram
+      - image: cimg/node:14.19.3
+      - image: cimg/postgres:9.6
         environment:
           POSTGRES_PASSWORD: odktest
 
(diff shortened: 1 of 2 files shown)


yanokwa (approved):

yanokwa:
https://github.com/CircleCI-Public/cimg-postgres#ram suggests that RAM variant doesn't make a difference.

matthew-white:
> For the Backend CI, probably fine to use the default I think

I haven't thought deeply about this, but I think so? The CircleCI [docs](https://circleci.com/docs/circleci-images#circleci-base-image) say that their base image is "a brand new Ubuntu-based image designed to install the bare minimum. The next-generation convenience images are based on this image." That seems reasonable to me for what these tests are supposed to be verifying.

PR #523, opened 2022-07-13
package.json: order dev dependencies alphabetically

alxndrsn:
`npm` re-orders these every time a dev dependency is added or removed.  Making the change permanently would save a bit of fiddling when changing dependencies.


issa-tseng:
i've resisted doing this myself since it'll nuke the blame but i don't feel strongly

matthew-white (approved):

PR #526, opened 2022-07-13
Generate pgbadger report while running something else

alxndrsn:
Usage:

```bash
./with-pgbadger make test
```

etc.

### DIFF
--- a/with-pgbadger
+++ b/with-pgbadger
@@ -1,0 +1,145 @@
+#!/bin/bash -eu
+set -o pipefail
+
+usage() {
+  this="$(basename "$0")"
+  cat <<EOF
+
+$(tput bold)NAME$(tput sgr0)
+	$this
+
+$(tput bold)SYONPSIS$(tput sgr0)
+	$(tput bold)$this$(tput sgr0) [options] [...command_string]
+
+$(tput bold)DESCRIPTION$(tput sgr0)
+	Run a command, and analyse the resulting PostgreSQL logs with pgBadger.
+
+$(tput bold)OPTIONS$(tput sgr0)
+	$(tput bold)--force-settings$(tput sgr0)
+		Forces changes to PostgreSQL settings.  If this option is not set, $this will prompt before making changes to PostgreSQL settings.
+	$(tput bold)--help$(tput sgr0)
+	$(tput bold)--usage$(tput sgr0)
+		Display this message.
+
+$(tput bold)EXAMPLES$(tput sgr0)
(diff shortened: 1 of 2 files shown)


with-pgbadger
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+reportsDir='./.pgbadger'
+mkdir -p "$reportsDir"
+
+start="$(badger_date)"
+
+log "Enabling required postgres options..."
+# see: https://github.com/darold/pgbadger#postgresql-configuration
+psql >/dev/null -c 'ALTER SYSTEM SET log_min_duration_statement = 0'
```
yanokwa, line 34: Add `ALTER SYSTEM SET log_line_prefix = '%m [%p] '` to guarantee that there is a log prefix.

yanokwa: It'd be nice to have an undo on this because the logs get heavy. On my default Postgres 9.6 install, these are the defaults...

```sql
ALTER SYSTEM SET log_checkpoints = off;
ALTER SYSTEM SET log_connections = off;
ALTER SYSTEM SET log_disconnections = off;
ALTER SYSTEM SET log_lock_waits = off;
ALTER SYSTEM SET log_temp_files = -1;
ALTER SYSTEM SET log_autovacuum_min_duration = -1;
ALTER SYSTEM SET log_error_verbosity = default;
ALTER SYSTEM SET log_duration = off;
ALTER SYSTEM SET log_statement = 'none';
ALTER SYSTEM SET log_line_prefix = '';
```

alxndrsn: To do reliably, this should read the initial values for each setting and then restore these at the end of the script, regardless of success.  Care should be take to handle failure within either reading the initial settings, or restoring.

@yanokwa would it be acceptable to just communicate that settings may change, and how to restore them, as part of the script?

yanokwa: Is it easy to say settings X,Y,Z with values A,B,C will change and ask for confirmation to continue? Then add a force flag for those who want to do this in CI.

lognaturel:
This seems like it'd be worth merging. Maybe with a couple of examples in the README so we don't forget that it exists and actually know what to do with it?

alxndrsn:
@lognaturel i've added inline docs (e.g. `with-pg-badger --help`)

yanokwa (approved):
LGTM. I didn't run it, but I've visually confirmed that it does tell you want your current settings are so it's easy to restore.

Issue #469, opened 2022-03-17
Add ability to preview form definitions and edit submissions for forms that are closed or closing


matthew-white:
There are three form states: [open, closing, closed](https://docs.getodk.org/central-forms/#managing-form-lifecycle).

# How form states operate with ODK Collect and OpenRosa

The three form states map very well to ODK Collect usage:
- open: data collectors can download the blank form and make submissions.
- closing: data collectors can't download the blank form (and it will disappear from their list of forms to fill) but if they already have submissions, they can send them in.
- closed: data collectors can't download the blank form and also can't submit existing submissions.

From an [OpenRosa API](https://docs.getodk.org/openrosa/) standpoint:
- open: form def available in `formList`, submissions accepted by `submission`
- closing: form def NOT available in `formList`, submissions accepted by `submission`
- closed: form def NOT available in `formList`, submissions NOT accepted by `submission`

# How form states currently operate with Enketo and what should change

The form states don't currently work well with Enketo because we use Enketo for different purposes:
1. Collecting new data, either using the Enketo links + Central Web User auth or using Enketo links + a special token (Public Links)
2. Showing previews of published forms
3. Enabling edits of submissions

Currently form-filling links (1 above) work as desired for the three states[^1] because that usage is very similar to Collect's. It's important that this behavior stays the same. It's NOT enough to block the buttons for submissions from Central Frontend because people distribute Enketo links directly. Specifically, form-filling links should work as follows:
- If a data collector opened a form-filling Enketo link in the open state and then the state moves to closing, that data collector should be able to submit
- If a data collector opened a form-filling Enketo link in the open state and then the state moves to closed, that data collector should NOT be able to submit
- If a data collector attempts to open a form-filling Enketo link in the closed or closing state, they should see an error message ("Oops, this form doesn't exist (anymore). Most likely the owner of the form has deleted it, archived it, or disabled it. Please contact the form's owner to confirm. Detailed error: Form with ID all-widgets not found in /formList.")

Currently, previews (2 above) and edits (3 above) are also affected by form state changes but they should not be:
- It should be possible to load previews and edits regardless of the form state. This doesn't work right now because the Enketo backend makes an OpenRosa `formList` request to Central. Currently Central always returns the same `formList` contents which is filtered by form state.
[...]

Issue #505, opened 2022-06-27
Add linting to CI and/or test targets?


alxndrsn:
This has been opened before at https://github.com/getodk/central-backend/issues/14.

Currently `eslint` is not run by any of the test-related make targets, or on CircleCI.  Would it be helpful to include it?  It should be an even quicker test/fail than running `mocha` tests, so adding it as a dependency to the `test` and `test-full` targets could be helpful.

matthew-white:
We run the linter in CircleCI for Frontend, and I think that's worked well for us: that setup has been successful in catching things. I'd support adding linting to CircleCI on Backend. @ktuite, what do you think?

I feel less sure about having test-related make targets also lint: the two things feel somewhat separate to me. Then again, maybe that'd be a good way to ensure that we're consistent about both testing and linting. And with #503, linting will often be quite speedy. I could go either way! @ktuite, do you have a preference?

alxndrsn:
> I feel less sure about having test-related make targets also lint: the two things feel somewhat separate to me.

I see `eslint` as a tool for testing the correctness of the code in the same way that `mocha` is.  From this perspective, there are 3 stages of testing:

1. lint
2. unit
3. integration

matthew-white:
I could see that! I also had the thought that we could add a new command to do both, something like `lint-and-test`. But that probably doesn't make sense given that we have a few different test commands: we wouldn't want to have a separate `lint-and-test` command corresponding to each of those.

Actually, thinking of those different test commands, I'm realizing that one is specifically for unit tests and one is for integration tests. Since those two each correspond to an individual stage in your list, maybe it makes sense for them to continue to be independent. Specifically, I'm thinking that `test-unit` and `test-integration` shouldn't also lint (and `test-integration` shouldn't run `test-unit`). But I could see the case for the more expansive `test` and `test-full` commands also linting.

One other variable to throw in here is that Frontend also separates linting and testing, and I think it'd be harder there to combine the two. Would it be confusing if running tests lints in Backend but doesn't in Frontend? But Frontend also doesn't use a makefile — it uses package.json + Vue CLI — so maybe Frontend and Backend are already different enough in this area that there wouldn't be much confusion.

PR #530, opened 2022-07-17
Allow submission edits on closed forms

alxndrsn:
Address case 21 of https://github.com/getodk/central-backend/issues/469#issuecomment-1184672088

TODO:

* [x] confirm if `!drafts` check is required (it is)
* [x] add a test demonstrating it's necessity

### DIFF
--- a/lib/resources/submissions.js
+++ b/lib/resources/submissions.js
@@ -68,55 +68,57 @@ module.exports = (service, endpoint) => {
         .then(always({ code: 204, body: '' }))));
 
     // Nonstandard REST; OpenRosa-specific API.
-    service.post(path, multipart.any(), endpoint.openRosa(({ Forms, Submissions, SubmissionAttachments }, { params, files, auth, query, headers }) =>
-      Submission.fromXml(findMultipart(files).buffer)
-        .then((partial) => getForm(auth, params, partial.xmlFormId, Forms, partial.def.version)
-          .catch(forceAuthFailed)
-          .then((form) => {
-            const deprecatedId = partial.deprecatedId.orNull();
-            return (deprecatedId != null)
-              // ((found a deprecatedId))
-              ? Submissions.getAnyDefByFormAndInstanceId(form.id, deprecatedId, draft)
-                .then(getOrReject(Problem.user.deprecatedIdNotFound({ deprecatedId })))
-                .then((deprecated) => ((deprecated.current !== true)
-                  // (even if deprecated is not current, if it is still n-1 we should allow upsert on it.
-                  // so, we get the instanceId referenced as the usurping id and see if /that/ is current.)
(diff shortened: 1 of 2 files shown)


lib/resources/submissions.js
```diff
@@ -110,6 +110,7 @@ module.exports = (service, endpoint) => {
                     return SubmissionAttachments.upsert(extant, files);
                   }
                   // NEW SUBMISSION REQUEST: create a new submission and attachments.
+                  if(!draft && !form.acceptsSubmissions()) throw Problem.user.notAcceptingSubmissions();
```
lognaturel, line 113: @matthew-white we talked about linting recently. I'm surprised there isn't something to catch the lack of space after `if`?

lognaturel:
Funny that this looked like it was going to be the hard part to me. I do think that's it!

> [Add new test cases for drafts(??)](https://github.com/getodk/central-backend/pull/530/commits/39d77e5c17263e0106bc42ee9bd5e8032c196f19)

Is there still a question around that or is the (??) stale?

I'd personally find it easier to review all of the state-related changes together once it's ready so my preference would be to squash this into one commit and cherry-pick it onto whatever branch closes the original issue. That'd be a good thing to check in with @matthew-white about.

alxndrsn:
> Is there still a question around that or is the (??) stale?

?? is stale :slightly_smiling_face: 

Those tests also pass on `master` - it's currently an untested path.  If it's helpful, I can open a separate PR to merge them first.

lognaturel:
Great! Makes sense to me to keep them here as you have them. As I see it, the change makes it into more of a noteworthy path. But again, pulling in @matthew-white since that touches on process for final review.

lib/resources/submissions.js
alxndrsn: https://github.com/getodk/central-backend/issues/505 maybe?  Although there isn't currently a rule for spaces after `if`.

I'll get this fixed.

matthew-white: It does seem like it'd be helpful to add linting to CircleCI! @alxndrsn, feel free to make that change once you get the chance.

matthew-white:
I'm realizing that a version of this issue is also present for the REST API endpoint to edit a submission (not just the OpenRosa endpoint):

https://github.com/getodk/central-backend/blob/fe1c0b78e53479285a244ec3d4b704b808af3bef/lib/resources/submissions.js#L209-L216

I'm guessing that addressing that would entail a similar change to what you've already done in this PR. What do you think about making that change to both the OpenRosa and REST APIs as part of this PR? At that point, I think it'd probably make sense to merge this PR on its own, unless @alxndrsn ends up finding it easier to put everything related to #469 in a single PR.

matthew-white (approved):

matthew-white:
> I'm realizing that a version of this issue is also present for the REST API endpoint to edit a submission (not just the OpenRosa endpoint)

I'll file a separate issue about this.

PR #531, opened 2022-07-17
Allow navigating to enketo for form submission editing

alxndrsn:
Address case 17 & 18 of https://github.com/getodk/central-backend/issues/469#issuecomment-1184672088

Should not be merged before https://github.com/getodk/central-backend/pull/530; maybe the two PRs should be combined.

### TODO

- [x] update `/projects/:projectId/forms/:formId/submissions/:instanceId/edit` to allow loading submission
- [ ] update `/projects/:projectId/formList` to include closing & closed forms _when appropriate_


alxndrsn:
Deprioritised.

lognaturel:
For future reference, I believe this would enable the link to Enketo but Enketo would show an error because the form would not be in the form list. So it’s definitely a necessary change but there’s more to figure out.

PR #535, opened 2022-07-21
projects: return combined verbs for users with multiple roles

alxndrsn:


matthew-white:
Closed in favor of #507.

PR #545, opened 2022-07-25
Run eslint on whole project

### LINKED ISSUE #544: Run lint against wider set of paths
Currently `make lint` will only run eslint against the `lib/` directory.  It might be useful to run against a wider set of source files, including e.g. `test/`.

alxndrsn:
Closes #544 

I think it would be useful to run eslint across all javascript source files in the repo so that simple mistakes etc. are caught even in non-prod code.

Currently the `test/` directory requires a lot of extra rules to be applied; these could be reduced in future as specific violations are cleaned up.

If `.json` is the preferred format for `.elsintrc` files in this project, then this can be converted.

The use of `private` as a variable name in `encryption.js` prevents eslint from running against this file, as otherwise eslint will fail with the error:
```bash
odk/backend/test/integration/other/encryption.js
  84:24  error  Parsing error: The keyword 'private' is reserved
```


matthew-white:
I think running ESLint on test/ would be useful! It's definitely tricky though to figure out how to handle previous ESLint issues. Rather than turning these rules off, what do you think about disabling ESLint (or the ESLint rule) for the specific lines with issues? (Could there be a way to do so automatically?) That way, we could keep these rules on for new tests while ignoring them for old tests.

Alternatively, what do you think about making these rules `'warn'` instead of `'off'`? That way, we wouldn't be ignoring them, even for old tests, but they also wouldn't fail in CircleCI. If we did that, maybe we'd also want to run ESLint on test/ before src/ so that warnings from test/ are shown above any issues in src/.

alxndrsn:
@matthew-white I've update the PR to separate lint runs on the test dir vs the rest of the repo, and change all the test rules to `warn`.  The `make lint` output is now _very_ noisy.  If you'd prefer individual disable tags for each current violation in the test dir, that would be simple to implement.

matthew-white:
> If you'd prefer individual disable tags for each current violation in the test dir, that would be simple to implement.

If that's not too hard, I think that would be quite nice!

alxndrsn:
@matthew-white done :+1: 

I cleaned up a few simple violations (e.g. newlines at end of files)

matthew-white (approved):
Thanks, @alxndrsn! Being able to lint test/ without a lot of noisy output feels like a nice win. I'm sure we'll try to actually resolve some of these lint issues over time, and I imagine that we may also adjust our linting rules for test/. But I think this PR makes sense as the next step.

I have a couple of questions about the makefile. However, since this is such a big diff, I think it makes sense to get this merged now, then continue discussing those questions. If need be, we can follow up with a change to the makefile in another PR.

I reviewed this PR by scanning through the files changed. I only noticed ESLint comments and small changes.

test/integration/api/forms/forms.js
```diff
@@ -600,25 +602,45 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
...
                 .expect(200)
+                  // eslint-disable-next-line indent
                 .then(({ body }) => {
+                    // eslint-disable-next-line indent
                   body.enketoId.should.equal('::abcdefgh');
+                    // eslint-disable-next-line indent
                   body.enketoOnceId.should.equal('::::abcdefgh');
+                  // eslint-disable-next-line indent
                 }));
+            // eslint-disable-next-line indent
```
ktuite, line 643: There are so many `// eslint-disable-next-line indent` comments inserted at each line here and elsewhere in this diff that i'd love to see the indentation just fixed. (I think it's only this way because no linter was yelling at the author in the first place!) _Or_ have the eslint comment applied to the whole test block or file. Then we can come back and fix the indention later, but not have to remove every one of these new lines.

test/integration/api/app-users.js
```diff
@@ -38,6 +40,7 @@ describe('api: /projects/:id/app-users', () => {
             .send(testData.instances.simple.one)
             .expect(403)))));
 
+    // eslint-disable-next-line no-unused-vars
```
issa-tseng, line 43: no-unused-vars is probably easy enough to fix in every case we should ideally do them rather than commit the lint exception.

alxndrsn:
I've removed violations of `no-unused-vars` and `indent`.

test/integration/worker/worker.js
```diff
@@ -16,15 +19,15 @@ describe('worker', () => {
...
     });
 
     it('should return false and do nothing if no jobs match the event', () => {
       let called = false;
       const reschedule = () => { called = true; };
       const event = { action: 'test.event' };
-      runner({}, { other: [ () => Promise.resolve(42) ] })(event, called).should.equal(false);
+      runner({}, { other: [ () => Promise.resolve(42) ] })(event, reschedule).should.equal(false);
       called.should.equal(false);
     });
```
matthew-white, line 32: I changed these two tests, passing `reschedule` rather than `called` to the function returned by `runner()`. I think that function expects a callback.

matthew-white (approved):
Thanks, @alxndrsn!

PR #550, opened 2022-07-28
bug #478: decode html entities in form title

### LINKED ISSUE #478: Form titles with `&` not decoded properly
If you put an `&` in the XLSForm title, Central’s UI/DB shows `&amp;` while Enketo and Collect show the correct `&`.

*xlsform*
xlsform

*central ui (and db)*
central

*xml*
Screen Shot 2022-05-05 at 09 42 21

*enketo*
enketo

@matthew-white says...
> Frontend just uses what it gets from the API, so I think it’d be an issue on central-backend. We probably just need to decode the text after we extract it from the XML.

If we don't already have tests, it'd also be good to add some tests for Unicode characters. 

Potentially relevant: I vaguely remember that HTML documents don't support Unicode in the `<title>` tag.

sadiqkhoja:
Closes #478 

### Change
I have decoded only title before saving it in database

### Verification
Created unit tests
Manually tested in ODK Collect and Enketo to ensure that change is not breaking functionality there

### Alternative
I could have decoded the title in GET APIs after retrieving the data from DB but there are few of them. 
Other option was to change the `text()` function but I am afraid of ripple effect it might create across the application

### DIFF
--- a/lib/util/util.js
+++ b/lib/util/util.js
@@ -32,14 +32,6 @@ const blankStringToNull = (x) => ((x === '') ? null : x);
 const sanitizeOdataIdentifier = (name) =>
   name.replace(/^([^a-z_])/i, '_$1').replace(/([^a-z0-9_]+)/gi, '_');
 
-
-// Decodes html entities into plain text
-// greedy approach, only converting entities that are return by xlsform to xform conversion process
-const decodeHtml = (input) =>
-  input.replace(/&amp;/g, '&')
-    .replace(/&lt;/g, '<')
-    .replace(/&gt;/g, '>');
-
 ////////////////////////////////////////
 // OBJECTS
 
@@ -82,7 +74,7 @@ const construct = (Type) => (x, y) => new Type(x, y);
 
 module.exports = {
   noop, noargs,
-  isBlank, isPresent, blankStringToNull, sanitizeOdataIdentifier, decodeHtml,
+  isBlank, isPresent, blankStringToNull, sanitizeOdataIdentifier,
   printPairs, without, pickAll,
   construct
 };
--- a/lib/util/util.js
+++ b/lib/util/util.js
@@ -33,6 +33,13 @@ const sanitizeOdataIdentifier = (name) =>
   name.replace(/^([^a-z_])/i, '_$1').replace(/([^a-z0-9_]+)/gi, '_');
 
 
+// Decodes html entities into plain text
+// greedy approach, only converting entities that are return by xlsform to xform conversion process
+const decodeHtml = (input) =>
+  input.replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>');
+
 ////////////////////////////////////////
 // OBJECTS
 
@@ -75,7 +82,7 @@ const construct = (Type) => (x, y) => new Type(x, y);
 
 module.exports = {
   noop, noargs,
-  isBlank, isPresent, blankStringToNull, sanitizeOdataIdentifier,
+  isBlank, isPresent, blankStringToNull, sanitizeOdataIdentifier, decodeHtml,
   printPairs, without, pickAll,
   construct
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -769,7 +769,7 @@
     "append-field": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/append-field/-/append-field-1.0.0.tgz",
-      "integrity": "sha1-HjRA6RXwsSA9I3SOeO3XubW0PlY="
+      "integrity": "sha512-klpgFSWLW1ZEs8svjfb7g4qWY0YS5imI82dTg+QahUvJ8YqAY0P10Uk8tTyh9ZGuYEZEMaeJYCF5BFuX552hsw=="
     },
     "append-transform": {
       "version": "2.0.0",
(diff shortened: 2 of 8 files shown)


lib/util/util.js
```diff
@@ -33,6 +33,13 @@ const sanitizeOdataIdentifier = (name) =>
   name.replace(/^([^a-z_])/i, '_$1').replace(/([^a-z0-9_]+)/gi, '_');
 
 
+// Decodes html entities into plain text
+// greedy approach, only converting entities that are return by xlsform to xform conversion process
+const decodeHtml = (input) =>
+  input.replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>');
```
matthew-white, line 41: Elsewhere we use this `decodeXML()` function:

https://github.com/getodk/central-backend/blob/321ac47d182e64cfffa9fb9e4ed421c5a614cb4a/lib/data/schema.js#L24

Would it be possible to use that instead of writing our own `decodeHtml()` function?

package-lock.json
```diff
@@ -769,7 +769,7 @@
     "append-field": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/append-field/-/append-field-1.0.0.tgz",
-      "integrity": "sha512-klpgFSWLW1ZEs8svjfb7g4qWY0YS5imI82dTg+QahUvJ8YqAY0P10Uk8tTyh9ZGuYEZEMaeJYCF5BFuX552hsw=="
+      "integrity": "sha1-HjRA6RXwsSA9I3SOeO3XubW0PlY="
```
matthew-white, line 772: Is it intentional that this changed?

matthew-white:
I suspect that there are other places where we should decode XML text. For example, I'm guessing we should decode the submission instance name:

https://github.com/getodk/central-backend/blob/321ac47d182e64cfffa9fb9e4ed421c5a614cb4a/lib/model/frames/submission.js#L36-L62

XML text isn't automatically decoded. I'm not sure, but I believe that XML attributes also aren't. If that's the case, there are probably places where we should decode XML attributes (for example, the form version string).

That said, I think it's reasonable to fix the specific issue filed. @lognaturel, let me know if you want me to file a separate issue about investigating other places to decode XML text and attributes.

package-lock.json
sadiqkhoja: I ran `npm install` and it modified it.

matthew-white: Interesting! I wonder what causes `integrity` to change like that? 🤔

Since it's a separate change from this issue, what do you think about creating a new PR for that particular change?

sadiqkhoja: done

lognaturel:
> for example, the form version string

Didn't we already make a fix specifically for this?

I don't think we need to file issues to look for other places to do this. Starting with Collect v2022.3, there will be no more HTML encoding of unicode so these kinds of issues would naturally phase out anyway.

matthew-white:
> > for example, the form version string
> 
> Didn't we already make a fix specifically for this?

I don't quite remember. Did we make a change to support Unicode? But it does look like we currently don't decode `&amp;`: https://test.getodk.cloud/#/projects/388/forms/ampersand_version

> Starting with Collect v2022.3, there will be no more HTML encoding of unicode so these kinds of issues would naturally phase out anyway.

Even without encoding Unicode, I think we'll still have issues with `&`, `<`, and `>`. Maybe those aren't super common characters though?

matthew-white:
Oh and since this came up when @sadiqkhoja and I chatted, I also wanted to mention explicitly that this fix won't undertake a backfill: this will be a fix going forward for newly published forms and form versions.

lognaturel:
> Did we make a change to support Unicode? But it does look like we currently don't decode &amp;: https://test.getodk.cloud/#/projects/388/forms/ampersand_version

Right, sorry, I conflated two things! Agreed they aren't super common. I think my preference is to wait and see if anyone complains so we can learn more about why they're putting those characters in version or formid strings.

matthew-white (approved):

matthew-white:
Oh wait, don't we now need to encode the form def name in the OpenRosa `formList`? For example, if the form title is `&amp;`, it should show up in the formList as `<name>&amp;amp;</name>`, not `<name>&amp;</name>`: I think the latter would be shown as `&`, not the original `&amp;`.

Or is this not really a problem in practice?

lognaturel:
> if the form title is &amp;

Well, more importantly, if the title is Crate & Barrel, wouldn't the form list have invalid XML of `<name>Crate & Barrel</name>`? So yes, this seems like it should be accompanied by a corresponding encode unfortunately.

sadiqkhoja:
When form title is `Crate & Barrel` in database, openRosa formList shows it as `<name>Crate &amp; Barrel</name>`.

lognaturel:
> formList shows it as <name>Crate &amp; Barrel</name>.
Great, then. For Matt's example, if you put &amp; as the title in an XLSForm, it will be &amp;amp; in the XForm, &amp; in the db and &amp;amp; in the formList?

Doesn't that mean there's an encode somewhere? Was it double-encoding before? (e.g. "Crate &amp;amp; Barrel" in formList?)

sadiqkhoja:
It does double encoding. 

Fill blank form show title correctly, probably it's reading it from xml form definition:
Screenshot_20220729-145830

Get blank form shows `&amp;` in the title, probably reading from `name` field in formList response
Screenshot_20220729-145816

lognaturel:
Great, so we get to fix 2 issues with this one. Looks ready to go to me.

matthew-white:
Just as a quick check (and to prevent any future regression), how about we add a test that the OpenRosa `formList` encodes the form title/name?

matthew-white (approved):
Thanks, @sadiqkhoja! It's great to have that test in there. I'm thinking that Mustache must be escaping for us.

PR #551, opened 2022-07-28
Return Problem response if user specifies too large number

### LINKED ISSUE #514: Return Problem response if user specifies too large number
Right now, if a number is specified to Backend that is too large, Backend can return a 500 error. For example, when I send a request to /v1/projects/10000000000 locally, I receive a 500 error response with the following body:

```json
{
  "message": "Completely unhandled exception: value \"10000000000\" is out of range for type integer",
  "details": {
    "stack": [
      "error: value \"10000000000\" is out of range for type integer",
      "at Parser.parseErrorMessage (/central-backend/node_modules/pg-protocol/dist/parser.js:278:15)",
      "at Parser.handlePacket (/central-backend/node_modules/pg-protocol/dist/parser.js:126:29)",
      "at Parser.parse (/central-backend/node_modules/pg-protocol/dist/parser.js:39:38)",
      "at Socket.<anonymous> (/central-backend/node_modules/pg-protocol/dist/index.js:10:42)",
      "at Socket.emit (events.js:400:28)",
      "at Socket.emit (domain.js:470:12)",
      "at addChunk (internal/streams/readable.js:290:12)",
      "at readableAddChunk (internal/streams/readable.js:265:9)",
      "at Socket.Readable.push (internal/streams/readable.js:204:10)",
      "at TCP.onStreamRead (internal/stream_base_commons.js:188:23)"
    ]
  }
}
```

This can be seen in [this](https://sentry.io/organizations/getodk/issues/3375582060/?environment=production) Sentry error.

Instead of returning a 500 error, I think it'd be better for Backend to return a 400 Problem response. We do that already for non-numeric input: see #135 (fixed in #182).

alxndrsn:
Closes #514

### DIFF
--- a/lib/util/db.js
+++ b/lib/util/db.js
@@ -336,7 +336,15 @@ const postgresErrorToProblem = (x) => {
   if (error.isProblem === true)
     return reject(error);
 
-  if (error.code === '23502') { // not_null_violation
+  console.log('error.code:', error.code);
+
+  if (error.code === '22003') {
+    const match = /value "([^"]+)" is out of range for type (\w+)$/.exec(error.message);
+    if (match != null) {
+      const [ , value, type ] = match;
+      return reject(Problem.user.valueOutOfRangeForType({ value, type }));
+    }
+  } else if (error.code === '23502') { // not_null_violation
     return reject(Problem.user.missingParameter({ field: error.column }));
   } else if (error.code === '23505') { // unique_violation
     const match = /^Key \(([^)]+)\)=\((.+)\) already exists.$/.exec(error.detail);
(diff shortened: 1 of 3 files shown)


lib/util/db.js
```diff
@@ -336,7 +336,15 @@ const postgresErrorToProblem = (x) => {
   if (error.isProblem === true)
     return reject(error);
 
-  if (error.code === '23502') { // not_null_violation
+  console.log('error.code:', error.code);
```
matthew-white, line 339: I think the linter will complain about this `console.log()`.

matthew-white (commented):
Looks right to me! Would be great to write a test around this as well!

lib/util/db.js
alxndrsn: :+1:  removed

alxndrsn:
@matthew-white tests added :+1:

matthew-white (approved):

Issue #499, opened 2022-06-24
Intermittent test failure: managed encryption > end-to-end > should reject with a reasonable message given incorrect passphrase


alxndrsn:
I've seen this test fail once, as the only failure in the whole suite:

```
  1) managed encryption
       end-to-end
         should reject with a reasonable message given incorrect passphrase:
     Error: expected 400 "Bad Request", got 500 "Internal Server Error"
      at Test._assertStatus (node_modules/supertest/lib/test.js:266:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:281:11)
      at Test.assert (node_modules/supertest/lib/test.js:171:18)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at Server.emit (domain.js:475:12)
      at emitCloseNT (net.js:1677:8)
      at processTicksAndRejections (internal/process/task_queues.js:81:21)
```

Scrolling back up the logs:

```
::ffff:127.0.0.1 - - [23/Jun/2022:08:39:04 +0000] "POST /v1/sessions HTTP/1.1" 200 228
::ffff:127.0.0.1 - - [23/Jun/2022:08:39:04 +0000] "POST /v1/projects/1/key HTTP/1.1" 200 16
::ffff:127.0.0.1 - - [23/Jun/2022:08:39:04 +0000] "GET /v1/projects/1/forms/simple.xml HTTP/1.1" 200 1285
::ffff:127.0.0.1 - - [23/Jun/2022:08:39:04 +0000] "POST /v1/projects/1/forms/simple/submissions HTTP/1.1" 200 127
::ffff:127.0.0.1 - - [23/Jun/2022:08:39:04 +0000] "POST /v1/projects/1/forms/simple/submissions/one/attachments/submission.xml.enc HTTP/1.1" 200 16
::ffff:127.0.0.1 - - [23/Jun/2022:08:39:04 +0000] "GET /v1/projects/1/forms/simple/submissions/keys HTTP/1.1" 200 492
Error: error:0909006C:PEM routines:get_name:no start line
    at createPrivateKey (internal/crypto/keys.js:351:10)
    at /x/central-backend/lib/util/crypto.js:151:14
    at async Promise.all (index 0) {
  library: 'PEM routines',
  function: 'get_name',
  reason: 'no start line',
  code: 'ERR_OSSL_PEM_NO_START_LINE'
}Error: error:0909006C:PEM routines:get_name:no start line
[...]

PR #552, opened 2022-07-28
move test reinit to mocha's after hooks

alxndrsn:
Closes https://github.com/getodk/central-backend/issues/499

### DIFF
--- a/test/integration/other/encryption.js
+++ b/test/integration/other/encryption.js
@@ -3,7 +3,7 @@ const { readFileSync } = require('fs');
 const should = require('should');
 const { sql } = require('slonik');
 const { toText } = require('streamtest').v2;
-const { testService, testContainerFullTrx, testContainer } = require(appRoot + '/test/integration/setup');
+const { testService, testContainerFullTrx, testContainer, reinitAfter } = require(appRoot + '/test/integration/setup');
 const testData = require(appRoot + '/test/data/xml');
 const { pZipStreamToFiles } = require(appRoot + '/test/util/zip');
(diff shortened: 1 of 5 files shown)


issa-tseng:
by doing it this way aren't you now repaving the database after every test? only some tests use reinit.

matthew-white:
It'd be pretty cool if we could get `testServiceFullTrx()` and `testContainerFullTrx()` to automatically clean up after the test has finished (succeeded, failed, or timed out). I think it'd be easy to forget to call `reinitAfter()`. What do you think about an approach that uses a flag? Something like:

```js
let reinitAfter = false;
afterEach(function() {
  if (reinitAfter) {
    this.timeout(5000);
    reinitAfter = false;
    return initialize();
  }
});

// ...

const testServiceFullTrx = (test) => {
  reinitAfter = true;
  return () => test(augment(request(service(baseContainer))), baseContainer);
};
```

alxndrsn:
@matthew-white maybe the most ergonomic would be to create a wrapper function which, given a test function, will wrap it with:

```javascript
describe('...', () => {
  before(appropriateSetup);
  after(appropriateTearDown);
  it('...', () => ...test code...);
});
```

alxndrsn:
Alternative implementation at https://github.com/getodk/central-backend/pull/557

alxndrsn:
@matthew-white PR with your suggestion of [automatic cleanup](https://github.com/getodk/central-backend/pull/552#issuecomment-1199700608) at https://github.com/getodk/central-backend/pull/558.

matthew-white:
Closing now that #558 is merged!

PR #554, opened 2022-07-29
Add timing for migrations

alxndrsn:
See how long is spent re-migrating the DB in CI.

### DIFF
--- a/test/integration/setup.js
+++ b/test/integration/setup.js
@@ -68,13 +68,19 @@ const populate = (container, [ head, ...tail ] = fixtures) =>
 //
 // this hook won't run if `test-unit` is called, as this directory is skipped
 // in that case.
+let timespent = 0;
 const initialize = () => migrator
   .raw('drop owned by current_user')
-  .then(() => migrator.migrate.latest({ directory: appRoot + '/lib/model/migrations' }))
+  .then(async () => {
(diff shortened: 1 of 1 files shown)


alxndrsn:
Interesting comparison of the time this takes on CI vs locally:

environment | time spent re-migrating DB | total test time | % time spent re-migrating
---|---|---|---
CI | 8s | 6m | 2%
local | 22s | 2m | 18%

alxndrsn:
@matthew-white I'd be interested what timings you get on your machine

alxndrsn:
Closing this, as it isn't for merging.

matthew-white:
I'm actually having trouble at the moment running some of the `testContainerFullTrx()` tests. I think it's some issue with my local setup. Maybe @ktuite could get these timings on her machine?

matthew-white:
I got these timings now that I'm no longer having trouble running tests locally. I'm getting 1m total test time, with 12s spent re-migrating the database.

alxndrsn:
@matthew-white thanks for sharing - I've updated the table above.

alxndrsn:
**The different ratio observed on CI is likely because CI is running `make test-full`, and locally we're running `make test`.**

matthew-white:
Ah good call, that makes a lot of sense!

PR #556, opened 2022-07-29
Parse entity field mappings and use in submission parsing

ktuite:
Followup to PR #555 

1. Makes a small change to `getFormFields(xml)` to get the entity property mappings from the bind tags while it is processing the form's fields. 
2. Uses new Dataset schema to save dataset name and field properties parsed from form XML.
3. Grabs the Dataset property/field mappings for a specific form definition and uses them to parse entity information from submission XML.

Tests!

I was thinking about putting the dataset property parsing logic on its own in the code (or with other entity-parsing code) but after looking into how submission XML is parsed and finding it makes heavy use of Form Fields, it made sense to put it with the field parsing and have the entity property names part of Fields.

There is still more work to be done in parsing everything entity-related from a submission, like the create tag or uuid (or making one if it doesn't exist) but this PR mainly focuses on the field mappings for now. 

The only user-facing change is that if a user uploads form XML that has an <entity> tag that is deemed invalid, they will see a new error about that. Otherwise, nothing should really change for users. 

I don't think the API docs have to change yet.


lognaturel:
> maybe even as a new column in the form_fields table?

I was also wondering about that when looking at @sadiqkhoja's schema ideas earlier! Makes a lot of high-level sense to me. 😄 I think when combined with his concept of a table for entity property names there would be good answers for entity creation when parsing submissions and for getting the current schema for exports.

lib/data/schema.js
```diff
@@ -38,6 +38,9 @@ const { stripNamespacesFromPath, traverseXml, findOne, findAll, and, or, root, n
 // Turns relpaths into abspaths; leaves abspaths alone.
 const absolutify = (path, rootName) => ((path.charAt(0) === '/') ? path : `/${rootName}/${path}`);
 
+// Look for entity saveto attribute in a binding
+const saveToAttr = (binding) => Object.keys(binding).find((k) => stripNamespacesFromPath(k) === 'saveto');
```
ktuite, line 42: Double-check `saveto` vs. other ways of writing it (`save_to`, `saveTo`)

lib/data/schema.js
```diff
@@ -78,6 +81,9 @@ const _recurseFormFields = (instance, bindings, repeats, selectMultiples, envelo
       const binding = bindings.find((bind) => absolutify(bind.nodeset, envelope) === bindingPath);
       if (binding != null) {
         field.type = binding.type || 'unknown'; // binding should have a type.
+        const prop = saveToAttr(binding);
+        if (prop)
+          field.entity_prop = binding[prop];
```
ktuite, line 86: `entity_prop` -> `propertyName`
will need to put in field aux in that other place in the code where this gets used

lib/data/schema.js
```diff
@@ -124,6 +130,8 @@ const oneOf = (f, g) => (e, x, y) => Option.of(f(e, x, y).orElseGet(() => g(e, x
 //   { name: "jobs", path: "/jobs", type: "repeat", order: 1 },
 //   { name: "title", path: "/jobs/title", type: "string", order: 2 }
 //   … ]
+// For fields that map to entities, there is an additional { entity_prop: "foo" } attribute.
```
ktuite, line 133: change comment to reflect new name

lib/data/submission.js
```diff
@@ -322,5 +322,31 @@ const diffSubmissions = (structurals, versions) => {
...
+//
+// Parse submission XML and use with form field -> entity property mappings to build entities.
+
+// n.b. this wants only entity fields with filled in entity_prop attribute!
+const getEntity = (entityFields, xml) => new Promise((resolve, reject) => {
+  const values = {};
+  const stream = submissionXmlToFieldStream(entityFields, xml);
+  stream.on('error', reject);
+  stream.on('data', ({ field, text }) => {
+    if (field.path === '/meta/entity/label')
```
ktuite, line 336: should look for entity label and entity create. these will be form fields so there's probably another way to get those fields and pass them in.

test/unit/data/schema.js
```diff
@@ -1848,6 +1848,14 @@ describe('form schema', () => {
       getDataset(testData.forms.simple).then((res) => {
         res.should.equal(Option.none());
       }));
+
+    it('should extract entity properties from form field bindings', () =>
+      getFormFields(testData.forms.simpleEntity).then((res) => {
+        // name->name, age->age, hometown->(not stored in entity)
```
ktuite, line 1854: might come back and make this test more readable

ktuite (commented):
Discussions with @sadiqkhoja

lib/data/schema.js
ktuite: This [random webpage](https://docs.seneca.nl/Formulierenserver-Docs/XForms/XForms-Reference/The-bind-element.html) has some attributes in xform bind tags like `readonly` and `nodeset` (maybe that is just one word) that are all lowercase so we will stick with `saveto`.

lib/model/query/forms.js
```diff
@@ -67,12 +67,16 @@ const createNew = (partial, project, publish = false) => ({ run, FormAttachments
...
         const ids = { formId: savedForm.id, formDefId: savedForm.def.id };
-        // TODO (entities): `dataset` is either Option.none or a string of the dataset name
-        // and can now be used to update dataset info in the database.
         return Promise.all([
           FormAttachments.createNew(partial.xml, savedForm, partial.xls.itemsets),
           run(insertMany(fields.map((field) => new Form.Field(Object.assign(field, ids)))))
         ])
+          .then(() => (dataset.isDefined()
+            ? Datasets.createOrMerge(
+              new Dataset({ name: dataset.get(), projectId: savedForm.projectId }),
```
ktuite, line 76: maybe use project.id instead of savedForm.projectId

lib/model/query/forms.js
```diff
@@ -67,12 +67,16 @@ const createNew = (partial, project, publish = false) => ({ run, FormAttachments
...
+              fields.filter((field) => (field.propertyName)).map((field) => new Form.Field(field, { propertyName: field.propertyName }))
+            )
+            : null))
```
ktuite, line 79: try to rewrite as just an if without the : null

test/integration/api/forms/datasets.js
```diff
@@ -0,0 +1,126 @@
...
+      return service.login('alice', (asAlice) =>
+        asAlice.post('/v1/projects/1/forms')
+          .send(xml)
+          .set('Content-Type', 'text/xml')
+          .expect(400)
+          .then(({ body }) => { body.code.should.equal(400.23); }));
+    }));
+
+    // TODO: fix Datasets.createOrMerge when no dataset property fields to use.
+    it.skip('should parse dataset information even when there are no binds', testService((service) => {
```
ktuite, line 74: unskip this test now

lib/model/query/datasets.js
```diff
@@ -118,4 +118,22 @@ const getById = (datasetId) => ({ all }) =>
...
+const getFieldsByFormDefId = (formDefId) => ({ all }) => all(sql`
+SELECT
+  form_fields.*, ds_properties."name" as "propertyName"
+FROM
+  form_fields
+LEFT OUTER JOIN ds_property_fields ON
+  ds_property_fields."formDefId" = form_fields."formDefId"
+    AND ds_property_fields."path" = form_fields."path"
+LEFT OUTER JOIN ds_properties ON
+  ds_properties."dsPropertyId" = ds_property_fields."dsPropertyId"
```
ktuite, line 132: change `dsPropertyId` to `id`

ktuite (commented):

sadiqkhoja (approved):

PR #559, opened 2022-08-03
bugfix: comment submission response format is bizarre

### LINKED ISSUE #516: Comment submission response format is bizarre
### Request

```bash
curl 'https://example.com/v1/projects/1/forms/some_form/submissions/uuid%3A3c4b658c-12f0-48c9-a22d-cec305be99a0/comments' \
    ...auth headers etc... \
    --data-raw $'{"body":"hi"}'
```

### Response

```json
{"id":{},"submissionId":{},"actorId":{},"body":{"0":"h","1":"i"},"createdAt":{}}
```

alxndrsn:
Closes #516

### DIFF
--- a/lib/model/query/comments.js
+++ b/lib/model/query/comments.js
@@ -15,7 +15,7 @@ const { construct } = require('../../util/util');
 
 const create = (actorId, submissionId, comment) => ({ one }) =>
   one(insert(comment.with({ submissionId, actorId })))
-    .then(map(construct(Comment)));
+    .then(construct(Comment));
 
 const _get = extender(Comment)(Actor)((fields, extend, options, submissionId) => sql`
 select ${fields} from comments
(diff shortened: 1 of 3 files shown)


alxndrsn:
If adding a test, can we just remove the comment body from the response?  There doesn't seem much benefit to echoing the data that the user just sent to the server.

matthew-white:
I agree that there's not a ton of benefit. I think that's what we do with POST responses in general though, so it seems nice to be consistent.

issa-tseng:
by REST conventions you should totally reply with the data as newly represented no? as opposed to RPC etc

alxndrsn:
@matthew-white test added :+1:

matthew-white (approved):

matthew-white:
Ah looks like CircleCI is failing on the `master` branch with some small linting issues.

matthew-white:
> Ah looks like CircleCI is failing on the `master` branch with some small linting issues.

Those are now resolved as of #550 — thanks, @sadiqkhoja!

PR #546, opened 2022-07-25
Include lint in make test & test-full targets

alxndrsn:
Closes #505

I have disabled 3 more eslint rules in this PR because these rules are currently violated on `master`.  If the violations need fixing, I'd suggest doing this separately.

PR #560, opened 2022-08-03
enforce-node-version: ignore lint violation

alxndrsn:
Ignores lint violation:

```
lib/bin/enforce-node-version.js
  1:16  error  'semver' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```

An alternative fix might be to move `enforce-node-version.js` to a directory specific to dev-only scripts.


matthew-white (approved):

matthew-white:
> An alternative fix might be to move `enforce-node-version.js` to a directory specific to dev-only scripts.

I'd be open to such a directory! We could also leave the script where it is, but turn off the rule for the entire directory. But for now I'll go ahead and merge so that we can get #546 merged soon.

Issue #565, opened 2022-08-07
db.stream() has no timeout


alxndrsn:
Currently there is no timeout on database stream connections.

This can be observed by patching https://github.com/gajus/slonik/issues/347 and then creating a large number of slow, streamed connections (e.g. submission zip export), until central-backend stops responding to requests.  You can check for open connections in postgres with `SELECT COUNT(*) FROM pg_stat_activity;`.

This _may_ be because the [version of Slonik currently in-use](https://github.com/getodk/central-backend/blob/f391b6c65bd94b99fbf7f01ed69b071ff296f48a/package-lock.json#L5996-L5997) [has various timeout settings disabled](https://github.com/gajus/slonik/blob/v23.6.0/src/factories/createPoolConfiguration.ts).

Slonik does not appear to implement its own stream timeout functionality.

It [seems like](https://github.com/gajus/slonik/blob/v23.6.0/src/QueryStream.ts#L14) slonik streams use [`pg-cursor`](https://github.com/brianc/node-postgres/tree/pg-cursor%402.7.3/packages/pg-cursor) under the hood, which may use [postgres cusrors](https://www.postgresql.org/docs/9.6/plpgsql-cursors.html).  I haven't found clear documentation of how/if postgres cursors are timed out.  However, when the lockup occurs there are clear, relevant entries in both `pg_stat_activity` and `pg_locks`, but there are no rows returned by `SELECT * FROM pg_cursors;`.  This may mean that PostgreSQL Cursors are not actually being used.

Timeouts might not be necessary if central-backend reliably destroys streams in all code paths.  It's hinted this may not be the case at https://github.com/getodk/central-backend/blob/master/lib/resources/submissions.js#L282.  Either way, it might be pragmatic to add timeout handling in `lib/util/db.js`.

Related:

* https://github.com/brianc/node-postgres/issues/2103
* https://github.com/gajus/slonik/issues/347
* https://github.com/getodk/central-backend/issues/485
* https://github.com/getodk/central-backend/issues/482

issa-tseng:
yeah, the current intended mechanism is that the client timeout will terminate all active queries, not just streams. of course, if the client decides to never time out that’s a problem but i think nginx will in most cases.

matthew-white:
> if the client decides to never time out that’s a problem but i think nginx will in most cases.

[...]

PR #566, opened 2022-08-08
db.stream(): add timeout

alxndrsn:
Possible fix for https://github.com/getodk/central-backend/issues/565.

Without fixing https://github.com/gajus/slonik/issues/347 (e.g. with https://github.com/gajus/slonik/pull/360), this will still result in hanging connections sometimes, but is still a significant improvement vs current `master`.

### DIFF
--- a/lib/util/db.js
+++ b/lib/util/db.js
@@ -239,7 +239,23 @@ const queryFuncs = (db, obj) => {
     for (let i = 0; i < xs.length; i += 1) result[i] = f(xs[i]);
     return result;
   };
-  obj.stream = (s) => new Promise((resolve, reject1) => db.stream(s, resolve).catch(reject1));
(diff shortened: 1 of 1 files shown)


matthew-white (commented):
It looks like we specify `proxy_read_timeout 2m;` in the production [nginx config](https://github.com/getodk/central/blob/63cfab894cd619a703a1a38a00756f864e58b42b/files/nginx/odk.conf.template#L48). From a look at the docs for `proxy_read_timeout`, it looks like that's two minutes between reads. Should we match those settings here?

Or thinking of a different approach entirely, does nginx indicate to Backend when nginx times out the request? If Backend could automatically detect that and clean up the request and its database connections then, that seems ideal. But I imagine that's easier said than done. 😅

matthew-white:
@alxndrsn, now that we have #599 and #604, is this PR ready to be closed?

PR #578, opened 2022-08-17
Check file headers in CircleCI

matthew-white:
I noticed that a few new files are missing file headers. I think it's easy to miss those, so I wrote a small script to check that in CircleCI.

The first version of the script didn't specify anything for `skip`. For each file that the script listed, I either added a file header to the file or added the file to `skip`. I verified that when the script lists files, it exits with an exit code of 1.

We do other checks in CircleCI, so this seemed like a potentially useful addition.

Not user-facing.

No change required.

### DIFF
--- a/lib/bin/check-file-headers.js
+++ b/lib/bin/check-file-headers.js
@@ -63,6 +63,6 @@ const checkDir = (dirPath) => {
 };
 checkDir('.');
 if (missing) {
-  console.error('\nThe files above are missing file headers.');
+  console.error('\nThe files above do not have the expected file header.');
   process.exit(1);
 }
--- a/lib/bin/check-file-headers.js
+++ b/lib/bin/check-file-headers.js
@@ -1,0 +1,68 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const fs = require('fs');
+
+const header = `// Copyright 2017 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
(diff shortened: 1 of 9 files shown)


lib/external/sentry.js
```diff
@@ -1,3 +1,12 @@
+// Copyright 2018 ODK Central Developers
```
matthew-white, line 1: It looks like this file was created in 2018: 3d1064060455bc60a193ad003bcde83d7d7ff392. (The file was moved in eca4b52e8eebe52f949ae0fda632d226402f439c.)

lib/bin/check-file-headers.js
```diff
@@ -0,0 +1,68 @@
...
+// This might not be the full list in local development: .gitignore includes
+// other files. However, this list should be sufficient for CircleCI.
+const skip = new Set([
+  '.circleci',
+  '.git',
+  '.github',
+  'config',
+  'docs',
+  'lib/bin/.eslintrc.js',
+  'lib/util/quarantine/pkcs7.js',
```
matthew-white, line 31: @issa-tseng, do you know whether this file needs a file header?

lib/bin/check-file-headers.js
```diff
@@ -0,0 +1,68 @@
...
+  '.eslintrc.json',
+  '.gitignore',
+  '.npmrc',
+  'CONTRIBUTING.md',
+  'LICENSE',
+  'Makefile',
+  'NOTICE',
+  'README.md',
+  'package-lock.json',
+  'package.json'
```
matthew-white, line 43: The last time I checked, it wasn't possible to add comments in package.json.

matthew-white, line 43: @lognaturel, let me know if you think any of these files needs a file header!

matthew-white:
@alxndrsn, tagging you for review in case you have time to take a look!

matthew-white:
@ktuite and I are taking a look now! This PR may conflict with #577, so I'm thinking we should try to merge this first.

ktuite (approved):

matthew-white:
I'm going to go ahead and merge, since I think this is an improvement, and I want to be able to resolve any conflicts with #577. @lognaturel and @issa-tseng, let me know if there are any other files that you think I should add a file header to!

lib/bin/check-file-headers.js
alxndrsn: I wouldn't expect any JSON file to allow comments.  In the current form, the header is probably only applicable to `.js` files

matthew-white: Comments in a JSON file would definitely be nonstandard (though I feel like I've seen it somewhere? 🤔). And agreed about the current header only really working for .js files. I'll probably add a similar script to Frontend at some point. But Frontend has .js and .vue files with slightly different headers, so the script will have to account for that somehow.

PR #589, opened 2022-08-30
new: Added hasDataset field on GET extended form

sadiqkhoja:
Added dataset_form_defs table that maintains relationship between datasets and form definitions
Added `entityRelated` flag in /forms/:id API, this flag will help frontend to know which forms has associated dataset i.e. it creates, updates or delete entities

Updated integration tests

I could have used ds_property_fields table to check if there's a mapping for a given form definition but that would not work for those forms which are related to a dataset without creating any properties.

GET form APIs now return `entityRelated` flag, ensure frontend doesn't break with this new flag

next PR

### DIFF
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -175,7 +175,7 @@ Form.Def = Frame.define(
   'sha256',      readable,              'draftToken',   readable,
   'enketoId',    readable,              'createdAt',
   'publishedAt', readable,              'xlsBlobId',
-  'name',        readable,              
+  'name',        readable
 );
 Form.Xml = Frame.define(into('xml'), 'xml');
 
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -144,11 +144,12 @@ Form.Extended = class extends Frame.define(
   'excelContentType',   readable,
   // counts of submissions in various review states
   'receivedCount',                      'hasIssuesCount',
-  'editedCount'
+  'editedCount',                        'hasDataset',  readable
 ) {
   forApi() {
     return {
       submissions: this.submissions || 0,
+      hasDataset: this.hasDataset || false,
       reviewStates: {
         received: this.receivedCount || 0,
         hasIssues: this.hasIssuesCount || 0,
@@ -174,7 +175,7 @@ Form.Def = Frame.define(
   'sha256',      readable,              'draftToken',   readable,
   'enketoId',    readable,              'createdAt',
   'publishedAt', readable,              'xlsBlobId',
-  'name',        readable,              'managesEntities',  readable
+  'name',        readable,              
 );
 Form.Xml = Frame.define(into('xml'), 'xml');
 
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -174,7 +174,7 @@ Form.Def = Frame.define(
   'sha256',      readable,              'draftToken',   readable,
   'enketoId',    readable,              'createdAt',
   'publishedAt', readable,              'xlsBlobId',
-  'name',        readable
+  'name',        readable,              'managesEntities',  readable
 );
 Form.Xml = Frame.define(into('xml'), 'xml');
 
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -144,12 +144,12 @@ Form.Extended = class extends Frame.define(
   'excelContentType',   readable,
   // counts of submissions in various review states
   'receivedCount',                      'hasIssuesCount',
-  'editedCount',                        'hasDataset',  readable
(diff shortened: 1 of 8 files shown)


lib/model/frames/form.js
```diff
@@ -144,11 +144,12 @@ Form.Extended = class extends Frame.define(
   'excelContentType',   readable,
   // counts of submissions in various review states
   'receivedCount',                      'hasIssuesCount',
-  'editedCount'
+  'editedCount',                        'hasDataset',  readable
```
ktuite, line 147: I like the new approach of `hasDataset` being part of the Form's extended metadata -- it's not too invasive of the existing Form infrastructure (e.g. migrating the entire form_defs table to add a new column) yet it is in an appropriate place for Frontend to access this info.

lib/model/frames/form.js
```diff
@@ -144,11 +144,12 @@ Form.Extended = class extends Frame.define(
...
 ) {
   forApi() {
     return {
       submissions: this.submissions || 0,
+      hasDataset: this.hasDataset || false,
```
ktuite, line 152: You will need to add `hasDataset` to the form extended metadata API docs, though. Doesn't have to be right now, but we might want to make a note so we don't forget.

lib/model/migrations/20220803-01-create-entities-schema.js
```diff
@@ -7,37 +7,41 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-const up = (db) =>
-  db.schema.createTable('datasets', (datasets) => {
+const up = async (db) => {
```
ktuite, line 10: I like the cleaner style of the async/await migrations :D

lib/model/migrations/20220830-01-create-dataset_def.js
```diff
@@ -0,0 +1,25 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.schema.createTable('dataset_defs', (t) => {
```
ktuite, line 11: Nit-picking the `dataset_defs` table name... I guess if there's no definition stored in that table, it could have a different name? `dataset_sources`? `dataset_form_sources`? `dataset_form_defs`? I'm not sure.

The closest example I can think of in the existing code is the `submission_attachments` table that stores the `blobId` <---> `submissionDefId` relation with some other stuff. Well, that stuff like attachment `name` is pretty important, maybe this is a bad example. the `assignments` table is also pretty minimal and just captures links between things.

If you can't think of a better name, the current table name is fine.

lognaturel:
Apologies for adding a nitpick. I wonder about the `has` prefix -- I'm not sure I'm seeing a "has-a" relationship here. Could we consider something like `datasetRelated` or `entityRelated`? If I understand correctly, that's how I would describe this category of forms: 
- "This registration form creates entities. This other registration form is not entity-related and we'll have to do some post-processing ourselves for the follow-up form."
- "This registration form populates a dataset. This other registration form is not dataset-related and we'll have to do some post-processing ourselves for the follow-up form."

lib/model/migrations/20220830-01-create-dataset_def.js
lognaturel: Without looking at code, I would expect `dataset_defs` do define the dataset's schema.

sadiqkhoja:
@lognaturel The intention of this flag is to identify if a particular form has `<entities:entity>` tag so that frontend can show "hint about testing Datasets in alpha". It will be true for all forms that can create/update/archive entities -- regardless of how many entity properties are defined in the form definition. 

Should I call this flag `entityRelated` or `datasetRelated`?

lognaturel:
> Should I call this flag entityRelated or datasetRelated?

I don't have a strong preference because they feel conceptually the same here. The form's role is to update one or more entity. That entity/those entities are contained in datasets. Given that you're using the presence of an `entity` tag to determine its value, `entityRelated` seems like a good bet.

lib/model/migrations/20220830-01-create-dataset_def.js
sadiqkhoja: renamed the table to `dataset_form_defs`

lib/model/frames/form.js
sadiqkhoja: will update API docs in next PR

PR #591, opened 2022-09-01
Node16

sadiqkhoja:
## Changes

- Upgraded node to v16
- Upgraded mocha to v10, fixes .zip related unit tests
- Replaced `Ramdajs.merge` (deprecated) with `Ramdajs.mergeRight`

## Status

- **ESLint**: no warnings/errors ✅
- **Unit tests**: All passed ✅
- **Integration tests**: All passed ✅

## Is it ready to be merged?
Yes

## History

- ~~Handle invalidInputError thrown by slonik~~ 
  Not required with slonik v23.6.0 (v23.9.0 throws SlonikError on invalid input)
- ~~Rearranged `streamForExport` invocation in `odata`, `.csv` and `.csv.zip` endpoints~~ 
  Something is wrong with slonik v23.9.0, using previously known good version i.e. v23.6.0

### DIFF
--- a/lib/http/endpoint.js
+++ b/lib/http/endpoint.js
@@ -18,7 +18,7 @@
 const { inspect } = require('util');
 const { pipeline } = require('stream');
 const { DateTime } = require('luxon');
-const { reduce } = require('ramda');
+const { reduce, pluck } = require('ramda');
 const { openRosaError } = require('../formats/openrosa');
 const { odataXmlError } = require('../formats/odata');
 const { noop, isPresent } = require('../util/util');
@@ -76,6 +76,24 @@ class Context {
   with(other) { return new Context(this, other); }
 }
 
+// pulls out the data we want from the node request object. we do this as of node
+// v16 both because it saves data weight and because in v16 they changed the
+// request.headers property to be calculated, so we have to copy it manual.
+const getRequestContext = (request) => ({
+  method: request.method,
+  url: request.url,
+  originalUrl: request.originalUrl,
+  headers: request.headers,
+  length: request.length,
+  body: request.body,
+  params: request.params,
+  query: request.query,
+  files: request.files,
+
+  apiVersion: request.apiVersion,
+  fieldKey: request.fieldKey
+});
+
 
 ////////////////////////////////////////////////////////////////////////////////
 // ENDPOINT
@@ -129,7 +147,7 @@ const endpointBase = ({ preprocessor = noop, before = noop, resultWriter, errorW
     reduce(
       (chain, f) => chain.then((context) =>
         ensurePromise(f(localContainer, context, request)).then((result) => result || context)),
-      resolve(new Context(request)),
+      resolve(new Context(getRequestContext(request))),
       [ preprocessor ].concat(preprocessors)
     )
 
--- a/lib/http/endpoint.js
+++ b/lib/http/endpoint.js
@@ -18,7 +18,7 @@
 const { inspect } = require('util');
 const { pipeline } = require('stream');
 const { DateTime } = require('luxon');
-const { reduce, pluck } = require('ramda');
+const { reduce } = require('ramda');
 const { openRosaError } = require('../formats/openrosa');
 const { odataXmlError } = require('../formats/odata');
 const { noop, isPresent } = require('../util/util');
--- a/lib/resources/odata.js
+++ b/lib/resources/odata.js
@@ -64,11 +64,11 @@ module.exports = (service, endpoint) => {
           const options = QueryOptions.fromODataRequest(params, query);
           return Promise.all([
             Forms.getFields(form.def.id),
-            (((params.table === 'Submissions') && options.hasPaging())
-              ? Submissions.countByFormId(form.id, draft, options) : resolve(null)),
-            Submissions.streamForExport(form.id, draft, undefined, options) // Node16 upgrade: streaming promise has to be last in the array, I don't understand the reason.
+            Submissions.streamForExport(form.id, draft, undefined, options),
+            ((params.table === 'Submissions') && options.hasPaging())
+              ? Submissions.countByFormId(form.id, draft, options) : resolve(null)
           ])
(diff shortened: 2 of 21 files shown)


lib/resources/odata.js
```diff
@@ -64,11 +64,11 @@ module.exports = (service, endpoint) => {
           const options = QueryOptions.fromODataRequest(params, query);
           return Promise.all([
             Forms.getFields(form.def.id),
-            Submissions.streamForExport(form.id, draft, undefined, options),
-            ((params.table === 'Submissions') && options.hasPaging())
-              ? Submissions.countByFormId(form.id, draft, options) : resolve(null)
+            (((params.table === 'Submissions') && options.hasPaging())
+              ? Submissions.countByFormId(form.id, draft, options) : resolve(null)),
+            Submissions.streamForExport(form.id, draft, undefined, options) // Node16 upgrade: streaming promise has to be last in the array, I don't understand the reason.
```
sadiqkhoja, line 69: hi @issa-tseng, do you know what could be the reason of this?

alxndrsn: @sadiqkhoja it may be unrelated to order.  To debug could it help to unroll this array and run the queries sequentially?

sadiqkhoja: Slonik v23.9.0 is the culprit, using v23.6.0. Thanks @alxndrsn  for helping me out

package.json
```diff
@@ -38,18 +38,18 @@
...
     "slonik-sql-tag-raw": "1.0.3",
     "tmp-promise": "~3",
     "uuid": "~3",
-    "yauzl": "~2.9"
+    "yauzl": "~2.10"
   },
   "devDependencies": {
     "app-root-path": "~2",
-    "eslint": "~8",
+    "eslint": "~7",
```
alxndrsn, line 49: This one looks like it's decreased :thinking:

package.json
```diff
@@ -38,18 +38,18 @@
     "prompt": "~1",
     "ramda": "~0",
     "sanitize-filename": "~1",
-    "slonik": "~23",
+    "slonik": "^23.6.0",
```
alxndrsn, line 41: This might be safer as
```
    "slonik": "23.6.0",
```

lib/resources/submissions.js
```diff
@@ -300,7 +300,7 @@ module.exports = (service, endpoint) => {
             Keys.getDecryptor(passphrases),
             draft ? null : Audits.log(auth.actor, 'form.submission.export', form)
           ])
-            .then(([ fields, rows, selectValues, decryptor ]) => {
+            .then(([fields, rows, selectValues, decryptor]) => {
```
alxndrsn, line 303: Is this just changing the formatting?

lib/http/endpoint.js
```diff
@@ -76,6 +76,24 @@ class Context {
...
 }
 
+// pulls out the data we want from the node request object. we do this as of node
+// v16 both because it saves data weight and because in v16 they changed the
+// request.headers property to be calculated, so we have to copy it manual.
+const getRequestContext = (request) => ({
+  method: request.method,
+  url: request.url,
+  originalUrl: request.originalUrl,
+  headers: request.headers,
```
alxndrsn, line 86: I'm struggling to understand how this works... does `request` now have a getter which returns the entire headers array?

package.json
sadiqkhoja: `eslint-config-airbnb-base` has dependency on eslint: "^5.16.0 || ^6.8.0 || ^7.2.0"

alxndrsn (approved):

lib/resources/submissions.js
sadiqkhoja: yes

lib/http/endpoint.js
sadiqkhoja: that's right

lib/resources/odata.js
matthew-white: Do we know what the issue with Slonik v23.9.0 is? Some issue with streams that appears in Node 16 but not Node 14? (Or does v23.9.0 not even work in Node 14?)

Or more importantly, do we know whether the issue is fixed in later versions of Slonik? Just want to check that merging this PR wouldn't affect our ability to upgrade Slonik as part of the next release.

package.json
matthew-white: It looks like the latest version of `eslint-config-airbnb-base` supports ESLint v8. What do you think about upgrading `eslint-config-airbnb-base` instead? That could also happen in a follow-up PR.

lib/resources/odata.js
sadiqkhoja: I have narrowed down the issue to a commit level: https://github.com/gajus/slonik/commit/72f97820f9fe5f83a8820029d8ead64189f48360

It was released in v23.6.3, so we can't use anything latest than that. Since we have a PR for migrating postgres.js, I don't think I should invest more time in slonik

package.json
sadiqkhoja: `eslint-config-airbnb-base@15` has added new rules that impact lots of files in our codebase, I can handle that in a separate PR

lib/resources/odata.js
matthew-white: Looking at that commit, it sort of seems like Slonik is waiting for the active query to complete before executing the next query. That would explain why shuffling the promise order had an effect: the stream won't complete until it's processed, after the `Promise.all()`, which means that the stream is blocking the query following it. The .csv.zip endpoint has multiple streams, so that seems potentially tricky.

Could it be that this is fixed in a later version of Slonik? If the explanation above is accurate, it seems like that would also be happening in Node 14. But from @alxndrsn's benchmarking, it seems like later versions of Slonik are working for us in Node 14.

sadiqkhoja: Slonik v23.6.3 doesn't work with Node 14 as well. I think issue has been fixed in later versions but those versions require other changes as well, we can upgrade Slonik in separate branch or merge postgresql.js branch

Issue #521, opened 2022-07-12
Uniquely identify servers in Sentry


yanokwa:
There are certain self-hosted servers we support that have known problems (e.g., premature close). In the Sentry UI, we typically ignore these errors, but it would be better if we could ignore them per server (or in Sentry terminology, per user) so we are alerted when new servers have the problem.

https://docs.sentry.io/platforms/javascript/enriching-events/identify-user/ suggests that we can add an internal identifier for the user. I would propose we use the Docker hostname of the service because it is randomly generated and unique (e.g., `3154baf7bA72`) on servers we don't support and explicitly set on servers (`nafundi-01`) we do support.

matthew-white:
Right now, `server_name` is sometimes set in Sentry and sometimes not (for example, #483). When `server_name` is set, I believe it's the Docker hostname, right? If we found a way to consistently set `server_name`, would that also work? It looks like you can filter on `server_name` within Sentry.

yanokwa:
Yes, `server_name` is the Docker hostname. Setting it consistently would be helpful for sure, but I think that's a separate issue. What I'm asking about in this issue is to be able to ignore an issue until it affects $n users or an additional user.

matthew-white:
Closed by #593.

PR #593, opened 2022-09-01
sentry: set 'user' to differntiate deployments

alxndrsn:
Ref https://github.com/getodk/central-backend/issues/521

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -1,7 +1,6 @@
 const { inspect } = require('util');
 const { isBlank } = require('../util/util');
 
-
 // Endpoints where query strings are sensitive and should be removed
 const sensitiveEndpoints = [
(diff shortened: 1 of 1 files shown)


lib/external/sentry.js
```diff
@@ -128,6 +127,11 @@ const init = (config) => {
       return null; // we have a user-space problem.
     }
   });
+
+  const { DOMAIN } = process.env;
+  const username = DOMAIN?.endsWith('.getodk.cloud') ? DOMAIN : require('os').hostname();
```
matthew-white, line 132: We're trying where possible to avoid having logic specific to ODK Cloud in the codebase. Could we use the hostname all the time? It's set to a useful value in ODK Cloud, I believe.

alxndrsn: :+1:  updated

matthew-white:
Rebased before merging into the `release` branch.

PR #599, opened 2022-09-07
Patch slonik to close streams more reliably

### LINKED ISSUE #482: 10 concurrent requests for OData feed results in database connection leak
To reproduce:

- Run `select count(*) from pg_stat_activity where state <> 'idle'`. The count should be 1 (corresponding to this query itself).
- Send 10 concurrent requests for a large OData feed, specifying the `$skip` query parameter. I used the [Survey](https://test.getodk.cloud/#/projects/149/forms/Survey) form on the QA server. (See below for a simple Bash script to send these requests.)
- 9 responses will be 200s. One will be a 500. Specifically, the one returns the following Problem response:
```
{"message":"Completely unhandled exception: timeout exceeded when trying to connect","details":{"stack":["ConnectionError: timeout exceeded when trying to connect","at Object.createConnection (/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js:54:23)","at runMicrotasks (<anonymous>)","at runNextTicks (internal/process/task_queues.js:60:5)","at listOnTimeout (internal/timers.js:526:9)","at processTimers (internal/timers.js:500:7)","at async Promise.all (index 0)"]}}
```
- Run the count query again. The count is now 2.
- Each time you repeat the above, the count will increase by 1. Once the count is 11, the database connection pool will be full, and all requests will result in a 500 response.

When I added more logging to the endpoint for the OData feed, I observed that in the request that failed, the first two queries in the following `Promise.all()` returned resolved Promises. However, the third query was unable to acquire a database connection (see #477) and returned a rejected promise:

https://github.com/getodk/central-backend/blob/d71a438f6c4ca823e46977481d26abc2a22ebd6b/lib/resources/odata.js#L65-L72

Because the third query returned a rejected promise, the `Promise.all()` was rejected, which means that the `then()` callback was never called. That means that a stream is successfully created, but the `then()` callback never processes it. As a result, I'm thinking the stream is never cleaned up and won't let go of its connection. If that's the case, I think we would need to find a way to end the stream even if the `Promise.all()` is rejected.

<hr>

Simple Bash script. Before running it, set `$ODK_CENTRAL_TOKEN` to a session token.

```bash
function request-odata() {
	echo "sending request for \$skip=$1..."
	local url="https://test.getodk.cloud/v1/projects/149/forms/Survey.svc/Submissions?\$skip=$1"
	local response
	response=$(curl -H "Authorization: Bearer $ODK_CENTRAL_TOKEN" --include --silent --show-error "$url") || return $?
	echo "response for \$skip=$1:"
	local head=$(head -1 <<< "$response")
	if grep -q '200 OK' <<< "$head"; then
		echo "$head"
	else
		echo "$response"
		echo
	fi
}

for i in {1..10}; do
	request-odata $i &
done
```

### LINKED ISSUE #485: Request soon after premature close can result in database connection leak
A client can cause a premature close. For example, if you request an OData feed using `curl`, then kill `curl` as the response is being received, then there may be a premature close. (The OData feed may have to be on the larger side for this to work.) A premature close logs an error, but by itself, it doesn't seem to cause an issue. For example, you can take the following steps:

- Run `select count(*) from pg_stat_activity where state <> 'idle'`. The count should be 1 (corresponding to this query itself).
- Request an OData feed and cause a premature close.
- Immediately run the count query again. The count should be 2.
- Wait a few seconds, then run the query again. The count should be back to 1.

However, if something (either a request or a worker) attempts to acquire a database connection soon after the premature close, things can go wrong. The OData feed uses a Slonik stream, and it looks like if the connection pool gives out the same connection that the stream was using, that becomes an issue. To see this, specify the Slonik `interceptors` option to `createPool()`:

```js
interceptors: [{
  afterPoolConnection: ({ connectionId }) => {
    console.log('using connection', connectionId);
  },
  beforePoolConnectionRelease: ({ connectionId }) => {
    console.log('releasing connection', connectionId);
  }
}]
```

Now take the following steps:

- Request an OData feed and cause a premature close.
- Immediately after, send another request. (You may find it easiest to send the same request, but any request should reproduce.)
- If the stream's connection has been given out, the request will not receive a response.

You may have to try a few times to see this behavior. You'll see logs that look like this:

```
using connection HSEiX93XRJK1gDyV70K_Cg-22
releasing connection HSEiX93XRJK1gDyV70K_Cg-22
using connection HSEiX93XRJK1gDyV70K_Cg-22
releasing connection HSEiX93XRJK1gDyV70K_Cg-22
using connection HSEiX93XRJK1gDyV70K_Cg-22
releasing connection HSEiX93XRJK1gDyV70K_Cg-22
using connection HSEiX93XRJK1gDyV70K_Cg-22
releasing connection HSEiX93XRJK1gDyV70K_Cg-22
using connection HSEiX93XRJK1gDyV70K_Cg-27
using connection HSEiX93XRJK1gDyV70K_Cg-29
releasing connection HSEiX93XRJK1gDyV70K_Cg-29
::1 - - [14/May/2022:14:51:29 +0000] "GET /v1/projects/1/forms/foo.svc/Submissions HTTP/1.1" 200 -
releasing connection HSEiX93XRJK1gDyV70K_Cg-27
Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at ServerResponse.onclose (internal/streams/end-of-stream.js:121:38)
    at ServerResponse.emit (events.js:412:35)
    at ServerResponse.emit (domain.js:470:12)
    at Socket.onServerResponseClose (_http_server.js:221:44)
    at Socket.emit (events.js:412:35)
    at Socket.emit (domain.js:470:12)
    at TCP.<anonymous> (net.js:675:12)
using connection HSEiX93XRJK1gDyV70K_Cg-27
```

In the logs above, `HSEiX93XRJK1gDyV70K_Cg-27` is the ID of the connection that `stream()` used. The connection is given out to the next request, which seems to be unable to use it.

You may have to try the steps above several times before reproducing this behavior. Once you do, run the count query from above. The count is now 2 and as far as I can tell, never goes back down, indicating a connection leak. Even if you kill the second request, the count remains at 2.

If you repeat this process enough times such that the count reaches 11, the connection pool will be full. At that point, Problem responses will be returned with the message `Completely unhandled exception: timeout exceeded when trying to connect`.

I believe that this is an upstream issue and have filed an issue with Slonik.

(By the way, I think that the latest version of Slonik may pass slightly different information to the interceptor methods, so I'm not sure that the interceptor above would work as-is in the latest version.)

### LINKED ISSUE #565: db.stream() has no timeout

alxndrsn:
Closes #482
Closes #485 
Closes #565

Incorporates backported fixes:

* https://github.com/gajus/slonik/pull/360
* https://github.com/gajus/slonik/pull/414
* https://github.com/gajus/slonik/pull/420

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - checkout
       - run: make check-file-headers
-      - run: npm ci
+      - run: npm ci --legacy-peer-deps
       - run: node lib/bin/create-docker-databases.js
       - run: make test-full
 
--- a/benchmarker/scripts/ci-benchmark
+++ b/benchmarker/scripts/ci-benchmark
@@ -35,22 +35,22 @@ if ! curl -s -o /dev/null "$serverUrl"; then
   fail_job
 fi
 
-log 'Checking open DB query count...'
-timeout 120 bash -c "cd ..; while ! node lib/bin/check-open-db-queries.js; do sleep 1; done"
-
-log 'Checking backend is serving requests...'
 responseLog="$(mktemp)"
 requestBody='{"email":"'"$userEmail"'","password":"'"$userPassword"'"}'
 loginStatus="$(curl -s -o "$responseLog" -w '%{http_code}' \
   --header 'Content-Type: application/json' --data "$requestBody" \
   "$serverUrl/v1/sessions"
 )"
-if [[ "$loginStatus" != "200" ]]; then
+if [[ "$loginStatus" = "200" ]]; then
+  log 'Backend survived; job should pass.'
+else
   log 'Backend behaving badly:'
   log "$(cat "$responseLog")"
   fail_job
 fi
-log 'Backend survived; job passed.'
+
+log 'Checking open DB query count...'
+timeout 120 bash -c "cd ..; while ! node lib/bin/check-open-db-queries.js; do sleep 1; done"
 
 # TODO upload results to getodk.cloud for graphing.  Include:
 #
--- a/benchmarker/scripts/ci-benchmark
+++ b/benchmarker/scripts/ci-benchmark
@@ -35,22 +35,22 @@ if ! curl -s -o /dev/null "$serverUrl"; then
   fail_job
 fi
 
+log 'Checking open DB query count...'
+timeout 120 bash -c "cd ..; while ! node lib/bin/check-open-db-queries.js; do sleep 1; done"
+
+log 'Checking backend is serving requests...'
 responseLog="$(mktemp)"
 requestBody='{"email":"'"$userEmail"'","password":"'"$userPassword"'"}'
 loginStatus="$(curl -s -o "$responseLog" -w '%{http_code}' \
   --header 'Content-Type: application/json' --data "$requestBody" \
   "$serverUrl/v1/sessions"
 )"
-if [[ "$loginStatus" = "200" ]]; then
-  log 'Backend survived; job should pass.'
-else
+if [[ "$loginStatus" != "200" ]]; then
   log 'Backend behaving badly:'
   log "$(cat "$responseLog")"
   fail_job
 fi
-
-log 'Checking open DB query count...'
-timeout 120 bash -c "cd ..; while ! node lib/bin/check-open-db-queries.js; do sleep 1; done"
+log 'Backend survived; job passed.'
 
 # TODO upload results to getodk.cloud for graphing.  Include:
 #
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "prompt": "~1",
     "ramda": "~0",
     "sanitize-filename": "~1",
-    "slonik": "23.6.2",
+    "slonik": "npm:@getodk/slonik@23.6.2-3",
     "slonik-sql-tag-raw": "1.0.3",
     "tmp-promise": "~3",
     "uuid": "~3",
--- a/test/integration/other/db-stream.js
+++ b/test/integration/other/db-stream.js
@@ -5,10 +5,7 @@ const FakeTimers = require('@sinonjs/fake-timers');
 const { slonikPool } = require('../../../lib/external/slonik');
 const { queryFuncs } = require('../../../lib/util/db');
 
-// REVIEW: these tests are now skipped, as the mocking no longer works.
-// The tests pass fine if mocha is run without a timeout, and the fake
-// sleeps are replaced with real ones.
-describe.skip('db.stream()', () => {
+describe('db.stream()', () => {
   let db;
   let pool;
 
@@ -33,7 +30,12 @@ describe.skip('db.stream()', () => {
       await clock.tickAsync(60 * 1000);
     };
 
-    beforeEach(() => { clock = FakeTimers.install({ shouldClearNativeTimers: true }); });
+    beforeEach(() => {
+      clock = FakeTimers.install({
+        shouldClearNativeTimers: true,
+        toFake: ['setTimeout', 'clearTimeout'],
+      });
+    });
     afterEach(() => clock?.uninstall());
 
     it('should time out after 2 mins if no activity at all', async () => {
--- a/test/integration/other/db-stream.js
+++ b/test/integration/other/db-stream.js
@@ -5,7 +5,10 @@ const FakeTimers = require('@sinonjs/fake-timers');
 const { slonikPool } = require('../../../lib/external/slonik');
 const { queryFuncs } = require('../../../lib/util/db');
 
-describe('db.stream()', () => {
+// REVIEW: these tests are now skipped, as the mocking no longer works.
+// The tests pass fine if mocha is run without a timeout, and the fake
+// sleeps are replaced with real ones.
+describe.skip('db.stream()', () => {
   let db;
   let pool;
 
@@ -32,25 +35,6 @@ describe('db.stream()', () => {
 
     beforeEach(() => { clock = FakeTimers.install({ shouldClearNativeTimers: true }); });
     afterEach(() => clock?.uninstall());
-
-    // TODO this doesn't seem like the correct behaviour - no Error is thrown
-    // here.  At some point, this should be fixed elsewhere!
-    it('should close streams given an invalid sql statement', async () => {
-      // when
-      const stream = await db.stream(sql`NOT A SQL STATEMENT`);
-
-      // then
-      pool.getPoolState().activeConnectionCount.should.equal(1);
-      stream.destroyed.should.equal(false);
-
-      // when
-      await oneMinute();
-      await oneMinute();
-      // then
-      pool.getPoolState().activeConnectionCount.should.equal(0);
-      stream.destroyed.should.equal(true);
-    });
-
 
     it('should time out after 2 mins if no activity at all', async () => {
       // given
--- a/.github/workflows/benchmarker.yml
+++ b/.github/workflows/benchmarker.yml
@@ -30,7 +30,7 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
         cache: 'npm'
-    - run: npm ci
+    - run: npm ci --legacy-peer-deps
     - run: node lib/bin/create-docker-databases.js
     - name: Benchmark
       timeout-minutes: 10
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 default: base
 
 node_modules: package.json
-	npm install
+	npm install --legacy-peer-deps
 	touch node_modules
 
 .PHONY: node_version
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "prompt": "~1",
         "ramda": "~0",
         "sanitize-filename": "~1",
-        "slonik": "23.6.2",
+        "slonik": "npm:@getodk/slonik@23.6.2-3",
         "slonik-sql-tag-raw": "1.0.3",
         "tmp-promise": "~3",
         "uuid": "~3",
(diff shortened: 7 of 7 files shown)


matthew-white (commented):
I think we're hoping to ship the patched Slonik with the upcoming patch release. We'll release off the `release` branch so that we don't ship other commits in `master`. After the release, we'll merge `release` into `master`. (For what it's worth, patch releases are usually simpler than this. Usually patch releases come pretty soon after the last major release, so there aren't as many differences between the release branch and `master`.) To add this PR to the patch release, could you target the `release` branch? You'll probably first need to rebase onto `origin/release` locally, then force-push.

lib/util/db.js
```diff
@@ -239,7 +241,33 @@ const queryFuncs = (db, obj) => {
...
+
+      stream.on('close', () => {
+        clearTimeout(streamTimeout);
+      });
+
+      resolve(wrappedStream);
+    } catch (error) {
+      reject1(error);
+    }
+  }).catch(reject1));
```
matthew-white, line 269: Is there a reasonable way for us to unit-test this change? For the `setTimeout()` part, I've used [fake timers](https://sinonjs.org/releases/v14/fake-timers/) with Sinon when I've written Frontend tests of functionality that uses `setTimeout()`.

lib/util/db.js
```diff
@@ -239,7 +241,65 @@ const queryFuncs = (db, obj) => {
     for (let i = 0; i < xs.length; i += 1) result[i] = f(xs[i]);
     return result;
   };
-  obj.stream = (s) => new Promise((resolve, reject1) => db.stream(s, resolve).catch(reject1));
+
+  obj.stream = (s) => new Promise((resolve, reject1) => db.stream(s, stream => {
+    try {
```
matthew-white, line 246: What below might throw an error that would be caught here? `stream.pipe()`? Attaching one of the listeners? Or is this more of a safeguard?

matthew-white (commented):
The unit tests look good to me!

lib/util/db.js
alxndrsn: With all the surprising error handling required for streams, I wanted to make sure that _any_ errors were caught, even if they don't seem likely.

lib/model/query/comments.js
```diff
@@ -7,6 +7,7 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
matthew-white: @alxndrsn prefers not to turn off the rule entirely.

lib/util/db.js
```diff
@@ -239,7 +241,65 @@ const queryFuncs = (db, obj) => {
...
+      const twoMinutes = 2 * 60 * 1000;
+      const streamTimeout = setTimeout(() => {
+        stream.destroy(new Error('Stream timed out.'));
```
matthew-white, line 249: @alxndrsn [has](https://github.com/getodk/central-backend/pull/566/files#diff-36d887fd172376101537bad122756c3090c233494f460ef975c1f4ad0fb9868aR244-R249) [mentioned](https://github.com/gajus/slonik/pull/360#issuecomment-1206303334) that passing an error to `stream.destroy()` can sometimes be an issue. However, he says that the error is infrequent and (IIRC) that if it is thrown, that could provide information to us.

package.json
```diff
@@ -38,7 +38,7 @@
     "prompt": "~1",
     "ramda": "~0",
     "sanitize-filename": "~1",
-    "slonik": "23.6.2",
+    "slonik": "npm:@getodk/slonik@23.6.2-1",
```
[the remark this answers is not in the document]
matthew-white: #604 targets the `release` branch and uses a Slonik patch based on 23.6.0.

lib/util/db.js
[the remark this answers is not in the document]
matthew-white: We're going to include this change in the patch release (as part of #604) because that ended up seeming like the best way to address the database connection leaks for now.

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     steps:
       - checkout
       - run: node lib/bin/check-file-headers.js
-      - run: npm ci
+      - run: npm ci --legacy-peer-deps
```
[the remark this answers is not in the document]
matthew-white: #604 targets the release branch; this PR does not.

alxndrsn:
@matthew-white this has been rebased onto current `master` to re-introduce the slonik `QueryStream._destroy()` patch with minimal other changes.

alxndrsn:
Maybe worth including https://github.com/gajus/slonik/pull/414?

lognaturel:
> Maybe worth including https://github.com/gajus/slonik/pull/414?

Unless it's really complicated for some reason I think that's a good idea.

alxndrsn:
https://github.com/gajus/slonik/pull/414 has been incorporated

lognaturel:
+ this will fix `master`

Took a quick look, all looks like I expect, but I think it would be good for @matthew-white to take one last look since you two have all the context.

benchmarker/scripts/ci-benchmark
```diff
@@ -50,7 +50,7 @@ else
 fi
 
 log 'Checking open DB query count...'
-(cd .. && node lib/bin/check-open-db-queries.js)
+timeout 120 bash -c "cd ..; while ! node lib/bin/check-open-db-queries.js; do sleep 1; done"
```
matthew-white, line 53: This change was approved in #635.

test/integration/other/db-stream.js
```diff
@@ -35,6 +38,7 @@ describe('db.stream()', () => {
 
     // TODO this doesn't seem like the correct behaviour - no Error is thrown
     // here.  At some point, this should be fixed elsewhere!
+    // REVIEW: this behaviour has been fixed in Slonik
     it('should close streams given an invalid sql statement', async () => {
```
matthew-white, line 42: If this behavior has been fixed, should we remove this test?

test/integration/other/db-stream.js
```diff
@@ -5,7 +5,10 @@ const FakeTimers = require('@sinonjs/fake-timers');
 const { slonikPool } = require('../../../lib/external/slonik');
 const { queryFuncs } = require('../../../lib/util/db');
 
-describe('db.stream()', () => {
+// REVIEW: these tests are now skipped, as the mocking no longer works.
+// The tests pass fine if mocha is run without a timeout, and the fake
+// sleeps are replaced with real ones.
+describe.skip('db.stream()', () => {
```
matthew-white, line 11: Do you know why the mocking no longer works? I'm curious about what's breaking that!

matthew-white:
@alxndrsn, would you mind creating a PR to the `central` repo that specifies `--legacy-peer-deps` to `npm ci`? I think we'll need that in order to deploy this PR to a server.

test/integration/other/db-stream.js
alxndrsn: I'd happily remove them all...

test/integration/other/db-stream.js
alxndrsn: No idea - I had a look and it wasn't obvious.  Given how delicate stream behaviour seems to be, I'm not sure how much value tests will bring if they involve mocks.  A more robust approach might be to decrease the stream timeout and test against the real clock...

matthew-white (approved):
I see that the PR description mentions that gajus/slonik#420 has been backported as well. That Slonik fix hasn't been approved yet, so I feel less sure about including that one. But it also seems OK to go ahead and merge: if we need to, we can always update our Slonik patch. Plus, we're hoping to ship #564 for the release.

I'm also unsure why `import/no-extraneous-dependencies` is being tripped. Do we need to configure or update `eslint-plugin-import` somehow? But also, there doesn't seem to be a lot of risk there, so that might not be worth digging into.

alxndrsn:
> I'm also unsure why `import/no-extraneous-dependencies` is being tripped.

I've double-checked, and it doesn't seem to be necessary :+1:

test/integration/other/db-stream.js
alxndrsn: This is almost certainly related to https://github.com/gajus/slonik/pull/414.  It looks like the callback for a `setImmediate()` call in `PgCursor._setRows()` is never triggered when fake timers are in use.  I will dig in further.

alxndrsn:
According to this github-actions failure, the stream/connection problems are not resolved by this PR: https://github.com/alxndrsn/odk-central-backend/actions/runs/3226712671/jobs/5280574409

alxndrsn:
Won't reliably pass github-actions checks until #628 is merged

matthew-white (approved):

PR #603, opened 2022-09-09
Remove ability to create backups config

matthew-white:
Background:

- https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html
- https://forum.getodk.org/t/backups-to-google-drive-from-central-will-stop-working-after-oct-3/38895

Tests. We should probably also try running the migration on the QA server.

Most of the changes to lib/ are removing the two endpoints that will stop working. I also wrote a migration to delete data that is no longer useful. I don't think that migration is strictly necessary, but I think it improves clarity around what's in the database. For example, we won't have to wonder in the future what the `initbkup` role is.

The removal of the endpoints should be low-risk. (I would be a little surprised if those are used anywhere outside Frontend.) The migration is higher-risk because it touches multiple tables and deletes data. However, the SQL itself isn't very complex, and I've written tests of the migration.

I've updated the API documentation.


lib/model/migrations/20220908-01-remove-google-backups.js
```diff
@@ -0,0 +1,30 @@
...
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw(`delete from sessions where "actorId" in (
+  select id from actors
+  where
+    type = 'singleUse' and
+    "displayName" = 'Backup creation token' and
```
matthew-white, line 15: I'm not completely sure what the best way is to identify single-use backup creation tokens among actors, but it feels like this should be pretty robust. I also thought about checking that the `meta` column has a `keys` property and that the actor has an associated `initbkup` assignment. But ultimately I don't think those checks add much. I also don't think that there's a lot of risk here of unintentionally deleting other kinds of single-use actors (they'd have to be named "Backup creation token"!).

lib/model/migrations/20220908-01-remove-google-backups.js
```diff
@@ -0,0 +1,30 @@
...
+    "deletedAt" is null
+)`);
+  await db.raw(`update actors set "deletedAt" = now()
+where
+  type = 'singleUse' and
+  "displayName" = 'Backup creation token' and
+  "deletedAt" is null`);
+  await db.raw(`delete from assignments
+where "roleId" in (select id from roles where system = 'initbkup')`);
```
matthew-white, line 24: I'm intending the three statements above to do the same thing as `Actors.consume()`, which was used in /v1/config/backups/verify:

https://github.com/getodk/central-backend/blob/78a75c0dce31b4b897fbcae81d64b6b3ee7628cd/lib/resources/config.js#L91
https://github.com/getodk/central-backend/blob/78a75c0dce31b4b897fbcae81d64b6b3ee7628cd/lib/model/query/actors.js#L36-L43

test/integration/other/migrations.js
```diff
@@ -227,3 +229,78 @@ describe('datbase migrations: removing default project', function() {
     projCount.should.equal(0);
   }));
 });
+
+describe('database migrations: 20220908-01-remove-google-backups', () => {
+  // eslint-disable-next-line space-before-function-paren, func-names
+  beforeEach(function() {
```
matthew-white, line 235: I wouldn't be surprised if we end up wanting to turn off these rules, at least in testing. For now, I'm just disabling these rules in a similar way as #545 so that this commit passes once it's merged into the `master` branch.

test/integration/other/migrations.js
```diff
@@ -227,3 +229,78 @@ describe('datbase migrations: removing default project', function() {
...
+        displayName: 'Backup creation token',
+        expiresAt,
+        meta: {
+          keys: { some: 'data' }
+        }
+      }));
+      await Assignments.grantSystem(actor, 'initbkup', Config.species);
+      await Sessions.create(actor);
+      return actor.id;
+    };
```
matthew-white, line 275: Much of this was copied from /v1/config/backups/verify.

matthew-white:
@alxndrsn, I've tagged you for review in case you have time. It'd be great to be able to ship this as part of the upcoming patch release!

docs/api.md
```diff
@@ -26,12 +26,18 @@ Forms and their submissions are also accessible through two **open standards spe
...
 
 ## Changelog
 
 Here major and breaking changes to the API are listed by version.
 
+### ODK Central v1.5.3
+
+**Removed**:
+
+* It is no longer possible to initiate a new backups configuration (`POST /v1/config/backups/initiate`) or to verify one (`POST /v1/config/backups/verify`). However, if there is an existing configuration, it is still possible to get it or terminate it. If the existing configuration is terminated, it will not be possible to set up a new configuration. Note that it is still possible to download a [Direct Backup](/reference/system-endpoints/direct-backup). For more information about this change, please see [this topic](https://forum.getodk.org/t/backups-to-google-drive-from-central-will-stop-working-after-oct-3/38895) in the ODK Forum.
```
lognaturel, line 39: Thanks for doing this @matthew-white. 👍 Any clarity around whether backups will keep happening for existing configurations? That doesn't need to be highlighted here but it would be good to know and to update the forum thread.

test/integration/api/config.js
```diff
@@ -221,7 +221,7 @@ describe('api: /config', () => {
                 }))))));
     });
 
-    describe('/initiate POST', () => {
+    describe.skip('/initiate POST', () => {
```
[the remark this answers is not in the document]
alxndrsn: Is it worth adding a comment to this effect?

test/integration/other/migrations.js
alxndrsn: I'd be up for disabling everywhere.  I'm not sure `func-names` has much value if there's no transpilation applied to the code.

test/integration/other/migrations.js
```diff
@@ -227,3 +229,78 @@ describe('datbase migrations: removing default project', function() {
     projCount.should.equal(0);
   }));
 });
+
+describe('database migrations: 20220908-01-remove-google-backups', () => {
+  // eslint-disable-next-line space-before-function-paren, func-names
+  beforeEach(function() {
+    this.timeout(10000);
```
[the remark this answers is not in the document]
alxndrsn: Also does running this inside `beforeEach()` actually affect the test timeout?  I'd have expected it to be run directly in `describe()`.  Ref https://mochajs.org/#hook-level

alxndrsn (approved):

docs/api.md
matthew-white: My best guess is that backups will keep happening for existing configurations, but I'm not sure. Here's what I've learned so far:

- An app may have one or more OAuth clients, which may use one or more redirect URIs. We use a desktop client, which in contrast to a client for web apps, does not seem to require redirect URIs to be preconfigured. [This page](https://support.google.com/cloud/answer/6158849) has more information about different OAuth client types.
- The redirect URI that we use is for the deprecated OOB flow. Again, that's not preconfigured, but it's specified when we set up Google things in Backend ([here](https://github.com/getodk/central-backend/blob/78a75c0dce31b4b897fbcae81d64b6b3ee7628cd/lib/external/google.js#L17)). That said, I believe that the redirect URI is only used to acquire the authorization code that's exchanged for the first set of access and refresh tokens. ([This page](https://developers.google.com/identity/protocols/oauth2/) has more information about the OAuth process as a whole.) In other words, our redirect URI will stop working, which means that it'll stop being possible to acquire a new authorization code, but I haven't seen much to indicate that the parts of the OAuth process that don't use the redirect URI will be affected.
- Just for the record, it doesn't look like there's an easy way for us to specify a different redirect URI. For desktop clients, the OOB migration guide [recommends](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration#desktop-client) a redirect URI for the loopback IP address flow. But @issa-tseng says that that flow wouldn't work us, since Central isn't actually a desktop client.
- At least one Google employee [thinks](https://groups.google.com/g/adwords-api/c/idMbRt1Zg38) that existing tokens will continue to work (but isn't sure). Elsewhere, a different Google employee [responded](https://groups.google.com/g/google-cloud-dev/c/2OvJZehxYYA) to the same question, but they think that existing tokens won't continue to work.
- Another open-source community seems to be dealing with this issue ([here](https://forum.rclone.org/t/google-oauth-migration-for-rclone/30545)). From what I can tell, they believe that existing tokens will continue to work.
- [Some](https://wordpress.org/support/topic/google-oauth-oob-flow-warning/) [folks](https://github.com/r-lib/gargle/issues/202#issuecomment-1147768090) say that they received an email from Google suggesting that an extension of the deadline to January 31, 2023 was possible. That email apparently said, "Keep in mind that all OOB authorization requests will be blocked on February 1, 2023." But to me, that's a pretty different message from something like "Any activity derived from an OOB flow, including access tokens, will be blocked."
- If you try to set up a new backups configuration in Central today, you'll see the warning message [mentioned](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html#warning-message) in the Google blog post. That message includes the somewhat scary text, "Google will soon block apps that don’t comply with Google’s security policies." Presumably old installations of ODK Central will continue sending OOB authorization requests for a while to come. However, the relevant [help page](https://support.google.com/accounts/answer/12379384) doesn't seem to be saying that the app will be removed from Google Cloud or anything like that. More it just seems like any OOB authorization request will fail.

To me, these sources weigh in the direction of new OOB authorization failing while existing access tokens continue to work. However, I don't feel certain about that.

In terms of next steps:

- If backups will keep happening for existing configurations, this PR should be the only change needed for Backend.
- If backups won't keep happening, we should remove lib/bin/backup.js and probably some other things.
- Either way, we should modify Frontend.
- I also think we should update the forum thread.

I would lean toward the following:

- Tell users that we think backups will keep happening, but aren't sure. It might be reassuring for us to say explicitly that backup restore will continue to work.
- Update Backend/Frontend/`central` along those lines.
- If that turns out to be wrong and everything stops working on October 3, we could probably release another patch that removes more things.

That said, I'd greatly appreciate a second opinion on all this!

test/integration/api/config.js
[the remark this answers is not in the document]
matthew-white: I didn't end up adding a comment, but I'll plan to completely remove these tests later this week!

matthew-white:
We decided to split this PR into two. This PR will now just remove the endpoints and update the API docs. This PR will ship in v1.5.3. A later PR will complete the migration.

test/integration/other/migrations.js
```diff
@@ -227,3 +229,78 @@ describe('datbase migrations: removing default project', function() {
...
+    return upToMigration('20220309-01-add-project-description.js');
+  });
+
+  describe('initbkup role', () => {
+    it('deletes the role that grants backup.verify', testContainerFullTrx(async ({ all }) => {
+      const getRoles = () => all(sql`select * from roles where verbs @> '["backup.verify"]'`);
+      const rolesBefore = await getRoles();
+      rolesBefore.length.should.equal(1);
+      rolesBefore[0].system.should.equal('initbkup');
+      await migrator.migrate.up({ directory: appRoot + '/lib/model/migrations' });
```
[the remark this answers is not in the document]
matthew-white: I've changed how `migrator` is used in #618.

test/integration/api/config.js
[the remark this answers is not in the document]
matthew-white: Done in #617.

test/integration/other/migrations.js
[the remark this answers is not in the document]
matthew-white: > Also does running this inside `beforeEach()` actually affect the test timeout? I'd have expected it to be run directly in `describe()`.

I think you're right that `describe()` is the right place for this timeout. Running it in `beforeEach()` did seem to have an effect: the tests failed before I added this timeout, and they succeeded after. I'm surprised by that now though, given the write-up in the Mocha docs.

> I'd vote for increasing globally if we think it's necessary. Or is there something specific to this test which makes it extra slow?

I think I'd say yes to both. Many of these migration tests are extra slow for some reason, so I think it makes sense for them to have a higher timeout. But we also do see a fair number of intermittent timeouts in CircleCI, which I think would be a reason to increase the timeout globally. Maybe we could start by increasing it globally just for `make test-full`, which I think we expect to take longer anyway.

docs/api.md
matthew-white: We ended up receiving a 3-month extension to the deadline: see https://forum.getodk.org/t/backups-to-google-drive-from-central-will-stop-working-after-jan-31st/38895/5.

PR #605, opened 2022-09-13
new: API Get form/formXmlId/draft/dataset

sadiqkhoja:
### Changes
Added a new API that return dataset information for a draft form
Endpoint: GET /v1/projects/1/forms/formXmlId/draft/dataset 
Returns an array of all the datasets and their properties that will be created by publishing the given draft form. If the dataset is already there then `properties` field will contained only properties delta i.e. only properties that will be created by publishing the given draft form. Response also returns `datasetStatus` ; possible values are 1) `Draft` means dataset will be created 2) `Published` means dataset is already there and given properties will be added.

This API will be consumed by frontend to display dialog box when user is publishing a draft form that has entityRelated definition.

Integration tests

Alternative was to return current and future state of dataset definition and let frontend do the heavy lifting. Implemented approach does the comparison at database level and returns simple response that makes the life of frontend quite easy.

none

In a separate PR

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -11,7 +11,7 @@ const { sql } = require('slonik');
 const { raw } = require('slonik-sql-tag-raw');
 const { unjoiner } = require('../../util/db');
 const { Dataset, Form } = require('../frames');
-const { map, reduce, compose, pickBy, startsWith, nthArg, assoc, keys, curry, nth, isEmpty, isNil, either } = require('ramda');
+const { map, reduce, compose, pickBy, startsWith, nthArg, assoc, keys, curry, nth, isEmpty, isNil, either, prop, reduceBy } = require('ramda');
 const { construct } = require('../../util/util');
 
 // It removes prefix from all the key of an object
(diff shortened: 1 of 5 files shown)


test/integration/api/forms/datasets.js
```diff
@@ -112,3 +112,86 @@ describe('api: /projects/:id/forms (entity-handling)', () => {
...
+              {
+                datasetName: 'people',
+                datasetStatus: 'Draft',
+                properties: ['name', 'age']
+              }
+            ]);
+          })));
+  }));
+
+  it('should return nothing if dataset and all properties already exist', testService(async (service) => {
```
sadiqkhoja, line 138: May be return the dataset and properties that will be updated by publishing this form

lib/resources/forms.js
```diff
@@ -232,6 +232,14 @@ module.exports = (service, endpoint) => {
       .then(getOrNotFound)
       .then(ensureDef));
 
+  ////////////////////////////////////////
+  // DATASET RELATED ENDPOINTS
+  service.get('/projects/:projectId/forms/:id/draft/dataset', endpoint(({ Forms, Datasets }, { params, auth }) =>
```
sadiqkhoja, line 237: Move endpoint to separate resource file

Change `dataset` to `dataset-diff`

lib/model/query/datasets.js
```diff
@@ -145,4 +145,46 @@ ORDER BY form_fields."order"
 `)
   .then((fields) => fields.map((field) => new Form.Field(field, { propertyName: field.propertyName }))); // TODO augment this field so it also has the property name
 
-module.exports = { createOrMerge, getById, getAllByProjectId, getFieldsByFormDefId };
+const getDatasetDelta = (formDefId) => ({ all }) => all(sql`
```
ktuite, line 148: Maybe call this function `getDatasetDiff` so it matches the resource path?

test/integration/api/forms/datasets.js
```diff
@@ -112,3 +112,106 @@ describe('api: /projects/:id/forms (entity-handling)', () => {
     }));
   });
 });
+
+describe('api: /projects/:id/forms/draft/dataset', () => {
```
ktuite, line 116: looks good - tests seem to cover all cases i can think of. 
actually, what happens on a form with no dataset defined? does it return [] or does it crash?

ktuite (approved):

PR #604, opened 2022-09-12
Minimal slonik patch for closing stream queries (release)

alxndrsn:
WIP

PR #553, opened 2022-07-29
test: handle errors in zipStreamToFiles()

alxndrsn:
Should prevent failures in zip streams from cascading into other tests.  See e.g. https://app.circleci.com/pipelines/github/getodk/central-backend/776/workflows/728fbdeb-7ff2-4b26-a4af-5561c76e53d1/jobs/1518

PR #608, opened 2022-09-15
Create streams only after Promise.all()

matthew-white:
Part of a possible solution for #482, copied in part from changes in #564. See the commit message of the fourth commit for notes. The first commit is from #604. I've also cherry-picked from #542 and #553 in order to prevent a conflict with the `master` branch: this PR touches some of the same files as those PRs. When reviewing this PR, I recommend focusing on the fourth commit.

In addition to tests, I've deployed this PR to the QA server and followed the reproduction steps in #482. I was successfully able to send 10 concurrent requests. When I sent 11 concurrent requests, one quickly resulted in an error response ("Completely unhandled exception: timeout exceeded when trying to connect"). That is expected given that the maximum number of database connections is 10. However, what's exciting is that after the 11 responses, there was no database connection leak. 🎉

I still want to write a couple more tests to get at the Promise-related behavior of `zipStreamFromParts()` that I've introduced.

I'm actually not sure that this is the best solution or that we want to ship this with v1.5.3. I've created this PR in part so that we can consider this solution further. See https://github.com/getodk/central-backend/pull/604#issuecomment-1244913740 for the three options we've been discussing for v1.5.3.

The main alternative to this solution is a utility function like the one at https://github.com/getodk/central-backend/issues/482#issuecomment-1126784318. In most places, I think this solution is clearer to reason about than what we would get with a utility function. I think the main exception is `zipStreamFromParts()`, which has become more complicated. I also think a utility function could be more performant (though I doubt by much). That said, some of the changes that #564 makes are similar to those here, so we might be heading in this direction for multiple reasons.

The hope is that this PR fixes connection leak issues. However, there's a risk of regression with each of the three streaming endpoints that the PR modifies. I think running @alxndrsn's benchmarker against this PR would help us gain confidence.

No changes needed.

### DIFF
--- a/lib/util/zip.js
+++ b/lib/util/zip.js
@@ -14,6 +14,7 @@
 const { Readable } = require('stream');
 const { PartialPipe } = require('./stream');
 const archiver = require('archiver');
+const { resolve } = require('./promise');
 
 // Returns an object that can add files to an archive, without having that archive
 // object directly nor knowing what else is going into it. Call append() to add a
@@ -36,8 +37,7 @@ const zipPart = () => {
 // if the final component in the pipeline emitted the error, archiver would then
 // emit it again, but if it was an intermediate component archiver wouldn't know
 // about it. by manually aborting, we always emit the error and archiver never does.
-const zipStreamFromParts = (...zipParts) => {
-  let completed = 0;
+const zipStreamFromParts = (...zipPartFns) => {
   const resultStream = archiver('zip', { zlib: { level: 9 } });
 
   // track requested callbacks and call them when they are fully added to the zip.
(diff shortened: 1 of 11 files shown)


lib/util/zip.js
```diff
@@ -36,8 +37,7 @@ const zipPart = () => {
 // if the final component in the pipeline emitted the error, archiver would then
 // emit it again, but if it was an intermediate component archiver wouldn't know
 // about it. by manually aborting, we always emit the error and archiver never does.
-const zipStreamFromParts = (...zipParts) => {
-  let completed = 0;
+const zipStreamFromParts = (...zipPartFns) => {
```
matthew-white, line 40: I recommend hiding whitespace when viewing the diff for this file.

lib/util/zip.js
```diff
@@ -47,29 +47,35 @@ const zipStreamFromParts = (...zipParts) => {
...
-      }
-    });
-    part.stream.on('error', (err) => { resultStream.emit('error', err); });
-    part.stream.on('end', () => { // eslint-disable-line no-loop-func
-      completed += 1;
-      if (completed === zipParts.length)
-        resultStream.finalize();
-    });
-  }
+    resolve(zipPartFns.shift()())
```
matthew-white, line 56: This is the main difference from the approach in #564. In this PR, `stream()` returns a Promise of a stream, so `zipStreamFromParts()` needs some Promise-related logic. However, in #564, `stream()` simply returns a stream.

matthew-white:
I've marked this PR as a draft while we finish working on #564.

matthew-white:
Closing now that we've decided to close #634.

PR #610, opened 2022-09-20
Entity schema, submission parsing, and CSV streaming

ktuite:
This PR is mainly about getting the entity structure in place and validating that structure through code that can create entities and stream them out as a CSV. 

- Defines `Entity` and `Entity.Def` structures along with `entities` and `entity_defs` tables (we plan to support updating entities in the future)
- Parsing of submissions to create entities, triggered by approving a submission
- Preliminary streaming of entities as a CSV file

Some things that are **NOT** in this PR and coming in a follow up:
1. Updated entity validation (in the form definition and the submission XML) to match the spec
2. Audit logging of entity creation
3. Working out the worker logic and ensuring good error handling - e.g. gracefully handling cases when:
    - submission validation fails
    - entity creation fails (e.g. uuid already exists)
    - a submission is re-approved

Tests, as well as a first pass "end-to-end" scenario

It's modeled after the `form` / `form_def` and `submission` / `submission_def` structures.

Most of this is behind-the-scenes entity work that users wont see until we start work on the frontend.
Before I figure out better error handling, one risk (for devs because this will get fixed before this all gets merged and released!) is if you approve a submission twice, the second time a worker tries to create an entity from that submission, it will come up with a uniqueness violation and fail, but the worker will keep retrying. One of the things to fix in the next PR!

There is a new endpoint for streaming the entity/dataset CSV but we will add documentation later when that is more formalized. 


lib/data/entity.js
```diff
@@ -0,0 +1,124 @@
...
+const { PartialPipe } = require('../util/stream');
+const { SchemaStack } = require('./schema');
+const { noop } = require('../util/util');
+
+////////////////////////////
+// ENTITY PARSING
+//
+// Parse submission XML and use with form field -> entity property mappings to build entities.
+
+const submissionXmlToEntityStream = (fields, xml) => {
```
sadiqkhoja, line 22: this function is quite similar to `submissionXmlToFieldStream`, only `opentag` is different, can we refactor to reduce duplication?

lib/model/query/entities.js
```diff
@@ -0,0 +1,76 @@
...
+  JOIN datasets ON datasets."projectId" = f."projectId"
+WHERE datasets."name" = ${dsName} AND sd."id" = ${subDefId}`;
+
+const createNew = (partial) => ({ one }) => {
+  const json = JSON.stringify(partial.def.data);
+
+  return one(sql`
+with def as (${_defInsert(nextval, partial, json)}),
+ds as (${_getDataset(partial.aux.dataset, partial.def.submissionDefId)}),
+ins as (insert into entities (id, "datasetId", "uuid", "label", "createdAt")
```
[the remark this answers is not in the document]
ktuite: I'm going to leave this as is because it's inserting a mix of values from different sources (a previous query, the partial entity, the functions arguments).

lib/model/migrations/20220829-01-create-entities-defs.js
```diff
@@ -0,0 +1,39 @@
...
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) =>
+  db.schema.createTable('entities', (entities) => {
+    entities.increments('id').primary();
+    entities.string('uuid').notNull();
+    entities.integer('datasetId');
+    entities.text('label').notNull();
+    entities.dateTime('createdAt');
+    entities.integer('createdBy').notNull();
```
sadiqkhoja, line 17: I think we should add `updatedAt` as well, it will help us identify if dataset has been changed

ktuite:
demo form that works with this PR:

(this form works with the current version of the PR but doesn't follow the spec we've decided on where certain entity fields are attributes rather than nodes)

```
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:entities="http://www.opendatakit.org/xforms">
    <h:head>
        <h:title>Shovel Registration</h:title>
        <model odk:xforms-version="1.0.0">
            <instance>
                <data id="shovel_registration" version="2022.09.29_01">
                    <color/>
                    <size/>
                    <meta>
                        <entities:entity entities:dataset="shovels">
                            <entities:label/>
                            <entities:id/>
                            <entities:create/>
                        </entities:entity>
                        <instanceID/>
                        <instanceName/>
                    </meta>
                </data>
            </instance>
            <bind nodeset="/data/color" type="string" entities:saveto="color"/>
            <bind nodeset="/data/size" type="integer" entities:saveto="length"/>
            <bind jr:preload="uid" nodeset="/data/meta/instanceID" readonly="true()" type="string"/>
            <bind jr:preload="uid" nodeset="/data/meta/entities:entity/entities:id" readonly="true()" type="string"/>
            <bind calculate="concat( /data/color , &quot; &quot;,  /data/size , &quot;cm &quot;)" nodeset="/data/meta/instanceName" type="string"/>
            <bind calculate="concat( /data/size , &quot;cm &quot;,  /data/color , &quot; shovel &quot;)" nodeset="/data/meta/entities:entity/entities:label" type="string"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/color">
            <label>Color of shovel</label>
        </input>
        <input ref="/data/size">
            <label>Length of shovel in cm</label>
        </input>
    </h:body>
</h:html>
```
example output `shovels.csv` downloaded from `http://localhost:8989/v1/projects/1/datasets/shovels/download` 
```
name,label,color,length
uuid:81fb9ac0-b4d3-4508-b175-61a6bc46cf75,123cm red shovel,red,123
uuid:3b0571f2-6ba3-4bbb-b687-5e110b18ceac,34cm pink shovel,pink,34
```

PR #612, opened 2022-09-21
Add benchmarker running on Github Actions

alxndrsn:
Example runs can be seen at https://github.com/alxndrsn/odk-central-backend/actions


matthew-white (commented):
I think @sadiqkhoja is going to take a look at the benchmarker, but I was wondering whether it'd be helpful to move the changes to lib/bin/check-file-headers.js to a separate PR. I'd be happy to take a look at those changes!

Also, I see the new file benchmarker/scripts/ci-initdb.js, but could lib/bin/create-docker-databases.js be used instead?

benchmarker/index.js
```diff
@@ -0,0 +1,317 @@
...
+  const headers = {
+    'Content-Type': 'multipart/form-data; boundary=foo',
+    'X-OpenRosa-Version': '1.0',
+  };
+
+  const body = `--foo\r
+Content-Disposition: form-data; name="xml_submission_file"; filename="submission.xml"\r
+Content-Type: application/xml\r
+\r
+<data id="250_questions">
```
sadiqkhoja, line 288: I don't see any advantage of having random values here, only unique instanceID should suffice

Then we can move this to a separate xml file

benchmarker/index.js
```diff
@@ -0,0 +1,317 @@
+// Copyright 2017 ODK Central Developers
```
sadiqkhoja, line 1: 2022

sadiqkhoja (commented):
This is superb, really like it

sadiqkhoja (approved):

lib/bin/check-file-headers.js
```diff
@@ -19,50 +19,52 @@ const header = `// Copyright 2017 ODK Central Developers
...
+      throw new Error(`Invalid skip entry: '${skipEntry}'`);
+    }));
+
+  const mode = process.argv[2] === '--fix' ? 'fix' : 'check';
+
+  let missing = false;
+
+  for (const entryPath of files) {
+    const contents = fs.readFileSync(entryPath).toString();
+    const withConsistentYear = contents.replace(/(Copyright 20)\d\d/, '$117');
```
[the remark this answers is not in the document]
matthew-white: I've thought a little about this, and I think that might be tricky. We could try checking that all new files specify the current year in their file header — that'd work in most cases. But files are sometimes moved, which makes things hard: a file that was moved may look like a new file, yet specify a previous year in its file header. Things also get challenging around the year boundary, because a single PR may have commits from both years.

So I don't think CircleCI should fail if the year in the file header isn't the current year. If there's an easy way for it just to provide a warning in that case, I could see that being helpful in a follow-up PR — not sure what the possibilities there are.

benchmarker/package-lock.json
```diff
@@ -0,0 +1,1480 @@
+{
+  "name": "benchmarker",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "benchmarker",
+      "version": "1.0.0",
+      "license": "MIT",
```
matthew-white, line 10: This doesn't seem to match package.json for some reason.

matthew-white:
It looks like there's a conflict now that #619 is merged. Might be worth rebasing?

benchmarker/package-lock.json
alxndrsn: :thinking: weird

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - checkout
-      - run: node lib/bin/check-file-headers.js
+      - run: git ls-files | node lib/bin/check-file-headers.js
```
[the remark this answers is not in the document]
alxndrsn: Or a bit annoying when developing locally?

.github/workflows/benchmarker.yml
```diff
@@ -0,0 +1,40 @@
...
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - run: npm ci
+    - run: node lib/bin/create-docker-databases.js
+    - name: Benchmark
+      timeout-minutes: 10
+      run: ./benchmarker/scripts/ci-benchmark
+    - name: Backend Logs
+      if: always()
```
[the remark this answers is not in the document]
alxndrsn: I think so...

benchmarker/index.js
alxndrsn: Non-random values will likely affect behaviour and response size for zipped streams.

benchmarker/index.js
alxndrsn: :+1: fixed.

benchmarker/index.js
alxndrsn: I think this could be moved to a separate file regardless, if that would be helpful.

PR #614, opened 2022-09-22
Update package-lock slonik version

alxndrsn:
Every time `npm install` runs on my machine, it makes this change to `package-lock.json`.

Is anyone else seeing the same behaviour?  If so, perhaps we can make the change permanent.


matthew-white (approved):
I see the same thing!

PR #619, opened 2022-09-23
check-file-headers: use git ls-files to simplify ignore list

alxndrsn:
Originally written in https://github.com/getodk/central-backend/pull/612, to avoid updating the check-file-headers ignore list with benchmarker `node_mdoules` directory.

Simplify diff by ignoring whitespace: https://github.com/getodk/central-backend/pull/619/files?w=1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


matthew-white (approved):
Good idea to use `git ls-files`!

PR #620, opened 2022-09-23
check-file-headers: add --fix

alxndrsn:
Depends on: https://github.com/getodk/central-backend/pull/619

### DIFF
--- a/lib/bin/check-file-headers.js
+++ b/lib/bin/check-file-headers.js
@@ -18,6 +18,7 @@ const header = `// Copyright 2017 ODK Central Developers
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 `;
+const thisYearHeader = header.replace('2017', new Date().getFullYear());
 
 const skip = [
   'test/',
@@ -40,6 +41,8 @@ process.stdin.on('end', () => {
       throw new Error(`Invalid skip entry: '${skipEntry}'`);
     }));
 
+  const mode = process.argv[2] === '--fix' ? 'fix' : 'check';
+
   let missing = false;
 
   for (const entryPath of files) {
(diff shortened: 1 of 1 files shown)


lib/bin/check-file-headers.js
```diff
@@ -48,11 +50,21 @@ process.stdin.on('end', () => {
     if (!withConsistentYear.startsWith(header)) {
       console.error(entryPath);
       missing = true;
+
+      if (mode === 'fix') {
+        fs.writeFileSync(entryPath, `${header}\n${contents}`);
```
matthew-white, line 55: I think it'd be good to add a file header with the current year (rather than 2017).

lib/bin/check-file-headers.js
```diff
@@ -48,11 +50,21 @@ process.stdin.on('end', () => {
...
 
   if (missing) {
-    console.error('\nThe files above do not have the expected file header.');
-    process.exit(1);
+    if (mode === 'fix') {
+      console.error('\nThe files above have been changed to include the required header.');
+    } else {
+      console.error('\nThe files above do not have the expected file header.');
+      console.error('\nPlease re-run this script with the --fix flag, or add the following header manually:');
+      console.error(`\n${header}`);
```
matthew-white, line 66: Similarly, if we're going to display a file header here, I think it'd be good to display the current year.

alxndrsn:
@matthew-white the script has been updated to use the current year when suggesting/inserting fresh headers.

lib/bin/check-file-headers.js
```diff
@@ -40,19 +42,32 @@ process.stdin.on('end', () => {
...
+
   let missing = false;
 
+  const hasHeader = new RegExp('^' + header.replace('{{year}}', '20\\d\\d'));
+
   for (const entryPath of files) {
     const contents = fs.readFileSync(entryPath).toString();
-    const withConsistentYear = contents.replace(/(Copyright 20)\d\d/, '$117');
-    if (!withConsistentYear.startsWith(header)) {
+    if (!hasHeader.test(contents)) {
```
matthew-white, line 53: When I first wrote the script, I thought about using a regular expression for the file header. But if we took that approach, I think we'd want to escape the regular-expression special characters in `header`. Ultimately that felt more complicated to me than replacing each year with 2017.

lib/bin/check-file-headers.js
```diff
@@ -19,6 +19,8 @@ const header = `// Copyright 2017 ODK Central Developers
 // except according to the terms contained in the LICENSE file.
 `;
 
+const newHeader = header.replace('{{year}}', new Date().getFullYear());
```
matthew-white, line 22: What do you think about keeping 2017 in the header, then replacing 2017 with the current year here?

alxndrsn:
@matthew-white I've reverted to use 2017.

lib/bin/check-file-headers.js
matthew-white: Ah I think this comment is applicable again. Could we use `thisYearHeader` here as well (or do something along those lines)?

matthew-white (approved):
Just one small change to make I think! Feel free to merge after taking a look at that.

PR #622, opened 2022-09-23
Add script: check-open-db-queries

alxndrsn:
This could be added at the end of the benchmark to confirm there are no hanging queries.


matthew-white (approved):
Seems quite useful!

PR #626, opened 2022-09-28
server name not needed now that user has hostname information.

yanokwa:
Not sure if we use server_name else where, but I checked central and central-frontend and didn't see it.


matthew-white:
@yanokwa, it looks like ESLint is failing in CircleCI.

alxndrsn (approved):

PR #627, opened 2022-09-28
new: link and unlink dataset to a form

sadiqkhoja:
### Changes:
- Added a new endpoint `PATCH /projects/:id/forms/:formId/draft/attachment/:name`, it expects following request body:`{ dataset: <boolean> }` and returns attachment object
- Updated `GET /projects/:id/forms/:formId/attachments`, it now returns `blobExists` (previously `exists`) and `datasetExists`
- `form_attachments` table has a new column `datasetId`, a constraint is added that makes sure that only `blobId` or `datasetId` is filled

Note: `PATCH ../attachment/:name`, `POST ../attachment` and `DELETE ..attachment` tries to perform requested action aggressively, which means existing `blobId` will be set to null if link dataset request is received using PATCH and similarly `datasetId` will be set to null if blob is uploaded using POST, and DELETE sets both to null.

Integration test

Had discussions with the team and decided to implement PATCH endpoint for linking dataset to a form

`exists` field has been renamed to `blobExists`, frontend should make the appropriate changes

In a separate PR

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -27,7 +27,8 @@ const makeHierarchy = reduce((result, item) => {
   const property = new Dataset.Property(pickFrameFields(Dataset.Property, item)).forApi();
   const propertyField = new Dataset.PropertyField(pickFrameFields(Dataset.PropertyField, item)).forApi();
 
-  return { ...result,
+  return {
+    ...result,
     [dataset.id]: {
       ...dataset,
       properties: !property.id ? { ...(result[dataset.id]?.properties || {}) } : {
@@ -40,7 +41,8 @@ const makeHierarchy = reduce((result, item) => {
           ]
         }
       }
-    } };
+    }
+  };
 }, {});
 
 const asArray = compose(map(d => ({ ...d, properties: Object.values(d.properties) })), Object.values);
@@ -65,7 +67,7 @@ const _insertDatasetDef = (dataset, withDsDefsCTE) => sql`
     DO UPDATE SET "revisionNumber" = datasets."revisionNumber" + 1
     RETURNING *
   )
-  ${raw(withDsDefsCTE ? ', ds_defs AS (': '')}
+  ${raw(withDsDefsCTE ? ', ds_defs AS (' : '')}
     INSERT INTO dataset_form_defs
     SELECT ds.id, ${dataset.aux.formDefId} FROM ds
     ON CONFLICT ON CONSTRAINT dataset_form_defs_datasetid_formdefid_unique 
@@ -203,9 +205,13 @@ const getDatasetDiff = (formDefId) => ({ all }) => all(sql`
   .then(reduceBy((acc, { propertyName, isPropertyNew }) => (propertyName ? acc.concat({ name: propertyName, isNew: isPropertyNew }) : acc), [], (row) => `${row.datasetName},${row.isDatasetNew}`))
   .then(r => Object.keys(r).map(k => ({ name: k.split(',')[0], isNew: k.split(',')[1] === 'true', properties: r[k] })));
 
+  const getByProjectAndName = (projectId, name) => ({ maybeOne }) => maybeOne(sql`
+SELECT * FROM datasets WHERE "projectId" = ${projectId} AND name = ${name}`)
+  .then(map(construct(Dataset)));
+
 module.exports = {
   createOrMerge,
   getById, getAllByProjectId, getByName,
   getFieldsByFormDefId,
-  getDatasetDiff
+  getDatasetDiff, getByProjectAndName
 };
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -205,7 +205,7 @@ const getDatasetDiff = (formDefId) => ({ all }) => all(sql`
   .then(reduceBy((acc, { propertyName, isPropertyNew }) => (propertyName ? acc.concat({ name: propertyName, isNew: isPropertyNew }) : acc), [], (row) => `${row.datasetName},${row.isDatasetNew}`))
   .then(r => Object.keys(r).map(k => ({ name: k.split(',')[0], isNew: k.split(',')[1] === 'true', properties: r[k] })));
 
-  const getByProjectAndName = (projectId, name) => ({ maybeOne }) => maybeOne(sql`
+const getByProjectAndName = (projectId, name) => ({ maybeOne }) => maybeOne(sql`
 SELECT * FROM datasets WHERE "projectId" = ${projectId} AND name = ${name}`)
   .then(map(construct(Dataset)));
 
--- a/lib/model/query/form-attachments.js
+++ b/lib/model/query/form-attachments.js
@@ -20,7 +20,6 @@ const { unjoiner, insertMany } = require('../../util/db');
 const { ignoringResult, resolve } = require('../../util/promise');
 const { construct } = require('../../util/util');
 const Option = require('../../util/option');
-const Problem = require('../../util/problem');
 
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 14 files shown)


lib/resources/datasets.js
```diff
@@ -7,11 +7,33 @@
...
 
 module.exports = (service, endpoint) => {
   service.get('/projects/:id/datasets', endpoint(({ Projects, Datasets }, { auth, params, queryOptions }) =>
     Projects.getById(params.id, queryOptions)
       .then(getOrNotFound)
       .then((project) => auth.canOrReject('project.read', project)) // TODO change the verb
       .then(() => Datasets.getAllByProjectId(params.id))));
+
+
+  service.post('/projects/:projectId/forms/:id/draft/attachments/:name/link-dataset', endpoint(({ Datasets, FormAttachments, Forms }, { auth, params }) =>
```
lognaturel, line 21: I believe this would be our first major deviation from REST since `link-dataset` is an action (so more RPC-like). I haven't given this deep thought yet but I wonder whether there might be a version with `PATCH` and a new property on attachments that would stay within REST. Maybe it's a combination of a new `type` version `dataset` and `datasetid`?

I see this is a draft so don't mean to derail your chain of thought, I'm excited to see this addition and the route jumped out at me as I took a peek!

lib/util/problem.js
```diff
@@ -193,6 +193,8 @@ const problems = {
     // returned in case no external ODK Analytics service is configured.
     analyticsNotConfigured: problem(501.9, () => 'This ODK Central has not been configured to report Analytics. Please contact your server administrator.'),
 
+    invalidParameters: problem(501.10, (details) => `Invalid parameters passed. ${details}`),
```
ktuite, line 196: Because of some string/number issue (I forget where), we skip `.10` errors and go straight to `.11` errors so they don't wind up getting mixed up with `.1` errors  (if we wind up keeping this error)

test/integration/api/datasets.js
```diff
@@ -37,3 +38,68 @@ describe('projects/:id/datasets GET', () => {
...
+
+  it('should match dataset name with attachment name', testService((service) =>
+    service.login('alice', (asAlice) =>
+      asAlice.post('/v1/projects/1/forms')
+        .send(testData.forms.withAttachments)
+        .set('Content-Type', 'application/xml')
+        .expect(200)
+        .then(() => asAlice.post('/v1/projects/1/forms?publish=true')
+          .send(testData.forms.simpleEntity))
+        .then(() => asAlice.post('/v1/projects/1/forms/withAttachments/draft/attachments/goodone.csv/link-dataset')
```
ktuite, line 78: The invalid parameter problem above where blob id and dataset id can't both be set, is there a way to trigger that through the API? 

Could/would it happen when someone defines `goodone.csv` in the form, attaches a real CSV attachment, and then links it with a dataset?

Separately, in that case, would we replace the CSV with the dataset link? What happens to the CSV - does it get lost?

lib/resources/datasets.js
ktuite: I like the idea of using `PATCH` too (with `type: dataset`), since it's about doing something to the attachment "slot".

`DELETE` could unlink if it is a dataset.

test/integration/api/forms/draft.js
```diff
@@ -1167,7 +1169,9 @@ describe('api: /projects/:id/forms (drafts)', () => {
                           formDefId: form.draftDefId,
                           name: attachment.name,
                           oldBlobId: attachment.blobId,
-                          newBlobId: attachment2.blobId
+                          newBlobId: attachment2.blobId,
+                          oldDatasetId: attachment.datasetId,
```
ktuite, line 1173: This is probably fine, just thinking about how it will show up in the audit logs and if people will wonder about dataset IDs showing up for all their attachment changes.

Or maybe i'm thinking about the fact that these tests aren't about dataset attachments so the dataset IDs will always be null. We might want more tests of what it looks like when the attachments are updated to link to datasets.

lib/model/query/datasets.js
```diff
@@ -145,4 +147,8 @@ ORDER BY form_fields."order"
 `)
   .then((fields) => fields.map((field) => new Form.Field(field, { propertyName: field.propertyName }))); // TODO augment this field so it also has the property name
 
-module.exports = { createOrMerge, getById, getAllByProjectId, getFieldsByFormDefId };
+const getByProjectAndName = (projectId, name) => ({ maybeOne }) => maybeOne(sql`
```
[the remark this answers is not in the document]
sadiqkhoja: formatting changes + addition of a new function `getByProjectAndName`

lib/resources/datasets.js
ktuite: Talked to @matthew-white a bit about this and he pointed that `type` has multiple other meanings in/near this context:
- content type of the blob (sent in the existing attachment `POST` [request](https://odkcentral.docs.apiary.io/#reference/forms/draft-form/uploading-a-draft-form-attachment) and then served back with the blob data)
- type of the attachment as figured out from the form XML (parsed around [here](https://github.com/getodk/central-backend/blob/master/lib/data/schema.js#L326) somewhere?) e.g. `image`, `audio`, `video`, and the only one that is valid for datasets: `file`

So another alternative could be something like `PATCH` with `{ dataset: true }` and avoid the word "type".

sadiqkhoja: implemented PATCH

lib/model/query/form-attachments.js
```diff
@@ -98,11 +99,12 @@ const createVersion = (xml, form, savedDef, itemsets, publish = false) => ({ Blo
 ////////////////////////////////////////////////////////////////////////////////
 // CRUD
 
-const update = (_, fa, blobId) => ({ run }) => run(sql`
-update form_attachments set "blobId"=${blobId}, "updatedAt"=clock_timestamp()
-  where "formId"=${fa.formId} and "formDefId"=${fa.formDefId} and name=${fa.name}`);
-update.audit = (form, fa, blobId) => (log) => log('form.attachment.update', form,
-  { formDefId: form.draftDefId, name: fa.name, oldBlobId: fa.blobId, newBlobId: blobId });
+const update = (_, fa, blobId, datasetId = null) => ({ run }) =>
+  (blobId && datasetId ? Problem.internal.invalidParameters('blobId and datasetId can\'t be set simultaneously') : run(sql`
```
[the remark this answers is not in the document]
sadiqkhoja: added database constraint

test/integration/api/datasets.js
sadiqkhoja: Invalid parameter problem can never occur through API, only by developers mistake. I have added integration test to ensure that never happens design time as well

lib/model/query/datasets.js
```diff
@@ -203,9 +205,13 @@ const getDatasetDiff = (formDefId) => ({ all }) => all(sql`
...
+SELECT * FROM datasets WHERE "projectId" = ${projectId} AND name = ${name}`)
+  .then(map(construct(Dataset)));
+
 module.exports = {
   createOrMerge,
   getById, getAllByProjectId, getByName,
   getFieldsByFormDefId,
-  getDatasetDiff
+  getDatasetDiff, getByProjectAndName
```
ktuite, line 216: Eventually we will want to consolidate/rename one or both of `getByName` and `getByProjectAndName`. In a dataset download resource, we need to use one of these to get a dataset Option that we can call `getOrReject` on.

lib/util/db.js
```diff
@@ -490,6 +490,11 @@ const postgresErrorToProblem = (x) => {
     if (match != null) {
       return reject(Problem.user.valueTooLong({ maxLength }));
     }
+  } else if (error.code === '23514') {
```
ktuite, line 493: what happens / what error gets surfaced if we don't try to catch this specific error like this? do we need to wrap it in a message like this or can we keep it more generic? we might have other integrity constraint violations that now get this message.

test/integration/api/datasets.js
```diff
@@ -1,5 +1,11 @@
 const { testService } = require('../setup');
 const testData = require('../../data/xml');
+const config = require('config');
+const { Form } = require('../../../lib/model/frames');
+const { getOrNotFound } = require('../../../lib/util/promise');
+const { omit } = require('ramda');
+
+// TODO merge with test/integration/api/forms/dataset.js
```
ktuite, line 8: this sounds good!
in one of my PRs (https://github.com/getodk/central-backend/pull/633) i moved something with a form draft path from dataset into forms/dataset.

ktuite (commented):
A few more comments! I think this is really close. I feel compelled to *try* out the API (I know there are tests) and hook up an attachment to a dataset. I'll do that on wednesday.

The current CSV entity streaming download URL is dataset-specific, not form-attachment-specific, so I think there's one extra step not reflected in this PR to change the attachment resource to stream the CSV in the same way.

ktuite (approved):
Ok, I've gotten a chance to try setting the attachment to a dataset via the API and it seems good!

My comments are still
1. Change the database DOWN migration (I have been using it when switching between entity branches even though it will hopefully not be used in other circumstances)
2. Still nervous about explicitly catching that DB constraint error. it's nice that the constraint is there (and that it has been tested at some point) but i don't think it needs to be in this code especially because it can't be triggered through the API. my worry is that by adding it, that error message would pop up in another unexpected situation.

PR #630, opened 2022-09-30
improve: move ensureDef to util/db.js to eliminate code duplication

sadiqkhoja:
Code Refactor: moved ensureDef to util/db.js as ensureFormDef to eliminate code duplication

All tests are passing

NA

NA

NA


alxndrsn:
It looks like `util/db.js` does not currently contain domain-specific code, so perhaps this function would be better elsewhere, and/or changed to not reference forms?

alxndrsn (approved):

PR #632, opened 2022-10-03
Dataset Validation and Worker Logic/Error Handling

ktuite:
Improves parsing and validation of both datasets (from form XML) and entities (from submission XML).

Restructures the parsing/validation code a bit to take things out of `schema.js` where they were starting to accumulate and put them in `lib/data/entity.js` and `lib/data/dataset.js`. There are now also tests that match this file pattern.

There is also additional error handling including specific problems related to dataset validation errors and entity processing errors. The dataset validation errors should get surfaced to central-frontend if a user uploads a problematic form. The entity errors will eventually get logged (but that will come in a follow-up PR).

The error handling works with the worker that processes submissions to create entities. If the entity is invalid and there is an error, the worker will catch the error but still exit successfully (and avoid attempting to reprocess the invalid entity). There is likely room for improvement in this part of the code, though. 

I also changed some of the test data. `name` is not an allowed property name so the simpleEntity test form changed. The uuids that are used in tests are valid v4 uuids now. We have discussed certain entity system properties (uuid, create, label) being attributes rather than elements but the tests and validation code still uses elements because that is what enketo is reliably able to handle at the moment.

Tests

This is an improvement (more thorough and better organized) than the preliminary validation code.

Dataset/entity handling is a bit more robust with this PR, with better error handling.

Will need to change docs to mention how certain new dataset validation errors can show up in the form upload API.


lib/model/query/entities.js
```diff
@@ -56,12 +58,67 @@ select ins.*, def.id as "entityDefId" from ins, def;`)
...
+          ((fields.length === 0)
+            ? resolve()
+            : Entity.fromSubmissionXml(submissionDef.xml, fields)
+              .then((partial) =>
+                Entities.canInsertEntity(partial.aux.dataset, submissionDefId, partial.uuid)
+                  .then((okToInsert) =>
+                    (okToInsert
+                      ? Entities.createNew(partial, submissionDef)
+                        .then((entity) => resolve(entity))
+                        .catch((err) => reject(Problem.user.entityViolation({ reason: 'Entity could not be created', err: err.problemDetails })))
```
sadiqkhoja, line 106: I think this should be caught in worker, can be refactored in next PR

lib/model/query/entities.js
```diff
@@ -56,12 +58,67 @@ select ins.*, def.id as "entityDefId" from ins, def;`)
 
 // TODO: add and tests entity creation audit logging and error handling (failed validation, etc.)
 
+// HACK: This is a function to check if the entity *can* successfully be inserted
+// because without it... in the case of processing an entity within a worker,
+// the database hits its constraints and then breaks the rest of the transaction.
+// Then it won't do things like update the processing time on the relevant audit log event,
+// and I can't log any more errors :(
+const canInsertEntity = (datasetName, submissionDefId, entityUuid) => ({ one }) =>
```
sadiqkhoja, line 66: as discussed on slack, we can try `ON CONFLICT do something` to get rid of this function in another PR

lib/data/dataset.js
```diff
@@ -0,0 +1,54 @@
...
+// for dealing with the XForms XML schema. When a form us uploaded to Central,
+// we check the XML for an <entity> block, which will contain information
+// about datasets and mappings from form fields to
+// dataset properties.
+
+const Option = require('../util/option');
+const Problem = require('../util/problem');
+const { traverseXml, findOne, and, root, node, hasAttr, tree, attr } = require('../util/xml');
+
+const validateDatasetName = (name) => {
```
sadiqkhoja, line 20: This doesn't handle all the invalid XML identifiers. For example: `×people` (first character is multiplication sign #xd7) is invalid xml identifier but this function returns true.

I would reverse the regex to check valid string and if there's a match then return true.

May be try this: `^[\p{L}_][\p{L}.\d_-]*$. N.B. p{L}`. ref: https://stackoverflow.com/questions/3158274/what-would-be-a-regex-for-valid-xml-names

lib/data/entity.js
```diff
@@ -31,7 +36,29 @@ const getEntity = (entityFields, xml) => new Promise((resolve, reject) => {
...
       entity.data[field.propertyName] = text;
   });
-  stream.on('end', () => resolve(entity));
+  stream.on('end', () => {
+    // After collecting the entity data from the submission and form fields,
+    // we validate the data (properties must include dataset, label, and uuid)
+    // and assemble a valid entity data object. System properties like dataset
+    // label, uuid are in entity.system while the contents from the bound form
+    // fields are in enttiy.data.
+    const { dataset, label, id } = entity.system;
```
sadiqkhoja, line 45: can we move following lines to separate function `validatedEntity(entity)`? it will make our unit tests simpler

sadiqkhoja (approved):
All are minor comments, and can be incorporated in a separate PR. 

Regarding validation of dataset name, we can discuss about handling of unicodes in our weekly meeting with rest of the team

lib/data/dataset.js
ktuite: Thanks for the tip! I finally settled on: `/^(?!__)[\p{L}:_][\p{L}:\d_-]*$/u`

PR #633, opened 2022-10-04
Add Dataset verbs and add Dataset count to Project extended metadata

ktuite:
Changes to backend to support Frontend PR: https://github.com/getodk/central-frontend/pull/636 (first version of Datasets tab)

* Adds dataset count to Project extended metadata (so Frontend can show datasets tab and page only if some exist)
* Adds `dataset.list` and `dataset.read` verbs to the permissions for the following roles: admin, manager, and viewer.
* Adds some other verbs like `dataset.create` and `dataset.update` to admin and manager only (not viewer)

Still todo in this PR or a separate one:
- [x] Use these permissions in endpoints about datasets.
- [x] Write tests checking the permission guards work.

Tests.

Needed by frontend.

- [ ] Documentation needs to be updated to include this extra dataset count in the Project extended metadata description.

### DIFF
--- a/lib/model/migrations/20221003-01-add-dataset-verbs.js
+++ b/lib/model/migrations/20221003-01-add-dataset-verbs.js
@@ -7,11 +7,6 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-// backup.create and backup.terminate will be combined into a single verb,
-// config.set. If a role permitted backup.create but not backup.terminate (or
-// vice versa), that role will now permit both. However, that shouldn't be an
-// issue, because the two verbs should go hand-in-hand.
-
 const { without } = require('ramda');
 
 /* eslint-disable no-await-in-loop */
@@ -26,7 +21,7 @@ const up = async (db) => {
 };
 
 const down = async (db) => {
-  // grant rights.
+  // revoke rights.
   for (const system of [ 'admin', 'manager', 'viewer' ]) {
     const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
     const newVerbs = without([ 'dataset.list', 'dataset.read' ], verbs);
--- a/lib/model/migrations/20221003-01-add-dataset-verbs.js
+++ b/lib/model/migrations/20221003-01-add-dataset-verbs.js
@@ -15,7 +15,7 @@ const up = async (db) => {
   // grant rights.
   for (const system of [ 'admin', 'manager', 'viewer' ]) {
     const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
-    const newVerbs = verbs.concat([ 'dataset.list', 'dataset.read' ]);
+    const newVerbs = verbs.concat([ 'dataset.list', 'entity.list' ]);
     await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
   }
 };
@@ -24,7 +24,7 @@ const down = async (db) => {
   // revoke rights.
   for (const system of [ 'admin', 'manager', 'viewer' ]) {
     const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
-    const newVerbs = without([ 'dataset.list', 'dataset.read' ], verbs);
+    const newVerbs = without([ 'dataset.list', 'entity.list' ], verbs);
     await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
   }
 };
--- a/lib/model/migrations/20221003-01-add-dataset-verbs.js
+++ b/lib/model/migrations/20221003-01-add-dataset-verbs.js
@@ -1,0 +1,50 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+// backup.create and backup.terminate will be combined into a single verb,
+// config.set. If a role permitted backup.create but not backup.terminate (or
+// vice versa), that role will now permit both. However, that shouldn't be an
+// issue, because the two verbs should go hand-in-hand.
+
+const { without } = require('ramda');
+
+/* eslint-disable no-await-in-loop */
+
+const up = async (db) => {
+  // grant rights.
+  for (const system of [ 'admin', 'manager', 'viewer' ]) {
+    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
+    const newVerbs = verbs.concat([ 'dataset.list', 'dataset.read' ]);
+    await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
+  }
+
+  for (const system of [ 'admin', 'manager' ]) {
+    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
+    const newVerbs = verbs.concat([ 'dataset.create', 'dataset.update' ]);
+    await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
+  }
+};
+
+const down = async (db) => {
+  // grant rights.
+  for (const system of [ 'admin', 'manager', 'viewer' ]) {
+    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
+    const newVerbs = without([ 'dataset.list', 'dataset.read' ], verbs);
+    await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
+  }
+
+  for (const system of [ 'admin', 'manager' ]) {
+    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
+    const newVerbs = without([ 'dataset.create', 'dataset.update' ], verbs);
+    await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
+  }
+};
+
+module.exports = { up, down };
+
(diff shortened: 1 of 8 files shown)


lib/resources/forms.js
```diff
@@ -136,6 +136,18 @@ module.exports = (service, endpoint) => {
       .then(((form) => Promise.all([ Forms.publish(form), Submissions.clearDraftSubmissions(form.id) ])))
       .then(success)));
 
+  // Entity/Dataset-specific endpoint that is used to show how publishing
+  // a form will change any datasets mentioned in the form.
+  // Even though there are dataset-related permissions, we will use the form modification verbs
+  // for authentication because it seems weird to mix the ability/inability to alter a dataset
+  // with the ability to modify forms -- they should just be the same.
```
matthew-white, line 143: Above, I suggest using a different verb than `form.update`, but I agree with the thoughts in this comment. It seems like updating a form and updating a dataset go hand-in-hand.

lib/resources/datasets.js
```diff
@@ -9,30 +9,22 @@
...
-      .then((form) => auth.canOrReject('form.update', form)) // what verb is required here? a new one
-      .then(ensureFormDef)
-      .then(() => Datasets.getDatasetDiff(params.id))));
-
   service.get('/projects/:projectId/datasets/:name/download', endpoint(({ Datasets, Entities, Projects }, { params, auth }, _, response) =>
-    Projects.getById(params.projectId) // TODO: different lookup/auth scheme for datasets
+    Projects.getById(params.projectId)
       .then(getOrNotFound)
-      .then((project) => auth.canOrReject('project.read', project))
+      .then((project) => auth.canOrReject('dataset.read', project))
```
matthew-white, line 25: Are datasets actees? Should the actee here be the dataset rather than the project?

test/integration/api/projects.js
```diff
@@ -349,7 +355,7 @@ describe('api: /projects', () => {
           .then(({ body }) => {
             body.verbs.should.be.an.Array();
             body.verbs.length.should.be.greaterThan(39);
-            body.verbs.should.containDeep([ 'user.password.invalidate', 'project.delete' ]);
+            body.verbs.should.containDeep([ 'user.password.invalidate' ]);
```
matthew-white, line 358: Why remove `project.delete`? I'm thinking that it's a stand-in for "powerful project permissions."

test/integration/api/projects.js
```diff
@@ -359,12 +365,34 @@ describe('api: /projects', () => {
           .expect(200)
           .then(({ body }) => {
             body.verbs.should.be.an.Array();
-            body.verbs.length.should.be.lessThan(26);
-            body.verbs.should.containDeep([ 'assignment.create', 'project.delete' ]);
+            body.verbs.length.should.be.lessThan(30);
```
matthew-white, line 368: Since you ended up adding 2 verbs instead of 4, should this be 28 instead of 30?

test/integration/api/projects.js
```diff
@@ -359,12 +365,34 @@ describe('api: /projects', () => {
...
+              .set('X-Extended-Metadata', 'true')
+              .expect(200)
+              .then(({ body }) => {
+                body.verbs.should.eqlInAnyOrder([
+                  // eslint-disable-next-line no-multi-spaces
+                  'project.read',      // from role(s): formfill, viewer
+                  // eslint-disable-next-line no-multi-spaces
+                  'form.list',         // from role(s): formfill, viewer
+                  // eslint-disable-next-line no-multi-spaces
+                  'form.read',         // from role(s): formfill, viewer
```
matthew-white, line 390: Remove "viewer" from these comments?

matthew-white (approved):

lib/resources/datasets.js
matthew-white: Datasets aren't currently actees.

lib/resources/forms.js
```diff
@@ -136,6 +136,18 @@ module.exports = (service, endpoint) => {
...
+  service.get('/projects/:projectId/forms/:id/draft/dataset-diff', endpoint(({ Forms, Datasets }, { params, auth }) =>
+    Forms.getByProjectAndXmlFormId(params.projectId, params.id, false, Form.DraftVersion)
+      .then(getOrNotFound)
+      .then((form) => auth.canOrReject('form.update', form)) // use form verb
```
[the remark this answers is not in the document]
matthew-white: This would be something like this:

```js
.then((form) => Promise.all([
  auth.canOrReject('form.read', form)),
  auth.canOrReject('dataset.list', project))
]))
```

You'd need to get the project first.

lib/resources/forms.js
```diff
@@ -136,6 +136,18 @@ module.exports = (service, endpoint) => {
...
+      .then(ensureFormDef)
+      .then(() => Datasets.getDatasetDiff(params.id))));
```
matthew-white, line 149: Should this pass in the project ID? The `xmlFormId` won't be unique on its own.

lib/model/migrations/20221003-01-add-dataset-verbs.js
```diff
@@ -0,0 +1,38 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+// backup.create and backup.terminate will be combined into a single verb,
```
sadiqkhoja, line 10: this comment can be removed or updated, I guess it is copied from existing migration file

lib/model/migrations/20221003-01-add-dataset-verbs.js
```diff
@@ -0,0 +1,38 @@
...
+
+const { without } = require('ramda');
+
+/* eslint-disable no-await-in-loop */
+
+const up = async (db) => {
+  // grant rights.
+  for (const system of [ 'admin', 'manager', 'viewer' ]) {
+    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
+    const newVerbs = verbs.concat([ 'dataset.list', 'dataset.read' ]);
```
sadiqkhoja, line 23: Should we add `entity.list`, `entity.create`, `entity.update` in here as well? Just like we have form.[verb] and submission.[verb], I think dataset's verbs and entities' verbs should be different

lib/resources/forms.js
sadiqkhoja: agree, I missed adding `projectId` in the query

sadiqkhoja (approved):
Overall looks good, can be merged after resolving comments

lib/model/migrations/20221003-01-add-dataset-verbs.js
matthew-white: Maybe we could replace the one place we use `dataset.read` with `entity.list`? I've suggested making that endpoint &hellip;/datasets/:name/entities.csv rather than &hellip;/datasets/:name/download to make the path more RESTful and to match the similar endpoint /projects/:projectId/forms/:xmlFormId/submissions.csv. Especially if we make that change, `entity.list` feels like a good verb to use.

If we introduced `entity.list`, I don't think we'd need `dataset.read` at all. As a parallel, there's a `field_key.list` verb but no `field_key.read`, since (at least right now) you can't read app users individually.

Above, I commented that I agree that it makes sense that there's no separate `dataset.update` verb, since that goes hand-in-hand with `form.update`. I think something similar is probably true of `entity.create` and `entity.update`. Entities are created via submissions, so I don't think there needs to be a verb other than `submission.create`. We don't check verbs in workers I believe, so it's also not clear to me where we'd check for a verb like `entity.create`.

lib/resources/forms.js
matthew-white: Probably that change should come in a separate PR, I just thought I'd leave a comment here as I noticed it.

lib/model/migrations/20221003-01-add-dataset-verbs.js
lognaturel: > Entities are created via submissions

At some point we may want to introduce those verbs to give certain roles the ability to create entities directly (e.g. bulk upload through Excel doc) but seems like we can add them then.

lib/resources/forms.js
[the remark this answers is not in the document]
ktuite: I realized I can use the form here without another query to look up the project and the auth will implicitly look up the actee parent chain to the form's parent, which is the project or higher.

lib/model/migrations/20221003-01-add-dataset-verbs.js
ktuite: I like the idea of adding and using `entity.list` instead of `dataset.read` now. It fits now and if we really need a different `dataset.read`, we could add that later, too, along with all the other eventual entity verbs. 

It'll address the RESTfulness of the csv download url in another PR!

lib/resources/forms.js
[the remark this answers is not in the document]
matthew-white: Oh interesting! I can see the appeal of avoiding an extra query, though making an optimization around auth makes me a little nervous. I think this approach would work well with typical Central use cases (and everything in Frontend), but wouldn't work as well with behavior that we expose over the API. In the API docs, we advertise the existence of a:

> Form-specific Assignments resource at `projects/…/forms/…/assignments`, allowing granular role assignments on a per-Form basis.

For example, Frontend only lets you create `viewer` users on a project level, but I believe the API allows you to create a `viewer` to a single form. But now that the `viewer` role grants `dataset.list`, checking for `dataset.list` on the form rather than the project would provide unintentional access to information about datasets:

- Alice is an admin and makes Chelsea a `viewer` on a form in a project that has a dataset
- Chelsea shouldn't be able to access &hellip;/dataset-diff for the form, because she doesn't have any project-level access, so doesn't have access to the dataset.
- However, checking for `dataset.list` on the form will allow her to access the endpoint, because the `viewer` role grants her that verb.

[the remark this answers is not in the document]
ktuite: good point! did some TDD to add this test case, saw it in action, and changed the code.

PR #635, opened 2022-10-06
ci-benchmark: add final timeout for stream queries

alxndrsn:
Streams should time out after 2 minutes of inactivity, so we should wait for that before deciding if there are unterminated/hanging connections left open after requests have been served.


sadiqkhoja (approved):
Looks good, can be merged.

PR #636, opened 2022-10-06
Update docs for OData `$select`

lognaturel:
Fixing a few small omissions I noticed.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -32,6 +32,12 @@ Finally, **system information and configuration** is available via a set of spec
 
 Here major and breaking changes to the API are listed by version.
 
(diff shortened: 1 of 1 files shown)


sadiqkhoja (approved):
Looks good, can be merged.

We will update the version to v1.6.0 or v2.0.0 when we are closer to release, right?

lognaturel:
> We will update the version to v1.6.0 or v2.0.0 when we are closer to release, right?

We've changed the naming scheme! You can read more about the reasoning at https://forum.getodk.org/t/proposal-change-collect-versioning-from-major-minor-patch-to-year-release/32572

PR #637, opened 2022-10-10
Set Sentry tags

### LINKED ISSUE #547: Add version information to Sentry to identify versions
"I would love to send Sentry the Central version somehow so that we can filter on that. There are so many non-Cloud users on previous versions of Central, so most Sentry errors seem to be noise."
– @matthew-white 

I believe this can be as simple as setting the environment and release identifier
* https://docs.sentry.io/platforms/javascript/configuration/environments/
* https://docs.sentry.io/platforms/javascript/configuration/releases/#setting-a-release

Related: #521

matthew-white:
This PR sets [Sentry tags](https://docs.sentry.io/platforms/node/enriching-events/tags/) that are specified in an environment variable named `SENTRY_TAGS`. The approach is generic — any sort of tags could be set — but it was developed as part of getodk/central#317 in order to pass Central version information to Sentry. Together with getodk/central#317, this PR closes #547.

Nothing yet. Once this PR has been merged, I'll verify it on the QA server.

We discussed this broad strategy in getodk/central#317.

This change only touches Sentry. When I verify this change, that'll also verify that Sentry is still working.

No change needed.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { inspect } = require('util');
-const { isBlank } = require('../util/util');
+const { isBlank, isPresent } = require('../util/util');
 
 
 // Endpoints where query strings are sensitive and should be removed
@@ -132,6 +132,12 @@ const init = (config) => {
 
   const username = require('os').hostname();
   Sentry.setUser({ username });
+
+  if (isPresent(process.env.SENTRY_TAGS)) {
+    const tags = JSON.parse(process.env.SENTRY_TAGS);
+    for (const [key, value] of Object.entries(tags))
+      Sentry.setTag(key, value);
+  }
 
   return Sentry;
 };


lib/external/sentry.js
```diff
@@ -133,6 +133,11 @@ const init = (config) => {
   const username = require('os').hostname();
   Sentry.setUser({ username });
 
+  if (isPresent(process.env.SENTRY_TAGS)) {
+    for (const [key, value] of JSON.parse(process.env.SENTRY_TAGS))
```
matthew-white, line 137: At https://github.com/getodk/central/pull/317#discussion_r985143621, I was thinking of wrapping the `JSON.parse()` in a try/catch. However, I've ended up thinking that that'd be premature. We don't expose `SENTRY_TAGS` to users. And if there's a problem with `SENTRY_TAGS`, that would indicate a bug that we'd want to know about.

lib/external/sentry.js
```diff
@@ -133,6 +133,11 @@ const init = (config) => {
   const username = require('os').hostname();
   Sentry.setUser({ username });
 
+  if (isPresent(process.env.SENTRY_TAGS)) {
```
matthew-white, line 136: `SENTRY_TAGS` won't typically be present during local development.

lib/external/sentry.js
```diff
@@ -133,6 +133,12 @@ const init = (config) => {
...
+    const tags = JSON.parse(process.env.SENTRY_TAGS);
```
matthew-white, line 137: At https://github.com/getodk/central/pull/317#discussion_r985143621, I was thinking of wrapping the `JSON.parse()` in a try/catch. However, I've ended up thinking that that'd be premature. We don't expose `SENTRY_TAGS` to users. And if there's a problem with `SENTRY_TAGS`, that would indicate a bug that we'd want to know about.

alxndrsn (approved):

PR #638, opened 2022-10-11
ci-benchmark: switch final checks

alxndrsn:
`Streams` should time out after 2 minutes of inactivity, so we should wait for that before deciding if backend is correctly responding to login requests.


sadiqkhoja (approved):
looks good

PR #640, opened 2022-10-18
new: download dataset csv as attachment

sadiqkhoja:
# Changes:

- Updated `GET attachment/:name` endpoint to return dataset csv if it is linked
- Fixed dataset-diff SQL to include projectId


ktuite (approved):
Excited to see entities used in other forms!

PR #641, opened 2022-10-18
Audit logging of dataset and entity creation

ktuite:
Logs `dataset.create`, `dataset.update`, `entity.create` and `entity.create.error` events.

Some changes to make this happen:
* adding an `acteeId` on Datasets
* adding the above events to the listed types of audit log events under new categories of Dataset and Entity events

Tests

This will add new audit log events but shouldn't have any regression risks.

Need to document new types of events.

### DIFF
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -33,7 +33,7 @@ values (${actorId}, ${action}, ${acteeId}, ${(details == null) ? null : JSON.str
 // TODO: some sort of central table for all these things, not here.
 const actionCondition = (action) => {
   if (action === 'nonverbose')
-    return sql`action not in ('entity.create', 'submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'analytics')`;
+    return sql`action not in ('entity.create', 'entity.create.error', 'submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'analytics')`;
   else if (action === 'user')
     return sql`action in ('user.create', 'user.update', 'user.delete', 'user.assignment.create', 'user.assignment.delete', 'user.session.create')`;
   else if (action === 'field_key')
@@ -49,7 +49,7 @@ const actionCondition = (action) => {
   else if (action === 'dataset')
     return sql`action in ('dataset.create', 'dataset.update')`;
   else if (action === 'entity')
-    return sql`action in ('entity.create')`;
+    return sql`action in ('entity.create', 'entity.create.error')`;
 
   return sql`action=${action}`;
 };
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -9,7 +9,7 @@
 
 const { sql } = require('slonik');
 const { map } = require('ramda');
-const { Actee, Actor, Audit, Form, Project } = require('../frames');
+const { Actee, Actor, Audit, Dataset, Form, Project } = require('../frames');
 const { extender, equals, page, QueryOptions } = require('../../util/db');
 const Option = require('../../util/option');
 const { construct } = require('../../util/util');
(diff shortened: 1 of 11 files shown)


lib/model/query/datasets.js
```diff
@@ -128,8 +128,27 @@ const _getByNameSql = ((fields, datasetName, projectId) => sql`
...
+        ? datasetOption.get().acteeId
+        : Actees.provision('dataset', project).then((actee) => (actee.id))))
+    .then((acteeId) =>
+      all(_createOrMerge(dataset, fields, acteeId)))
+    .then(() => Datasets.getByProjectAndName(dataset.projectId, dataset.name))
+    .then((ds) => ds.get());
+
+createOrMerge.audit = (dataset, _, fields) => (log) =>
+  ((dataset.revisionNumber === 0)
+    ? log('dataset.create', dataset, { fields: fields.map((f) => [f.path, f.propertyName]) })
```
matthew-white, line 149: Since datasets are a new kind of actee, I think we'll need to modify this query as well:

https://github.com/getodk/central-backend/blob/c200de1d0a8883f5126cfb348287096f8d86500f/lib/model/query/audits.js#L69-L93

Frontend will request /v1/audits with extended metadata, and I think we'll need that to return dataset names.

lib/model/query/datasets.js
```diff
@@ -128,8 +128,27 @@ const _getByNameSql = ((fields, datasetName, projectId) => sql`
...
+    : log('dataset.update', dataset, { fields: fields.map((f) => [f.path, f.propertyName]) }));
```
matthew-white, line 150: New audit log actions generally need to be added to `actionCondition()` in lib/model/query/audits.js. There's no `if` there for datasets, so I think that'll need to be added. It should be possible to request /v1/audits?action=dataset and have the API return audit log entries for both `dataset.create` and `dataset.update`.

lib/worker/entity.js
```diff
@@ -7,10 +7,24 @@
...
+      .then((entity) => {
+        // TODO: figure out target for this audit. Could be the form of the
+        // original event or it could be the entity's dataset.
+        Audits.log({ id: event.actorId }, 'entity.create', null,
+          { entityUuid: entity.uuid,
+            submissionId: event.details.submissionId,
+            submissionDefId: event.details.submissionDefId });
+      })
+      .catch((problem) => {
+        Audits.log({ id: event.actorId }, 'entity.create.error', null,
```
matthew-white, line 22: Similarly here, I think there'll need to be a new `if` in `actionCondition()` for entities, in order to support /v1/audits?action=entity. We don't show submission audit log entries in the Frontend audit log, and I think we'll probably want to exclude entity audit log entries as well. To accomplish that, these new actions will need to be added to the `nonverbose` condition, which Frontend requests by default.

matthew-white (commented):
I didn't take a look at the code, I just wanted to make a note about a couple of things that I know will be needed for Frontend. Also, just FYI, we list all the audit log actions in the API docs, so that'll probably need to be updated.

lib/model/query/audits.js
```diff
@@ -33,7 +33,7 @@ values (${actorId}, ${action}, ${acteeId}, ${(details == null) ? null : JSON.str
 // TODO: some sort of central table for all these things, not here.
 const actionCondition = (action) => {
   if (action === 'nonverbose')
-    return sql`action not in ('submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'analytics')`;
+    return sql`action not in ('entity.create', 'submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'analytics')`;
```
ktuite, line 36: also `entity.create.error`

lib/model/query/audits.js
```diff
@@ -46,6 +46,10 @@ const actionCondition = (action) => {
     return sql`action in ('form.create', 'form.update', 'form.delete', 'form.restore', 'form.purge', 'form.attachment.update', 'form.submission.export', 'form.update.draft.set', 'form.update.draft.delete', 'form.update.publish')`;
   else if (action === 'submission')
     return sql`action in ('submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update')`;
+  else if (action === 'dataset')
+    return sql`action in ('dataset.create', 'dataset.update')`;
+  else if (action === 'entity')
+    return sql`action in ('entity.create')`;
```
ktuite, line 52: also add `entity.create.error`

test/integration/api/forms/datasets.js
```diff
@@ -123,6 +123,41 @@ describe('api: /projects/:id/forms (entity-handling)', () => {
...
+  describe('dataset audit logging', () => {
+    it('should log dataset creation in audit log', testService(async (service, { Audits }) => {
+      await service.login('alice', (asAlice) =>
+        asAlice.post('/v1/projects/1/forms')
+          .send(testData.forms.simpleEntity)
+          .set('Content-Type', 'text/xml')
+          .expect(200));
+
+      const audit = await Audits.getLatestByAction('dataset.create').then((o) => o.get());
+      audit.details.fields.should.eql([[ '/name', 'first_name' ], [ '/age', 'age' ]]);
```
ktuite, line 136: maybe test other things about this audit log entry like actor and actee

ktuite (commented):

PR #646, opened 2022-10-24
new: autolink attachment to dataset if name matches

sadiqkhoja:
# Changes:
- Autolink attachments to dataset if name is same


ktuite (commented):
Could you write some tests for:
1. if a dataset is linked to attachment, making a new version of a form preserves that attachment
2. if you unlink the dataset (if you are undoing this automatic linking) from the attachment, that making a new version of the form does *not* re-link it?

ktuite (approved):
Brief interactive review with @sadiqkhoja

PR #647, opened 2022-10-24
Fix: update dataset form attachment on publish

ktuite:
Fixes a bug I noticed while trying to make use of entities.


sadiqkhoja (approved):
I made this change in my PR as well, but you can merge this. I'll rebase

PR #648, opened 2022-10-25
new: dataset-diff endpoint for published forms

sadiqkhoja:
#Changes:
- Response now returns all the properties of the dataset, regardless of which form defined it
- Added `inForm` fields in the response, it distinguishes if the a property is defined by the current form. Had to make this change to cater the need of the frontend design
- Added `dataset-diff` endpoint for published forms, it doesn't return `isNew` flag

### DIFF
--- a/test/integration/api/forms/datasets.js
+++ b/test/integration/api/forms/datasets.js
@@ -190,7 +190,6 @@ describe('api: /projects/:id/forms/draft/dataset-diff', () => {
             .expect(403))))));
 
   it('should return all properties of dataset', testService(async (service) => {
-    // Upload a form and then create a new draft version
     await service.login('alice', (asAlice) =>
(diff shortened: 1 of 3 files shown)


test/integration/api/forms/datasets.js
```diff
@@ -325,8 +319,56 @@ describe('api: /projects/:id/forms/draft/dataset-diff', () => {
...
+              body.should.be.eql([{
+                name: 'people',
+                properties: [
+                  { name: 'age', inForm: true },
+                  { name: 'first_name', inForm: false },
+                  { name: 'last_name', inForm: true }
+                ]
+              }]);
+            }))));
+  }));
```
sadiqkhoja, line 373: add a test to ensure this endpoint doesn't unpublished properties

ktuite (approved):
interactive review

PR #650, opened 2022-10-28
Revert "Allow submission edits on closed forms (#530)"

alxndrsn:
This reverts commit 8377a0f1f62607098148af77a939f2cce3710994.

This commit may be re-introduced alongside similar changes to the OpenRosa API.


matthew-white (approved):
Sorry for the delay in review!

What do you think about `.skip`ping the tests rather than removing them? We've done that in other cases when we've reverted work, and it makes for a smaller diff. Feel free to remove though if that seems better!

PR #655, opened 2022-11-01
Fix: Reprocessing doubly-approved entity submission

ktuite:
There was a bug with reapproving an entity-creating submission where when the entity worker would just crash the backend server.

Now it cleanly fails to make a new entity without crashing.

Tests

Small fix of promise chain bug.

It's good -- Less crashing!

No.


sadiqkhoja (approved):

lognaturel:
Yay, less crashing!

PR #659, opened 2022-11-09
Add label and dataset name to entity.create audit log

ktuite:
Frontend needs to have the entity **label** and **dataset name** in the details of the `entity.create` audit log in order to display this info on the submission activity feed.

Tests and seeing it in action.

It's pretty straightforward. It does mean duplicating the data of the entity label, but saves us from changing how the audit log pulls in extra info. 

Lets the right information come through the API to show on Frontend.

No, we don't describe the details of the different types of audit events in the API docs.


sadiqkhoja (approved):
looks good

PR #683, opened 2022-11-15
fixes #666: dataset and property has publishedAt

### LINKED ISSUE #666: Purging forms and submissions that create entities
When a form is purged, it purges the submissions associated with that form. Entities have a `submissionDefId` foreign key that links back to the source submission def, but that foreign key constraint will be violated when the submission is purged. We will change this by allowing the submission def ID to be set to null on the entity. Matt left a comment and example test [here](https://github.com/getodk/central-backend/issues/649#issuecomment-1301611258).

Form purging also affects how the dataset draft endpoints work. Currently, much of the state about a dataset and its properties is inferred from the state of the forms that contribute properties to that dataset. We will add `publishedAt` flags to the datasets themselves and the dataset properties to be a bit more explicit about when properties or whole datasets became available, and rely less on the state of the forms themselves in case they get deleted purged. More notes [here](https://github.com/getodk/central-backend/issues/649#issuecomment-1301627820).

Conceptually, datasets and entities exist in a different realm than forms and submissions. While a form and its submissions get deleted, datasets and entities wont get deleted, though the link between them will be broken. We'll come back to dataset and entity deletion soon.

sadiqkhoja:
Closes #666 

- Added integration tests
- Performed manual testing with frontend

Previous solution of having implicit status for dataset and its properties was not working, especially when form gets purged

I guess whole dataset/entities feature should be tested

NA

### DIFF
--- a/lib/model/migrations/20221114-01-explict-dataset-publish.js
+++ b/lib/model/migrations/20221114-01-explict-dataset-publish.js
@@ -10,6 +10,8 @@
 const up = async (db) => {
   await db.schema.table('entity_defs', (eDef) => {
     eDef.integer('submissionDefId').alter({ nullable: true });
+    eDef.dropForeign('submissionDefId');
+    eDef.foreign('submissionDefId').references('submission_defs.id').onDelete('SET NULL');
   });
 
   await db.schema.table('datasets', (ds) => {
@@ -24,6 +26,8 @@ const up = async (db) => {
 const down = async (db) => {
   await db.schema.table('entity_defs', (eDef) => {
     eDef.integer('submissionDefId').alter({ nullable: false });
+    eDef.dropForeign('submissionDefId');
+    eDef.foreign('submissionDefId').references('submission_defs.id').onDelete('NO ACTION');
   });
 
   await db.schema.table('datasets', (ds) => {
--- a/lib/model/migrations/20221114-01-explict-dataset-publish.js
+++ b/lib/model/migrations/20221114-01-explict-dataset-publish.js
@@ -1,0 +1,38 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.schema.table('entity_defs', (eDef) => {
+    eDef.integer('submissionDefId').alter({ nullable: true });
+  });
+
+  await db.schema.table('datasets', (ds) => {
+    ds.dateTime('publishedAt');
+  });
+
+  await db.schema.table('ds_properties', (dsProp) => {
+    dsProp.dateTime('publishedAt');
+  });
+};
+
+const down = async (db) => {
+  await db.schema.table('entity_defs', (eDef) => {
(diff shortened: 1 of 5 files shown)


lib/model/query/forms.js
```diff
@@ -258,6 +262,13 @@ with redacted_audits as (
     left outer join form_defs on coalesce(forms."currentDefId", forms."draftDefId") = form_defs.id
     where actees.id = forms."acteeId"
     and ${_trashedFilter(force, id, projectId)}
+  ), update_entities as (
```
ktuite, line 265: if we change the migration above to "on delete set null", then wont this just happen automatically?

lib/model/migrations/20221114-01-explict-dataset-publish.js
```diff
@@ -0,0 +1,38 @@
...
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.schema.table('entity_defs', (eDef) => {
+    eDef.integer('submissionDefId').alter({ nullable: true });
```
ktuite, line 12: In addition to setting this to nullable, I was thinking you could set this onDelete command. This is like my strategy in `./20220121-01-form-cascade-delete.js` but using `.onDelete('set null')` instead of `.onDelete('cascade')`:
```
eDef.dropForeign('submissionDefId');
eDef.foreign('submissionDefId').references('submission_defs.id').onDelete('set null');
```

lib/model/query/forms.js
```diff
@@ -159,12 +161,14 @@ createVersion.audit.withResult = true;
...
-const publish = (form) => ({ Forms }) => {
+const publish = (form) => ({ Forms, Datasets }) => {
   if (form.draftDefId !== form.def.id) throw Problem.internal.unknown();
 
+  const publishedAt = (new Date()).toISOString();
   return Promise.all([
     Forms._update(form, { currentDefId: form.draftDefId, draftDefId: null }),
-    Forms._updateDef(form, { draftToken: null, enketoId: null, publishedAt: (new Date()).toISOString() })
+    Forms._updateDef(form, { draftToken: null, enketoId: null, publishedAt }),
+    Datasets.publish(form.def.id, publishedAt)
```
ktuite, line 171: A question that might be better for an interactive review: this will be called for every form when it gets published so I just wanted to make sure this wouldn't have any negative effects if called on a form without any datasets. Maybe it could have a more non-committal name like "Datasets.publishIfPresent", "maybePublish", "possiblyPublish"... or could we check `entityRelated` on the form and only conditional run this?

ktuite (commented):
Took a quick async look late Tuesday night and left a few comments. Will do a more thorough review on Wedneday.

matthew-white:
I'll let @ktuite take a look, but let me know if there's anything you want me to look at here!

One question I had was about backfilling. If a database already has a dataset, then after this PR, I think that dataset will have a `null` `publishedAt`, so won't be listed in &hellip;/datasets. I think that's reasonable, but I just wanted to make sure that that's the expectation.

lib/model/query/forms.js
sadiqkhoja: just realized, I have to add audit logs for publish as well

matthew-white: You two may have discussed this already, but I think we'll also need to add this new audit action to Frontend.

sadiqkhoja: yes, I have added it in https://github.com/getodk/central-frontend/pull/641

sadiqkhoja: renamed publish function to publishIfExists

PR #685, opened 2022-11-17
fixes #669: cleaning up code for entity worker

### LINKED ISSUE #669: Entity Worker Improvements
We're going to use a nested transaction inside the entity worker so that we can rely on the database constraints to flag any errors about entity creation. The "current" (before this issue gets fixed) version of the entity worker does some weird stuff like has a complicated pre-check query to test if the entity can successfully be inserted. We'll be able to do away with that extra query with this new approach.

We also need to clearly handle each of the possible submission processing outcomes:
1. it could return null (the submission doesn't create an entity)
2. it could successfully create an entity and return it
3. there could be a validation error
4. there could be a database constraint error 
    a. there could be an error that comes from attempting to reprocess a submission that has already been processed, and we could do this with a db constraint or with some other pre-check
5. there could be a validation error that comes from the fact that we're trying to process an encrypted submission (see more #668)
6. there could be some other kind of internal error, e.g. a database connection problem, that is a kind of error we'd like to have the worker retry

ktuite:
Closes #669 by cleaning up the entity creation code used by the worker that processes submission update events. 

- went back to relying on postgres to flag entity creation errors due to constraint violations (use a nested transaction to do this and properly catch the errors and then be able to continue logging)
- made the processing code use async/await to be easier to follow
- moved the audit logging code for entity creation and error events into the processing code, not the worker
- refactored the tests to be clearer and handle different kinds of failure/non-failure cases:
    * submission should not be processed
    * submission successfully makes entity
    * submission fails to make entity because of validation error
    * submission fails to make entity because of constraint violation error (duplicate uuid or something)

Tests.

It's a lot better than it was! This change gets rid of a big weird hacky code block that was trying to carefully catch specific kind of errors.

n/a

no

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -10,7 +10,6 @@
 const { sql } = require('slonik');
 const { Entity } = require('../frames');
 const { unjoiner } = require('../../util/db');
-const Problem = require('../../util/problem');
 const { map } = require('ramda');
 const { construct } = require('../../util/util');
 
@@ -65,28 +64,21 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
 
 // Entrypoint to where submissions (a specific version) become entities
 const _processSubmissionDef = (submissionDefId) => async ({ Datasets, Entities, Submissions }) => {
-  try {
-    const existingEntity = await Entities.getDefBySubmissionDefId(submissionDefId);
-    // If the submission has already been used to make an entity, don't try again
-    // and don't log it as an error.
-    if (existingEntity.isDefined())
-      return null;
-    const submissionDef = await Submissions.getDefById(submissionDefId).then((s) => s.get());
-    const fields = await Datasets.getFieldsByFormDefId(submissionDef.formDefId);
-    // No fields found for this formDefId means there is nothing entity-related to parse out
-    // of this submission. Even entity system properties like ID and dataset name would be here
-    // if the form def is about datasets and entities.
-    if (fields.length === 0)
-      return null;
-    const partial = await Entity.fromSubmissionXml(submissionDef.xml, fields);
-    const entity = await Entities.createNew(partial, submissionDef);
-    return entity;
-  } catch (problem) {
-    if (problem.problemCode === 409.14)
-      throw problem;
-    else
-      throw Problem.user.entityViolation({ reason: 'Something went wrong validating or inserting' });
-  }
+  const existingEntity = await Entities.getDefBySubmissionDefId(submissionDefId);
+  // If the submission has already been used to make an entity, don't try again
+  // and don't log it as an error.
+  if (existingEntity.isDefined())
+    return null;
+  const submissionDef = await Submissions.getDefById(submissionDefId).then((s) => s.get());
+  const fields = await Datasets.getFieldsByFormDefId(submissionDef.formDefId);
+  // No fields found for this formDefId means there is nothing entity-related to parse out
+  // of this submission. Even entity system properties like ID and dataset name would be here
+  // if the form def is about datasets and entities.
+  if (fields.length === 0)
+    return null;
+  const partial = await Entity.fromSubmissionXml(submissionDef.xml, fields);
+  const entity = await Entities.createNew(partial, submissionDef);
+  return entity;
 };
 
 const processSubmissionEvent = (event) => (container) =>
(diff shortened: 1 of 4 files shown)


lib/model/query/entities.js
```diff
@@ -62,55 +62,49 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
...
-                  .catch((problem) => reject(problem)))))));
-  });
+const _processSubmissionDef = (submissionDefId) => async ({ Datasets, Entities, Submissions }) => {
+  try {
+    const existingEntity = await Entities.getDefBySubmissionDefId(submissionDefId);
+    // If the submission has already been used to make an entity, don't try again
+    // and don't log it as an error.
+    if (existingEntity.isDefined())
+      return null;
+    const submissionDef = await Submissions.getDefById(submissionDefId).then((s) => s.get());
```
ktuite, line 74: Possibly add test: What if the submission def doesn't exist?

lib/model/query/entities.js
```diff
@@ -62,55 +62,49 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
...
+    // No fields found for this formDefId means there is nothing entity-related to parse out
+    // of this submission. Even entity system properties like ID and dataset name would be here
+    // if the form def is about datasets and entities.
+    if (fields.length === 0)
+      return null;
+    const partial = await Entity.fromSubmissionXml(submissionDef.xml, fields);
+    const entity = await Entities.createNew(partial, submissionDef);
+    return entity;
+  } catch (problem) {
+    if (problem.problemCode === 409.14)
```
ktuite, line 85: Clarify with comment that these are all mainly entity validation errors

ktuite (commented):
Interactive review

Question: can we make a test to emulate DB failure to see worker try again?

matthew-white:
> Question: can we make a test to emulate DB failure to see worker try again?

Could you modify the `container` within the test? Something like:

```js
container.Entities.getDefBySubmissionDefId = () => reject(new Error('completely unhandled error'));
```

lib/model/query/entities.js
```diff
@@ -62,55 +62,49 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
...
+      throw problem;
+    else
+      throw Problem.user.entityViolation({ reason: 'Something went wrong validating or inserting' });
```
ktuite, line 88: classify errors a bit more to distinguish between system errors and data errors
- uuid clash
- value of something is not in another table as expected
- some piece of data is missing

ktuite (commented):
Interactive review

Question: can we make a test to emulate DB failure to see worker try again?

lib/model/query/entities.js
```diff
@@ -62,55 +62,49 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
...
+          { entity: { uuid: entity.uuid, label: entity.label, dataset: entity.aux.dataset.name },
+            submissionId: event.details.submissionId,
+            submissionDefId: event.details.submissionDefId });
+      }
+    })
+    .catch((problem) =>
+      container.Audits.log({ id: event.actorId }, 'entity.create.error', null,
+        { submissionId: event.details.submissionId,
+          submissionDefId: event.details.submissionDefId,
+          problem }));
```
ktuite, line 107: does this log the stack trace?

lib/model/query/entities.js
ktuite: if we don't make this new problem, but we just pass whatever other error we get through, make sure that in the outer logging step, we strip out the stack trace if it is there.

matthew-white (approved):

ktuite (commented):

lib/model/query/entities.js
ktuite: I've switched to catching the errors more generically so this comment isn't relevant anymore.

lib/model/query/entities.js
ktuite: it does not log the stack trace.

ktuite:
I've updated the code to catch and log whatever kind of error the inner `processSubmissionEvent` might throw.  I've also changed the tests to reflect this.

Now you can see when a database uniqueness Problem comes through, or even an error that's not one of our Problems but something like a slonik error.

I haven't figured out an even deeper way of messing with the container to trigger an error. Deleting database tables or breaking the submission event all cause errors that get caught in a reasonable way.

I just wanted to write out another thought behind this error handling. In many other cases of workers doing things in Central, there's a clear reason to retry the task if there is a problem. With entities, if there is any sort of problem processing the submission, it seems okay to not retry, and to rely on the user to notice and re-approve the submission to get it to reprocess. At least for this first pass of entities while we discover what kinds of failure modes are possible.

lib/model/query/entities.js
```diff
@@ -62,55 +61,52 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
...
+      // known Problems, just in case there are weird error details we don't want to
+      // expose. In experimenting with breaking the code, it seems that non-Problem errors
+      // convert to empty objects through the JSON.stringify transformation within audit logging
+      // so this probably isn't even an issue.
+      // The JSON.stringify appears to strip out error.stack so that doesn't make it to the
+      // log details even for our Problems.
+      container.Audits.log({ id: event.actorId }, 'entity.create.error', null,
+        { submissionId: event.details.submissionId,
+          submissionDefId: event.details.submissionDefId,
+          errorMessage: err.message,
```
matthew-white, line 108: It'd be nice to show this message in Frontend at some point in case the error isn't a Problem. I don't think we need to do that for v2022.3 though. Let me know if it'd be useful for me to file a Frontend issue about this!

ktuite: Yes! we should be able to update frontend to check for `errorMessage` instead. It will have slightly more detail (include the name of the problem concatenated with the reason) than just the currently displayed Problem reason.

PR #687, opened 2022-11-17
fixes 681: added constrainst to check datasetId is null for non-data attachments

### LINKED ISSUE #681: Only allow form attachment of type file to be linked to dataset
I don't think I saw this tracked elsewhere, though maybe I missed it: https://github.com/getodk/central-backend/pull/576#discussion_r1012541043

sadiqkhoja:
Closes #681 

Adds a database constraint on `form_attachments` table to check datasetId is null for non-file/non-data type attachments

I could have added condition to existing constraint but to keep things segregated I have added a new constraint

NA

NA

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -326,9 +326,9 @@ describe('datasets and entities', () => {
                   }))))));
     });
 
-    describe('check only blobId or datasetId is set', () => {
-      // this scenario will never happen by just using APIs, adding this test for safety
-      it('should throw problem 501.11 if both are being set', testService((service, { Forms, FormAttachments, Datasets }) =>
+    // these scenario will never happen by just using APIs, adding following tests for safety
+    describe('check datasetId constraints', () => {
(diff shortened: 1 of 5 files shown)


ktuite (approved):

test/integration/api/datasets.js
```diff
@@ -349,6 +349,24 @@ describe('datasets and entities', () => {
                 .catch(error => {
                   error.constraint.should.be.equal('check_blobId_or_datasetId_is_null');
                 }))))));
+
+      it('should throw problem if datasetId is being set for non-data type', testService((service, { Forms, FormAttachments, Datasets }) =>
```
sadiqkhoja, line 353: modify this test to call PATCH attachment

sadiqkhoja: Didn't modify this test, instead added two additional tests to ensure that dataset can only be linked to attachments with type='file'

ktuite (approved):
New changes and tests look good!

PR #656, opened 2022-11-04
updated API.md to reflect dataset/entities changes

sadiqkhoja:
Update API doc to reflect dataset / entities related changes.

You can view the draft version of the document at https://centralbackendsk.docs.apiary.io/#introduction/changelog

PR #689, opened 2022-11-17
fixes #671: only published dataset and its published properties should be accessible

### LINKED ISSUE #671: Access to Datasets (unpublished, non-existent, and more)
There are a number of smaller issues/comments from code review that fall under how datasets are accessed: 

- [x] Downloading a non-existent dataset and getting a 500 error instead of a 404 error. To fix this, a related query that fetches datasets needs to return an option instead of an actual dataset. More notes [here](https://github.com/getodk/central-backend/issues/649#issuecomment-1301629912).
- [x] Change download path to `/projects/:projectId/datasets/:name/entities.csv` Comment [here](https://github.com/getodk/central-backend/pull/576#discussion_r1012103011).
- [x] Dataset CSV streaming endpoints return data even if dataset is not yet published. See notes [here](https://github.com/getodk/central-backend/issues/649#issuecomment-1309046061).
- [x] You can link a form attachment to a dataset that is not yet published. See notes [here](https://github.com/getodk/central-backend/issues/649#issuecomment-1309055161).
- [x] Dataset property included in CSV download even if form for that dataset is not yet published. Note [here](https://github.com/getodk/central-backend/issues/649#issuecomment-1309167680). This might be addressed by issue #666  that involves adding `publishedAt` to dataset properties.

sadiqkhoja:
Closes #671

Added integration tests

none

APIs should now return only published datasets with its published properties. 
On the frontend we have to update download dataset link and test that linking dataset to attachment is working - especially it links only published datasets

We should mention that dataset related APIs only returns published datasets, explicitly mention what do we mean by "published datasets" and "published property"

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -13,6 +13,7 @@ const { unjoiner } = require('../../util/db');
 const { Dataset, Form } = require('../frames');
 const { map, reduce, compose, pickBy, startsWith, nthArg, assoc, keys, curry, nth, isEmpty, isNil, either, reduceBy, groupBy } = require('ramda');
 const { construct } = require('../../util/util');
+const Option = require('../../util/option');
 
 // It removes prefix from all the key of an object
 const removePrefix = curry((prefix, obj) => compose(reduce((acc, key) => assoc(key.replace(prefix, ''), obj[key], acc), {}), keys)(obj));
@@ -122,6 +123,8 @@ const _getByNameSql = ((fields, datasetName, projectId) => sql`
         ds_properties.id = ds_property_fields."dsPropertyId"
     WHERE datasets.name = ${datasetName}
       AND datasets."projectId" = ${projectId}
+      AND datasets."publishedAt" IS NOT NULL
+      AND ds_properties."publishedAt" IS NOT NULL
  `);
 
 
@@ -161,11 +164,13 @@ const getById = (datasetId) => ({ all }) =>
     .then(asArray)
     .then(nth(0));
 
+// Returns only published dataset with its published properties
 const getByName = (datasetName, projectId) => ({ all }) =>
   all(_getByNameSql(unjoiner(Dataset, Dataset.Property, Dataset.PropertyField).fields, datasetName, projectId))
     .then(makeHierarchy)
     .then(asArray)
-    .then(nth(0));
+    .then(nth(0))
+    .then(Option.of);
 
 // Returns list of dataset for a given projectId
 // Properties and field mappings are not returned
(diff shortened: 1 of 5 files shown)


lib/model/query/datasets.js
```diff
@@ -161,11 +164,13 @@ const getById = (datasetId) => ({ all }) =>
     .then(asArray)
     .then(nth(0));
 
+// Returns only published dataset with its published properties
 const getByName = (datasetName, projectId) => ({ all }) =>
```
sadiqkhoja, line 168: `Datasets.getByName` is called from a single place and it expects to receive only published dataset with its published properties

lib/model/query/datasets.js
```diff
@@ -222,8 +227,8 @@ const getDatasetDiff = (projectId, xmlFormId, forDraft) => ({ all }) => all(sql`
   .then(reduceBy((acc, { propertyName, isPropertyNew, inForm }) => (propertyName ? acc.concat({ name: propertyName, isNew: forDraft ? isPropertyNew : undefined, inForm }) : acc), [], (row) => `${row.datasetName},${row.isDatasetNew}`))
   .then(r => Object.keys(r).map(k => ({ name: k.split(',')[0], isNew: forDraft ? k.split(',')[1] === 'true' : undefined, properties: r[k] })));
 
-const getByProjectAndName = (projectId, name) => ({ maybeOne }) => maybeOne(sql`
-SELECT * FROM datasets WHERE "projectId" = ${projectId} AND name = ${name}`)
+const getByProjectAndName = (projectId, name, excludeUnpublished = false) => ({ maybeOne }) => maybeOne(sql`
```
sadiqkhoja, line 230: if `excludeUnpublished` is true then we only return published properties

lib/resources/datasets.js
```diff
@@ -19,12 +19,12 @@ module.exports = (service, endpoint) => {
       .then((project) => auth.canOrReject('dataset.list', project))
       .then(() => Datasets.getAllByProjectId(params.id))));
 
-  service.get('/projects/:projectId/datasets/:name/download', endpoint(({ Datasets, Entities, Projects }, { params, auth }, _, response) =>
+  service.get('/projects/:projectId/datasets/:name.csv', endpoint(({ Datasets, Entities, Projects }, { params, auth }, _, response) =>
```
sadiqkhoja, line 22: I like the `:name.csv` endpoint, it works with integration tests that we have, do you see any problem if dataset name is strange? I can definitely change this to `entities.csv`

ktuite (approved):

test/integration/api/datasets.js
```diff
@@ -62,15 +62,15 @@ describe('datasets and entities', () => {
                   }))))));
     });
 
-    describe('projects/:id/datasets/:dataset/download GET', () => {
```
sadiqkhoja, line 65: add a test for checking unpublished dataset can't be linked to a form attachment

lognaturel (changes requested):
See https://github.com/getodk/central-backend/pull/656#discussion_r1026761828 for one more revision to the dataset download endpoint!

lognaturel (approved):

PR #694, opened 2022-11-18
Make changes related to linking an attachment to a dataset

### LINKED ISSUE #672: Permission check when linking form attachment to dataset
Quote from [Matt](https://github.com/getodk/central-backend/issues/649#issuecomment-1302158032):

> In #633, we accounted for the case where a user can `form.update` but not `dataset.list` (see https://github.com/getodk/central-backend/pull/633#discussion_r994086934). I think we should also think about the case where a user can `form.update` but not `entity.list`. Right now, a user who can `form.update` can effectively `entity.list` a dataset if the dataset name is known, by linking an attachment in the form to the dataset. I'm thinking that we should have PATCH /v1/projects/:projectId/forms/:id/draft/attachments/:name check for `entity.list` in addition to `form.update` (if `datasetId` is specified): before we allow you to provide entity access to all submitters, we should check that you yourself have that access.

We want to be careful not to leak entity info to users who shouldn't have access to entities. Another way of phrasing the statement above: if a user doesn't have access to a project (doesn't have access to datasets/entities in that project) but is a manager of a form (permission which can be granted through the API), they shouldn't be allowed to link form attachments to a dataset. BUT if someone with greater (project-level) permissions links the form attachment to a dataset first, then they should be able to see it.

matthew-white:
Closes #672.

### DIFF
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -189,7 +189,7 @@ module.exports = (service, endpoint) => {
       .then(getOrNotFound)
       .then((form) => auth.canOrReject('form.update', form))
       .then((form) => Promise.all([
-        Datasets.getByProjectAndName(params.projectId, params.name.replace(/\.csv$/, '')).then(getOrNotFound),
+        Datasets.getByProjectAndName(params.projectId, params.name.replace(/\.csv$/i, '')).then(getOrNotFound),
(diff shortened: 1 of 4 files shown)


ktuite (approved):

sadiqkhoja (approved):
perfect!

PR #695, opened 2022-11-18
Don't return meta/entity fields for form def not associated with dataset

### LINKED ISSUE #686: Bug about not always using the right form def fields
I noticed this situation where I had uploaded an entity form, created a submission, but then rolled back all the dataset/entity database migrations and reapplied them, leaving empty dataset and entity tables.

When I republished the form to reinitialize the dataset, I tried approving an old submission for the *previous* version of the form.
To my surprise, it made an entity, even though that submission was linked to an earlier version of the form that no longer had any datasets or dataset fields associated with it. I've recreated that situation in the following test.

I'm thinking the query in `Datasets.getFieldsByFormDefId` is too liberal - I wanted to get the dataset property fields and the entity meta-data fields which may not be dataset properties, but in trying to do that, the joins are probably wrong and the fields for the wrong form version are also coming through.

```js
  it.only('should test out this weird case i saw', testService(async (service, container) => {
    await service.login('alice', (asAlice) =>
      asAlice.post('/v1/projects/1/forms?publish=true')
        .expect(200)
        .then(() => asAlice.post('/v1/projects/1/forms/simpleEntity/submissions')
          .send(testData.instances.simpleEntity.one)
          .expect(200)));

    // delete all the dataset stuff
    await container.run(sql`delete from dataset_form_defs`);
    await container.run(sql`delete from ds_property_fields`);
    await container.run(sql`delete from ds_properties`);
    await container.run(sql`delete from datasets`);

    // re-draft and re-publish the form
    await service.login('alice', (asAlice) =>
      asAlice.post('/v1/projects/1/forms/simpleEntity/draft')
        .expect(200));

    // dataset should come back, even as a draft
    const { count } = await container.one(sql`select count(*) from datasets`);
    count.should.equal(1);

    await service.login('alice', (asAlice) =>
      asAlice.post('/v1/projects/1/forms/simpleEntity/draft/publish?version=new')
        .expect(200)
        .then(() => asAlice.patch('/v1/projects/1/forms/simpleEntity/submissions/one')
          .send({ reviewState: 'approved' })
          .expect(200)));

    const forms = await container.all(sql`select "id", "xmlFormId" from forms`);
    // eslint-disable-next-line no-console
    console.log(forms);

    const row = await container.all(sql`select "id", "formId", "version", "createdAt", "publishedAt" from form_defs`);
    // eslint-disable-next-line no-console
    console.log(row);

    await exhaust(container);

    // this should NOT create an entity because the submission is for a form that had all
    // of its dataset information stripped away.
    const createEvent = await container.Audits.getLatestByAction('entity.create');
    // eslint-disable-next-line no-console
    console.log(createEvent);
  }));
```

matthew-white:
Closes #686.

I think this is pretty tricky to test in a way that's not brittle. I believe this issue only comes up if you upload a form def with a `meta/entity` field in an earlier version of Central, then upgrade to a version that supports datasets/entities. That's not something that we're really set up to test. The query that I'm modifying is only used in one place, and I think we can be confident that this join is helpful (or at least not harmful!). This area might get more complicated once a single form can contribute to multiple datasets, but we're not there yet.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -172,7 +172,7 @@ const getByName = (datasetName, projectId) => ({ all }) =>
 const getAllByProjectId = (projectId) => ({ all }) => all(_getAllByProjectId(projectId)).then(map(construct(Dataset)));
 
 // Returns list of Fields augmented by dataset property name OR entity creation information
-// including create, label, and id.
+// including label.
 const getFieldsByFormDefId = (formDefId) => ({ all }) => all(sql`
 SELECT
   form_fields.*, ds_properties."name" as "propertyName"
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -178,6 +178,8 @@ SELECT
   form_fields.*, ds_properties."name" as "propertyName"
 FROM
   form_fields
+JOIN dataset_form_defs ON
+  dataset_form_defs."formDefId" = form_fields."formDefId"
 LEFT OUTER JOIN ds_property_fields ON
   ds_property_fields."formDefId" = form_fields."formDefId"
     AND ds_property_fields."path" = form_fields."path"


ktuite (approved):

matthew-white:
Try running the test in the issue.

matthew-white:
I ran the test in the issue and added an extra assertion at the end:

```js
createEvent.isEmpty().should.be.true();
```

That assertion passes with this change and fails without it.

PR #696, opened 2022-11-21
fixes #679: added not null constraints on several columns

### LINKED ISSUE #679: Set some dataset/entity database fields to Not Null
Make a migration that adds all these suggested `.notNull()`s to relevant database fields. 

Here's [one](https://github.com/getodk/central-backend/pull/576#discussion_r1012196005), and there are more in the comments for this PR #576.

sadiqkhoja:
Closes #679

#### Changes:
- added not null constraints on several columns
- Updated unique constraint on ds_property_fields table, added a test to ensure that we don't allow multiple fields to be mapped to a single dataset property

All tests are passing

NA

NA

NA

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -897,12 +897,12 @@ describe('datasets and entities', () => {
       it('should not let multiple fields to be mapped to a single property', testService(async (service) => {
         await service.login('alice', (asAlice) =>
           asAlice.post('/v1/projects/1/forms')
-            .send(testData.forms.simpleEntity.replace(/first_name/g, 'name'))
+            .send(testData.forms.simpleEntity.replace(/first_name/g, 'age'))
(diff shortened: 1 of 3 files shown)


test/integration/api/datasets.js
```diff
@@ -848,6 +848,18 @@ describe('datasets and entities', () => {
                   body.entityRelated.should.equal(true);
                 }))));
       }));
+
+      it('should not let multiple fields to be mapped to a single property', testService(async (service) => {
```
sadiqkhoja, line 852: revisit this test

ktuite (approved):
Looks great! Saw the error come through to frontend like expected! 

Screen Shot 2022-11-21 at 11 13 20 AM

lognaturel:
Could we add an "a" so it's "to a single Dataset Property"? 🙏😊

Issue #678, opened 2022-11-14
Long tail of smaller changes in merged Entities PR


ktuite:
There are smaller changes that don't exactly fit into bigger issues that are suggested in the comments for PR #576.

Things like... 

1. Removing `readable`/`writable` from some places in the API where they don't make sense
    * (done in #700)
3. [Clunky code](https://github.com/getodk/central-backend/pull/576#discussion_r1012891214) in the entity CSV streaming that was copied from somewhere else but doesn't apply here 
    * (done in #769 PR that closes this issue)
5. Unnecessary lines of code
    * (probably addressed in #700, otherwise unclear what this refers to)
7. Duplicate tests 
    * (Not sure what this refers to anymore)
9. Etc.

sadiqkhoja:
Created sub-issue to address "before release" items #700. Keeping this issue open and moving it to next release

ktuite:
I've gone through all the comments in https://github.com/getodk/central-backend/pull/576 and it will be safe to close this issue with the linked PR (#769) because everything is either done or in its own issue at this point.

Issue #674, opened 2022-11-14
Remove "revisionNumber" from datasets


ktuite:
Comment from [Matt](https://github.com/getodk/central-backend/issues/649#issuecomment-1308296503):

> I think I'm seeing that simply creating a form draft has the effect of increasing the `revisionNumber` of a dataset. For example:
> 
> - Create a new form that creates entities for a new dataset named `people`. Publish the form. &hellip;/datasets returns `people` with a `revisionNumber` of 0.
> - Create a new draft for the form, but don't publish it. Now &hellip;/datasets returns `people` with a `revisionNumber` of 1.
> - Abandon the draft, then create a new draft. &hellip;/datasets returns `people` with a `revisionNumber` of 2.
> 
> I'd expect that creating but not publishing a form draft wouldn't change an existing dataset in a way that's visible over the API.
> 
> <hr>
> 
> Maybe we could remove `revisionNumber`? Maybe we'd want it in the future, but probably OK to remove for now? For now we could even just remove it from the API (remove `readable` from the field in the frame).

matthew-white:
We can't just remove this field entirely, because we use it here:

https://github.com/getodk/central-backend/blob/d19b72a8f4881218814144dcff229ca0045f2b54/lib/model/query/datasets.js#L147-L150

However, I suspect that #671 may update that code, in which case we may be able to remove the field.

sadiqkhoja:
I have removed `revisionNumber` from projects/:id/datasets API response in #697 . Will remove the column from the table during QA testing period or after release

PR #697, opened 2022-11-21
Improve code at several places for dataset/entities feature

sadiqkhoja:
Fixes several items of #678
and partial fix for #674

## Changes
- removed unnecessary readables writables
- removed duplicate tests
- renamed CTE expressions
- removed  from makeHierarchy
- changed {} with Object.create
- added deletedAt is null to getDatasetDiff sql
- projects/:id/datasets doesn't return revisionNumber

All tests are passing

NA

NA

`projects/:id/datasets` doesn't return `revisionNumber` anymore


ktuite (approved):
Looks good!

PR #704, opened 2022-11-28
fix: dataset-diff should return dataset even if there's no properties

sadiqkhoja:
Added condition in dataset-diff SQL query to return dataset even if there is no published properties. 
Fixes issue # 8 in QA's "ODK Central" doc

- Integration tests

- Dataset summary on Form Draft Status page and Publish modal should be tested as well

NA

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -833,6 +833,21 @@ describe('datasets and entities', () => {
 
       }));
 
+      it('should let the user download even if there are no properties', testService(async (service) => {
+        const asAlice = await service.login('alice', identity);
+
+        await asAlice.post('/v1/projects/1/forms?publish=true')
+          .send(testData.forms.simpleEntity.replace(/entities:saveto[^/]+/g, ''))
+          .set('Content-Type', 'application/xml')
+          .expect(200);
+
+        await asAlice.get('/v1/projects/1/datasets/people/entities.csv')
+          .expect(200)
+          .then(({ text }) => {
+            text.should.equal('name,label\n');
+          });
+      }));
+
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/integration/api/datasets.js
```diff
@@ -813,6 +813,26 @@ describe('datasets and entities', () => {
...
+          .expect(200)
+          .then(({ body }) => {
+            body.should.be.eql([{
+              name: 'people',
+              properties: []
+            }]);
+          });
+
+      }));
+
```
ktuite, line 835: Here's another test (that currently fails) for the other part of the issue (a published dataset with no published properties gives a 404 when downloading the CSV):

```js
      it('should let the user download even if if there are no properties', testService(async (service) => {
        const asAlice = await service.login('alice', identity);

          .send(testData.forms.simpleEntity.replace(/entities:saveto[^/]+/g, ''))
          .expect(200);

        await asAlice.get('/v1/projects/1/datasets/people/entities.csv')
          .expect(200)
          .then(({ text }) => {
            text.should.equal('name,label\n');
          });
      }));
```

It's calling `Datasets._getByNameSql`. I tried removing the line `AND ds_properties."publishedAt" IS NOT NULL` from that function but then it lets unpublished properties through. 

I think changing that line to something like this works (but neater):
```sql
AND ((ds_properties."publishedAt" IS NOT NULL AND ds_properties."id" IS NOT NULL) OR ds_properties."id" IS NULL)
```
where if the property exists, it must be published, otherwise it's okay to be null from the left join.

ktuite (commented):
Looks good for the dataset-diff part of the issue! I left a comment about the other part mentioned by QA.

ktuite (approved):

PR #711, opened 2022-12-05
Enhance: don't parse dataset schema of an encrypted form

### LINKED ISSUE #698: Revisit encrypted submissions and entities
This is a followup to issue #668. Currently, an encrypted submission from an entity-creating form can't be used to create an entity and when it tries to do that, it logs an entity create error saying "There was a problem with entity processing: Dataset empty or missing."

We will want to either come up with better messaging or figure out a way to possibly make entities from encrypted submissions.

sadiqkhoja:
## Changes:
* don't parse dataset schema of an encrypted form
* If we have keyId on a Form then we can skip dataset parsing logic
* For Form with keyId, we can return empty array for dataset-diff endpoints

Closes #698

- Integration tests and manual testing

- For now, I have short-circuited dataset parsing logic on form upload. Later we can parse complete form and return warnings, for that we have to validate form xml.
 
- Initially I was thinking that frontend needs to be changed as well for this but it turns out nothing is required there. I have manually tested and all looks fine.

NA

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -62,7 +62,7 @@ const createNew = (partial, project, publish = false) => ({ run, Datasets, FormA
   Promise.all([
     partial.aux.key.map(Keys.ensure).orElse(resolve(null)),
     getFormFields(partial.xml),
-    getDataset(partial.xml)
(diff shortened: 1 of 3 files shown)


ktuite (approved):
Reviewed async. Looks like a low risk change!

PR #716, opened 2022-12-08
Fix #459: reduce timestamptz precision

### LINKED ISSUE #459: OData filter: date/time fields more precise than millisecond
We are seeing cases where filtering OData does not exclude all the submissions it's expected to. @lognaturel observed while using the `gt` operator:

> If my last update timestamp is `2022-02-23T23:28:50.058Z`, it includes a submission made at timestamp `2022-02-23T23:28:50.058Z`. If I use `2022-02-23T23:28:50.059Z` in the query, that submission is excluded.

I think this is because Postgres timestamp columns are more precise than millisecond. This also came up in #358 (which led to b15086db66af98c39d7d9dd2a9a505c27760421c).

This comes up when fetching any submissions created since the latest known submission. One way to handle this is to ignore duplicate submissions. If that's not possible, 1ms could be added to the latest timestamp.

We could attempt to fix this in the code, probably by truncating any timestamp column. We could truncate a timestamp column in queries in which it is used to filter, though we would probably have to add an index for that. We could also consider truncating the timestamp when it is stored. It looks like the precision of Postgres timestamp columns can be configured (as part of the data type of the column: [docs](https://www.postgresql.org/docs/9.6/datatype-datetime.html)).

Alternatively, perhaps we don't change the code and simply document the issue in the API docs, along with the different strategies to handle it.

sadiqkhoja:
Closes #459 

#### Changes:
- Reduce precision of timestamptz columns to milliseconds across the database

- Added integration test

- see the discussion on the issue #459 

NA

NA 

### DIFF
--- a/lib/model/migrations/20221208-01-reduce-tz-precision.js
+++ b/lib/model/migrations/20221208-01-reduce-tz-precision.js
@@ -1,0 +1,46 @@
+// Copyright 2022 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+// Issue:        #cb459 - `gt` filter for submissionDate is not working as expected because of tz precision
+// Root cause:   Default timestamptz precision in postgres is microseconds and node/js has just milliseconds
+// Solution:     Let's change precision to milliseconds in database, since there is no value in having higher precision
+//               in the database when application can't use/handle it + typical usage of ODK Central doesn't demand higher
+//               precision.
+
+const getTimestampColumns = (db) => db.raw(`
+  SELECT
+    table_name, column_name, is_nullable
+  FROM
+    information_schema.columns
+  WHERE
+    table_schema = 'public'
+    AND udt_name = 'timestamptz'`)
+  .then(data => data.rows);
+
+const changePrecision = (db, columnMeta, precision) => db.schema.alterTable(columnMeta.table_name, (table) => {
+  if (columnMeta.is_nullable === 'YES') {
+    table.dateTime(columnMeta.column_name, { useTz: true, precision }).alter();
+  } else {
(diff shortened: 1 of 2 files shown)


sadiqkhoja:
There is still a separate but related issue: Only using submissionDate > xxx filter criterion can skip submision(s)

**Scenario:**
Assume user has ETL proceduce where they are fetching new submissions periodically by using filter submissionDate greater than submissionDate of last known submission. Then consider following sequence of events: 
```
t0: lastTimestamp is t0
t1: new submission(1)
t1: get submissions with submissionDate > t0 -- submission(1) is returned and lastTimestamp is t1
t1: new submission(2)
t2: get submissions with submissionDate > t1 -- no submission is returned
```
This can always happen no matter what the database timestamp precision is.  Following test captures the scenario.

```
it.only('should not skip submission', testService(async (service, { run }) => {
  const asAlice = await service.login('alice', identity);
  
  let lastTimestamp = '2022-01-01T09:00:00.015Z';
  
  // ****************************************************************************************
  // Simulate concurrent requests
  // Assume following three requests are executed at 2022-01-01T09:00:00.020Z
  // But there is an order at higher precision i.e. microsecond or picosecond
  // ****************************************************************************************
  {
    await asAlice.post('/v1/projects/1/forms/withrepeat/submissions')
    .send(testData.instances.withrepeat.one)
    .expect(200)
    .then(() => run(sql`update submissions set "createdAt"='2022-01-01T09:00:00.020Z'`));
  
    lastTimestamp = await asAlice.get(`/v1/projects/1/forms/withrepeat.svc/Submissions?$filter=__system/submissionDate gt ${lastTimestamp}`)
      .expect(200)
      .then(({ body }) => {
        body.value.length.should.be.eql(1);
        body.value[0].__id.should.be.eql('rone');
        return  body.value[0].__system.submissionDate;
      });
  
    await asAlice.post('/v1/projects/1/forms/withrepeat/submissions')
      .send(testData.instances.withrepeat.two)
      .expect(200)
      .then(() => run(sql`update submissions set "createdAt"='2022-01-01T09:00:00.020Z'`));
  }
  // ****************************************************************************************
  
  
  await asAlice.get(`/v1/projects/1/forms/withrepeat.svc/Submissions?$filter=__system/submissionDate gt ${lastTimestamp}`)
    .expect(200)
    .then(({ body }) => {
      body.value.length.should.be.eql(1);
      body.value[0].__id.should.be.eql('rtwo');
    });
}));
```

**Possible solution:**  
Recommend users to use submissionDate > xxx and submissionDate < xxx in ETL and similar workflows

PR #717, opened 2022-12-12
Enhance: complex type / groups can be '$select'ed

sadiqkhoja:
## Change:

- OData `$select` now support selecting complex type / group fields

- Added integration tests
- Tested manually on Excel

NA

- All possible OData endpoints should be tested

NA

### DIFF
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -444,6 +444,7 @@ const singleRowToOData = (fields, row, domain, originalUrl, query) => {
 // Where each node has reference to its parent and children
 // Returns an object with key=path and value=fieldNode
 // so that we can quickly get the field node instead of traversing tree from top to bottom
+// Assumption: formFields are in order
 const getFieldTree = (formFields) => {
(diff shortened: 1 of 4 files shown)


ktuite (approved):
Interactive review with @sadiqkhoja - makes sense to me!

@lognaturel had some performance-related questions on slack about the need to get the field hierarchy with `getFieldTree`, including:

> Going a little further regarding my comment about the field hierarchy: I might naively have expected that the code that writes out the Odata document could be told to go recursive for groups instead of having to explicitly identify and select all children. Basically it seems like it should be able to do what it does when building a non-filtered result.

Maybe there are some optimizations to be made in the future but I'm not sure what they would be right now. There are thorough tests in here and I think it'll solve the overall issue for now.

lognaturel:
> Maybe there are some optimizations to be made in the future

Sounds great to me! We know we want to do a systematic perf review either way.

ktuite (approved):

Issue #607, opened 2022-09-14
Entity type of repeats with hyphens cannot be resolved from metadata in OData feed


yanokwa:
If you have a repeat that has a hyphen in the name, you get this error in Power BI. 
> Cannot get feed for 'Submissions.r-1' as we could not resolve its entity type from metadata

<img width="875" alt="Screen Shot 2022-09-14 at 16 27 30" src="https://user-images.githubusercontent.com/32369/190280212-28c2780e-329b-4e48-b108-9dccbce477c6.png">

**Service document**
https://test.getodk.cloud/v1/projects/102/forms/bug.svc with the header (Accept:application/json)
```json
{
   "@odata.context":"https://test.getodk.cloud/v1/projects/102/forms/bug.svc//$metadata",
   "value":[
      {
         "name":"Submissions",
         "kind":"EntitySet",
         "url":"Submissions"
      },
      {
         "name":"Submissions.r-1",
         "kind":"EntitySet",
         "url":"Submissions.r-1"
      }
   ]
}
```

**Metadata document**
https://test.getodk.cloud/v1/projects/102/forms/bug.svc/$metadata
```xml
<EntityType Name="Submissions.r_1">
	<Key>
		<PropertyRef Name="__id"/>
	</Key>
	<Property Name="__id" Type="Edm.String"/>
	<Property Name="__Submissions-id" Type="Edm.String"/>
	<Property Name="t_3" Type="Edm.String"/>
</EntityType>
```

**Parent data document**
https://test.getodk.cloud/v1/projects/102/forms/bug.svc/Submissions with the header (Accept:application/json)
```json
{
   "value":[
      {
         "__id":"uuid:c2858e84-349b-420d-a25e-1e6978339759",
         "t_1":"A",
         "t_2":"B",
         "meta":{
            "instanceID":"uuid:c2858e84-349b-420d-a25e-1e6978339759"
         },
         "__system":{
            "submissionDate":"2022-09-14T23:26:18.944Z",
            "updatedAt":null,
            "submitterId":"15415",
            "submitterName":"yanokwa@getodk.org",
            "attachmentsPresent":0,
            "attachmentsExpected":0,
            "status":null,
            "reviewState":null,
            "deviceId":null,
            "edits":0,
            "formVersion":""
         },
         "r_1@odata.navigationLink":"Submissions('uuid%3Ac2858e84-349b-420d-a25e-1e6978339759')/r-1"
      }
   ],
[...]

PR #720, opened 2022-12-14
Support repeat names that are not valid OData identifiers

### LINKED ISSUE #469: Add ability to preview form definitions and edit submissions for forms that are closed or closing

### LINKED ISSUE #496: OData: Can't request table if repeat group name isn't valid OData name
ODK field names might not be valid OData names, so we sanitize field names for OData. However, I think I'm seeing that if the name of a repeat group isn't a valid OData name, it's not possible to request the table for the repeat group. For example, I've uploaded a form ([`odata_sanitize_repeat_name`](https://test.getodk.cloud/#/projects/388/forms/odata_sanitize_repeat_name)) with a repeat group named `q1.8`. If I request either https://test.getodk.cloud/v1/projects/388/forms/odata_sanitize_repeat_name.svc/Submissions.q1_8 or https://test.getodk.cloud/v1/projects/388/forms/odata_sanitize_repeat_name.svc/Submissions.q1.8, a 404.1 Problem is returned. I think that's because `verifyTablePath()` verifies against the original field name, not the sanitized OData name:

https://github.com/getodk/central-backend/blob/bdc73e91be2529242b5518b5c16c5c559ffefec1/lib/formats/odata.js#L336
https://github.com/getodk/central-backend/blob/bdc73e91be2529242b5518b5c16c5c559ffefec1/lib/formats/odata.js#L79-L90

Relatedly, if I request https://test.getodk.cloud/v1/projects/388/forms/odata_sanitize_repeat_name.svc/Submissions, I see:

```json5
{
  "value": [
    {
      "__id": "uuid:bf17f620-c56a-4533-8f89-da3f5c9bf37a",
      "meta": {
        "instanceID": "uuid:bf17f620-c56a-4533-8f89-da3f5c9bf37a"
      },
      "__system": {
        // ...
      },
      "q1_8@odata.navigationLink": "Submissions('uuid%3Abf17f620-c56a-4533-8f89-da3f5c9bf37a')/q1.8"
    }
  ],
  "@odata.context": "https://test.getodk.cloud/v1/projects/388/forms/odata_sanitize_repeat_name.svc/$metadata#Submissions"
}
```

Here `q1_8@odata.navigationLink` ends with `/q1.8`, but I'd expect the sanitized `/q1_8`.

I think the same issues come up if the field name of the repeat group is a valid OData name, but the field name of a group in which the repeat group is nested isn't a valid OData name.

sadiqkhoja:
Closes #496 and #607

- Integration Test
- Manually tested in Excel and PowerBI

### DIFF
--- a/lib/data/odata.js
+++ b/lib/data/odata.js
@@ -37,7 +37,7 @@ const { sanitizeOdataIdentifier } = require('../util/util');
 // compares fieldStack to a target tablename and returns whether we are:
 // out of (-1), at (0), or in (1) our target branch.
 const getBranchState = ({ fieldStack }, table) => {
-  const impliedTableName = [ 'Submissions' ].concat(fieldStack.map((field) => field.name)).join('.');
+  const impliedTableName = [ 'Submissions' ].concat(fieldStack.map((field) => sanitizeOdataIdentifier(field.name))).join('.');
   if (impliedTableName === table) return 0;
   if (impliedTableName.startsWith(table)) return 1;
   return -1;
@@ -95,9 +95,9 @@ const navigationLink = (schemaStack, instanceId, slicer = identity) => {
     // don't output an ID for the very last repeat, since we want the whole table:
     if ((field.type === 'repeat') && (i < fieldStack.length - 1)) {
       const id = hashId(schemaStack, instanceId, (stack) => stack.slice(0, i + 1));
-      result.push(`${field.name}('${id}')`);
+      result.push(`${sanitizeOdataIdentifier(field.name)}('${id}')`);
     } else {
-      result.push(field.name);
+      result.push(sanitizeOdataIdentifier(field.name));
     }
   }
   return result;
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -120,7 +120,7 @@ const serviceDocumentFor = (fields, domain, serviceUrl) => {
   const tables = [];
   for (const field of fields)
     if (field.type === 'repeat')
-      tables.push(field.path.slice(1).replace(/\//g, '.'));
+      tables.push(field.path.slice(1).split('/').map(sanitizeOdataIdentifier).join('.'));
 
   return {
     '@odata.context': `${domain}${urlPathname(serviceUrl)}/$metadata`,
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -86,7 +86,7 @@ const verifyTablePath = (tableParts, fields) => {
   if (tableParts[0] !== 'Submissions') return false;
   if (tableParts.length === 1) return true;
 
-  const stack = new SchemaStack(fields);
+  const stack = new SchemaStack(fields.map(f => ({ ...f, path: f.path.split('/').map(sanitizeOdataIdentifier).join('/') })));
   stack.push();
   for (let idx = 1; idx < tableParts.length; idx += 1)
     if (stack.push(tableParts[idx]) == null)
--- a/test/integration/api/odata.js
+++ b/test/integration/api/odata.js
@@ -1381,6 +1381,100 @@ describe('api: /forms/:id.svc', () => {
 
   });
 
+    // bug cb#496 and cb#607
+    it('should return results even when repeat name is not a valid OData name ', testService(async (service) => {
+      const asAlice = await service.login('alice', identity);
+
+      await asAlice.post('/v1/projects/1/forms?publish=true')
+        .send(`<?xml version="1.0"?>
+        <h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+            <h:head>
+                <h:title>odata_sanitize_repeat_name</h:title>
+                <model odk:xforms-version="1.0.0">
+                    <instance>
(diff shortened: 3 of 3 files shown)


lognaturel:
I think we should consider this as a candidate for a point release. Not significant enough to trigger one, I think, but if we have another reason to release it would be good for this to be in it.

ktuite (approved):

PR #721, opened 2022-12-16
fix #459: use for-loop instead of Promise.all in migration

### LINKED ISSUE #459: OData filter: date/time fields more precise than millisecond

sadiqkhoja:
Issue: `20221208-01-reduce-tz-precision.js` is timing out if there is too much data in DB.
Change: Replaced `Promise.all` with for-loop in migration file

### DIFF
--- a/lib/model/migrations/20221208-01-reduce-tz-precision.js
+++ b/lib/model/migrations/20221208-01-reduce-tz-precision.js
@@ -7,13 +7,15 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
+const { groupBy } = require('ramda');
+
 // Issue:        #cb459 - `gt` filter for submissionDate is not working as expected because of tz precision
 // Root cause:   Default timestamptz precision in postgres is microseconds and node/js has just milliseconds
 // Solution:     Let's change precision to milliseconds in database, since there is no value in having higher precision
 //               in the database when application can't use/handle it + typical usage of ODK Central doesn't demand higher
 //               precision.
 
-const getTimestampColumns = (db) => db.raw(`
+const getTablesWithTimestampColumns = (db) => db.raw(`
   SELECT
     table_name, column_name, is_nullable
   FROM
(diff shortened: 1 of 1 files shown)


matthew-white:
@alxndrsn, do you have time to take a look at this or try it out locally? It'd be great to resolve this before we deploy to the QA server.

matthew-white:
I'm going to be updating the QA server tomorrow, so let's try to get this merged within the next 24 hours or so so that we can exercise the migration that we intend to deploy. @alxndrsn, the previous version of this migration timed out for you, right? Have you been able to try out this new version? Or @sadiqkhoja, have you been able to validate this new version of the migration another way?

lib/model/migrations/20221208-01-reduce-tz-precision.js
```diff
@@ -7,40 +7,45 @@
...
 // Issue:        #cb459 - `gt` filter for submissionDate is not working as expected because of tz precision
 // Root cause:   Default timestamptz precision in postgres is microseconds and node/js has just milliseconds
 // Solution:     Let's change precision to milliseconds in database, since there is no value in having higher precision
 //               in the database when application can't use/handle it + typical usage of ODK Central doesn't demand higher
 //               precision.
 
-const getTimestampColumns = (db) => db.raw(`
+const getTablesWithTimestampColumns = (db) => db.raw(`
   SELECT
     table_name, column_name, is_nullable
```
alxndrsn, line 20: is `is_nullable` still used?

lib/model/migrations/20221208-01-reduce-tz-precision.js
```diff
@@ -7,40 +7,45 @@
...
   FROM
     information_schema.columns
   WHERE
     table_schema = 'public'
     AND udt_name = 'timestamptz'`)
-  .then(data => data.rows);
+  .then(data => groupBy(r => r.table_name, data.rows));
```
alxndrsn, line 26: Is `groupBy()` equivalent to `GROUP BY`?

e.g.

```sql
SELECT table_name, JSON_AGG(column_name) 
  FROM information_schema.columns
  WHERE table_schema = 'public'
    AND udt_name = 'timestamptz'
  GROUP BY table_name
```

alxndrsn (approved):
The code looks fine to me, but I no longer have a database in a known-broken state to test this against.

PR #723, opened 2022-12-23
Add request duration in Sentry logs

sadiqkhoja:
# Change:
- Added request duration in the event object of Sentry

- Tested manually 

- I have added `x-request-received-at` in the request header, I could have created a different property in the request object but that would have forced me to modify RequestHandler of Sentry. Implemented solution is the simplest one.

- NA

- NA

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -92,9 +92,7 @@ const sanitizeEventRequest = (event) => {
       event.request.url = filterTokenFromUrl(event.request.url);
     }
 
-    event.extra = {
-      duration: `${Math.round((Date.now() - event.request.headers['x-request-received-at']) / 1000)} sec`
-    };
+    event.extra.duration = Date.now() - event.extra.startTimestamp;
   }
 
   /* eslint-enable no-param-reassign */
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { inspect } = require('util');
-const { isBlank, isPresent } = require('../util/util');
+const { isBlank, isPresent, noop } = require('../util/util');
 
 
 // Endpoints where query strings are sensitive and should be removed
@@ -92,7 +92,6 @@ const sanitizeEventRequest = (event) => {
       event.request.url = filterTokenFromUrl(event.request.url);
     }
 
-    event.extra.duration = Date.now() - event.extra.startTimestamp;
   }
 
   /* eslint-enable no-param-reassign */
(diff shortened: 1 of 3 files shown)


lib/http/service.js
```diff
@@ -18,6 +18,11 @@ module.exports = (container) => {
   ////////////////////////////////////////////////////////////////////////////////
   // PRERESOURCE MIDDLEWARE
 
+  service.use((req, res, next) => {
+    req.headers['x-request-received-at'] = Date.now();
```
alxndrsn, line 22: Is storing this as a request header idiomatic?  I'd expect request headers to have come from the client.

alxndrsn:
It's surprising this isn't supported out of the box.  The Sentry docs make references to `event.time_spent`, but I can't find references to it in the JS libs; is this something you came across @sadiqkhoja?

lib/external/sentry.js
```diff
@@ -91,6 +91,10 @@ const sanitizeEventRequest = (event) => {
       // filter out access tokens embedded in URLs
       event.request.url = filterTokenFromUrl(event.request.url);
     }
+
+    event.extra = {
+      duration: `${Math.round((Date.now() - event.request.headers['x-request-received-at']) / 1000)} sec`
```
alxndrsn, line 96: Is rounding this useful?  I'd expect milliseconds to be a more useful resolution here.  Hopefully most requests would `round()` to 0 seconds :grin:

sadiqkhoja:
I don't see event.time_spent in javascript, here is the event object that I see while debugging:



Even in the payload document I don't see it: https://develop.sentry.dev/sdk/event-payloads/

There is `start_timestamp` but it is part of Sentry's performance monitoring which we are not using: https://develop.sentry.dev/sdk/event-payloads/span/

matthew-white:
Also wanted to mention that I like the approach of using `Sentry.configureScope()`!

lib/external/sentry.js
```diff
@@ -122,6 +124,10 @@ const init = (config) => {
     beforeSend(event, hint) {
       const sanitizedEvent = sanitizeEventRequest(event);
 
+      if (sanitizedEvent.extra && sanitizedEvent.extra.startTimestamp) { // just an extra safety check. this should be always true
```
matthew-white, line 127: I think this won't be true for an error thrown by a worker. So a good check I think!

matthew-white (approved):

matthew-white:
@sadiqkhoja, what do you think about verifying this change on the dev server? You would need to cause an error to be sent to Sentry, then check that the error has the expected information in Sentry. The way I've done that in the past is by adding a new endpoint that just throws an error, then sending a request to it.

sadiqkhoja:
I had verified it by sending logs from local environment to Sentry.

Just checked for test environment, I can see duration and startTimestamp in Sentry as well:

image


PS: easier way to produce error is to cancel zip download of large submission data, we get `Premature close`. I used this form: https://test.getodk.cloud/#/projects/220/forms/repeat-export-stress-2/submissions

matthew-white:
> easier way to produce error is to cancel zip download of large submission data, we get `Premature close`.

Ah nice idea!

PR #726, opened 2023-01-06
Added indices for entities related tables

### LINKED ISSUE #660: Add indexes to entities and entity_defs tables
For after v2022.3:

Right now the `entities` and `entity_defs` tables don't have indexes. However, we know that those tables could become large, so it'd be good to add some.

sadiqkhoja:
Closes #660

- Executed migration locally

- We can potentially add indices for dataset tables, would that be overkill?
 
NA

NA


lib/model/migrations/20230104-01-add-entity-indices.js
```diff
@@ -0,0 +1,34 @@
...
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.schema.table('entities', (en) => {
+    en.index('datasetId');
+  });
```
matthew-white, line 13: For the entities OData feed / the entities table in Frontend, are we going to sort by `"createdAt" desc, id desc` like we do for submissions? If so, then maybe we should also add an index involving `createdAt`? It looks like we did that for the `submissions` table:

https://github.com/getodk/central-backend/blob/9d767ff84ffc8fef1fa36fd97a723139091bb72a/lib/model/migrations/20181012-01-add-submissions-createdat-index.js#L10-L12

(To be clear, this would be in addition to the index you've added here.)

Definitely feel free to skip for now and circle back to this if/when it's clear whether we'll need an index on `createdAt`.

lib/model/migrations/20230104-01-add-entity-indices.js
```diff
@@ -0,0 +1,34 @@
...
+
+  await db.schema.table('entity_defs', (ed) => {
+    ed.index('entityId');
+    ed.index('submissionDefId');
+  });
```
matthew-white, line 18: I imagine that at some point, we'll want some index involving `current`. Here's what I think we do for `submission_defs`:

https://github.com/getodk/central-backend/blob/9d767ff84ffc8fef1fa36fd97a723139091bb72a/lib/model/migrations/20210118-01-add-current-flag-to-submission-defs.js#L11-L15

That's also something we could circle back to once we support entity versioning/editing.

matthew-white (approved):
Looks good! ✅

> We can potentially add indices for dataset tables, would that be overkill?

It looks like we already have an index on `(name, "projectId")` from the uniqueness constraint, and that seems like the main one to me. I took a quick look at the columns of the `datasets` table, and I didn't see anything else for which we clearly need an index.

lib/model/migrations/20230104-01-add-entity-indices.js
sadiqkhoja: I have updated the index to include `createdAt` and `id`, one index is sufficient to handle all of our use cases.

lib/model/migrations/20230104-01-add-entity-indices.js
sadiqkhoja: updated the index to include `current`

lib/model/migrations/20230104-01-add-entity-indices.js
```diff
@@ -0,0 +1,34 @@
+// Copyright 2023 ODK Central Developers
```
ktuite, line 1: I think the file should be named to have a date that comes after the latest existing migration otherwise knex might get confused (mainly for us devs updating between different commits in master, not necessarily someone doing a migration that includes all the 2023.1 files).

ktuite: Tried it, knex doesn't actually complain, but it will generally try to apply the migrations in the filename order and since it's getting committed after `20230106-01-remove-revision-number.js` and `20230109-01-add-form-schema.js` it would be safer to have it come after.

ktuite (approved):
Aside from my comment on the filename, it looks good!

PR #730, opened 2023-01-17
Enhance: See more context on Datasets list

sadiqkhoja:
Added `extended` option for datasets API that returns entities count and timestamp of latest entity

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -95,6 +95,176 @@ describe('datasets and entities', () => {
                   }))))));
     });
 
+    describe('projects/:id/datasets GET extended', () => {
+
+      it('should return the 0 for entities', testService(async (service) => {
+        const asAlice = await service.login('alice', identity);
+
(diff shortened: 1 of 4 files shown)


test/integration/api/datasets.js
```diff
@@ -41,6 +41,39 @@ describe('datasets and entities', () => {
...
+        await asAlice.get('/v1/projects/1/datasets')
+          .set('X-Extended-Metadata', 'true')
+          .expect(200)
+          .then(({ body }) => {
+            body.map(({ createdAt, lastEntity, ...d }) => {
+              createdAt.should.not.be.null();
+              lastEntity.should.not.be.null();
+              return d;
+            }).should.eql([
+              { name: 'people', projectId: 1, entities: 1 }
```
ktuite, line 72: A couple other test cases that might be nice to check are:
- no entities in a given dataset
- counting more than 1 entity in a given dataset
- multiple datasets in the same project that are each getting counted independently

ktuite (approved):

PR #731, opened 2023-01-17
new: /datasets/:name API

sadiqkhoja:
Added new API `datasets/:name` that returns metadata of a dataset i.e. properties, forms that created properties and linked forms that consume the dataset.


ktuite (approved):
Looks good. I added an issue (https://github.com/getodk/central-backend/issues/733) for tracking API doc changes including this new endpoint.

PR #737, opened 2023-01-21
Create audit log entry when password is invalidated

### LINKED ISSUE #736: Create audit log entry when password is invalidated
/v1/users/reset/initiate initiates a password reset. If ?invalidate=true is specified, then the existing password is also immediately invalidated. The results of /v1/users/reset/initiate don't appear in the audit log. I think it makes sense that initiating a password reset isn't logged. However, if a password is invalidated, I think it'd be a good idea to log that, since that results in an immediate change to an existing resource. Only sitewide administrators can invalidate another user's password.

In terms of which action is logged, I think we could use `user.update`. Changing a password also logs `user.update`.

matthew-white:
Closes #736.

New integration test.

The biggest question to me was which action to use. I used `user.update` since changing a password also logs that action. This is how we describe `user.update` in the API docs:

> * `user.update` when User information is updated, like email or password.

I think that fits its new use in this PR.

This is a targeted change, so regression risks should be low.

I think the current description of `user.update` in the API docs already fits its use in this PR.

### DIFF
--- a/lib/model/query/users.js
+++ b/lib/model/query/users.js
@@ -43,6 +43,9 @@ const invalidatePassword = (user) => ({ Sessions, run }) => Promise.all([
   run(sql`update users set password=null where "actorId"=${user.actor.id}`),
   Sessions.terminateByActorId(user.actorId)
 ]);
+invalidatePassword.audit = (user) => (log) => log('user.update', user.actor, {
+  data: { password: null }
+});
 
 const provisionPasswordResetToken = (user) => ({ Actors, Assignments, Sessions }) => {
   const expiresAt = new Date();
--- a/test/integration/api/users.js
+++ b/test/integration/api/users.js
@@ -300,6 +300,21 @@ describe('api: /users', () => {
         .expect(401)));
     }));
 
+    it('should log action in audit log if password is invalidated', testService(async (service) => {
+      const asAlice = await service.login('alice');
+      await asAlice.post('/v1/users/reset/initiate?invalidate=true')
+        .send({ email: 'bob@getodk.org' })
+        .expect(200);
+      const { body: audits } = await asAlice.get('/v1/audits?action=user.update')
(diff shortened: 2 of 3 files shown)


lib/model/query/users.js
```diff
@@ -43,6 +43,9 @@ const invalidatePassword = (user) => ({ Sessions, run }) => Promise.all([
   run(sql`update users set password=null where "actorId"=${user.actor.id}`),
   Sessions.terminateByActorId(user.actorId)
 ]);
+invalidatePassword.audit = (user) => (log) => log('user.update', user.actor, {
+  data: { password: null }
+});
```
matthew-white, line 48: This is similar to `updatePassword.audit` just a few lines above. You can also see in `update.audit` above that when a user's basic details are changed, `user.update` logs the user's new data: the `data` property of the audit log details is the user's new data. `updatePassword.audit` does something similar, logging `{ data: { password: true } }`. It doesn't log the new password of course, but rather just `true` to mean "changed to something". I thought about logging `{ data: { password: false } }` here. However, I think `{ password: null }` makes more sense, as the `password` field actually is changed to `null`, and it's OK to log that. And if `{ password: true }` means "changed", I think `{ password: false }` would mean "not changed".

test/integration/api/users.js
```diff
@@ -300,6 +300,21 @@ describe('api: /users', () => {
         .expect(401)));
     }));
 
+    it('should log action in audit log if password is invalidated', testService(async (service) => {
+      const asAlice = await service.login('alice');
```
matthew-white, line 304: This is taking advantage of the new usage of `service.login()` that doesn't require a callback. Since I'm passing a single string to `service.login()`, not an array, I wanted `service.login()` to return a single authed service, not an array of a single authed service. I made a small change to test/integration/setup.js to facilitate that.

alxndrsn (approved):

PR #738, opened 2023-01-22
Update analytics for v2023.1

matthew-white:
See getodk/central#350. The only change to the metrics is removing the `backups_configured` metric. I'm removing that metric because we're removing Google Drive backups (getodk/central#323), which means that backups will never be configured. (Direct backups will continue to work, but those aren't configured.)

Tests. During regression testing, the QA team will verify that analytics are shown in Frontend and sent to data.getodk.cloud. I don't think that the team usually verifies specific changes to metrics, but they probably could do so here. If that sounds like it'd be useful, I could mark getodk/central#350 as "needs testing".

I think this is the first metric that we're removing, so I'm not totally sure what the best approach is. I think it makes sense to remove the `backups_configured` field entirely, since we will no longer collect data for it. I don't expect that the field will continue to be useful going forward. Another approach would be to keep the field, but send an empty string for it. However, I think it's cleaner to remove the field, and as we continue to add and remove metrics, I think we want the ability to decrease the size of the form and reduce noise. We will continue to be able to retrieve past submissions' values for this field if we need to. If we end up deciding we want the field back for any reason, we can reverse its removal in a new form version.

We should verify that analytics as a whole continue working. Users shouldn't be relying on the existence of a particular metric in analytics, so I don't think there's much risk of regression in removing this metric.

The specific metrics we send for analytics aren't part of the stable API and aren't documented. There was an example response that included `backups_configured`, so I removed that there.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -34,7 +34,7 @@
       "analytics": {
         "url": "https://data.getodk.cloud/v1/key/eOZ7S4bzyUW!g1PF6dIXsnSqktRuewzLTpmc6ipBtRq$LDfIMTUKswCexvE0UwJ9/projects/1/submission",
         "formId": "odk-analytics",
-        "version": "2022.11.29.01"
+        "version": "2023.01.22.01"
       }
     }
   },
--- a/docs/api.md
+++ b/docs/api.md
@@ -3924,7 +3924,6 @@ An Administrator can use this endpoint to preview the metrics being sent. The pr
                   "recent":1,
                   "total":1
                 },
-                "backups_configured":1,
                 "database_size":12345,
                 ...
               },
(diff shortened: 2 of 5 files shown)


config/default.json
```diff
@@ -34,7 +34,7 @@
       "analytics": {
         "url": "https://data.getodk.cloud/v1/key/eOZ7S4bzyUW!g1PF6dIXsnSqktRuewzLTpmc6ipBtRq$LDfIMTUKswCexvE0UwJ9/projects/1/submission",
         "formId": "odk-analytics",
-        "version": "2022.11.29.01"
+        "version": "2023.01.22.01"
```
matthew-white, line 37: I haven't created this form version yet, but I will! First I wanted to validate my thinking that the field should be removed.

ktuite (approved):
Looks good! I've come around to thinking that deleting the field is a fine idea. 

I *believe* this will work right away on the frontend - since backups are removed from the dict that frontend uses to render the preview table, it will just skip that row. But there will be an 'extra' unused translation floating around until it gets removed.

PR #740, opened 2023-01-23
new: OData endpoint for entities

sadiqkhoja:


test/integration/api/odata-entities.js
```diff
@@ -0,0 +1,107 @@
...
+          .send({ reviewState: 'approved' })
+          .expect(200);
+
+      }
+      await exhaust(container);
+    };
+    /* eslint-enable no-await-in-loop*/
+
+    it('should return all entities', testService(async (service, container) => {
+      const asAlice = await service.login('alice', identity);
```
matthew-white, line 39: I just merged a PR that makes the test callback optional. So you can now remove `identity` here (if you rebase)!

lib/data/entity.js
```diff
@@ -124,5 +127,56 @@ const streamEntityCsvs = (inStream, properties) => {
...
+        if (isFirstEntity) { // header or fencepost.
+          this.push('{"value":[');
+          isFirstEntity = false;
+        } else {
+          this.push(',');
+        }
+
+        this.push(JSON.stringify(mergeLeft({
+          name: entity.uuid,
+          createdAt: entity.createdAt,
```
matthew-white, line 159: `createdAt` isn't a reserved property name, is it? Should we put it in a `__system` `ComplexType` similar to what we do with submissions? I imagine that we'll continue accumulating metadata/summary properties like we have with submissions, so a `__system` object might be a good pattern to put in place now.

lib/data/entity.js
```diff
@@ -124,5 +127,56 @@ const streamEntityCsvs = (inStream, properties) => {
...
+  const { limit, offset, shouldCount } = extractPaging(query);
+
+  const schema = {
+    name: null,
+    label: null
+  };
+
+  for (const property of properties) {
+    schema[property.name] = null;
+  }
```
matthew-white, line 142: I don't think it's a huge deal here, but since the keys here are user-defined, I think it'd be better to create the object with `Object.create(null)`.

matthew-white (commented):
I was just glancing over what's in the PR and thought I'd leave a couple of one-off comments.

test/integration/api/odata-entities.js
```diff
@@ -0,0 +1,248 @@
...
+        .set('Content-Type', 'application/xml')
+        .send(testData.forms.simpleEntity)
+        .expect(200);
+
+      await createSubmissions(asAlice, container, 2);
+
+      await asAlice.get('/v1/projects/1/datasets/people.svc/Entities')
+        .expect(200)
+        .then(({ body }) => {
+          body.value.length.should.be.eql(2);
```
ktuite, line 50: I'm not seeing the entity data when I use this endpoint. The values of the properties are blank strings.
Can there be a test that checks that the properties (like first name and age, in this case) are present?

lib/resources/odata-entities.js
```diff
@@ -0,0 +1,54 @@
...
+    const dataset = await Datasets.getByName(params.name, params.projectId).then(getOrNotFound);
+    return xml(edmxForEntities(params.name, dataset.properties));
+  }));
+
+  // serves the odata feed for the entities
+  service.get('/projects/:projectId/datasets/:name.svc/Entities', endpoint.odata.json(async ({ Datasets, Entities, Projects, env }, { auth, params, originalUrl, query }) => {
+    const project = await Projects.getById(params.projectId).then(getOrNotFound);
+    await auth.canOrReject('entity.list', project);
+    const dataset = await Datasets.getByName(params.name, params.projectId).then(getOrNotFound);
+    const count = await Entities.countByDatasetId(dataset.id);
```
sadiqkhoja, line 46: pass filter criteria here as well

lib/resources/odata-entities.js
```diff
@@ -0,0 +1,54 @@
...
+    const options = QueryOptions.fromODataRequestEntities(query);
+    return Entities.streamForExport(dataset.id, options)
```
sadiqkhoja, line 48: split this line into two using await

lib/model/query/entities.js
```diff
@@ -131,19 +135,29 @@ const getByUuid = (uuid) => ({ maybeOne }) =>
...
   entities."datasetId" = ${datasetId}
-  and entity_defs.current=true
-order by entities.id`)
+  AND entity_defs.current=true
+  AND  ${odataFilter(options.filter, odataToColumnMap)}
+ORDER BY entities.id
+${page(options)}`)
     .then(stream.map(_exportUnjoiner));
 
+const countByDatasetId = (datasetId) => ({ oneFirst }) => oneFirst(sql`
```
sadiqkhoja, line 154: use the filter criteria

lib/util/odata.js
```diff
@@ -0,0 +1,71 @@
+// Copyright 2017 ODK Central Developers
```
sadiqkhoja, line 1: 2023

test/integration/api/odata-entities.js
```diff
@@ -0,0 +1,255 @@
...
+          body.value.forEach(r => {
+            r.first_name.should.be.eql('Alice');
+            r.age.should.be.eql('88');
+          });
+
+          body.value[0].name.should.not.be.eql(body.value[1].name);
+        });
+    }));
+
+    it('should return count of entities', testService(async (service, container) => {
```
sadiqkhoja, line 61: add a test where count is correct for the $filter request

test/integration/api/odata-entities.js
```diff
@@ -0,0 +1,255 @@
...
+      await createSubmissions(asAlice, container, 2);
+
+      await asAlice.get('/v1/projects/1/datasets/people.svc/Entities?$top=1')
+        .expect(200)
+        .then(({ body }) => {
+          body['@odata.nextLink'].should.be.equal('http://localhost:8989/v1/projects/1/datasets/people.svc/Entities?%24skip=1');
+        });
+    }));
+
+    it('should return filtered entities', testService(async (service, container) => {
```
sadiqkhoja, line 112: add a test with invalid filter column

Issue #692, opened 2022-11-18
Retry worker in case of transient failure


sadiqkhoja:
Related to https://github.com/getodk/central-backend/pull/685

Entity worker is currently catching all the errors and logging them in audit table. I think in case of transient errors (common in cloud environments) we should try to reprocess the event.

I have written a test to emulate database failure, see https://github.com/sadiqkhoja/central-backend/commit/36cd848d507276eb8059de5c3e45e8d61ac24b48 where Entities.createNew throws `Database is down` error and we are marking original event as processed and logging error details in audit table.

matthew-white:
I think @ktuite convinced me that the error would have to be very transient for this to be an issue. If the database is down, or the connection pool is full, or something similar, then it seems very unlikely that `Entities.createNew()` would fail, yet the `Audits.log()` logging `entity.create.error` would be successful. I think it will be much more common for either both to succeed or neither to succeed. (And if neither succeeds, then the worker will try again.) A very transient error is a possibility, but I'm not sure it's worth us circling back to the worker. What do you think? Do you think we should try to be resilient to that possibility? Is the possibility more likely than I'm imagining?

sadiqkhoja:
I agree that it is highly likely that Entities.createNew() and Audit.log() will succeed or fail together. However, if both fail due to transient error then the event will be reprocessed after 2 hours, which might cause out of order event processing.

One safety net that we have against transient errors is the default retry mechanism of Slonik, if I understand correctly, it retries 5 times before throwing the error.

Having said that I don't think this issue should be high risk item, we can observe how things play out during QA testing. I just wanted to log this just in case we have to revisit it.

matthew-white:
@sadiqkhoja, do you think we could go ahead and close this issue? I don't think we've seen this be a problem in the last couple of years.

> if both fail due to transient error then the event will be reprocessed after 2 hours, which might cause out of order event processing.

These days, we have multiple workers working in parallel on events. That means that already, events aren't guaranteed to be processed in order. For `entity.create` events in particular, I don't think it matters what order the events are processed in. Order is potentially more of an issue for `entity.update` events, but it shouldn't be a problem going forward now that we support offline updates.

sadiqkhoja:
Sure, makes sense.

Issue #661, opened 2022-11-09
Inform user about failure to process submission for entities


matthew-white:
For after v2022.3:

When a submission is approved, the worker job `createEntityFromSubmission` calls `Entities.processSubmissionDef()`, which checks whether the submission creates a new entity and tries to create the entity if so. `Entities.processSubmissionDef()` can return a rejected promise for a variety reasons. If there's a user error (for example, invalid XML), then `Entities.processSubmissionDef()` will generally reject with an `entityViolation` Problem. In that case, the job logs an `entity.create.error` audit log entry and marks the submission as having been processed. The `entity.create.error` will be returned to the user over the API and shown in Frontend. However, `Entities.processSubmissionDef()` can also be rejected for other reasons, for example, if there's an error connecting to the database. In that case, the job won't log an `entity.create.error` audit log entry and will just fail. That's intentional, because in that case, we want Backend to try the job again. However, if the job fails enough times, then Backend will stop trying. In that case, it's possible that an entity should have been created, but won't be — and there's no indication of that to the user. The user isn't able to detect that case in Frontend or over the API.

I think we should account for that case somehow:

- Maybe the job should check whether it's on its last try. In that case, if `Entities.processSubmissionDef()` is rejected, then the job should log an `entity.create.error` audit log entry no matter what the reason for the rejection was.
- Or maybe we should separately list to the user all `submission.update` audit log entries that haven't been successfully processed for entities (failed too many times, the job hasn't been tried yet, etc.).

ktuite:
We changed the entity processing mechanism in this PR(https://github.com/getodk/central-backend/pull/685) since this issue was created. The only other entity problem we saw after release was addressed in #742.

I'm going to close this issue now. When we have more entity-processing errors (maybe when we get to updates) we will have to revisit a lot of things about entity errors, events, and audit logs anyway.

PR #742, opened 2023-01-24
Check for existing entity across submission defs for same submission

### LINKED ISSUE #729: Don't attempt to re-create entity from edited submission
On the [forum](https://forum.getodk.org/t/entities-and-datasets-in-odk-central/40051/9), a user ran into an error where they created an entity from a submission, edited the submission hoping to be able to edit the entity, reapproved the submission, and then saw a "Problem creating Entity" error message with no explanation. See [frontend issue](https://github.com/getodk/central-frontend/issues/701) for a fix to make sure this error message is visible.

What is happening is backend is attempting to create an Entity from the re-approved submission, but is running to the entity UUID uniqueness constraint "A resource already exists with uuid value(s) of [some uuid]", and logging this as an `entity.create.error` audit event.

We haven't implemented any entity-updating logic yet, and when we do, we are likely to do it through a different entity-specific editing mechanism, not submission-editing. In the mean time, we shouldn't give the impression that editing a submission *might* work to update an entity. 

In this case where an entity has already been created for this submission, backend shouldn't even try to make another entity and it shouldn't log an error event. The unique UUID error message should only come up if a *separate* submission has a UUID collision with an existing entity. 

Example of this situation on the [QA server](https://test.getodk.cloud/#/projects/422/forms/trees_registration_demo/submissions/uuid%3A8fbd68c9-974a-44b4-8248-4145dd635c26). 
Screen Shot 2023-01-17 at 10 31 52 AM

ktuite:
Closes #729. This is to make the server not log an error when reapproving an edited submission that was already made into an entity.

Wrote a test to mimic scenario where this came up. It failed and then with the code change, it passes.

It's ok for now.

Not very risky.

No.

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -114,10 +114,6 @@ const processSubmissionEvent = (event) => (container) =>
 ////////////////////////////////////////////////////////////////////////////////
 // GETTING ENTITIES
 
-const getDefBySubmissionDefId = (submissionDefId) => ({ maybeOne }) =>
-  maybeOne(sql`select * from entity_defs where "submissionDefId" = ${submissionDefId} limit 1`)
-    .then(map(construct(Entity.Def)));
-
 // This will check for an entity related to any def of the same submission
 // as the one specified. Used when trying to reapprove an edited submission.
 const getDefBySubmissionId = (submissionId) => ({ maybeOne }) =>
@@ -150,4 +146,4 @@ order by entities.id`)
 
 
 module.exports = { createNew, _processSubmissionDef, processSubmissionEvent, streamForExport,
-  getDefBySubmissionDefId, getDefBySubmissionId, getByUuid };
+  getDefBySubmissionId, getByUuid };
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -63,8 +63,8 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
 
 
 // Entrypoint to where submissions (a specific version) become entities
-const _processSubmissionDef = (submissionDefId) => async ({ Datasets, Entities, Submissions }) => {
-  const existingEntity = await Entities.getDefBySubmission(submissionDefId);
+const _processSubmissionDef = (submissionDefId, submissionId) => async ({ Datasets, Entities, Submissions }) => {
+  const existingEntity = await Entities.getDefBySubmissionId(submissionId);
   // If the submission has already been used to make an entity, don't try again
   // and don't log it as an error.
   if (existingEntity.isDefined())
@@ -86,7 +86,7 @@ const _processSubmissionDef = (submissionDefId) => async ({ Datasets, Entities, 
 
 const processSubmissionEvent = (event) => (container) =>
   container.db.transaction((trxn) =>
-    container.with({ db: trxn }).Entities._processSubmissionDef(event.details.submissionDefId))
+    container.with({ db: trxn }).Entities._processSubmissionDef(event.details.submissionDefId, event.details.submissionId))
     .then((entity) => {
       if (entity != null) {
         return container.Audits.log({ id: event.actorId }, 'entity.create', { acteeId: entity.aux.dataset.acteeId },
@@ -120,12 +120,11 @@ const getDefBySubmissionDefId = (submissionDefId) => ({ maybeOne }) =>
 
 // This will check for an entity related to any def of the same submission
 // as the one specified. Used when trying to reapprove an edited submission.
-const getDefBySubmission = (submissionDefId) => ({ maybeOne }) =>
+const getDefBySubmissionId = (submissionId) => ({ maybeOne }) =>
   maybeOne(sql`select ed.* from submissions as s
   join submission_defs as sd on s."id" = sd."submissionId"
-  join submission_defs as sd2 on s."id" = sd2."submissionId"
-  join entity_defs as ed on ed."submissionDefId" = sd2."id"
-  where sd.id = ${submissionDefId} limit 1`)
+  join entity_defs as ed on ed."submissionDefId" = sd."id"
+  where s.id = ${submissionId} limit 1`)
     .then(map(construct(Entity.Def)));
 
 const getByUuid = (uuid) => ({ maybeOne }) =>
@@ -151,4 +150,4 @@ order by entities.id`)
 
 
 module.exports = { createNew, _processSubmissionDef, processSubmissionEvent, streamForExport,
-  getDefBySubmissionDefId, getDefBySubmission, getByUuid };
+  getDefBySubmissionDefId, getDefBySubmissionId, getByUuid };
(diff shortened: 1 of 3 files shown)


lib/model/query/entities.js
```diff
@@ -118,6 +118,16 @@ const getDefBySubmissionDefId = (submissionDefId) => ({ maybeOne }) =>
   maybeOne(sql`select * from entity_defs where "submissionDefId" = ${submissionDefId} limit 1`)
     .then(map(construct(Entity.Def)));
 
+// This will check for an entity related to any def of the same submission
+// as the one specified. Used when trying to reapprove an edited submission.
```
lognaturel, line 122: Is this necessary? It looks expensive, especially once we have long-lived entities with a rich history. Any chance we could instead do this when receiving the submission? If it's an edit, don't consider it for entity processing?

ktuite: Yeah, I didn't feel great about this query.

Looking at the event data, we have the submission ID, too, so I'm adding a commit that uses that instead. That's more of the intent, anyway, that we'll only allow one entity per submission.

matthew-white: > If it's an edit, don't consider it for entity processing?

If approval comes after the edit, then we should use the latest submission version as the basis of the entity, right?

> Any chance we could instead do this when receiving the submission?

The event that's being hooked into isn't the request, it's the `submission.update` audit log action. After each change to a submission's review state, a worker will check whether an entity needs to be created from the submission.

lib/model/query/entities.js
```diff
@@ -63,8 +63,8 @@ select ins.*, def.id as "entityDefId", ds."acteeId" as "dsActeeId", ds."name" as
 
 
 // Entrypoint to where submissions (a specific version) become entities
-const _processSubmissionDef = (submissionDefId) => async ({ Datasets, Entities, Submissions }) => {
-  const existingEntity = await Entities.getDefBySubmissionDefId(submissionDefId);
```
matthew-white, line 67: This was the only place where `Entities.getDefBySubmissionDefId()` was used, right? Is it worth removing it?

lib/model/query/entities.js
lognaturel: > If approval comes after the edit, then we should use the latest submission version as the basis of the entity, right?

Yes.

> The event that's being hooked into isn't the request

Right, I'm getting ahead of myself here and thinking of the create on receive flow.

It's good to see some simplification and it does seem like that's probably as good as it gets.

matthew-white:
What about the following sequence?

- A submission to an entity-creating form is submitted.
- The submission is approved.
- The worker isn't able to create an entity from the submission. (The database goes down.)
- Some time later, the submission is edited and reapproved. (The database is back up.)
- An entity will be created from the submission edit, not the original submission version.

Of course, this sequence is pretty unlikely, and I think the ultimate outcome is fine and maybe even desirable. But as we continue to think through and communicate the relationship between the submission lifecycle and the entity lifecycle, I just wanted to be explicit that I think Central won't necessarily create the entity from the first submission version that's approved.

matthew-white:
(The alternative would be for Central to determine which submission version was the first to be approved for the submission, then check whether the current version being processed matches that one, bailing if not.)

lognaturel:
> The worker isn't able to create an entity from the submission. (The database goes down.)

There are some number of retries, right? So this would be an extended database outage? Certainly possible but feels pretty unlikely as you say.

We do edits with the latest form version, right? What if I have a form version without entities, I approve the submission, then I update the form to add entities, I edit the submission, and I approve again?

ktuite:
> The worker isn't able to create an entity from the submission. (The database goes down.)

I think if the worker isn't able to process the submission because the db went down, the first event won't be marked as processed, and when it comes back online, it'll still try to process that event first -- the approve event on the earlier version of the submission.

And yes, if someone edits the submission before approving, the edited one will be used for the entity. 

Things will definitely change when entity creation isn't tied so closely to the submission approval event and using the audit log as a job queue.

ktuite:
> We do edits with the latest form version, right? What if I have a form version without entities, I approve the submission, then I update the form to add entities, I edit the submission, and I approve again?

😅 oh man, I don't know. I'd have to double-check what the editing looks like, but if that new def thinks it's for the new form version, which I think it should, then yes, the second approval of the edited form would create an entity.

lib/model/query/entities.js
ktuite: I'll take it out - It's not used, but I was thinking "this feels like it may be useful some day!" If so, I guess I'll bring it back then.

matthew-white:
> There are some number of retries, right?

Usually there are 5 attempts to process an audit log entry, but if I'm remembering correctly, I think it's just one for entity creation. Unless we changed that? I know there are ways we want to adjust that job (#661, #692).

> I think if the worker isn't able to process the submission because the db went down, the first event won't be marked as processed, and when it comes back online, it'll still try to process that event first

Ah right, I keep forgetting that if the database goes down, you can't write anything, including that a job has failed.

I think you're right that the next available worker will claim the first event. But doesn't the existence of multiple workers mean that there's still a race condition? There will be a backlog of events that multiple workers are attempting to process. If another worker claims the second event and finishes it before the first worker finishes the first event, then won't the second worker win? Again, probably not extremely likely in practice.

> Things will definitely change when entity creation isn't tied so closely to the submission approval event and using the audit log as a job queue.

Agreed! So maybe we can wait to come back to some of these questions then.

test/integration/worker/entity.js
```diff
@@ -140,6 +140,47 @@ describe('worker: entity', () => {
...
+          .send(testData.forms.simpleEntity)
+          .set('Content-Type', 'application/xml')
+          .expect(200)
+          .then(() => asAlice.post('/v1/projects/1/forms/simpleEntity/submissions')
+            .send(testData.instances.simpleEntity.one)
+            .set('Content-Type', 'application/xml')
+            .expect(200))
+          .then(() => asAlice.patch('/v1/projects/1/forms/simpleEntity/submissions/one')
+            .send({ reviewState: 'approved' })
+            .expect(200)));
```
matthew-white, line 155: I just merged a PR that allows you to call `service.login()` without passing it a callback, which I think will reduce nesting. So this could be something like:

```js
  .expect(200);
await asAlice.post('/v1/projects/1/forms/simpleEntity/submissions')
  .send(testData.instances.simpleEntity.one)
  .expect(200);
await asAlice.patch('/v1/projects/1/forms/simpleEntity/submissions/one')
  .send({ reviewState: 'approved' })
  .expect(200);
```

No need to change anything of course, just wanted to put the new possibility on your radar!

test/integration/worker/entity.js
```diff
@@ -140,6 +140,47 @@ describe('worker: entity', () => {
...
+          .expect(201)
+          .then(() => asAlice.patch('/v1/projects/1/forms/simpleEntity/submissions/one')
+            .send({ reviewState: 'approved' })
+            .expect(200)));
+
+      await exhaust(container);
+
+      const secondApproveEvent = await container.Audits.getLatestByAction('submission.update').then((o) => o.get());
+      firstApproveEvent.id.should.not.equal(secondApproveEvent.id);
+      should.exist(firstApproveEvent.processed);
```
matthew-white, line 177: Should `firstApproveEvent` be `secondApproveEvent` here?

matthew-white (approved):

test/integration/worker/entity.js
ktuite: oh, thanks! i'll use it.

matthew-white:
> But doesn't the existence of multiple workers mean that there's still a race condition?

Hmm I guess in the case of the database having been down, the worker will almost certainly have a chance to process the first event before the new edit is even made.

PR #746, opened 2023-01-26
Allows task frequencies for purging and analytics to be set in config

### LINKED ISSUE #713: Add ability to adjust frequencies of certain tasks
We have two tasks that happen every 30 days:

- A deleted form is purged after 30 days.
- If usage information metrics are configured, a usage report is sent every 30 days.

However, these frequencies make it hard to verify this functionality during regression testing. Instead, I think it'd be good to add the ability to adjust these frequencies.

One idea I had was to use the `SENTRY_ENVIRONMENT` environment variable for this. Maybe if that environment variable is something specific (`test`? `testing`? anything other than production?), we automatically change the frequencies to every day or every 2 days. The `SENTRY_ENVIRONMENT` variable seems like a convenient variable to reach for, though part of me isn't sure it's a good idea to use that variable for multiple purposes like this.

Alternatively, we could add one or more additional environment variables or Backend config properties for this. Nothing user-facing. Whatever is easiest that allows the QA team to verify this functionality.

ktuite:
To enable, add to `default` in config:
```
"taskSchedule": {
  "purge": 1,
  "analytics": 1
}
```

Otherwise default remains set to 30 days.

Closes #713

I tried setting the numbers through the config and printing them out? Not sure how to write a test or if it's necessary. We don't test other server config settings. 

It's pretty simple.

Not really for users, more for QA to use. Since the default time ranges aren't even being added to the config, most people won't even notice this.

Should there be some sanitization of these values? 

no.

### DIFF
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -590,8 +590,11 @@ const previewMetrics = () => (({ Analytics }) => Promise.all([
   return metrics;
 }));
 
-// Get most recent 'analytics' audit from the past 30 days
-const AUDIT_SCHEDULE = 30;
+// Get most recent 'analytics' audit from the past N days
+const AUDIT_SCHEDULE = config.has('default.taskSchedule.analytics')
+  ? config.get('default.taskSchedule.analytics')
+  : 30; // Default is 30 days
+
 const getLatestAudit = () => ({ maybeOne }) => maybeOne(sql`select * from audits
   where action='analytics' and "loggedAt" >= current_date - cast(${AUDIT_SCHEDULE} as int)
   order by "loggedAt" desc limit 1`);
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

matthew-white:
> Not sure how to write a test or if it's necessary. We don't test other server config settings.

If it were easy to write a test, I think that wouldn't hurt. But I don't see an easy way to mock the config in testing. One part of the config, `config.get('env')`, is in the container as `env`, so it'd be easier to mock all or part of that if we ever wanted to. But I don't think the config as a whole is in the container.

We will only use this config property on the QA server, and the logic here is pretty straightforward, so I don't think it's a big deal not to have a test for it.

> Should there be some sanitization of these values?

I'm not sure, but I don't believe we generally sanitize config properties. These properties in particular are pretty hidden away, so I don't think we need to.

PR #749, opened 2023-01-27
Ignore structural change warning on form publish

sadiqkhoja:
Ignore structural change warning on form publish

Added a test

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -129,7 +129,14 @@ const _getDraftToken = (form) => {
   return generateToken();
 };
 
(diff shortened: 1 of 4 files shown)


ktuite (approved):
Looks good for now, but @sadiqkhoja and I have been discussing how `createVersion`, and the `publish` argument in particular, could be refactored at a later time. 

The meaning of this arguments is unclear and is used in very specific circumstances (E.g. setting the encryption key or updating a form's version) and not how one might assume based on the name.

We've added more logic to `createVersion` that is more thorough about checking for conflicts with previous versions and that doesn't need to run when some of these simpler updates happen.

Issue #439, opened 2022-01-04
Warn if new form definition has structural change


matthew-white:
Deleting a form field can add complications to a data analysis workflow, because the user may still need to access the data from the deleted field. This comes up in particular when a user tries to rename a field or moves a field into or out of a group, because Backend understands that as two operations: the deletion of the old field and the addition of a new field.

In v1.4, we are adding the ability to include deleted fields in the submissions export. @yanokwa has suggested for a later release that we also warn the user if they upload a form definition that deletes a field. The text of the warning would point to the docs. The user experience would be similar to how warnings are currently shown for an XLSForm. One tricky case to consider is if an uploaded form definition generates XLSForm warnings and also deletes a field: how do we return both sets of warnings from Backend?

Related Forum topic: https://forum.getodk.org/t/data-loss-scare-when-updating-to-a-new-form-version/39777

matthew-white:
@yanokwa, you mentioned showing a warning about structural changes to the form. Are there structural changes other than field deletion that should result in a warning? If the user adds a field but doesn't delete a field, no warning should be shown, right?

yanokwa:
Adding a field should not trigger a warning. It's only deletion (direct or indirect through renaming).

lognaturel:
Here's a user describing the current experience: https://forum.getodk.org/t/data-loss-scare-when-updating-to-a-new-form-version/39777/2

dbemke:
A user is unable to publish a draft version of o form with a deleted field. After clicking the “proceed” button the dialog is shown again.
In the warning there is “?ignoreWarnings=true”



![Image](https://user-images.githubusercontent.com/56479916/214815905-1050b8b7-446a-4ba6-b378-5ef749139a4c.png)

srujner:
Does deleting the whole Column form the xml file should also trigger a warning, because right now it doesn't?

matthew-white:
> A user is unable to publish a draft version of o form with a deleted field.

This should be resolved now.

> Does deleting the whole Column from the xml file should also trigger a warning, because right now it doesn't?

A warning should be shown whether the user uploads an XLSForm or an XML XForm. Is that what you mean? If you have an example that would be easy to share, it'd be helpful to see what file you're trying.

matthew-white:
> Does deleting the whole Column from the xml file should also trigger a warning, because right now it doesn't?

[...]

PR #757, opened 2023-01-30
Fix compare to not match select and select1 fields

ktuite:
Related to a bug found by QA (and @matthew-white) when looking into issue #439. 

A "select" and a "selectMultiple" field were considered to be the same type of field, so if a user uploaded a form with a single select (`select1`) and then changed it to a selectMultiple (`select` in xml), the intermediate form schema would not think the schema had changed, continue to treat it as a single select, and thus any multi select behavior (mainly involving exports) would fail. 

This PR adds a condition in to the form schema `compare()` utility to also look at the selectMultiple attribute (or lack thereof) on a field. 

It also adds a test for the scenario above and some additional unit tests. 

Because the migration uses this `compare` function, this should fix any issues with that, too.

Tests.

It's the best approach.

Fixes a bug.

No.

### DIFF
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -432,7 +432,11 @@ const compare = (prevFields, fields) => {
   for (let i = 0; i < fields.length; i += 1) {
     if (!(prevFields[i].path === fields[i].path &&
       prevFields[i].order === fields[i].order &&
-      prevFields[i].type === fields[i].type)) {
+      prevFields[i].type === fields[i].type &&
+      // field.selectMultiple may be true, null, or undefined.
+      // null should be seen as equal to undefined here.
+      // eslint-disable-next-line eqeqeq
+      prevFields[i].selectMultiple == fields[i].selectMultiple)) {
       return false;
     }
   }
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -434,9 +434,8 @@ const compare = (prevFields, fields) => {
       prevFields[i].order === fields[i].order &&
       prevFields[i].type === fields[i].type &&
       // field.selectMultiple may be true, null, or undefined.
-      // null should be seen as equal to undefined here.
-      // eslint-disable-next-line eqeqeq
-      prevFields[i].selectMultiple == fields[i].selectMultiple)) {
+      // should match if both are true, or both are something other than true.
+      (prevFields[i].selectMultiple === true) === (fields[i].selectMultiple === true))) {
       return false;
     }
   }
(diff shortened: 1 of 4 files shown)


test/unit/data/schema.js
```diff
@@ -1157,6 +1157,46 @@ describe('form schema', () => {
...
+          </model>
+        </h:head>
+        <h:body>
+          <select1 ref="/data/q1"><label>one</label></select1>
+          <group ref="/data/g1">
+            <label>group</label>
+            <select1 ref="/data/g1/q2"><label>two</label></select1>
+          </group>
+        </h:body>
+      </h:html>`;
```
matthew-white, line 1191: Is this the same XML as what's in the integration test? Is it worth including it in `testData.forms`? (Maybe nearby `testData.forms.selectMultiple` in case that form ever changes?)

test/integration/other/select-many.js
```diff
@@ -83,5 +83,72 @@ describe('select many value processing', () => {
             { formId, submissionDefId: one2, path: '/g1/q2', value: 'xyz' }
           ]);
         }))));
+
+  it('should handle when field changes from select1 to select (selectMultiple)', testService(async (service, container) => {
```
matthew-white, line 87: Yay, thanks for writing this test! I'm happy it passes. I forgot that I think we handle unprocessed values on the fly.

matthew-white (approved):
Perfect! 💯

What do you think about adding a comment around the `Form.Field` frame? Something to the effect of "if you add a new field here, you may need to modify compare() to account for it".

lib/data/schema.js
```diff
@@ -432,7 +432,11 @@ const compare = (prevFields, fields) => {
   for (let i = 0; i < fields.length; i += 1) {
     if (!(prevFields[i].path === fields[i].path &&
       prevFields[i].order === fields[i].order &&
-      prevFields[i].type === fields[i].type)) {
+      prevFields[i].type === fields[i].type &&
+      // field.selectMultiple may be true, null, or undefined.
+      // null should be seen as equal to undefined here.
+      // eslint-disable-next-line eqeqeq
+      prevFields[i].selectMultiple == fields[i].selectMultiple)) {
```
matthew-white, line 439: Could also do something like

```js
(prevFields[i].selectMultiple === true) === (fields[i].selectMultiple === true)
```

or

```js
(!!prevFields[i].selectMultiple) === (!!fields[i].selectMultiple)
```

if you also wanted to account for `false`. Though it sounds like we don't currently ever set `selectMultiple` to `false`.

matthew-white: (`false == null` is `false`.)

test/unit/data/schema.js
ktuite: I thought about this, and even though it's used twice, I'm not sure it should go there. One main reason is the id is still called 'selectMultiple'. Maybe once there's a >3x usage of the form or something!

PR #759, opened 2023-02-02
Check for meta group on form upload

ktuite:
This PR addresses a situation (maybe this should become an issue but it's also not that big of a deal) where the submission view page on Frontend failed to load when showing a submission for a form without a `<meta/>` tag (generally containing the instance ID). We expect forms in the present day to all include meta information, but in case they don't, this change will now reject the uploading of forms that are missing the meta group.

Tests.

Another option is leaving the code as is and not checking. I think the biggest risk is for us as developers - it's more code to maintain. And many of the old tests didn't have meta information so I had to go through and update them. 

*Edited to add*: For forms that are already in the system, this might cause problems when trying to make a new draft of a meta-less form. 

It should prevent the uploading of a form without `<meta><instanceID/></meta>`, which was a condition that would break a small part of frontend (the submission viewing page).

I don't think so; I think this is part of the xform spec already. 

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -170,9 +170,6 @@ const createVersion = (partial, form, publish = false, duplicating = false) => a
     if (match)
       schemaId = prevFields[0].schemaId;
     else {
-      // Check that meta field (group containing instanceId and name) exists
-      await Forms.checkMeta(fields);
-
       // Only need to check new fields against old fields if the structure does not match
       const allFields = await Forms.getMergedFields(form.id);
       await Forms.checkFieldDowncast(allFields, fields);
@@ -623,8 +620,10 @@ order by actors."displayName" asc`)
 ////////////////////////////////////////////////////////////////////////////////
 // CHECKING CONSTRAINTS, STRUCTURAL CHANGES, ETC.
 
+// This will check if a form contains a meta group (for capturing instanceID).
+// It is only to be used for newly uploaded forms.
 const checkMeta = (fields) => () =>
-  (fields.some((f) => f.name === 'meta')
+  (fields.some((f) => (f.name === 'meta' && f.type === 'structure'))
     ? resolve()
     : reject(Problem.user.missingMeta()));
 
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -75,6 +75,9 @@ const createNew = (partial, project, publish = false) => async ({ run, Datasets,
     (partial.aux.key.isDefined() ? resolve(Option.none()) : getDataset(partial.xml)) // Don't parse dataset schema if Form has encryption key
   ]);
 
+  // Check that meta field (group containing instanceId and name) exists
+  await Forms.checkMeta(fields);
+
   // Check for xmlFormId collisions with previously deleted forms
   await Forms.checkDeletedForms(partial.xmlFormId, project.id);
   await Forms.rejectIfWarnings();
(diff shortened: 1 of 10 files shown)


test/integration/api/forms/draft.js
```diff
@@ -406,6 +406,32 @@ describe('api: /projects/:id/forms (drafts)', () => {
               .set('Content-Type', 'application/xml')
               .expect(200)))));
 
+
+      it('should reject new draft with missing meta group', testService(async (service) => {
```
matthew-white, line 410: What happens if you have a form that doesn't have a `meta` group and you want to create a new form version that includes `meta`? Are you able to use Frontend to create a draft to do so, or will that request fail?

lib/model/query/forms.js
```diff
@@ -617,6 +623,11 @@ order by actors."displayName" asc`)
 ////////////////////////////////////////////////////////////////////////////////
 // CHECKING CONSTRAINTS, STRUCTURAL CHANGES, ETC.
 
+const checkMeta = (fields) => () =>
+  (fields.some((f) => f.name === 'meta')
```
matthew-white, line 627: While we're at it, how about we check that `f.type === 'structure'`?

matthew-white (commented):
I think this is a useful check! I didn't look at all the code, but I wanted to leave a couple of thoughts.

test/integration/api/forms/draft.js
ktuite: Oh, good point. It will probably be bad. I'm not sure how to even write a test for this.

One way around it is only check for the missing meta field on a new form. Then if they delete it in a later version, they'll see a warning about the structure changing, though they won't be blocked from it.

matthew-white: I think that would be a nice improvement!

lib/model/query/forms.js
```diff
@@ -75,6 +75,9 @@ const createNew = (partial, project, publish = false) => async ({ run, Datasets,
     (partial.aux.key.isDefined() ? resolve(Option.none()) : getDataset(partial.xml)) // Don't parse dataset schema if Form has encryption key
   ]);
 
+  // Check that meta field (group containing instanceId and name) exists
+  await Forms.checkMeta(fields);
```
sadiqkhoja, line 79: do we need this in `createVersion`?

ktuite: I took it out of createVersion because of Matt's comment above: 

> What happens if you have a form that doesn't have a `meta` group, and you want to create a new form version that includes `meta`? Are you able to use Frontend to create a draft to do so, or will that request fail?

It would likely break the ability to create a new draft of any existing meta-less form. But at least catching the problem when the form is new is a good improvement.

test/integration/api/forms/forms.js
```diff
@@ -438,6 +438,59 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
         });
     }));
 
+    it('should reject form with missing meta group', testService(async (service) => {
```
sadiqkhoja, line 441: can you please add another test case:
- publish a form with meta group
- upload another draft for the same form without meta group, it should be rejected

ktuite:
I've been thinking about whether or not to add this meta check back to `createVersion` (when a form is updated) and some reasons against doing this are:
- not possible to write a test for the case where the first version of a form doesn't contain meta information, because we now require it on form upload
- if someone has a meta-less form already, they may need to update it without also being required to add a meta group
- if someone tries to remove the meta group from a form that had one already, they will see a warning about it being removed 
- not able to have QA try it
- most modern forms probably do have meta group already so this is probably an edge case

sadiqkhoja (approved):

PR #765, opened 2023-02-13
Disallow string downcast for groups and repeats

### LINKED ISSUE #381: Disallow string downcast for groups and repeat groups
Right now, Backend will allow any field to be downcast to string. I just checked, and it looks like that includes groups and repeat groups. Is that something we want to prevent? My instinct is that it would make it easier to reason about types (and maybe also XML traversal?) if we could assume that structures never become strings.

ktuite:
Closes #381 

Unskipping some tests that existed before.

It was easy to add this to the downcast checking function! 

Shouldn't have a big impact on users unless they're changing their field types a lot. Should help people keep their data cleaner. 

I don't think so.

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -666,11 +666,21 @@ const rejectIfWarnings = () => ({ context }) => {
   }
 };
 
+// This function replaces a database trigger called check_field_collisions that
+// prevents certain kinds of field type changes and downcasts.
+// - Most types can be downcast to a string, but they cannot change between non-string types.
+//    e.g. A date can become a string, but a string can't become a date, int, etc. because if
+//    the data in that string is not a valid date/int, it will cause problems when exporting.
+// - A group or repeat cannot be downcast to a string.
 const checkFieldDowncast = (allFields, fields) => () => {
   const lookup = {};
   for (const f of allFields) lookup[f.path] = f;
-  for (const f of fields.filter(ff => (ff.type !== 'string'))) {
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

ktuite:
Hi QA! 
Here are two forms that represent the change from a group field to a string with the same name as a group:

[group downcast.xlsx](https://github.com/getodk/central-backend/files/10845613/group.downcast.xlsx)

[group downcast (group is now string).xlsx](https://github.com/getodk/central-backend/files/10845612/group.downcast.group.is.now.string.xlsx)

I just tried on https://test.getodk.cloud/#/projects/445 (not updated? at v2023.1.) and it is allowed.

Also, here's the message that mentions 'now in different groups or repeats' that it detected but didn't stop:
Screen Shot 2023-02-27 at 5 08 58 PM

 It should *not* be allowed in staging.

dbemke:
Hi. Thank you for the information, it was really helpful to know where the change to string happened. I changed the forms a bit and tried to upload them to Central. I've noticed that if I change the group/repeat field to a string and put them inside a repeat/group I'm still able to upload the one changed to a string. I don't know if it's a valid case - should it also be prevented.
insiderepeatgroup
[group.downcast.xlsx](https://github.com/getodk/central-backend/files/10848839/group.downcast.xlsx)
[group.downcast.group.is.now.string.xlsx](https://github.com/getodk/central-backend/files/10848843/group.downcast.group.is.now.string.xlsx)

PR #766, opened 2023-02-13
Return specific Problem when form version in submission not found

### LINKED ISSUE #629: Return specific Problem if form version in submission XML is not found
From the forum: https://forum.getodk.org/t/central-api-docs-example-submission-xml-issue/39139

If you send submission XML whose `version` attribute does not match any version of the form, then Backend returns a 404.1 Problem ("Could not find the resource you were looking for."):

https://github.com/getodk/central-backend/blob/a2ca89a0744d34fc1f68e052b51b4d0179aadb1a/lib/resources/submissions.js#L123-L127
https://github.com/getodk/central-backend/blob/a2ca89a0744d34fc1f68e052b51b4d0179aadb1a/lib/resources/submissions.js#L209-L212

However, especially if the user doesn't know about the `version` attribute or that Backend is searching on it, they might not understand why they're receiving a 404 response and might not know how to proceed. (The user might wonder: Does the URL not match a current endpoint? Is the project ID wrong? The form ID?) I think I've gotten confused around this myself after forgetting to specify the `version` attribute.

I think it'd be useful to return a specific 404 Problem in this case. I think it'd be useful for the Problem to indicate the project ID + `xmlFormId` + form version string that Backend is looking for. The forum post mentions the REST endpoint specifically, but I think it'd be useful to update both the REST and OpenRosa endpoints.

CC @lindsay-stevens

ktuite:
Closes #629, where if the submission XML is missing the version string or has a version that doesn't match any known version for that form, the response returns a generic 404 error message. Now it returns a more informative 404 error message. 

The 404 resource not found is an artifact of how when we have a specific version to look for, our form query has these specific constrains and it's easy to reject if we don't find what we are looking for.

The approach of this PR changes that flow a bit: if we don't find the right form and form def/version that the submission is targeting, we look again _without_ the specific version constraint. If we find something when we look that way, then it's a problem with the form version. It feels like extra work, but it'll lead to clearer error messages.

It seems like we only check for the form version when dealing with a published form. For draft forms, the version in the submission doesn't actually have to match (maybe it should? it would be more code..) It may not be that necessary because all the submissions for that draft will be purged when the draft is purged (via publishing or deleting).  

Tests.

Open to discussing other options... this is very focused on the version. I think if the user gets the project ID or xmlFormId wrong, the original rejection is appropriate.

Clearer error messages, which will especially help API users and ourselves when writing tests!

I don't think we document the nuances of such errors.

### DIFF
--- a/lib/util/problem.js
+++ b/lib/util/problem.js
@@ -136,6 +136,9 @@ const problems = {
     // deprecated id does not exist.
     deprecatedIdNotFound: problem(404.5, ({ deprecatedId }) => `You tried to update an existing submission, but we can't find that submission to update (${deprecatedId})`),
 
+    // form version in submission does not exist
+    formVersionNotFound: problem(404.6, ({ formVersion }) => `The form version specified in this submission (${formVersion}) does not exist.`),
+
     // { allowed: [ acceptable formats ], got }
     unacceptableFormat: problem(406.1, ({ allowed }) => `Requested format not acceptable; this resource allows: ${allowed.join()}.`),
 
--- a/lib/util/problem.js
+++ b/lib/util/problem.js
@@ -137,7 +137,7 @@ const problems = {
     deprecatedIdNotFound: problem(404.5, ({ deprecatedId }) => `You tried to update an existing submission, but we can't find that submission to update (${deprecatedId})`),
 
     // form version in submission does not exist
-    formVersionNotFound: problem(404.6, ({ formVersion }) => `The form version specified in this submission (${formVersion}) does not exist.`),
+    formVersionNotFound: problem(404.6, ({ formVersion }) => `The form version specified in this submission '${formVersion}' does not exist.`),
 
     // { allowed: [ acceptable formats ], got }
     unacceptableFormat: problem(406.1, ({ allowed }) => `Requested format not acceptable; this resource allows: ${allowed.join()}.`),
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -10,13 +10,13 @@
 const config = require('config');
 const { sql } = require('slonik');
 const { map, compose } = require('ramda');
-const { Frame, into } = require('../frame');
+const { Frame, into, ensureDef } = require('../frame');
 const { Actor, Blob, Form, Dataset } = require('../frames');
 const { getFormFields, merge, compare } = require('../../data/schema');
 const { getDataset } = require('../../data/dataset');
 const { generateToken } = require('../../util/crypto');
 const { unjoiner, extender, updater, equals, insert, insertMany, markDeleted, markUndeleted, QueryOptions } = require('../../util/db');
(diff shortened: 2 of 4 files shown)


matthew-white:
> It seems like we only check for the form version when dealing with a published form. For draft forms, the version in the submission doesn't actually have to match (maybe it should? it would be more code..)

Ah very interesting, it's good to know it works that way. I don't have strong feelings about whether it's important to check that. Maybe it'd be sensible to do so, but it also feels low-priority.

> Open to discussing other options... this is very focused on the version. I think if the user gets the project ID or xmlFormId wrong, the original rejection is appropriate.

If the user gets the project ID or xmlFormId wrong, then I think they'll get stopped before they reach the code that you're modifying. I think in that case today, they'll get stopped by the `getOrNotFound` before `ensureDef`.

Right now, `Forms.checkWrongFormVersion()` calls `Forms.getByProjectAndXmlFormId()` again in order to decide whether to reject with a `Problem.user.formVersionNotFound()` or with the Problem from `ensureDef()` (a generic `Problem.user.notFound()`). But I'm wondering whether the latter case is a real possibility / whether there are two different cases here. If `getOrNotFound()` is catching cases where the project ID or xmlFormId is wrong, then I think the only way that `form.def.id == null` is if the form version string in the submission doesn't match a published form version, which is the `Problem.user.formVersionNotFound()` case. So could we skip the second `Forms.getByProjectAndXmlFormId()` and just reject with a `Problem.user.formVersionNotFound()` if `form.def.id == null`? Or maybe I'm misunderstanding how the flow here works. 😅

> I don't think we document the nuances of such errors.

Not yet! Though hopefully we will at some point once we've completed #753.

matthew-white:
> if the submission XML is missing the version string or has a version that doesn't match any known version for that form, the response returns a generic 404 error message.

I don't think it's inherently a problem if the submission is missing the form version string. If it's missing the form version string, then I think it falls back to an empty string:

https://github.com/getodk/central-backend/blob/c90025b65f3232c1b6b61f9ee99eecf43a23a910/lib/model/frames/submission.js#L63

Then if there's a published form version whose version string is an empty string, then the submission will match that.

lib/util/problem.js
```diff
@@ -136,6 +136,9 @@ const problems = {
     // deprecated id does not exist.
     deprecatedIdNotFound: problem(404.5, ({ deprecatedId }) => `You tried to update an existing submission, but we can't find that submission to update (${deprecatedId})`),
 
+    // form version in submission does not exist
+    formVersionNotFound: problem(404.6, ({ formVersion }) => `The form version specified in this submission (${formVersion}) does not exist.`),
```
matthew-white, line 140: In the case where `formVersion` is an empty string, there won't be anything between the parentheses. What do you think about putting single quotes around `formVersion`? We seem to do something similar for other Problems.

matthew-white (approved):

PR #771, opened 2023-02-15
fixes 707: use sql.unnest instead of sql.join for bulk insert

### LINKED ISSUE #707: Error not handled when uploading a big form
[25000q-form.xml.txt](https://github.com/getodk/central-backend/files/10112142/25000q-form.xml.txt)

When the big form above is uploaded (with `.txt` extension removed), I see the following error:

```
RangeError: Maximum call stack size exceeded
    at sql (/odk/backend/node_modules/slonik/dist/src/factories/createSqlTag.js:36:29)
    at insertMany (/odk/backend/lib/util/db.js:157:13)
    at /odk/backend/lib/model/query/forms.js:72:15
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.transaction (/odk/backend/node_modules/slonik/dist/src/connectionMethods/transaction.js:22:24)
    at async Object.createConnection (/odk/backend/node_modules/slonik/dist/src/factories/createConnection.js:97:18)RangeError: Maximum call stack size exceeded
    at sql (/odk/backend/node_modules/slonik/dist/src/factories/createSqlTag.js:36:29)
    at insertMany (/odk/backend/lib/util/db.js:157:13)
    at /odk/backend/lib/model/query/forms.js:72:15
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.transaction (/odk/backend/node_modules/slonik/dist/src/connectionMethods/transaction.js:22:24)
    at async Object.createConnection (/odk/backend/node_modules/slonik/dist/src/factories/createConnection.js:97:18)::ffff:127.0.0.1 - - [29/Nov/2022:10:13:26 +0000] "POST /v1/projects/973/forms HTTP/1.1" 500 739
Error: Connection terminated
    at Connection.<anonymous> (/odk/backend/node_modules/pg/lib/client.js:132:36)
    at Object.onceWrapper (node:events:627:28)
    at Connection.emit (node:events:525:35)
    at Connection.emit (node:domain:489:12)
    at Socket.<anonymous> (/odk/backend/node_modules/pg/lib/connection.js:57:12)
    at Socket.emit (node:events:513:28)
    at Socket.emit (node:domain:489:12)
    at TCP.<anonymous> (node:net:757:14)node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
```

Locally, this crashes my node process.  On the server, the process does not crash, but the error is sent to Sentry.

An `odk-central-frontend` user sees:

Screenshot_2022-11-29_13-27-02

The response body contains the full stacktrace (as per https://github.com/getodk/central-backend/issues/549).

sadiqkhoja:
Closes #707 

Unit test, integration test and manually tested with the form with 25K questions.

Postgresql has a limit of parameters that can be passed in a query. To insert large number of rows it provides `unnest` function. Alternative is to break the rows into batches which is not performant and would require a lot of code changes.

Test cases related to multiselect question, client audits, form attachments and submission attachments should be executed.

no

### DIFF
--- a/lib/model/frame.js
+++ b/lib/model/frame.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { raw } = require('slonik-sql-tag-raw');
-const { pick, without } = require('ramda');
+const { pick, without, remove, indexOf } = require('ramda');
 const uuid = require('uuid/v4');
 const { pickAll, noargs } = require('../util/util');
 const Option = require('../util/option');
@@ -26,6 +26,10 @@ const table = (name, to) => (def) => {
   def.from = name;
   def.to = to || name.substring(0, name.length - 1); // take off the s
 };
+const fieldTypes = (types) => (def) => {
+  def.fieldTypes = types;
+};
+
 const from = (fro) => (def) => { def.from = fro; };
 const into = (to) => (def) => { def.to = to; };
 
@@ -66,6 +70,8 @@ class Frame {
       const Frame = class extends this { static get def() { return def; } };
       Frame.fieldlist = raw(def.fields.map((s) => `"${s}"`).join(','));
       Frame.insertfields = without([ 'id' ], def.fields);
+      const indexOfId = indexOf('id', def.fields);
+      Frame.insertFieldTypes = indexOfId > -1 ? remove(indexOf('id', def.fields), 1, def.fieldTypes ?? []) : def.fieldTypes;
       Frame.insertlist = raw(Frame.insertfields.map((s) => `"${s}"`).join(','));
       Frame.hasCreatedAt = def.fields.includes('createdAt');
       Frame.hasUpdatedAt = def.fields.includes('updatedAt');
@@ -156,5 +162,5 @@ Frame.def = { fields: [], readable: [], hasCreatedAt: false, hasUpdatedAt: false
 // util func to ensure the frame we fetched successfully joined to the intended def.
 const ensureDef = rejectIf(((f) => f.def.id == null), noargs(Problem.user.notFound));
 
-module.exports = { Frame, table, from, into, aux, species, embedded, readable, writable, ensureDef };
+module.exports = { Frame, table, fieldTypes, from, into, aux, species, embedded, readable, writable, ensureDef };
 
--- a/lib/util/db.js
+++ b/lib/util/db.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { inspect } = require('util');
-const { mergeRight, pick, always } = require('ramda');
+const { mergeRight, pick, always, without, remove, indexOf } = require('ramda');
 const { sql } = require('slonik');
 const { raw } = require('slonik-sql-tag-raw');
 const { reject } = require('./promise');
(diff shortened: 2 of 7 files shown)


lib/model/frame.js
```diff
@@ -26,6 +26,10 @@ const table = (name, to) => (def) => {
   def.from = name;
   def.to = to || name.substring(0, name.length - 1); // take off the s
 };
+const fieldTypes = (types) => (def) => {
+  def.fieldTypes = types;
+};
+
```
sadiqkhoja, line 32: Being pragmatic here, don't want to change a whole lot of things. With this function, we can define fieldTypes for only those frames that are bulk inserted using `insertMany` function.

Alternative was to change the def.fields from array of string to array of object with name, type, readable, writeable, etc or have associated array of types and define type for all fields in every frame.

lib/util/db.js
```diff
@@ -151,15 +151,35 @@ values (${sql.join(keys.map(_assign(obj)), sql`,`)})
...
   if (objs.length === 0) return sql`select true`;
   const Type = objs[0].constructor;
+  if (!Type.def.fieldTypes) throw Problem.internal.fieldTypesNotDefined(Type.from);
+
+  let columns; let rows; let columnTypes; let
+    selectExp;
+
+  // we need to set clock_timestamp if there's createdAt column
+  // Slonik doesn't support setting sql identitfier for sql.unnest yet
+  if (Type.hasCreatedAt) {
```
sadiqkhoja, line 168: this special treatment of createdAt field is due to an existing unit test, we don't have any frame with createdAt that we bulk insert today. This is nice to have for future use cases.

lib/model/frame.js
sadiqkhoja: we can remove this when and if https://github.com/gajus/slonik/issues/455 is implemented

ktuite (approved):
Writing out my understanding after an interactive code review with @sadiqkhoja:

We use a function called `insertMany` to do bulk inserting of certain things including Form Fields, Submission Attachments, Comments, Audits, and Client Audits.

`insertMany` was using `sql.join` internally, but broke down in extreme cases such as a very large form with 25K fields. In such a form, we store ~10 columns about each form field, so `sql.join` was forming a query with 250,000 placeholders that was too much (exceeding some stack size in node). Sadiq pinpointed the line in slonik (in sql.join) that was the problem, where there was a spread operator that was failing on those 250K values... but rather than trying to change slonik, he found a different `sql.unnest` slonik function to use instead.

`sql.unnest` works by taking arrays of placeholder values, e.g., when inserting 25K of 10-column form fields, it will have 10 placeholders where each is a 25K length array of every value for that one column. The one caveat of using this function is that it takes a second required argument to list out the types of each column.

Most of the code in this PR is about modifying our frames to also define field types on any frame that needs to use `insertMany`. If field types are missing and insertMany is called anyway, there is a 500 Problem explaining that.

To summarize
- `insertMany` changed to use `sql.unnest` instead of `sql.join`.
- Frames that use `insertMany` must now also specify a `fieldTypes` array for the properties in that frame.

alxndrsn:
Another approach to this could be using ` json_to_recordset()`/`jsonb_to_recordset()`, although not available in postgres 9.6.

PR #773, opened 2023-02-21
Use postgres 14

lognaturel:
Match what the standard setup will be starting with 2023.2. There are no changes needed to the code so we might as well start doing this now.

Nothing, CI should be enough.

We could wait until we're ready to release backend but I don't see any advantage in that. I used 14.6 which is the [latest available from CircleCI](https://circleci.com/developer/images/image/cimg/postgres). This matches our [recommendations in docs](https://github.com/getodk/docs/pull/1577/files#diff-c3ac24ca7534cc83beec0c8af79032b0aff1d9010803759e057192dcd4739952R205).

None, this is CI only.

No.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     docker:
       - image: cimg/node:16.19.1
-      - image: cimg/postgres:9.6
+      - image: cimg/postgres:14.6
         environment:
           POSTGRES_PASSWORD: odktest
 
--- a/.github/workflows/soak-test.yml
+++ b/.github/workflows/soak-test.yml
@@ -12,7 +12,7 @@ jobs:
     services:
       # see: https://docs.github.com/en/enterprise-server@3.5/actions/using-containerized-services/creating-postgresql-service-containers
       postgres:
-        image: postgres:9.6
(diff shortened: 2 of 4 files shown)


matthew-white:
I searched the codebase for "9.6" and see a few other references:

- .github/workflows/soak-test.yml
- `run-docker-postgres` in the makefile. @ktuite, I know you use `run-docker-postgres` in development. How easy do you think it would be to change the version of Postgres there?
- We link to the Postgres 9.6 docs from the API docs in one place.

ktuite:
Good catches, @matthew-white!

I updated the postgres docker image for the soak test:
> * .github/workflows/soak-test.yml

Following the convention in the overall postgres update of calling the new docker container `postgres14` instead of just `postgres`, I renamed the container made by this command. This lets me keep my old container around and figure out how to copy over data later or just deal with having a fresh dev database.
> * `run-docker-postgres` in the makefile. @ktuite, I know you use `run-docker-postgres` in development. How easy do you think it would be to change the version of Postgres there?

Updated this link
> * We link to the Postgres 9.6 docs from the API docs in one place.

matthew-white (approved):

PR #776, opened 2023-02-23
ETag for entities.csv

### LINKED ISSUE #770: Add etags to dataset CSV endpoints
Have the dataset request endpoint produce an `ETag` in the response header and also read and handle the corresponding value `If-None-Match` in the request header, returning `304` if there is a match.

The value of the etag will be the hash of the last updated at timestamp, the same value we send as the md5 hash in the manifest.

Handling the incoming etag value in `If-None-Match` will look like fetching the last updated timestamp of the dataset, hashing it, comparing it to the provided value, and returning `304 Not Modified` if there is a match, otherwise returning the dataset CSV as before (with generated etag).

While Collect will still make a request to Central, and Central will still have to look at the dataset/entities to check the etag, Collect will be able to use the 304 response from Central to not re-download the dataset CSV, saving on bandwidth between the two.

Central does not yet return any 304s, but it does return a bunch of custom 400 and 500 error http codes, so it should hopefully be straightforward to return a 304 in this case.

sadiqkhoja:
Closes #770 

## Changes:
- Added ETag in response of entity.csv and logic to return 304 if `If-None-Match` matches serverEtag
- Fix a bug: form manifest crashes if the form has attached dataset with entities in it. Root cause md5sum function expects a string and we were passing a Date object
- Fix: use coalesce of updatedAt and createdAt for timestamp of latest entity

- Added integration tests
- Manually tested ETag logic

NA

Test ODK Collect's behaviour for forms with dataset attached that has entities in it.

No

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -50,7 +50,7 @@ const _getAllByProjectId = (fields, extend, options) => sql`
   SELECT ${fields} FROM datasets
   ${extend|| sql`
   LEFT JOIN (
-    SELECT "datasetId", count(1) "entities", max("createdAt") "lastEntity" 
+    SELECT "datasetId", COUNT(1) "entities", MAX(COALESCE("updatedAt", "createdAt")) "lastEntity" 
     FROM entities e 
     GROUP BY "datasetId" 
   ) stats on stats."datasetId" = datasets.id`}
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -24,7 +24,7 @@ const pickFrameFields = (frame, obj) => compose(
 )(obj);
 
 const makeHierarchy = reduce((result, item) => {
-  const dataset = new Dataset(pickFrameFields(Dataset, item));
+  const dataset = new Dataset(pickFrameFields(Dataset, item)).with({ lastEntity: item.lastEntity });
   const property = new Dataset.Property(pickFrameFields(Dataset.Property, item));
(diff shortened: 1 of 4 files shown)


lib/model/query/datasets.js
```diff
@@ -121,9 +121,14 @@ const _getByIdSql = ((fields, datasetId) => sql`
 
 const _getByNameSql = ((fields, datasetName, projectId, includeForms) => sql`
     SELECT
-        ${fields}
+        ${fields}, stats."lastEntity" "datasets!lastEntity"
```
ktuite, line 124: Can we do this with the `Dataset.Extended` frame like in `_getAllByProjectId`?

lib/model/query/datasets.js
```diff
@@ -121,9 +121,14 @@ const _getByIdSql = ((fields, datasetId) => sql`
...
     FROM
         datasets
+    LEFT OUTER JOIN (
+      SELECT "datasetId", MAX(COALESCE("updatedAt", "createdAt")) "lastEntity"
```
lognaturel, line 128: `coalesce` always returns a single value (first non-null), right? So this would always call `max` on a single value and return that?

lognaturel, line 128: Should have edited my comment rather than deleting. `MAX(COALESCE("updatedAt", "createdAt"))` wasn't looking right to me but I have figured it out.

ktuite (approved):

Issue #767, opened 2023-02-14
Guarantee entity order in CSV


ktuite:
Order entities in the same way submissions are ordered `entities."createdAt" desc, entities.id desc` and note it in the API docs. 

Comments from [original entity PR here](https://github.com/getodk/central-backend/pull/576/files#r1014975206).

Issue #675, opened 2022-11-14
Guarantee dataset property order in CSV


ktuite:
Comment from [Matt](https://github.com/getodk/central-backend/issues/649#issuecomment-1312601181):

> Do we want an `order` column on `ds_properties`? I might be missing something, but right now when we write the entities CSV file, I don't see that there's a guarantee about the order of the columns after `name` and `label` (the order of the properties). I'd expect that the order of the properties would be stable.
> 
> I'm not sure how much the column order matters when the entities CSV file is served to an OpenRosa client. The column order isn't user-facing there, right? But as users start downloading the entities CSV file for use in data management or analysis, I think it'll be important that the column order be stable.
> 
> I don't think we can rely on `form_fields.order` for this, as multiple forms can feed into the dataset: two forms may save the same fields to the same dataset properties, yet have those fields in different relative orders. Even for a single form, the field order may change from version to version. Also, given that forms and datasets are allowed to have separate lifecycles, I think that the order of properties should remain stable even if a contributing form is deleted.
> 
> Instead, I think we should independently track the property order using a column on `ds_properties`. Maybe for now we just order by `ds_properties.id` and have that be our guaranteed order.

lognaturel:
> I'm not sure how much the column order matters when the entities CSV file is served to an OpenRosa client. The column order isn't user-facing there, right?

That's right, it doesn't matter for OpenRosa clients.

> order by ds_properties.id and have that be our guaranteed order.

That feels like a reasonable long term solution to me. We know those ids are auto-incrementing. Why introduce a separate way to capture order?

PR #780, opened 2023-02-23
Return dataset metadata properties and entities in order

### LINKED ISSUE #779: Guarantee dataset property order in metadata
Return the properties of a given dataset in order when fetching the metadata of the project. This is very related to #675 as both use `Datasets.getByName`.

ktuite:
Closes #779, #767 and #675 

Dataset properties when returned for a specific dataset (not in a diff) or for the header of a CSV are ordered first in published order, and then the order they are found in the form. 

Things get a little complicated when multiple forms share properties, but as much as possible, properties are grouped with the form fields they came from. 

Entities in the CSV and in the OData endpoint are returned in reverse chronological order like submissions with the newest entity at the top.

This PR also fixes an issue with setting the publishedAt field on relevant dataset properties when they already exist but are unpublished.

Tests. 

One original suggestion was to order the properties by their property id, but there can be a situation where
- form 1 has properties A, B, C, D
- form 2 has properties A, B, (doesn't make new properties for C and D because they are already there and have lower id)
- form 1 gets deleted and purged before being published
- form 2 gets published
- property id order is C, D, A, B

Makes entities table look cleaner. 

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -438,11 +438,11 @@ describe('datasets and entities', () => {
               publishedAt.should.not.be.null();
               return p;
             }).should.be.eql([
-              { name: 'age', forms: [ { name: 'simpleEntity', xmlFormId: 'simpleEntity' }, ] },
               { name: 'first_name', forms: [
                 { name: 'simpleEntity', xmlFormId: 'simpleEntity' },
                 { name: 'simpleEntity2', xmlFormId: 'simpleEntity2' }
               ] },
+              { name: 'age', forms: [ { name: 'simpleEntity', xmlFormId: 'simpleEntity' }, ] },
               { name: 'address', forms: [ { name: 'simpleEntity2', xmlFormId: 'simpleEntity2' }, ] }
             ]);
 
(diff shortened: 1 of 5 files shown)


sadiqkhoja (approved):

test/integration/api/datasets.js
```diff
@@ -415,6 +415,96 @@ describe('datasets and entities', () => {
...
+                'd_q2',
+                'a_q3',
+                'c_q4',
+                'f_q1',
+                'e_q2'
+              ]);
+          });
+      }));
+
+      // Test is broken because unpublished properties from previous form don't get properly published in new form.
```
sadiqkhoja, line 473: this comment can be removed

sadiqkhoja (approved):
Reviewed recent comments, they look fine.

ktuite:
Before I forget, I wanted to articulate some of the challenges with this PR:

Many tables have to be joined to get the state of a dataset's properties:
1. `datasets` (look up by dataset name, project) (can be published or unpublished)
2. `ds_properties` (name, when published)
--- the following are deleted when forms are purged deleted ---
4. `ds_property_fields` (has schema id which i thought would be helpful but turned out not to be. used to join with form fields but doesn't have useful info of its own)
5. `form_fields` (to get the order in the form)
6. `forms`, sometimes, when reporting what properties are linked to what forms

Also, dataset property can link to many ds_property_fields so it's hard to build the query of just the current, relevant dataset properties and get the grouping/aggregating of the fields correct.

PR #782, opened 2023-02-25
Speed up acquisition of enketoId in two cases

matthew-white:
See getodk/central#385.

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -16,7 +16,7 @@ const { getFormFields, merge, compare } = require('../../data/schema');
 const { getDataset } = require('../../data/dataset');
 const { generateToken } = require('../../util/crypto');
 const { unjoiner, extender, updater, equals, insert, insertMany, markDeleted, markUndeleted, QueryOptions } = require('../../util/db');
-const { resolve, reject } = require('../../util/promise');
+const { resolve, reject, timebound } = require('../../util/promise');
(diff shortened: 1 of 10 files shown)


matthew-white (commented):
Notes from interactive code review

lib/model/query/forms.js
```diff
@@ -117,6 +117,42 @@ createNew.audit = (form, partial, _, publish) => (log) =>
...
+  // generate a draft token and enketoId.
+  if (form.def.id == null || form.def.id !== form.draftDefId) {
+    const draftToken = generateToken();
+
+    // Try to push the draft to Enketo. If doing so fails or is too slow, then
+    // the worker will try again later.
+    const encodedId = encodeURIComponent(form.xmlFormId);
+    const path = `/v1/test/${draftToken}/projects/${form.projectId}/forms/${encodedId}/draft`;
+    const request = enketo.create(`${env.domain}${path}`, form.xmlFormId);
+    const enketoId = await timebound(request, 0.5).catch(() => null);
```
matthew-white, line 147: What if `Forms.createVersion()` fails for some reason? Then we will have pushed a form draft to Enketo that we will never use. Is that an issue?

ktuite (approved):

lib/model/query/forms.js
matthew-white: I asked about this in Slack. But I don't think it's a big issue.

Issue #590, opened 2022-08-30
REST GET .../submissions should return `instanceName` or update API docs


sadiqkhoja:
Currently REST endpoint `GET forms/id/submissions` doesn't return `instanceName`, we need to fix that or update API docs where we say that `instanceName` is only supported for ODATA endpoints

```
curl 'https://test.getodk.cloud/v1/projects/390/forms/Simple/Submissions' -H  ...

Response:
[
    {
        "instanceId": "uuid:08349870-0941-4a7c-8843-11fdd9a70908",
        "submitterId": 17150,
        "deviceId": null,
        "createdAt": "2022-08-30T15:34:22.162Z",
        "updatedAt": null,
        "reviewState": null
    }
]
```

REF: https://forum.getodk.org/t/odk-central-api-not-returning-instancename-of-submission/38729/6

matthew-white:
`instanceName` is also available using the REST API endpoints for a submission's versions or for a particular submission version. I think this relates to #481: the API distinguishes between submissions and submission versions (the two return different data), but the API docs don't always reflect that.

Issue #481, opened 2022-05-06
Improve documentation around submission versions and updating


ktuite:
Issue #417 was about the response for a PUT request to update a Submission not returning the fields that a full Submission is expected to have and what is described in the documentation.

We determined that what is returned makes sense because it includes the fields relevant to a specific Submission _version_ (internally called a Submission Def) and matches the response format for other Submission version endpoints.

The documentation needs to be changed to make this distinction between "Submission" and this lighter, more specific kind of submission "version" or "def" for the endpoints involving submission versions. 

- Updating Submission Data `[PUT /v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}]`
- Submission Versions `[/v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}/versions]`
- Getting Version Details `[GET /v1/projects/{projectId}/forms/{xmlFormId}/submissions/{instanceId}/versions/{versionId}]`

PR #783, opened 2023-02-25
Update Submissions response

### LINKED ISSUE #417: PUT .../submissions/:instanceId does not return expected fields

sadiqkhoja:
Closes #417, #481, #590

Updated integration tests

Detailed discussion on Slack

None

Updated.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -32,6 +32,11 @@ Finally, **system information and configuration** is available via a set of spec
 
 Here major and breaking changes to the API are listed by version.
 
+### ODK Central v2023.2
+
+**Changed**:
+- The response of `GET`, `POST`, `PUT` and `PATCH` methods of [Submissions](#reference/submissions/listing-all-submissions-on-a-form) endpoint has been updated to include metadata of the `currentVersion` of the Submission. 
+
 ### ODK Central v2023.1
 
 **Added**:
@@ -43,7 +48,6 @@ Here major and breaking changes to the API are listed by version.
 - [GET /projects/:id/datasets](#reference/datasets/datasets/datasets) now supports `X-Extended-Metadata` header to retrieve number of Entities in the Dataset and timestamp of the last Entity
 - `$select` in OData now supports selecting complex type(groups) 
 - [Creating a form](#reference/forms/forms/creating-a-new-form) can now return workflow warnings
-- The response of `GET`, `POST`, `PUT` and `PATCH` methods of [Submissions](#reference/submissions/listing-all-submissions-on-a-form) endpoint has been updated to include metadata of the `currentVersion` of the Submission. 
 
 **Removed**:
 
@@ -2106,7 +2110,7 @@ Like how `Form`s are addressed by their XML `formId`, individual `Submission`s a
 
 As of version 1.4, a `deviceId` and `userAgent` will also be returned with each submission. The client device may transmit these extra metadata when the data is submitted. If it does, those fields will be recognized and returned here for reference. Here, only the initial `deviceId` and `userAgent` will be reported. If you wish to see these metadata for any submission edits, including the most recent edit, you will need to [list the versions](/reference/submissions/submission-versions/listing-versions).
 
-As of version 2023.2, this API returns `currentVersion` that contains metadata of the most recent edit of the Submission.
+As of version 2023.2, this API returns `currentVersion` that contains metadata of the most recent version of the Submission.
 
 This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to return a `submitter` data object alongside the `submitterId` Actor ID reference.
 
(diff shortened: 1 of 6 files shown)


lib/model/query/submissions.js
```diff
@@ -244,10 +259,12 @@ const _exportUnjoiner = unjoiner(Submission, Submission.Def, Submission.Xml, Sub
 // the /only/ place we need to do this in the entire codebase right now. so for now
 // we just use the terrible hack.
 const { raw } = require('slonik-sql-tag-raw');
-const _exportFields = raw(_exportUnjoiner.fields.sql.replace(
-  'submission_defs."xml" as "submission_defs!xml"',
-  '(case when submission_defs."localKey" is null then submission_defs.xml end) as "submission_defs!xml"'
-));
+const _exportFields = raw(_exportUnjoiner.fields.sql
+  .replace(',submissions."userAgent" as "submissions!userAgent"', '')
```
sadiqkhoja, line 263: we don't need userAgent for OData

docs/api.md
```diff
@@ -43,6 +43,7 @@ Here major and breaking changes to the API are listed by version.
 - [GET /projects/:id/datasets](#reference/datasets/datasets/datasets) now supports `X-Extended-Metadata` header to retrieve number of Entities in the Dataset and timestamp of the last Entity
 - `$select` in OData now supports selecting complex type(groups) 
 - [Creating a form](#reference/forms/forms/creating-a-new-form) can now return workflow warnings
+- The response of `GET`, `POST`, `PUT` and `PATCH` methods of [Submissions](#reference/submissions/listing-all-submissions-on-a-form) endpoint has been updated to include metadata of the `currentVersion` of the Submission.
```
matthew-white, line 46: This is in the changelog for .1. What do you think about creating a new section for .2?

docs/api.md
```diff
@@ -2105,6 +2106,8 @@ Like how `Form`s are addressed by their XML `formId`, individual `Submission`s a
 
 As of version 1.4, a `deviceId` and `userAgent` will also be returned with each submission. The client device may transmit these extra metadata when the data is submitted. If it does, those fields will be recognized and returned here for reference. Here, only the initial `deviceId` and `userAgent` will be reported. If you wish to see these metadata for any submission edits, including the most recent edit, you will need to [list the versions](/reference/submissions/submission-versions/listing-versions).
 
+As of version 2023.2, this API returns `currentVersion` that contains metadata of the most recent edit of the Submission.
```
matthew-white, line 2109: How about "most recent version" instead of "most recent edit"? The API will return `currentVersion` even if there hasn't been an edit.

docs/api.md
```diff
@@ -4237,17 +4245,28 @@ These are in alphabetic order, with the exception that the `Extended` versions o
...
++ submitterId: `23` (number, required) - The ID of the `Actor` (`App User`, `User`, or `Public Link`) that originally submitted this `Submission`.
++ deviceId: `imei:123456` (string, optional) - The self-identified `deviceId` of the device that collected the data, sent by it upon submission to the server. The initial submission `deviceId` will be returned here. 
++ userAgent: `Enketo/3.0.4` (string, optional) - The self-identified `userAgent` of the device that collected the data, sent by it upon submission to the server. The initial submission `userAgent` will be returned here. 
 + reviewState: `approved` (Submission Review State, optional) - The current review state of the submission.
 + createdAt: `2018-01-19T23:58:03.395Z` (string, required) - ISO date format. The time that the server received the Submission.
 + updatedAt: `2018-03-21T12:45:02.312Z` (string, optional) - ISO date format. `null` when the Submission is first created, then updated when the Submission's XML data or metadata is updated.
++ currentVersion: (SubmissionVersion, optional) - The current version of the `Submission`
```
matthew-white, line 4254: Period at the end.

lib/model/frames/submission.js
```diff
@@ -24,7 +24,9 @@ class Submission extends Frame.define(
   'instanceId', readable,               'submitterId',  readable,
   'deviceId',   readable,               'createdAt',    readable,
   'updatedAt',  readable,               'reviewState',  readable, writable,
-  embedded('submitter')
+  'userAgent',  readable,
```
matthew-white, line 27: Since we didn't return `userAgent` on the logical submission before, could you add a test that verifies that this field is now returned? And that it's specifically the `userAgent` of the root version?

matthew-white (commented):
This is looking like a nice set of improvements to the API and the docs!

lib/model/query/submissions.js
```diff
@@ -134,20 +153,34 @@ order by path asc, value asc`)
...
-join projects on projects.id=${projectId} and projects.id=forms."projectId"
-${extend|| sql`left outer join actors on actors.id=submissions."submitterId"`}
-where ${equals(options.condition)} and draft=${draft} and submissions."deletedAt" is null
-order by submissions."createdAt" desc, submissions.id desc
-${page(options)}`);
+// Helper function to assign submission.currentVersionSubmitter to submission.currentVersion.submitter
+// Current there is no way to create such complex object using `extender` and `unjoiner` framework functions
+const assignCurrentVersionSubmitter = (x) => x.withAux('currentVersion', x.aux.currentVersion.withAux('submitter', x.aux.currentVersionSubmitter));
+
+const _get = extender(Submission, Submission.Def.into('currentVersion'))(Actor.into('submitter'), Actor.alias('current_version_actors', 'currentVersionSubmitter'))((fields, extend, options, projectId, xmlFormId, draft) => sql`
```
sadiqkhoja, line 160: I might be wrong here, do you know if there is a way to use `extender` to set `currentVersion.submitter`?

matthew-white: I'm not sure actually! I can't think of a similar example of nested extended metadata, so maybe not.

matthew-white (approved):
I'll take a closer look tomorrow, but this is looking good to me. I think we should go ahead and merge so that QA can start on regression testing.

PR #787, opened 2023-03-02
Enhance/dataset assertions

### LINKED ISSUE #768: Add should.be.a.Dataset test assertion
Add `Dataset` (and `ExtendedDataset`) assertion to `test/assertions.js` and use throughout tests.

sadiqkhoja:
Closes #768 

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -36,6 +36,7 @@ describe('datasets and entities', () => {
               asAlice.get('/v1/projects/1/datasets')
                 .expect(200)
                 .then(({ body }) => {
+                  body[0].should.be.a.Dataset();
                   body.map(({ createdAt, ...d }) => d).should.eql([
                     { name: 'people', projectId: 1 }
                   ]);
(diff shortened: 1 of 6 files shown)


test/integration/api/datasets.js
```diff
@@ -64,6 +65,7 @@ describe('datasets and entities', () => {
           .set('X-Extended-Metadata', 'true')
           .expect(200)
           .then(({ body }) => {
+            body[0].should.be.an.ExtendedDataset();
             body.map(({ createdAt, lastEntity, ...d }) => {
               createdAt.should.not.be.null();
               lastEntity.should.not.be.null();
```
ktuite, line 71: We don't really need these lines (the lines in the block following this line) if we have the ExtendedDataset assertion

test/integration/api/datasets.js
```diff
@@ -109,6 +112,7 @@ describe('datasets and entities', () => {
           .set('X-Extended-Metadata', 'true')
           .expect(200)
           .then(({ body }) => {
+            body[0].should.be.an.ExtendedDataset();
             body.map(({ createdAt, lastEntity, ...d }) => {
               createdAt.should.not.be.null();
               should(lastEntity).be.null();
```
ktuite, line 118: I don't think we need the `createdAt.should.not.be.null();` line but the last entity being null line can stay.

test/integration/api/datasets.js
```diff
@@ -188,6 +193,7 @@ describe('datasets and entities', () => {
           .set('X-Extended-Metadata', 'true')
           .expect(200)
           .then(({ body }) => {
+            body[0].should.be.an.ExtendedDataset();
```
ktuite, line 196: get rid of lines below
```
createdAt.should.not.be.null();
lastEntity.should.not.be.null();
```

ktuite (approved):
Looks good. I like that the `identity` part of logging in goes away in a lot of places!

Issue #793, opened 2023-03-20
Refactor Datasets.createOrMerge and Datasets.publishIfExists


ktuite:
This is used by `Forms.createNew` and `Forms.createVersion` and intersects with `Datasets.publishIfExists` (called by Forms.publish`).

It creates these audit events: `dataset.create`, `dataset.update`, and/or `dataset.update.publish`.

Changes:
1. Make inputs cleaner. Move work from Forms query module (to prepare Dataset object, filter form fields) into this function. 
2. Make it async/await!

There was more to this issue but it has been moved to a separate issue that can come later: https://github.com/getodk/central-backend/issues/811

PR #807, opened 2023-03-23
Rename getter functions in Dataset query module

### LINKED ISSUE #691: Rename getter functions in Dataset query module
Multiple functions in this file have very similar behaviors. See if there's a way to reorganize so that there is less duplication of behavior and more explanatory names. 

```js
module.exports = {
  createOrMerge, publishIfExists,
  getById, getAllByProjectId, getByName,
  getFieldsByFormDefId, getPublishedProperties,
  getDatasetDiff, getByProjectAndName,
  getDatasetMetadata
};
```

* `getById` is used in places where we have the dataset id and need to do something with the properties. 
* `getPublishedProperties` is only used internally by another function
    * but the places we use `getById` to get the properties, we usually only want the published properties there 
* `getFieldsByFormDefId` used in submission parsing to create entity. Similar to looking up relevant  
 
* `getByName` and `getByProjectAndName` used in similar situations (both take name and project id)
    * sometimes we have name and project id and just need to do a simple query for an auth check
    * sometimes we have name and project id but also need to get the properties

* `getDatasetMetadata` gets the properties but also links in the forms

* `getDatasetDiff` pretty specialized...

See #576 for some specific comments. For example:

- https://github.com/getodk/central-backend/pull/576#discussion_r1008419220 (about getByName and getByProjectAndName)
- https://github.com/getodk/central-backend/pull/576#discussion_r1008414409 (about dataset actee provisioning)

-----

How these functions seem to be categorized:
1. Get properties directly to use for exporting or parsing
2. Get a dataset by name and project id from a resource
    * maybe just an auth check
    * maybe the properties are needed
    * maybe even more details about the linked forms are needed
3. Getting the diff of a dataset seems like its own special thing
4. Creating / updating / publishing (_setting_) dataset information is its own thing

ktuite:
Closes #691 

The diff looks like a mess because I changed a bunch and moved things around. Here's the new organization:

#### Creating/updating

`Datasets.createOrMerge` and `Datasets.publishIfExists` used by the Forms query module.
This is mostly the same as before but there's an issue #793 for refactoring.

#### Simplified Dataset getters

Get all datasets for a project or get a single dataset. These return minimal information about Datasets -- just the metadata, really, or the "extended" metadata consisting of entity counts and last entity timestamp. They can be used for auth checks and listing basic information about Datasets on frontend. More complicated stuff about Datasets (including properties) is accessed below. 

One observation here is that these are called by a resources where the project id and dataset name is known. By having a simple call here to get and check the metadata, we can separate out the more complicated dataset queries after we know we have access to the dataset.

#### Complex Dataset getters

Another observation is  that certain things like getting the right list of dataset properties in order can be really tricky, so making these more modular will hopefully help with reusability, understandability, and maintainability, even if it means 2 or 3 queries instead of 1. I don't think any of these are too performance-critical.

**_Metadata_** 

Properties and lists of forms that produce and consume datasets.

**_Properties_**

This is probably the most important thing about any given dataset, and the most complicated to compute. I'm trying to have all the property SQL focused here so it doesn't have to be duplicated.

**_Diffs_**

Conveying how a form changes a dataset. This one is called based on the project and xmlformid so it doesn't really intersect with any of the other functions, except it also returns information about properties. 

Mostly the same tests as before.

I hope it's going to be easier to maintain and use.

Doesn't impact users.

No

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -10,8 +10,7 @@
 const { sql } = require('slonik');
 const { extender, QueryOptions, equals } = require('../../util/db');
 const { Dataset, Form } = require('../frames');
-const { isEmpty, isNil, either, reduceBy, groupBy } = require('ramda');
-const R = require('ramda');
+const { isEmpty, isNil, either, reduceBy, groupBy, uniqWith, equals: rEquals } = require('ramda');
 
 const Problem = require('../../util/problem');
 const { construct } = require('../../util/util');
@@ -216,9 +215,9 @@ const getByProjectAndName = (projectId, name, publishedOnly = false, extended = 
 
 // Gets the dataset information, properties (including which forms each property comes from),
 // and which forms consume the dataset via CSV attachment.
-const getDatasetMetadata = (datasetName, projectId) => async ({ all, Datasets }) => {
+const getMetadata = (dataset) => async ({ all, Datasets }) => {
 
-  const _getLinkedForms = () => sql`
+  const _getLinkedForms = (datasetName, projectId) => sql`
     SELECT DISTINCT f."xmlFormId", coalesce(fd.name, f."xmlFormId") "name" FROM form_attachments fa
     JOIN forms f ON f.id = fa."formId"
     JOIN form_defs fd ON f."currentDefId" = fd.id
@@ -228,9 +227,8 @@ const getDatasetMetadata = (datasetName, projectId) => async ({ all, Datasets })
       AND ds."publishedAt" IS NOT NULL
   `;
 
-  const dataset = await Datasets.getByProjectAndName(projectId, datasetName).then((o) => o.get());
   const properties = await Datasets.getPublishedProperties(dataset.id, true);
-  const linkedForms = await all(_getLinkedForms(datasetName, projectId));
+  const linkedForms = await all(_getLinkedForms(dataset.name, dataset.projectId));
 
   const propertiesMap = new Map();
   for (const property of properties) {
@@ -290,7 +288,7 @@ const getPublishedProperties = (datasetId, includeForms = false) => ({ all }) =>
 `)
   .then((rows) => (includeForms
     ? rows.map(({ formName, xmlFormId, ...row }) => new Dataset.Property(row, { formName, xmlFormId }))
-    : R.uniqWith(R.equals, rows).map(construct(Dataset.Property))));
+    : uniqWith(rEquals, rows).map(construct(Dataset.Property))));
 
 // Returns list of Fields augmented by dataset property name.
 // This it used by the submission XML parser to create an entity from a submission.
@@ -357,7 +355,7 @@ const getDatasetDiff = (projectId, xmlFormId, forDraft) => ({ all }) => all(sql`
 module.exports = {
   createOrMerge, publishIfExists,
   getAllByProjectId, getByProjectAndName,
-  getDatasetMetadata,
+  getMetadata,
   getPublishedProperties, getFieldsByFormDefId,
   getDatasetDiff
 };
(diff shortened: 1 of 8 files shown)


lib/model/query/datasets.js
```diff
@@ -206,77 +123,177 @@ createOrMerge.audit = (dataset, _, fields, publish) => (log) =>
...
+  LEFT JOIN form_defs ON
+      forms."currentDefId" = form_defs.id
+  ` : sql``}
+  WHERE ds_properties."datasetId" = ${datasetId}
+    AND ds_properties."publishedAt" IS NOT NULL
+  ORDER BY ds_properties."publishedAt", form_fields.order, ds_properties.id
+`)
+  .then((rows) => (includeForms
+    ? rows.map(({ formName, xmlFormId, ...row }) => new Dataset.Property(row, { formName, xmlFormId }))
+    : R.uniqWith(R.equals, rows).map(construct(Dataset.Property))));
```
ktuite, line 293: alias `R.equals` to use w/o fully importing ramda

lib/resources/datasets.js
```diff
@@ -20,21 +20,22 @@ module.exports = (service, endpoint) => {
...
-  service.get('/projects/:projectId/datasets/:name', endpoint(({ Datasets, Projects }, { params, auth }) =>
-    Projects.getById(params.projectId)
+  service.get('/projects/:projectId/datasets/:name', endpoint(({ Datasets }, { params, auth }) =>
+    Datasets.getByProjectAndName(params.projectId, params.name)
       .then(getOrNotFound)
-      .then((project) => auth.canOrReject('dataset.read', project))
-      .then(() => Datasets.getDatasetMetadata(params.name, params.projectId)
-        .then(getOrNotFound))));
+      .then((dataset) => auth.canOrReject('dataset.read', dataset))
+      .then(() => Datasets.getDatasetMetadata(params.name, params.projectId))));
```
ktuite, line 27: `getDatasetMetadata` contains another call to `getByProjectAndName` that we can find a way to optimize out - can pass the dataset frame directly to this function. 

Another option:
Data file for dataset can be responsible for compositing things together: dataset, properties, linked forms. 
Query module can just be response for sql to fetch this info. 
Resource here can also stay simple.

lib/model/query/datasets.js
```diff
@@ -330,69 +353,11 @@ const getDatasetDiff = (projectId, xmlFormId, forDraft) => ({ all }) => all(sql`
...
-
-publishIfExists.audit.withResult = true;
 
 module.exports = {
   createOrMerge, publishIfExists,
-  getById, getAllByProjectId, getByName,
-  getFieldsByFormDefId, getPublishedProperties,
-  getDatasetDiff, getByProjectAndName,
-  getDatasetMetadata
+  getAllByProjectId, getByProjectAndName,
```
ktuite, line 359: simplify these names
getAllByProjectId -> getList
getByProjectAndName -> get
getPublishedProperties -> getProperties 
getDatasetMetadata -> getMetadata
getFieldsByFormDefId -> getFields???
getDatasetDiff -> getDiff

test/integration/api/datasets.js
```diff
@@ -1580,32 +1580,6 @@ describe('datasets and entities', () => {
...
-                }))));
-
-        // Get all datasets by projectId
-        const datasetId = await Datasets.getAllByProjectId(1)
-          .then(result => result[0].id);
-
-        await Datasets.getById(datasetId)
-          .then(result => {
-            result.properties.length.should.be.eql(2);
-          });
```
ktuite, line 1606: bring back test but remove these calls.

ktuite (commented):
Interactive review with Sadiq

sadiqkhoja (approved):
changes look good

PR #808, opened 2023-03-24
Add dataset.read and entity.read verbs.

### LINKED ISSUE #794: Check dataset/entity verbs (backend)
We just have `dataset.list` and `entity.list`. I think we need `dataset.read` and `entity.read` too. See related frontend issue https://github.com/getodk/central-frontend/issues/662.

ktuite:
Closes #794 

Adds `dataset.read` and `entity.read` verbs for reading individual datasets and (eventually) entities to the list of verbs for Admins, Managers, and Project Viewers.

Also adds `entity.create` and `entity.update` verbs for Admins and Managers to be ready for entity creating/updating!

Tests.

Something we've needed for a little while.

No effect. In some places, `dataset.list` was used in place of `dataset.read` but the same roles have the same verbs so it doesn't really change anything.

No, verbs aren't individually listed in the docs.

### DIFF
--- a/lib/model/migrations/20230324-01-edit-dataset-verbs.js
+++ b/lib/model/migrations/20230324-01-edit-dataset-verbs.js
@@ -7,27 +7,25 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-const { without } = require('ramda');
+const up = (db) => db.raw(`
+UPDATE roles
+SET verbs = verbs || '["dataset.read", "entity.read"]'::jsonb
+WHERE system in ('admin', 'manager', 'viewer')
+`).then(() => db.raw(`
+UPDATE roles
+SET verbs = verbs || '["entity.create", "entity.update"]'::jsonb
+WHERE system in ('admin', 'manager')
+`));
 
-/* eslint-disable no-await-in-loop */
-
-const up = async (db) => {
-  // grant rights.
-  for (const system of [ 'admin', 'manager', 'viewer' ]) {
-    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
-    const newVerbs = verbs.concat([ 'dataset.read', 'entity.read' ]);
-    await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
-  }
-};
-
-const down = async (db) => {
-  // revoke rights.
-  for (const system of [ 'admin', 'manager', 'viewer' ]) {
-    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
-    const newVerbs = without([ 'dataset.read', 'entity.read' ], verbs);
-    await db.update({ verbs: JSON.stringify(newVerbs) }).into('roles').where({ system });
-  }
-};
+const down = (db) => db.raw(`
+UPDATE roles
+SET verbs = (verbs - 'dataset.read' - 'entity.read')
+WHERE system in ('admin', 'manager', 'viewer')
+`).then(() => db.raw(`
+UPDATE roles
+SET verbs = (verbs - 'entity.create' - 'entity.update')
+WHERE system in ('admin', 'manager')
+`));
 
 module.exports = { up, down };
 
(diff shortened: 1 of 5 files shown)


lib/model/migrations/20230324-01-edit-dataset-verbs.js
matthew-white, line 1: I don't think anything needs to change here, but thinking about #744, I think we should try to write migrations like this using raw SQL going forward. We even have a verbs-related migration that uses raw SQL that we can draw from: [20210721-01-add-config-set-verb.js](https://github.com/getodk/central-backend/blob/master/lib/model/migrations/20210721-01-add-config-set-verb.js).

lib/resources/forms.js
```diff
@@ -161,7 +161,7 @@ module.exports = (service, endpoint) => {
       .then(([form, project]) =>
         Promise.all([
           auth.canOrReject('form.update', form),
-          auth.canOrReject('dataset.list', project)
+          auth.canOrReject('dataset.read', project)
```
matthew-white, line 164: Right now the diff will only include a single dataset, right? In the future, might it include multiple datasets? If so, `dataset.list` actually feels more right to me than `dataset.read`. I think of `dataset.read` as being about an individual dataset.

lib/resources/forms.js
```diff
@@ -181,7 +181,7 @@ module.exports = (service, endpoint) => {
...
```
matthew-white, line 184: Same question here about whether the diff might include multiple datasets in the future.

lib/model/migrations/20230324-01-edit-dataset-verbs.js
```diff
@@ -0,0 +1,33 @@
...
+
+const { without } = require('ramda');
+
+/* eslint-disable no-await-in-loop */
+
+const up = async (db) => {
+  // grant rights.
+  for (const system of [ 'admin', 'manager', 'viewer' ]) {
+    const [{ verbs }] = await db.select('verbs').from('roles').where({ system });
+    const newVerbs = verbs.concat([ 'dataset.read', 'entity.read' ]);
```
matthew-white, line 18: Do you want to throw in `entity.create` and `entity.update` while you're here? Maybe we'd need one other one for updating the properties of an entity, though maybe not. Looking at the existing submission verbs, I only see list/read/create/update.

test/integration/api/projects.js
```diff
@@ -69,7 +69,7 @@ describe('api: /projects', () => {
               .set('X-Extended-Metadata', 'true')
               .expect(200)
               .then(({ body }) => {
-                body.verbs.length.should.be.greaterThan(39);
+                body.verbs.length.should.be.greaterThan(43);
```
matthew-white, line 72: Should be 41?

matthew-white (approved):

lib/model/migrations/20230324-01-edit-dataset-verbs.js
ktuite: I can! Just need to figure out what roles does `entity.create`/`entity.update` apply to? Probably not `viewer`. Does that just leave `admin` + `manager`?

test/integration/api/projects.js
ktuite: This is such an awkward test because it doesn't change when new verbs are added. I think the number was not incremented last time? I ran it and it was expecting 44 verbs.

lib/resources/forms.js
ktuite: I wasn't totally sure what to put here. Good point that there could be multiple datasets!

test/integration/api/projects.js
matthew-white: Oh very interesting! Would it be worth changing it to an equality check rather than `greaterThan` so that we remember  to update it more frequently?

lib/model/migrations/20230324-01-edit-dataset-verbs.js
matthew-white: I agree, seems like it would just be `admin` + `manager` for now. I guess that would maybe add complications since the migration also deals with viewers. Those verbs could also come in a separate PR, whatever's more convenient!

lib/model/migrations/20230324-01-edit-dataset-verbs.js
ktuite: I've gone ahead and done this and added a migration test just to be sure the raw sql worked.

lib/model/migrations/20230324-01-edit-dataset-verbs.js
ktuite: Doing it here so they can be used soon, even though there's nothing that really tests these new verbs yet!

PR #810, opened 2023-03-28
Move prep work from Forms to Datasets.createOrMerge

### LINKED ISSUE #793: Refactor Datasets.createOrMerge and Datasets.publishIfExists

ktuite:
Closes #793

Changes the interface between Forms.createNew and Forms.createVersion that updates Datasets. Now `Datasets.createOrMerge` takes raw arguments from Forms and builds and filters what it needs itself. 

Tests have remained the same.

It was a little awkward making the Forms query module filter out the form fields and build a partial Dataset object to pass the name and project id through. Now the code in the Forms module is simpler. If more changes need to happen on Dataset creation, they can happen within the Dataset query module. The function in the query module is also now async/await and a bit easier to follow.

No changes to users.

No changes to docs.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -89,39 +89,50 @@ SELECT "action" FROM ds
 `;
 
 // Creates or merges dataset, properties and field mapping.
-// Expects dataset:Frame auxed with `formDefId`, `schemaId`,
-// and array of field:Frame auxed with `propertyName`
-const createOrMerge = (dataset, fields, publish = false) => ({ one, Actees, Datasets, Projects }) =>
-  Promise.all([
-    Projects.getById(dataset.projectId).then((o) => o.get()),
-    Datasets.get(dataset.projectId, dataset.name)
-  ])
-    .then(([project, datasetOption]) =>
-      (datasetOption.isDefined()
-        ? datasetOption.get().acteeId
-        : Actees.provision('dataset', project).then((actee) => (actee.id))))
-    .then((acteeId) =>
-      one(_createOrMerge(dataset, fields, acteeId, publish))
-        .catch(error => {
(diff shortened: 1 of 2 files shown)


lib/model/query/datasets.js
```diff
@@ -89,40 +89,51 @@ SELECT "action" FROM ds
...
+  const result = await one(_createOrMerge(partial, dsPropertyFields, acteeId, publish))
+    .catch(error => {
+      if (error.constraint === 'ds_property_fields_dspropertyid_formdefid_unique') {
+        throw Problem.user.invalidEntityForm({ reason: 'Multiple Form Fields cannot be saved to a single Dataset Property.' });
+      }
+      throw error;
+    });
+
+  // Fetch dataset again with properties in order to write audit log
+  const ds = await Datasets.get(project.id, name).then((o) => o.get());
```
ktuite, line 119: Thinking about ways to remove extra queries, especially based on discussion from last review. There's probably a way to not refetch this information using `existingDataset` and/or `partial` above.

ktuite: Things needed by the audit function.. just construct an object with these things in it instead of refetching dataset.
- Dataset with actee id
- fields (filtered)
- properties (re-fetch from database)

ktuite (commented):

PR #812, opened 2023-03-31
Entity APIs specification

sadiqkhoja:
Updated API docs with the high level API specs

Documentation can be viewed at : https://centralbackendsk.docs.apiary.io/#

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -35,7 +35,15 @@ Here major and breaking changes to the API are listed by version.
 **Added**:
 
 - New endpoint [PATCH /projects/:id/datasets/:name](#reference/datasets/datasets/update-dataset-metadata) to change whether approval of Submission is required to create an Entity.
-- TODO: list new entity endpoints
+- New [Entities](#reference/entities) endpoints for richer Entity-Based Data Collection workflows. These endpoints provide ways of accessing Entities, as well as the ability to _create_, _update_ and _soft-delete_ Entities via the API!
+  * New endpoint [GET /projects/:id/datasets/:name/entities](#reference/entities/retrieving-entities/entities-metadata) for listing Entities within a Dataset.
+  * New endpoint [GET /projects/:id/datasets/:name/entities/:uuid](#reference/entities/retrieving-entities/getting-entity-details) for getting the metadata, or details, about a specific Entity.
+  * New endpoint [GET /projects/:id/datasets/:name/entities/:uuid/versions](#reference/entities/retrieving-entities/listing-versions) for listing the versions of an Entity.
+  * New endpoint [GET /projects/:id/datasets/:name/entities/:uuid/diffs](#reference/entities/retrieving-entities/getting-changes-between-versions) for getting the changse between versions of an Entity.
+  * New endpoint [GET /projects/:id/datasets/:name/entities/:uuid/audits](#reference/entities/retrieving-entities/entity-audit-log) for getting the changse between versions of an Entity.
+  * New endpoint [POST /projects/:id/datasets/:name/entities](#reference/entities/creating-an-entity/creating-an-entity) for creating an Entity.
+  * New endpoint [PATCH /projects/:id/datasets/:name/entities/:uuid](#reference/entities/updating-an-entity/updating-an-entity) for updating the data or label of an Entity.
+  * New endpoint [DELETE /projects/:id/datasets/:name/entities/:uuid](#reference/entities/deleting-an-entity/deleting-an-entity) for soft-deleting an Entity.
 
 **Changed**:
 
@@ -3140,6 +3148,28 @@ Returns the metadata and current data of an Entity
     This is the extended response
 
     + Attributes (Extended Entity)
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
+### Listing Versions [GET /projects/{projectId}/datasets/{name}/entities/{uuid}/versions]
+
+This returns the Entity metadata and data for every version of this Entity, in ascending creation order.
+
+This endpoint supports retrieving extended metadata; provide a header `X-Extended-Metadata: true` to return a `creator` data object alongside the `creatorId` Actor ID reference.
+
++ Parameters
+    + projectId: `16` (number, required) - The numeric ID of the Project
+    + name: `people` (string, required) - Name of the Dataset
+    + uuid: `54a405a0-53ce-4748-9788-d23a30cc3afa` (string, required) - UUID of the Entity
+
++ Response 200 (application/json)
+    + Attributes (array[Entity Version])
+
++ Response 200 (application/json; extended)
+    This is the extended response
+
+    + Attributes (array[Extended Entity Version])
 
 + Response 403 (application/json)
     + Attributes (Error 403)
--- a/docs/api.md
+++ b/docs/api.md
@@ -3091,6 +3091,218 @@ name,label,first_name,last_name,age,favorite_color
 + Response 403 (application/json)
     + Attributes (Error 403)
 
+
+# Group Entities
+
+An Entity is a data row of the dataset. Following API endpoints will let you perform CRUD (create, read, update and delete) operations on the entities.
+
+## Retrieving Entities [/projects/{projectId}/datasets/{datasetName}/entities]
+
+### Entities Metadata [GET]
+This endpoint returns list of entities of a Dataset. Please note that this endpoint only returns metadata of the entities not the data. If you want to get the data of all entities then please refer to [OData Dataset Service](#reference/odata-endpoints/odata-form-service)
+
+You can provide `?deleted=true` to get only deleted entities 
+
++ Parameters
+    + projectId: `16` (number, required) - The numeric ID of the Project
+    + name: `people` (string, required) - Name of the Dataset
+
++ Response 200 (application/json)
+    This is the standard response
+
+    + Attributes (array[Entity Summary])
+
++ Response 200 (application/json; extended)
+    This is the extended response
+
+    + Attributes (array[Extended Entity Summary])
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
+### Getting Entity Details  [GET /projects/{projectId}/datasets/{name}/entities/{uuid}]
+
+Returns the metadata and current data of an Entity
+
++ Parameters
+    + projectId: `16` (number, required) - The numeric ID of the Project
+    + name: `people` (string, required) - Name of the Dataset
+    + uuid: `54a405a0-53ce-4748-9788-d23a30cc3afa` (string, required) - UUID of the Entity
+
++ Response 200 (application/json)
+    + Attributes (Entity)
+
++ Response 200 (application/json; extended)
+    This is the extended response
+
+    + Attributes (Extended Entity)
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
+### Getting changes between Versions [GET /projects/{projectId}/datasets/{name}/entities/{uuid}/diffs]
+
+This returns the changes, or edits, between different versions of an Entity. These changes are returned as an array of arrays. Between two Entities, there is an array of objects representing how each property changed. This change object contains the old and new values, as well as the property name.
+
++ Parameters
+    + projectId: `16` (number, required) - The numeric ID of the Project
+    + name: `people` (string, required) - Name of the Dataset
+    + uuid: `54a405a0-53ce-4748-9788-d23a30cc3afa` (string, required) - UUID of the Entity
+
++ Response 200 (application/json)
+    + Response 200
+    + Attributes (array[array[Entity Diff Value]])
+
+    + Body
+
+            [
+              [
+                {
+                  "new": "John",
+                  "old": "Dana",
+                  "propertyName": "firstName"
+                },
+                {
+                  "new": "Doe",
+                  "old": "Roe",
+                  "propertyName": "lastName"
+                },
+                {
+                  "new": "John Doe",
+                  "old": "Jane Roe",
+                  "propertyName": "label"
+                }
+              ],
+              [
+                {
+                  "new": "Robert",
+                  "old": "Doe",
+                  "propertyName": "firstName"
+                },
+                {
+                  "new": "Robert Doe",
+                  "old": "Doe Doe",
+                  "propertyName": "label"
+                }
+              ]
+            ]
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
+### Entity Audit Log [GET /projects/{projectId}/datasets/{name}/entities/{uuid}/audits]
+
+Returns [Server Audit Logs](/reference/system-endpoints/server-audit-logs) relating to an Entity. They will be returned most recent first. 
+
++ Parameters
+    + projectId: `16` (number, required) - The numeric ID of the Project
+    + name: `people` (string, required) - Name of the Dataset
+    + uuid: `54a405a0-53ce-4748-9788-d23a30cc3afa` (string, required) - UUID of the Entity
+
++ Response 200 (application/json)
+    + Attributes (array[Audit])
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
+## Creating an Entity [/projects/{projectId}/datasets/{name}/entities]
+
+### Creating an Entity [POST]
+Creates an Entity in the Dataset. Request body takes the json representation of the entity. It should have `uuid` and `label` property in addition to the user-defined properties of the Dataset in `data` property. For e.g.
+```
+{
+    "uuid": "54a405a0-53ce-4748-9788-d23a30cc3afa",
+    "label": "John Doe",
+    "data": {
+        "firstName": "John",
+        "lastName": "Doe",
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -4552,4 +4814,50 @@ These are in alphabetic order, with the exception that the `Extended` versions o
...
++ SUBMISSION (string) - Entity was created through a Form Submission.
++ API (string) - Entity was created using the API.
++ FILE_UPLOAD (string) - Entity was created through uploading a file.
+
+## EntitySource (object)
++ type: `API` (EntitySourceType, required) - From what type of source this Entity was created.
++ id: `uuid:85cb9aff-005e-4edd-9739-dc9c1a829c44` (string, required) - Submission ID if the type is Submission otherwise value of `X-Client-Id` provided at the time of Entity creation.
+
+## EntityVersionFields (object)
++ versionNumber: `1` (number, required) - Numeric number to identify the version (auto incrementing)
```
sadiqkhoja, line 4838: Alternative: sha1 of the data

docs/api.md
```diff
@@ -3040,6 +3040,270 @@ name,label,first_name,last_name,age,favorite_color
...
+    "uuid": "54a405a0-53ce-4748-9788-d23a30cc3afa",
+    "label": "John Doe",
+    "firstName": "John",
+    "lastName": "Doe",
+    "city": "New York",
+    ...
+}
+```
+
+You can provide header `X-Client-Id` to identify the client/transaction of the consumer of the API. The value is logged and can be retrieved from the [Getting Entity Details](#reference/entities/getting-entity-details) and [Listing Entity Versions](#reference/entities/listing-entity-versions) APIs
```
sadiqkhoja, line 3193: returned in `currentVersion.source.id`

sadiqkhoja:
@matthew-white you can use this branch to begin frontend work

docs/api.md
matthew-white, line 3193: Is this something we want to support long-term? If we don't need to include it in .3, then we could wait and see how the entity space continues to develop. If we add API token actors at some point, then maybe this header wouldn't be needed anymore.

matthew-white (commented):
Thanks, @sadiqkhoja, I'm excited to use this on the frontend! I took a quick look through the API docs and wanted to leave a few questions/thoughts. (Also noting a few typos that I randomly noticed.)

docs/api.md
matthew-white: I like the idea of an ID like `versionNumber` rather than a SHA! Conceivably someone could change the entity data, then change it back again, so a SHA of the data wouldn't be a unique identifier.

test/assertions.js
```diff
@@ -363,3 +338,88 @@ should.Assertion.add('ExtendedDataset', function assertExtendedDataset() {
...
+});
+
+should.Assertion.add('Source', function() {
+  this.params = { operator: 'to be a Source' };
+
+  this.obj.should.have.property('type').which.is.a.String();
+  this.obj.should.have.property('id').which.is.a.String();
+});
+
+should.Assertion.add('EntityDefSummaryFields', function() {
```
ktuite, line 359: Was this `EntityDefSummaryFields` assertion made to be reused as an intermediate structure in the other assertions? Could you do more of a cascading assertion object approach instead? Using `this.obj.should.be.a.ParentAssertion()`? It might be shorter, easier to reason about, and easier to toggle certain properties on/off during preliminary development.

```
should.Assertion.add('Entity', function() {
  this.params = { operator: 'to be an Entity' };

  // root property / summary fields
  this.obj.should.have.property('uuid').which.is.a.String();
  this.obj.should.have.property('datasetName').which.is.a.String();
  this.obj.should.have.property('createdAt').which.is.a.isoDate();
  this.obj.should.have.property('updatedAt').which.is.nullOrIsoDate();
  this.obj.should.have.property('deletedAt').which.is.nullOrIsoDate();

  // current version
  this.obj.should.have.property('currentVersion').which.is.an.EntityDef();
});

should.Assertion.add('ExtendedEntity', function() {
  this.params = { operator: 'to be a Extended Entity' };

  this.obj.should.be.an.Entity();
  this.obj.creator.should.be.an.Actor();
});
```

(and a similar thing for the entityDef)

docs/api.md
```diff
@@ -4552,4 +4816,50 @@ These are in alphabetic order, with the exception that the `Extended` versions o
...
++ deletedAt: `2018-03-21T12:45:02.312Z` (string, optional) - Timestamp of the deletion in ISO date format. `null` if the Entity is not deleted.
+
+## EntitySourceType (enum)
++ SUBMISSION (string) - Entity was created through a Form Submission.
++ API (string) - Entity was created using the API.
++ FILE_UPLOAD (string) - Entity was created through uploading a file.
+
+## EntitySource (object)
++ type: `API` (EntitySourceType, required) - From what type of source this Entity was created.
++ id: `uuid:85cb9aff-005e-4edd-9739-dc9c1a829c44` (string, required) - `FormXmlId`/`SubmissionID` if the type is Submission otherwise value of `X-Client-Id` provided at the time of Entity creation.
```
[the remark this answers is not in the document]
matthew-white: Oh and it looks like the release criteria call for the submission instance name to be shown. Is it possible for the source here to include that as well?

docs/api.md
```diff
@@ -4552,4 +4816,50 @@ These are in alphabetic order, with the exception that the `Extended` versions o
...
-+ forms: (array[Form KeyValue]) - List of forms that create the property
\ No newline at end of file
++ forms: (array[Form KeyValue]) - List of forms that create the property
+
+## EntitySummaryFields (object)
++ uuid: `uuid:85cb9aff-005e-4edd-9739-dc9c1a829c44` (string, required) - The `uuid` of the Entity that uniquely identifies the Entity.
++ datasetName: `people` (string, required) - Name of the dataset
++ createdAt: `2018-01-19T23:58:03.395Z` (string, required) - ISO date format. The time that the server received the Entity.
++ updatedAt: `2018-03-21T12:45:02.312Z` (string, optional) - Timestamp of the last update in ISO date format. `null` when there is only one version of the Entity.
++ deletedAt: `2018-03-21T12:45:02.312Z` (string, optional) - Timestamp of the deletion in ISO date format. `null` if the Entity is not deleted.
```
matthew-white, line 4826: Could we include `creatorId` here as well? Right now, the individual entity endpoint will return the creator of the `currentVersion`, but it'd be useful if it could also return the original creator of the entity itself.

docs/api.md
```diff
@@ -4552,4 +4816,50 @@ These are in alphabetic order, with the exception that the `Extended` versions o
 ## Property Detailed (object)
 + name: `first_name` (string, required) - The name of the Property.
 + publishedAt: `2018-01-21T00:04:11.153Z` (string, required) - Publishing timestamp of the form that defined this property for the first time.
-+ forms: (array[Form KeyValue]) - List of forms that create the property
\ No newline at end of file
++ forms: (array[Form KeyValue]) - List of forms that create the property
+
+## EntitySummaryFields (object)
++ uuid: `uuid:85cb9aff-005e-4edd-9739-dc9c1a829c44` (string, required) - The `uuid` of the Entity that uniquely identifies the Entity.
++ datasetName: `people` (string, required) - Name of the dataset
```
matthew-white, line 4823: Do we need to include the dataset name? I don't think that submissions return the `xmlFormId`. Also, the dataset name by itself won't identify the dataset: the project ID is also needed. In other words, I'm thinking that we should return either more information (`projectId`) or less information (`datasetName`).

matthew-white (commented):
Just had a couple more thoughts about the API specification!

docs/api.md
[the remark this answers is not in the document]
sadiqkhoja: I have added `details` object but frontend would call `/diffs` API to generate entity feed.

Also `projectId` is not necessary because API consumer already knows it - it is required in URL

docs/api.md
sadiqkhoja: I like the simplicity of versionNumber, but we may need it to conflict resolution of offline entities. SHA could be an additional field, I haven't thought about it thoroughly though.

docs/api.md
```diff
@@ -4552,4 +4816,50 @@ These are in alphabetic order, with the exception that the `Extended` versions o
...
+
+## EntitySource (object)
++ type: `API` (EntitySourceType, required) - From what type of source this Entity was created.
++ id: `uuid:85cb9aff-005e-4edd-9739-dc9c1a829c44` (string, required) - `FormXmlId`/`SubmissionID` if the type is Submission otherwise value of `X-Client-Id` provided at the time of Entity creation.
+
+## EntityVersionFields (object)
++ versionNumber: `1` (number, required) - Numeric number to identify the version (auto incrementing)
++ label: `John Doe` (string, required) - Label of the Entity
++ current: `true` (boolean, required) - `true` if the version is the latest one
++ deleted: `false` (boolean, required) - `true` if the Entity is deleted
```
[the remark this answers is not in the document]
sadiqkhoja: I have added `deletedAt` in Entity. For EntityVersion if `deleted` is true then `createdAt` is actually timestamp of deletion. I think I should rename `createdAt` to `timestamp` in EntityVersion, what do you think?

test/assertions.js
sadiqkhoja: I started with cascading approach but that was making me duplicate fields because EntitySummary can Extended as well and that would contain ExtendedEntityDefSummary. So I can't have a single EntitySummary assertion that could be extended by other assertions.

docs/api.md
matthew-white: I feel like some unique ID is needed for each entity version for the diff response. I don't have any preferences though on what that unique ID looks like, whether it's auto-incrementing or a UUID that Central generates or something else.

docs/api.md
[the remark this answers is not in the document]
matthew-white: Does that mean that deleting or undeleting an entity creates a new entity version? With forms, I think you can move a form in and out of the trash, and the only thing that changes is the form's `deletedAt` property.

docs/api.md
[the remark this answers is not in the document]
matthew-white: I'm thinking that Frontend would use `entity.currentVersion.source.details` in the "Basic Information" section of the page to show "Creating Submission". Could we add a `deletedAt` timestamp to the `details` object as well in case the form is soft-deleted or purged?

If the form is purged, will we be able to return the submission's instance name? I don't think we save that anywhere now, and I'm not sure we'd want to. Sometimes data is purged because it's sensitive. Maybe Frontend should just show the instance ID in that case? 🤔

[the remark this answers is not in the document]
matthew-white: > Could we add a `deletedAt` timestamp to the `details` object as well in case the form is soft-deleted or purged?

There's `forms.deletedAt` and `actees.purgedAt`. Maybe we return them as two separate timestamps?

docs/api.md
matthew-white: In addition to `creatorId`, could it also return `creator` when extended metadata is requested?

docs/api.md
```diff
@@ -3040,6 +3040,270 @@ name,label,first_name,last_name,age,favorite_color
...
+    + Attributes
+        + projectId: `16` (number, required) - The numeric ID of the Project
+        + name: `people` (string, required) - Name of the Dataset
+        + uuid: `54a405a0-53ce-4748-9788-d23a30cc3afa` (string, required) - UUID of the Entity
+
+    + Body
+
+            { 
+                "firstName": "John"
+            }
```
matthew-white, line 3264: I'm wondering what would happen to this endpoint and also the PUT endpoint if one day we add editable metadata to entities, something like review state. With that possibility in mind, what do you think about nesting the entity data in a property named `data` in the request body? Here's an example request body:

```json
{
  "data": {
    "firstName": "John"
  }
}
```

Another idea: what if creating a new version happened from a POST to &hellip;/entities/:uuid/versions? Then PATCH to the entity itself could be reserved only for future metadata properties.

docs/api.md
matthew-white, line 1: Another thing that I think will need to change in this file is our documentation of the `__system` property of entity OData: it will need a new property for the number of edits (#815).

docs/api.md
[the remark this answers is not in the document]
sadiqkhoja: I meant to say `/entities/:uuid/audits` here

> I have added details object but frontend would call /diffs API to generate entity feed.

We can return `forms.deletedAt` and `actees.purgedAt` with the `/audits`. 

Related to this, which `instanceName` is required on the frontend. Instance name at the time of entity creation or current instance name?

docs/api.md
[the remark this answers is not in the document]
sadiqkhoja: My original idea was to create a new version for every delete and restore but now I am thinking not to do that. 

I will remove this flag

docs/api.md
```diff
@@ -3040,6 +3040,270 @@ name,label,first_name,last_name,age,favorite_color
...
 
+# Group Entities
+
+An Entity is a data row of the dataset. Following API endpoints will let you performance CRUD (create, read, update and delete) operations on the entities.
+
+## Entities Metadata [GET /projects/{projectId}/datasets/{datasetName}/entities]
+
+This endpoint returns list of entities of a Dataset. Please note that this endpoint only returns metadata of the entities not the data. If you want to get the data of all entiities then please refer to [OData Dataset Service](#reference/odata-endpoints/odata-form-service)
+
+You can provide `?deleted=true` to include deleted entities
```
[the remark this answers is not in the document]
sadiqkhoja: changed it

docs/api.md
sadiqkhoja: Concluded in the meeting to use `userAgent` instead

test/assertions.js
sadiqkhoja: Created a separate PR https://github.com/getodk/central-backend/pull/820

docs/api.md
sadiqkhoja: agree

docs/api.md
sadiqkhoja: `datasetName` removed

docs/api.md
sadiqkhoja: in that case I would like to have similar body for POST and PUT as well

docs/api.md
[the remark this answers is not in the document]
sadiqkhoja: In case of delete and restore, do we update the `updateAt` time as well?

docs/api.md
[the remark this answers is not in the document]
matthew-white: > We can return `forms.deletedAt` and `actees.purgedAt` with the `/audits`.

Sounds good to me. :+1:

> Related to this, which `instanceName` is required on the frontend. Instance name at the time of entity creation or current instance name?

I don't think the release criteria say, but my instinct is that it should be the current instance name. If the user clicks the link to the submission, that page is going to show the current instance name. If that's hard, I also think we could consider having Frontend send an extra/sequential request for the submission and its current instance name (and just show a loading message in the meantime for "Creating Submission").

docs/api.md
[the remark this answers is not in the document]
matthew-white: > In case of delete and restore, do we update the `updateAt` time as well?

I was wondering that myself. It looks like we don't:

https://github.com/getodk/central-backend/blob/ca92c4c078764692df57f90d512e17626ebd6c64/lib/model/query/forms.js#L280-L286
https://github.com/getodk/central-backend/blob/ca92c4c078764692df57f90d512e17626ebd6c64/lib/util/db.js#L198-L203

docs/api.md
matthew-white: That makes sense for the endpoints to be consistent. We've struggled a little with the PUT endpoint for a project because whenever we add a property to the project, we need to change our PUT requests to include the new property. I'm worried we could run into similar issues here (though it's not often that new properties are added). Do you think we could get by with just PATCH? A user could still use PATCH to specify all the properties at once, including blanking out a property.

docs/api.md
matthew-white: It'll also need a new property for `updatedAt` I think.

docs/api.md
sadiqkhoja: let's go with the PATCH only

docs/api.md
matthew-white: You could also return the diffs as an array rather than an object indexed by the version number/ID, and Frontend could figure it out. That'd be different from submission diffs, but it wouldn't require a version ID.

docs/api.md
matthew-white: How should the user update the entity label? As a top-level property next to `data` or only as a property within `data`? If the latter, I think it'd be helpful to note that in the docs.

sadiqkhoja:
Updated the API doc to reflect implementation. Updated doc can be reviewed at https://centralbackendsk.docs.apiary.io/#reference/entities

ktuite (approved):
I quickly looked through a diff of the master branch vs. the previous release to see which endpoints had changed or been added. I also used PyODK to go through the endpoints related to datasets and entities. The changelog has been updated/filled out to show all the things we did! I think it's all covered and looking good!

PR #820, opened 2023-04-05
Entity dummy response and assertions

sadiqkhoja:
- Dummy responses from the Entity API
- Integration tests and assertions


lib/resources/entities.js
```diff
@@ -0,0 +1,368 @@
+// Copyright 2022 ODK Central Developers
```
ktuite, line 1: 2023

test/integration/api/entities.js
```diff
@@ -0,0 +1,263 @@
...
+    it('should return extended metadata of the entities of the dataset', testService(async (service) => {
+      const asAlice = await service.login('alice');
+
+      await asAlice.get('/v1/projects/1/datasets/People/entities')
+        .set('X-Extended-Metadata', true)
+        .expect(200)
+        .then(({ body: people }) => {
+          people.forEach(p => {
+            p.should.be.an.Entity();
+            p.should.have.property('currentVersion').which.is.an.EntityDef();
```
ktuite, line 42: Should these be ExtendedEntity and ExtendedEntityDef?

test/integration/api/entities.js
```diff
@@ -0,0 +1,263 @@
...
+    // it should reject if uuid is not found
+    // it should reject if uuid in queryParam and body don't match
+    // it should reject if uuid or label is missing
+    // it should reject if property is not present in dataset.publishedProperties
+    // it should reject if user don't have permission
+  });
+
+  describe('PATCH /datasets/:name/entities/:uuid', () => {
+
+    it('should partially update an Entity', testService(async (service) => {
```
ktuite, line 193: I'm trying to understand the intended difference between PUT and PATCH here. Not that we have to figure it out now, though!

ktuite (approved):
Looks great! I have a much easier time reasoning about this breakdown - thanks for making the changes.

test/integration/api/entities.js
sadiqkhoja: PUT requires complete object. PATCH allows single field to be updated

Issue #818, opened 2023-04-05
Add "userAgent" to entity def table


ktuite:
Related to issue #817, could be done at the same time as part of the same migration?

This will be a way to capture info about what mechanism created or updated an entity in addition to the actor. It follows the same pattern as submission defs.

PR #824, opened 2023-04-06
Entity migrations

### LINKED ISSUE #816: Move "label" from root entity to entity def
Make database migration that moves entity label from root entity table to entity defs table because it is something that can be updated.

### LINKED ISSUE #817: Add "creatorId" to entity def table
We left this off before because it was implicitly the creator of the submission used to create the entity. When entities can be made *and updated* w/o a source submission, we will want to store the `creatorId` of each version/entity def on that specific def

ktuite:
Closes #817 and #818. Also closes #816. 

This PR includes 2 migrations that get the `entities` and `entity_defs` table better set up for ongoing entity work.
1. Adds `creatorId` to `entity_defs` to track who is responsible for each version / update of an entity
2. Adds `userAgent` as well to `entity_defs` to track more context about each version
3. Moves `label` from the root `entities` table to `entity_defs` because it is something that can be updated
4. Adds a `deletedAt` field to `entities` table.

The `creatorId` and `userAgent` are filled from the submissionDefId that we already have linked to each entity. In the future, entities will be able to be created without a submission, which is why we need to store these source fields separately and not implicitly keep them in the submission. Plus, a submission could get deleted when a form is deleted, losing this entity creator info! 

The `label` on `entities` is appropriately moved over to the corresponding `entity_defs` column. For now, before updates have been implemented, there is always only one version of an entity so it is clear what the label is.

Tests, including migration tests.

Lots of discussion. Also trying to keep these migrations as simple as possible.

Not going to change things for users too much except allow us to make progress on entity creates and updates!

It will require some doc changes which are being addressed in another PR.

### DIFF
--- a/lib/model/frames/entity.js
+++ b/lib/model/frames/entity.js
@@ -16,7 +16,7 @@ const { parseSubmissionXml } = require('../../data/entity');
 class Entity extends Frame.define(
   table('entities'),
   'id',                  'uuid',
-  'datasetId',           'label',
+  'datasetId',
   'createdAt',           'creatorId',
   embedded('creator')
 ) {
@@ -44,6 +44,7 @@ Entity.Def = Frame.define(
   'id',                  'entityId',
   'createdAt',           'current',
   'submissionDefId',
+  'label',
   'data'
 );
 
--- a/lib/model/migrations/20230406-01-add-entity-def-fields.js
+++ b/lib/model/migrations/20230406-01-add-entity-def-fields.js
@@ -34,8 +34,7 @@ WHERE entity_defs."submissionDefId" = submission_defs.id;
   // alter table to make new column not null
   await db.raw(`
 ALTER TABLE entity_defs
-ALTER COLUMN "creatorId" SET NOT NULL,
-ALTER COLUMN "userAgent" SET NOT NULL
+ALTER COLUMN "creatorId" SET NOT NULL
 `);
 };
 
--- a/lib/model/migrations/20230406-01-add-entity-def-fields.js
+++ b/lib/model/migrations/20230406-01-add-entity-def-fields.js
@@ -15,14 +15,28 @@ ADD COLUMN "creatorId" integer,
 ADD COLUMN "userAgent" varchar(255)
 `);
 
-  // Assign submission submitter id to entity def creator id
+  // Assign root entity creator id (same as submitter id) to entity def creator id
   await db.raw(`
 UPDATE entity_defs
-set "creatorId" = submission_defs."submitterId",
-"userAgent" = submission_defs."userAgent"
+SET "creatorId" = entities."creatorId"
+FROM entities
+WHERE entity_defs."entityId" = entities.id;
+  `);
+
+  // Assign submission user agent to entity def
+  await db.raw(`
+UPDATE entity_defs
+SET "userAgent" = submission_defs."userAgent"
 FROM submission_defs
 WHERE entity_defs."submissionDefId" = submission_defs.id;
   `);
+
+  // alter table to make new column not null
+  await db.raw(`
+ALTER TABLE entity_defs
+ALTER COLUMN "creatorId" SET NOT NULL,
+ALTER COLUMN "userAgent" SET NOT NULL
+`);
 };
 
 const down = (db) => db.raw(`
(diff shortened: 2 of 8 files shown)


lib/model/migrations/20230406-01-add-entity-def-fields.js
```diff
@@ -0,0 +1,35 @@
...
+  await db.raw(`
+ALTER TABLE entity_defs
+ADD COLUMN "creatorId" integer,
+ADD COLUMN "userAgent" varchar(255)
+`);
+
+  // Assign submission submitter id to entity def creator id
+  await db.raw(`
+UPDATE entity_defs
+set "creatorId" = submission_defs."submitterId",
```
sadiqkhoja, line 21:
```diff
- set
+ SET
```

lib/model/migrations/20230406-02-move-entity-label-add-deletedAt.js
```diff
@@ -0,0 +1,63 @@
...
+
+  await db.raw(`
+  ALTER TABLE entities
+  DROP COLUMN "label"
+  `);
+
+  await db.raw(`
+  ALTER TABLE entities
+  ADD COLUMN "deletedAt" timestamp
+  `);
```
sadiqkhoja, line 35: these two statements can be merged

lib/model/migrations/20230406-01-add-entity-def-fields.js
```diff
@@ -0,0 +1,35 @@
...
+"userAgent" = submission_defs."userAgent"
+FROM submission_defs
+WHERE entity_defs."submissionDefId" = submission_defs.id;
+  `);
+};
```
sadiqkhoja, line 26: I think we should alter `entity_defs` table again here to make new columns `NOT NULL`

lib/model/migrations/20230406-01-add-entity-def-fields.js
```diff
@@ -0,0 +1,35 @@
...
+
```
sadiqkhoja, line 27: We have to remove `submissionDefId` as well

lib/model/frames/entity.js
```diff
@@ -44,6 +44,7 @@ Entity.Def = Frame.define(
   'id',                  'entityId',
   'createdAt',           'current',
   'submissionDefId',
+  'label',
```
sadiqkhoja, line 47: Missing `creatorId` and `userAgent`, but that can be added in another PR

sadiqkhoja (approved):

lib/model/migrations/20230406-01-add-entity-def-fields.js
ktuite: Let's leave it for now - we can take it out when we have a better plan for entity source.

matthew-white:
Is there a possibility that the submission def is already deleted? How about we backfill `entity_defs.creatorId` based on `entities.creatorId` rather than `submission_defs.submitterId` so that `entity_defs.creatorId` is guaranteed to exist?

lib/model/migrations/20230406-01-add-entity-def-fields.js
matthew-white: I think `creatorId` should be `NOT NULL`, but I don't think `userAgent` can be. I think the user agent is optional in the request (it's `NOT NULL` on the `submissions` table). Also, the submission might have been deleted, in which case the user agent will be gone.

ktuite: Take a look at my last 2 commits where I change up the migrations a bit. Now the new `creatorId` column is NOT NULL (and it come from the entity rather than submission def) and `userAgent` can still be NULL.

lib/model/frames/entity.js
ktuite: will add in another PR or let whoever comes to this part of the code next add it.

lib/model/migrations/20230406-01-add-entity-def-fields.js
sadiqkhoja: did you forget to remove `ALTER COLUMN "userAgent" SET NOT NULL`?

sadiqkhoja (approved):

PR #828, opened 2023-04-13
Features/entities/api get diffs

### LINKED ISSUE #800: Get version diffs of an entity
`GET /v1/projects/:projectId/datasets/:name/entities/:uuid/diffs`

Related to issue https://github.com/getodk/central-backend/issues/799

sadiqkhoja:
Closes #800

Integration Test

Based on our discussions.

No

PR is open, will update it soon.

### DIFF
--- a/lib/resources/entities.js
+++ b/lib/resources/entities.js
@@ -10,7 +10,7 @@
 const { getOrNotFound, reject } = require('../util/promise');
 const { success } = require('../util/http');
 const Problem = require('../util/problem');
-const { parseUpdateJson } = require('../data/entity');
+const { parseUpdateJson, diffEntityData } = require('../data/entity');
 
 const getActor = () => ({
   id: 1,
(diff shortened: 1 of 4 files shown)


matthew-white:
I know I [said](https://github.com/getodk/central-backend/pull/812#discussion_r1162021450) I thought it'd work to return the diffs as an array rather than as an object indexed by a version ID. However, I'm starting to question how robustly I'd be able to link the diffs to the audit log entries. I was thinking that I'd match the latest diff to the audit log entry with the latest `loggedAt` timestamp, the next latest diff to the audit log entry with the next latest `loggedAt`, and so on. But I'm realizing that if there's enough concurrency, the two aren't guaranteed to have the same order: the order of `loggedAt` is not guaranteed to be the same order as the order in which concurrent entity updates are applied, i.e., the order in which the database transactions are committed. That's because we use `clock_timestamp()` to set `loggedAt`, which may differ from the timestamp of the transaction commit.

I'm not sure how likely that possibility is given the level of concurrency we're likely to see, but I thought I'd mention it. I've started to wonder though, other than for the submission events, does Frontend really even need to retrieve the list of audit log entries for the entity? Could it just fetch &hellip;/entities/:uuid/versions with extended metadata along with &hellip;/entities/:uuid/diffs?

matthew-white:
Actually, we'll probably prohibit concurrent updates of entity data, right? Using something like `SELECT FOR UPDATE`. So the two orders should match. I was getting confused because we're thinking of allowing two new updates to branch off the same entity version. But those two updates won't be requested concurrently.

sadiqkhoja:
Yes, order of audit and diffs will be same.

matthew-white:
If we're not planning to otherwise expose `versionNumber`, then I think an array works well. If we are going to expose `versionNumber`, then an object would probably be more consistent with the submission diffs endpoint. Either way works well for Frontend though.

lib/resources/entities.js
```diff
@@ -95,17 +95,19 @@ module.exports = (service, endpoint) => {
...
+    const dataset = await Datasets.get(params.projectId, params.name, true).then(getOrNotFound);
+
+    await auth.canOrReject('entity.read', dataset);
+
+    const defs = await Entities.getAllDefs(dataset.id, params.uuid, queryOptions);
+
+    // it means there's no entity with the provided UUID
+    if (defs.length === 0) return reject(Problem.user.notFound());
+
+    return diffEntityData(defs.map(d => d.data));
```
ktuite, line 109: I think it will not pick up changes to the label because the label is not part of data.

test/integration/api/entities.js
```diff
@@ -264,22 +262,65 @@ describe('Entities API', () => {
...
-          body[2][0].should.be.eql({ old: 'John', new: 'Jane', propertyName: 'firstName' });
-          body[2][1].should.be.eql({ old: 'Doe', new: 'Roe', propertyName: 'lastName' });
-          body[2][2].should.be.eql({ old: 'John Doe', new: 'Jane Roe', propertyName: 'label' });
+          body.should.be.eql([
+            [
+              { old: '88', new: '12', propertyName: 'age' },
+              { old: 'Alice', new: 'John', propertyName: 'first_name' }
+            ],
+            [
+              { new: 'Toronto', propertyName: 'city' }
```
ktuite, line 321: Is there no `old: undefined` for this?

ktuite: I think missing `old` key is fine

ktuite (approved):

PR #829, opened 2023-04-14
User: remove unused prop 'mfaSecret'

alxndrsn:
It looks like this prop was added speculatively in 2017 (938c75a43374878342593553ea324f53b90cbfcc), and has been unused since then.

### DIFF
--- a/lib/model/frames.js
+++ b/lib/model/frames.js
@@ -189,7 +189,7 @@ Submission.Attachment = class extends Frame.define(
 class User extends Frame.define(
   table('users'), aux(Actor),
   'actorId',                            'password',
-  'mfaSecret',                          'email',        readable, writable,
+  'email',        readable, writable,
   species('user')
 ) {
   forV1OnlyCopyEmailToDisplayName() {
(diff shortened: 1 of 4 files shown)


lib/model/frames.js
```diff
@@ -189,7 +189,7 @@ Submission.Attachment = class extends Frame.define(
 class User extends Frame.define(
   table('users'), aux(Actor),
   'actorId',                            'password',
-  'mfaSecret',                          'email',        readable, writable,
+  'email',        readable, writable,
```
alxndrsn, line 192: I'm not sure what the spacing here signifies, so this may no longer be correct.

matthew-white: I think it's meant to keep field attributes aligned across fields, sort of a column. This is the only field in the `User` frame with attributes, but I think that pattern is easier to see in other frames.

test/integration/api/audits.js
```diff
@@ -55,7 +55,6 @@ describe('/audits', () => {
               audits[0].details.should.eql({ data: {
                 actorId: david.actor.id,
                 email: 'david@getodk.org',
-                mfaSecret: null,
```
matthew-white, line 58: I guess this means that `mfaSecret` will continue to exist in old audit log entries. I think that's fine though: it doesn't seem useful to scrub it from the audit log.

matthew-white (approved):

lib/model/frames.js
alxndrsn: :+1: updated to match alignment in other frames

PR #840, opened 2023-04-20
Fix/833 null forms ds metadata

### LINKED ISSUE #833: deleted form, forms with duplicate titles update properties
https://test.getodk.cloud/#/projects/430/datasets/trees - deleted form, forms with duplicate titles update properties

### Steps to reproduce:

- Publish a form that defines a dataset
- Delete that form
- Recreate the same form, you will get a warning that form with same name exists in trash. Upload it anyways
- Publish the form
- Go to the datasets page, select the dataset that's created by the form
- You will see empty rows in front of properties

### Another way to reproduce it:

- Publish a form that defines a dataset
- Create a new draft by clicking "New Draft"
- Publish it as it is, no need to upload any new form definition 
- Go to the datasets page, select the dataset that's created by the form
- You will see empty rows in front of properties

sadiqkhoja:
Closes #833

Integration tests and manual verification

No

No

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -306,6 +306,7 @@ const getProperties = (datasetId, includeForms = false) => ({ all }) => all(sql`
   ${includeForms ? sql`
   LEFT OUTER JOIN forms ON
       ds_property_fields."formDefId" = forms."currentDefId"
+      AND forms."deletedAt" IS NULL
   LEFT JOIN form_defs ON
       forms."currentDefId" = form_defs.id
   ` : sql``}
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -266,13 +266,15 @@ const getMetadata = (dataset) => async ({ all, Datasets }) => {
     }
 
     // populate forms for each property
-    propertiesMap.get(property.name).forms.push({
-      xmlFormId: property.aux.xmlFormId,
-      // TODO: check that returning null name works ok in frontend.
-      name: property.aux.formName
-        ? property.aux.formName
-        : property.aux.xmlFormId
-    });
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
This looks like a nice improvement. :+1: The tests are quite clear. I've labeled the issue as "needs testing", because I think the QA team will be able to use Frontend to follow the repro steps.

I was a little confused by what exactly `Datasets.getProperties()` returns when `includeForms` is `true`. The comments above the function say that "Dataset.Properties will be duplicated for each form that writes that field in the dataset." But am I understanding correctly that they're actually duplicated for each form version? That is, properties are duplicated for each archived form version in addition to the current version. However, in the case of an archived version, `xmlFormId` is `null` — a fact that is used later to filter out archived versions in `Datasets.getMetadata()`.

What do you think about updating the comments above `Datasets.getProperties()` to mention what happens with archived form versions? Or maybe it doesn't matter that much because `Datasets.getMetadata()` is the only place that specifies `true` for `includeForms`.

Alternatively, what do you think about excluding archived form versions in the SQL rather than later on in JS? I think the smallest change would be an additional condition in the `WHERE` clause, something like:

```js
${includeForms
  // Exclude rows that match the JOIN on ds_property_fields but not the JOIN on forms.
  ? sql`AND (ds_property_fields."dsPropertyId" IS NULL OR forms.id IS NOT NULL)`
  : sql``}
```

Just some ideas! I think the PR could be merged as-is, but I wanted to throw out these thoughts in case any resonate.

sadiqkhoja:
Actually I started with updating the SQL but excluding null `form.id` causes other tests to fail, specifically where a form creates an unique property and gets deleted.

I am updating the comments on `getProperties` function

Issue #801, opened 2023-03-20
Revisit Entity CSV export


ktuite:
There's a version of this that exists already, but we should enhance it so it's more distinct from the csv form attachment entity export. 

- `GET /v1/projects/:projectId/datasets/:name/entities.csv`
- Move this in the documentation from Download Dataset to Entity Export
- Change it to export all the data present in the odata entity feed
    - Data + system data

sadiqkhoja:
from #803:
> Eventually test that it IS included in the entity export? Or can be included? In the same way deleted submissions can be included?

sadiqkhoja:
Add condition to filter out deleted entities

matthew-white:
I think the QA team should take a look at this one. @sadiqkhoja, would you mind adding a little more context for the QA team so that they know what's expected?

dbemke:
**Tested with success!**

PR #841, opened 2023-04-21
new: added DELETE api for entities

### LINKED ISSUE #803: Allow soft-deleting entities
We've talked about archiving vs. deleting vs. purging. Archiving is something that could be managed in user-space by adding and editing a specific dataset property so we probably won't directly target that right now. We are aiming to have (soft) entity deletion via the API with this release in order to have full CRUD functionality for entities.

sadiqkhoja:
Closes #803 

Integration tests

No

We have a separate PR for that

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -9,7 +9,7 @@
 
 const { sql } = require('slonik');
 const { Actor, Entity, Submission, Form } = require('../frames');
-const { equals, extender, unjoiner, page } = require('../../util/db');
+const { equals, extender, unjoiner, page, markDeleted } = require('../../util/db');
 const { map } = require('ramda');
 const { blankStringToNull, construct } = require('../../util/util');
 const { QueryOptions } = require('../../util/db');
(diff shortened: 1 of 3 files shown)


ktuite (commented):
Looks good but I think we should have some more tests

Please add:
- Use this deletion mechanism in test above called `should return metadata of the entities of the dataset - only deleted`
- Test that you can't access the deleted entity (get a 404) once its deleted, unless it's through the `entities?deleted=true` endpoint.

Maybe add:
- Test that it's not included in the CSV attachment (I'm not sure if that should go in this API test area or in the CSV export/attachment area. There are some other things that are not part of this PR I think we should also test in the CSV attachments, like that the updated values are getting passed through.) Made an issue https://github.com/getodk/central-backend/issues/842 to capture these thoughts.
- Eventually test that it IS included in the entity export? Or can be included? In the same way deleted submissions can be included?

sadiqkhoja:
Updated `should return metadata of the entities of the dataset - only deleted` and added more assertion for `should delete an Entity`

----

> Eventually test that it IS included in the entity export? Or can be included? In the same way deleted submissions can be included?

can be done as part of #801

ktuite (approved):

PR #843, opened 2023-04-21
added etag feature for openrosa attachment endpoint

### LINKED ISSUE #826: Set ETag header for OpenRosa form attachment endpoint
https://github.com/getodk/central-backend/pull/776/files#diff-30634ec3e57ee3f29ffd8f157350f6cb947c0221536272b20cbb7a458efc76df sets the `ETag` header for the REST API endpoint to fetch an entity CSV. However, clients use the OpenRosa API so they're not getting caching benefits.

https://github.com/getodk/central-backend/blob/376e2bb3ea4894531138d8989d166b43a42632e5/lib/resources/forms.js#L275

Expected behavior:
* Collect requests a form with an entity attachment, that results in a 200
* Collect makes the same request again, that results in a 304

I identified this issue by looking at the `nginx` log on the server (`docker compose logs --tail 500 nginx`).

sadiqkhoja:
Closes #826

Added ETag feature for OpenRosa attachment endpoint. It works for both blobs and datasets

Integration tests and manual verification locally

NA

No

Updated.

### DIFF
--- a/test/integration/api/forms/forms.js
+++ b/test/integration/api/forms/forms.js
@@ -1117,6 +1117,38 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
             .expect(304);
 
         }));
+
+        it('should return latest content and correct ETag', testService(async (service) => {
+          const asAlice = await service.login('alice');
+
+          await asAlice.post('/v1/projects/1/forms')
+            .send(testData.forms.withAttachments)
+            .set('Content-Type', 'application/xml')
+            .expect(200);
+
+          await asAlice.post('/v1/projects/1/forms/withAttachments/draft/attachments/goodone.csv')
+            .send('test,csv\n1,2')
+            .set('Content-Type', 'text/csv')
+            .expect(200);
+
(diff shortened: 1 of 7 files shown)


test/integration/api/forms/forms.js
matthew-white, line 1: How about adding one or more tests asserting that the following happens if the form attachment is changed:

- The ETag header changes.
- Passing the previous ETag in `If-None-Match` doesn't result in a 304.

matthew-white (approved):
I don't think the QA team can really verify this, but if you haven't already, I think it'd be a good idea for someone to try this with Collect and make sure that everything is working as expected.

sadiqkhoja:
Previously I tested this in postman only. Now verified it using ODK Collect as well:

```
central-nginx-1  | 72.136.89.58 - - [24/Apr/2023:21:20:07 +0000] "GET /v1/key/<redacted>/projects/2/forms/trees_follow_up/attachments/trees.csv HTTP/1.1" 304 0 "-" "org.odk.collect.android/v2023.1.2 Dalvik/2.1.0 (Linux; U; Android 13;)" "-"
```

PR #844, opened 2023-04-21
Create entity via POST

### LINKED ISSUE #798: Create entity data with POST request
`POST /v1/projects/:projectId/datasets/:name/entities`

Can be anything in the dataset properties.

ktuite:
Closes #798 

Allows entities to be created via POST request to `/v1/projects/:projectId/datasets/:datasetName/entities`

Sample payload may look like this, with uuid, label, and data. Provided properties must be a subset of existing dataset properties. Values of label and data must be _strings_ (numbers must be in quotes and null/empty values must be empty strings: `''`)

```
uuid: '12345678-1234-4123-8234-111111111aaa',
label: 'Johnny Doe',
data: {
  first_name: 'Johnny',
  age: '22'
}
```

Tests.

Wanted to double-check the format of the incoming entity with other folks.
- Should label be in or out of the data block?
- Should the outer fields be in a meta block? It would match the submission xml a bit more but i don't know how much it matters.

I tried to reuse the same validation used by entities created through submissions.

New feature in the realm of entities!

Will need to make sure PR #812 is consistent with this code.

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -34,10 +34,14 @@ const odataToColumnMap = new Map([
 // and assemble a valid entity data object. System properties like dataset
 // label, uuid are in entity.system while the contents from the bound form
 // fields are in enttiy.data.
-const validateEntity = (entity) => {
+//
+// Since validateEntity is also used on entities parsed from JSON, which
+// don't include dataset or create flag (these are implicit) we have made
+// these specific validation checks optional (but turned on by default).
+const validateEntity = (entity, checkDataset = true, checkCreate = true) => {
   const { dataset, label, id } = entity.system;
 
-  if (!dataset || dataset.trim() === '')
+  if (checkDataset && (!dataset || dataset.trim() === ''))
     throw Problem.user.entityViolation({ reason: 'Dataset empty or missing.' });
 
   if (!label || label.trim() === '')
@@ -52,7 +56,7 @@ const validateEntity = (entity) => {
   const uuid = matches[2].toLowerCase();
 
   // If create is not an allowed value of true, don't return entity data
-  if (!(entity.system.create === '1' || entity.system.create === 'true' || entity.system.create === 1))
+  if (checkCreate && !(entity.system.create === '1' || entity.system.create === 'true'))
     return null;
 
   const newEntity = { ...entity };
@@ -88,6 +92,42 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject) 
     }
   });
 });
+
+// Can be used to parse new entities and entity updates from
+// request data in the form of JSON.
+const parseJson = (existingEntity, body, propertyNames) => {
+  const bodyKeys = ['label', 'uuid', 'data'];
+  if (Object.keys(body).filter(k => !bodyKeys.includes(k)).length > 0)
+    throw Problem.user.invalidEntity({ reason: 'Unrecognized fields included in request.' });
+
+  let entity;
+  if (existingEntity) {
+    entity = clone(existingEntity);
+    entity.system.id = entity.system.uuid;
+    delete entity.system.uuid;
+  } else {
+    entity = { system: Object.create(null), data: Object.create(null) };
+    entity.system.id = body.uuid;
+    entity.system.label = body.label; // label will come in under body for new entities
+  }
+
+  const { data } = body;
+
+  if (data == null)
+    throw Problem.user.invalidEntity({ reason: 'No entity data provided.' });
+
+  for (const [k, v] of Object.entries(data)) {
+    if (k === 'label') entity.system.label = v; // label will come in under data for updated entities
+    else if (propertyNames.includes(k)) {
+      if (typeof v !== 'string')
+        throw Problem.user.invalidEntity({ reason: `Property value for  [${k}] is not a string.` });
+      entity.data[k] = isBlank(v) ? '' : v;
+    } else
+      throw Problem.user.invalidEntity({ reason: `You specified the dataset property [${k}] which does not exist.` });
+  }
+
+  return validateEntity(entity, false, false);
+};
 
 const parseCreateJson = (body, properties, datasetName) => new Promise((resolve, reject) => {
   const entity = { system: Object.create(null), data: Object.create(null) };
@@ -314,6 +354,7 @@ const diffEntityData = (defData) => {
 
 module.exports = {
   parseSubmissionXml, validateEntity,
+  parseJson,
   parseCreateJson, parseUpdateJson,
   streamEntityCsv, streamEntityOdata,
   odataToColumnMap,
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -95,7 +95,7 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject) 
 
 // Can be used to parse new entities and entity updates from
 // request data in the form of JSON.
-const parseJson = (existingEntity, body, propertyNames) => {
+const extractEntity = (body, propertyNames, existingEntity) => {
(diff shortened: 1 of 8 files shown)


lib/data/entity.js
```diff
@@ -88,6 +88,35 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject)
...
+    const propSet = new Set(properties.map(p => p.name));
+    for (const [k, v] of Object.entries(data)) {
+      if (k === 'label') entity.system.label = v;
+      else if (propSet.has(k))
+        entity.data[k] = isBlank(v) ? '' : v;
+      else
+        throw Problem.user.invalidEntity({ reason: `You specified the dataset property [${k}] which does not exist.` });
+    }
+  } catch (err) {
+    throw Problem.user.invalidEntity({ reason: 'Problem parsing JSON.' });
```
ktuite, line 110: This error could be more specific: was the problem with the uuid, label, or create flag missing?

lib/data/entity.js
```diff
@@ -102,7 +131,7 @@ const parseUpdateJson = (entity, properties, body) => {
     if (k === 'label') label = v;
     else if (propSet.has(k))
       data[k] = isBlank(v) ? '' : v;
-    else
+    else // TODO: maybe this should be a different problem.
```
ktuite, line 134: In the submission entity code, I use this `entityViolation` error a bunch, which is a 409 conflict error. Some of the problems there, and here with JSON parsing, would probably be better as the new 400.28 error about `invalidEntity`, but I haven't gone and tried to change those yet.

lib/model/frames/entity.js
```diff
@@ -20,7 +20,8 @@ class Entity extends Frame.define(
   'createdAt', readable,    'creatorId', readable,
   'updatedAt', readable,    'deletedAt', readable,
   embedded('creator'),
-  embedded('currentVersion')
+  embedded('currentVersion'),
+  embedded('dataset') // used on occasion when entire dataset attached to entity
```
ktuite, line 24: Can remove: I added this when I was trying to reconstruct a full Entity w/ source and everything from the insert query. I also needed to add the dataset to it, so I could use the dataset actee id in the audit logging, and also have the `.forApi()` function happily format the entire Entity as the right API response. 

This part of the code was getting complex, there were a bunch of different ways to do things, none which felt great. I ended up just re-fetching the entity from the database so all the info would be there in the right way.

lib/model/frames/entity.js
```diff
@@ -37,6 +38,20 @@ class Entity extends Frame.define(
...
+
+  static fromJson(body, properties, dataset) {
+    return parseCreateJson(body, properties, dataset.name)
+      .then((entityData) => {
+        if (!entityData)
+          return null; // entity create was false
+        const { uuid, label, datasetName } = entityData.system;
+        const { data } = entityData;
+        return new Entity.Partial({ uuid, datasetId: dataset.id }, {
+          def: new Entity.Def({ data, label }),
```
ktuite, line 50: some places we stick the current def in `def` and other places (the user facing ones), we put it in `currentVersion`. Would like to make it more consistent, maybe get rid of this .def?

lib/model/query/entities.js
```diff
@@ -23,13 +23,16 @@ const { isTrue } = require('../../util/http');
 // Submission-defining entitiies contain the dataset name as a string
 // so we will look up the dataset id by name (unique within a project)
 // and project id (from submission def -> submission -> form def -> form)
-const _getDataset = (dsName, subDefId) => sql`
-SELECT datasets."id", sd."id" as "subDefId", datasets."acteeId", datasets."name"
-FROM submission_defs AS sd
+const _getDataset = (dsName, subDefId, dsId) => sql`
```
ktuite, line 26: Explanation for reviewer: this subquery is important for the case where the entity is parsed from the submission and we don't know the dataset until we look it up by name and project (found from subdef -> sub -> form -> project). Obviously in this other case, we have the dataset more directly and we don't have a submission def.

lib/model/query/entities.js
```diff
@@ -42,28 +45,39 @@ const nextval = sql`nextval(pg_get_serial_sequence('entities', 'id'))`;
...
           entityId: entityData.id,
-          submissionDefId: subDef.id,
-          createdAt: entityData.createdAt,
-          creatorId: subDef.submitterId,
+          submissionDefId: subDefId,
           data: partial.def.data,
           label: partial.def.label
         }),
-        dataset: { acteeId: dsActeeId, name: dsName }
+        dataset: new Dataset({ acteeId: dsActeeId, name: dsName })
```
ktuite, line 80: I don't quite know what to return here. Either
- just enough for audit logging
OR
- EVERYTHING the API response will need

Currently went with enough details for audit logging.

ktuite (commented):
Notes from myself to an async reviewer

lib/data/entity.js
```diff
@@ -88,6 +88,35 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject)
...
+    entity.system.create = create;
+
+    const propSet = new Set(properties.map(p => p.name));
+    for (const [k, v] of Object.entries(data)) {
+      if (k === 'label') entity.system.label = v;
+      else if (propSet.has(k))
+        entity.data[k] = isBlank(v) ? '' : v;
+      else
+        throw Problem.user.invalidEntity({ reason: `You specified the dataset property [${k}] which does not exist.` });
+    }
```
sadiqkhoja, line 108: these lines are duplicated with `parseUpdateJson`, can be extracted to a new function

matthew-white (commented):
I was glancing over the PR and had a few questions. I'll let Sadiq do the review, but I thought I'd leave these comments.

lib/model/query/entities.js
```diff
@@ -96,7 +109,7 @@ const processSubmissionEvent = (event) => (container) =>
     .then((entity) => {
       if (entity != null) {
         return container.Audits.log({ id: event.actorId }, 'entity.create', { acteeId: entity.aux.dataset.acteeId },
-          { entity: { uuid: entity.uuid, label: entity.def.label, dataset: entity.aux.dataset.name },
+          { entity: { uuid: entity.uuid, label: entity.aux.currentVersion.label, dataset: entity.aux.dataset.name },
```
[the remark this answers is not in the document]
ktuite: This entity creation path (via submissions) seems less tied to users and permissions than other paths. It feels like it would be more mysterious if someone approved a submission, and no entity was made when everything else (dataset, form design) pointed to creating an entity. I don't think we'd stop them from approving the submission, and i also don't think we'd surface the permission error, so it would be hard to track down that it was a permissions problem.

ktuite (commented):
notes from interactive review with Sadiq

test/integration/api/entities.js
```diff
@@ -350,35 +350,249 @@ describe('Entities API', () => {
...
         .send({
-          uuid: '10000000-0000-0000-0000-000000000001',
-          label: 'Johnny Doe',
-          firstName: 'Johnny',
-          lastName: 'Doe',
-          city: 'Toronto'
+          uuid: '12345678-1234-4123-8234-111111111aaa',
+          create: 'true',
+          data: {
+            label: 'Johnny Doe',
```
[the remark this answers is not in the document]
ktuite: Made an issue to make these consistent: https://github.com/getodk/central-backend/issues/856

sadiqkhoja (commented):
Two comments about the tests in here:

1- Most of the tests are sam, just the data is different, so we can create test "theory" and run it on list of input
2- If `extractEntity` function is tested in unit tests then we don't need so many negative integrations tests

sadiqkhoja (approved):
All comments are incorporated, we can merge this.

PR #846, opened 2023-04-25
new: GET /entities/:uuid/audits

### LINKED ISSUE #796: Entity audit log
Related to dataset audit log issue: https://github.com/getodk/central-backend/issues/795 

Add `GET /v1/projects/:projectId/datasets/:name/entities/:uuid/audits` to show events about creating and updating a specific entity.

Questions 
- what will frontend need to show?
- what info do we need to store?
- should come after/along with work to create and update entities

sadiqkhoja:
Closes #796

### DIFF
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -124,7 +124,7 @@ SELECT ${fields} FROM audits
   LEFT JOIN (
     (
       SELECT submissions.*, submission_defs."userAgent" FROM submissions
-      JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root
+      JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root AND submissions."deletedAt" IS NULL
     ) submissions
       JOIN forms ON forms.id = submissions."formId" AND forms."deletedAt" IS NULL  
   ) ON (triggering_event.details->'submissionId')::INTEGER = submissions.id
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -8,9 +8,9 @@
 // except according to the terms contained in the LICENSE file.
 
 const { sql } = require('slonik');
-const { map } = require('ramda');
-const { Actee, Actor, Audit, Dataset, Form, Project } = require('../frames');
-const { extender, equals, page, QueryOptions } = require('../../util/db');
+const { map, mergeLeft } = require('ramda');
+const { Actee, Actor, Audit, Dataset, Form, Project, Submission } = require('../frames');
+const { extender, equals, page, QueryOptions, unjoiner } = require('../../util/db');
 const Option = require('../../util/option');
 const { construct } = require('../../util/util');
 
@@ -110,5 +110,56 @@ const getBySubmissionId = (submissionId, options) => ({ all }) =>
   _getBySubmissionId(all, options, submissionId);
 
 
-module.exports = { log, getLatestByAction, get, getBySubmissionId };
+
+const _getByEntityId = (fields, options, entityId) => sql`
+SELECT ${fields} FROM audits
+  LEFT JOIN actors ON actors.id=audits."actorId"
+  LEFT JOIN (
+    SELECT submissions.*, submission_defs."userAgent" FROM submissions
+    JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root -- just for the sake of userAgent which we don't need, we should remove it
+  ) submissions ON (audits.details->'submissionId')::INTEGER = submissions.id AND action IN ('entity.create', 'entity.update.version')
+  LEFT JOIN actors submission_actor ON submission_actor.id = submissions."submitterId"
+  LEFT JOIN submission_defs on submissions.id = submission_defs."submissionId" and submission_defs.current
+  LEFT JOIN actors current_submission_actor on current_submission_actor.id=submission_defs."submitterId"
+  LEFT JOIN audits approval ON (approval.details->'submissionId')::INTEGER = submissions.id AND approval.action = 'submission.update' AND (approval.details->>'reviewState')::TEXT = 'approved'
+  LEFT JOIN actors approval_actor ON approval_actor.id = approval."actorId"
+  WHERE (audits.details->'entityId')::INTEGER = ${entityId}
+  ORDER BY audits."loggedAt" DESC, audits.id DESC
+  ${page(options)}`;
+
+const getByEntityId = (entityId, options) => ({ all }) => {
+
+  const _unjoiner = unjoiner(
+    Audit, Actor,
+    Option.of(Submission), Option.of(Actor.alias('submission_actor', 'submissionActor')),
+    Option.of(Submission.Def.into('currentVersion')), Option.of(Actor.alias('current_submission_actor', 'currentSubmissionActor')),
+    Option.of(Audit.alias('approval', 'approval')), Option.of(Actor.alias('approval_actor', 'approvalActor'))
+  );
+
+  return all(_getByEntityId(_unjoiner.fields, options, entityId))
+    .then(map(_unjoiner))
+    .then(map(audit => {
+      const submission = audit.aux.submission
+        .map(s => s.withAux('submitter', audit.aux.submissionActor.orNull()))
+        .map(s => s.withAux('currentVersion', audit.aux.currentVersion.map(v => v.withAux('submitter', audit.aux.currentSubmissionActor.orNull()))))
+        .map(s => s.forApi());
+
+      const approval = audit.aux.approval
+        .map(a => a.withAux('actor', audit.aux.approvalActor.orNull()))
+        .map(a => a.forApi());
+
+      const details = mergeLeft(audit.details, {
+        submission: submission.orElse(undefined),
+        approval: approval.orElse(undefined),
+      });
+
+      return new Audit({ ...audit, details }, { actor: audit.aux.actor });
+    }));
+};
+
+
+module.exports = {
+  log, getLatestByAction, get,
+  getBySubmissionId, getByEntityId
+};
 
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -122,13 +122,15 @@ SELECT ${fields} FROM audits
   LEFT JOIN actors submission_create_event_actor ON submission_create_event_actor.id = submission_create_event."actorId"
 
   LEFT JOIN (
-    SELECT submissions.*, submission_defs."userAgent" FROM submissions
-    JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root
-  ) submissions ON (triggering_event.details->'submissionId')::INTEGER = submissions.id
+    (
+      SELECT submissions.*, submission_defs."userAgent" FROM submissions
+      JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root
+    ) submissions
+      JOIN forms ON forms.id = submissions."formId" AND forms."deletedAt" IS NULL  
+  ) ON (triggering_event.details->'submissionId')::INTEGER = submissions.id
   LEFT JOIN actors submission_actor ON submission_actor.id = submissions."submitterId"
   LEFT JOIN submission_defs on submissions.id = submission_defs."submissionId" and submission_defs.current
   LEFT JOIN actors current_submission_actor on current_submission_actor.id=submission_defs."submitterId"
-  LEFT JOIN forms ON forms.id = submissions."formId"
   
   WHERE (audits.details->'entityId')::INTEGER = ${entityId}
   ORDER BY audits."loggedAt" DESC, audits.id DESC
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -114,15 +114,22 @@ const getBySubmissionId = (submissionId, options) => ({ all }) =>
 const _getByEntityId = (fields, options, entityId) => sql`
 SELECT ${fields} FROM audits
   LEFT JOIN actors ON actors.id=audits."actorId"
+
+  LEFT JOIN audits triggering_event ON (audits.details->'eventId')::INTEGER = triggering_event.id 
+  LEFT JOIN actors triggering_event_actor ON triggering_event_actor.id = triggering_event."actorId"
+
+  LEFT JOIN audits submission_create_event ON (triggering_event.details->'submissionId')::INTEGER = (submission_create_event.details->'submissionId')::INTEGER AND submission_create_event.action = 'submission.create'
+  LEFT JOIN actors submission_create_event_actor ON submission_create_event_actor.id = submission_create_event."actorId"
+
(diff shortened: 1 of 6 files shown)


sadiqkhoja:
Sample response:

```javascript
[
  {
    actorId: 6,
    action: "entity.update.version",
    acteeId: "a889f7cf-202c-4213-88a7-ca20dd61a1c7",
    details: {
      uuid: "12345678-1234-4123-8234-123456789abc",
      dataset: "people",
      entityId: 1,
      entityDefId: 3,
    },
    loggedAt: "2023-04-25T18:43:03.431Z",
    notes: null,
    actor: {
      id: 6,
      type: "user",
      displayName: "Bob",
      createdAt: "2023-04-25T18:43:02.560Z",
      updatedAt: null,
      deletedAt: null,
    },
  },
  {
    actorId: 5,
    action: "entity.create",
    acteeId: "a889f7cf-202c-4213-88a7-ca20dd61a1c7",
    details: {
      submission: {
        instanceId: "one",
        submitterId: 5,
        deviceId: null,
        createdAt: "2023-04-25T18:43:02.941Z",
        updatedAt: "2023-04-25T18:43:02.948Z",
        reviewState: "approved",
        userAgent: "node-superagent/3.8.3",
        submitter: {
          id: 5,
          type: "user",
          displayName: "Alice",
          createdAt: "2023-04-25T18:43:02.550Z",
          updatedAt: null,
          deletedAt: null,
        },
        currentVersion: {
          submitterId: 5,
          createdAt: "2023-04-25T18:43:02.941Z",
          instanceName: "one",
          instanceId: "one",
          current: true,
          deviceId: null,
          userAgent: "node-superagent/3.8.3",
          submitter: {
            id: 5,
            type: "user",
            displayName: "Alice",
            createdAt: "2023-04-25T18:43:02.550Z",
            updatedAt: null,
            deletedAt: null,
          },
        },
      },
      approval: {
        actorId: 5,
        action: "submission.update",
        acteeId: "3975f854-e47f-485e-b3a9-ef20cb4d9ed0",
        details: {
          reviewState: "approved",
          submissionId: 1,
          submissionDefId: 1,
        },
        loggedAt: "2023-04-25T18:43:02.949Z",
        notes: null,
        actor: {
          id: 5,
          type: "user",
          displayName: "Alice",
          createdAt: "2023-04-25T18:43:02.550Z",
          updatedAt: null,
          deletedAt: null,
        },
      },
      entity: {
        uuid: "12345678-1234-4123-8234-123456789abc",
        label: "Alice (88)",
        dataset: "people",
      },
      entityId: 1,
      submissionId: 1,
      submissionDefId: 1,
    },
    loggedAt: "2023-04-25T18:43:02.975Z",
    notes: null,
    actor: {
      id: 5,
      type: "user",
      displayName: "Alice",
      createdAt: "2023-04-25T18:43:02.550Z",
      updatedAt: null,
      deletedAt: null,
    },
  },
]
```

lib/model/query/audits.js
```diff
@@ -110,5 +110,56 @@ const getBySubmissionId = (submissionId, options) => ({ all }) =>
...
+SELECT ${fields} FROM audits
+  LEFT JOIN actors ON actors.id=audits."actorId"
+  LEFT JOIN (
+    SELECT submissions.*, submission_defs."userAgent" FROM submissions
+    JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root -- just for the sake of userAgent which we don't need, we should remove it
+  ) submissions ON (audits.details->'submissionId')::INTEGER = submissions.id AND action IN ('entity.create', 'entity.update.version')
+  LEFT JOIN actors submission_actor ON submission_actor.id = submissions."submitterId"
+  LEFT JOIN submission_defs on submissions.id = submission_defs."submissionId" and submission_defs.current
+  LEFT JOIN actors current_submission_actor on current_submission_actor.id=submission_defs."submitterId"
+  LEFT JOIN audits approval ON (approval.details->'submissionId')::INTEGER = submissions.id AND approval.action = 'submission.update' AND (approval.details->>'reviewState')::TEXT = 'approved'
```
matthew-white, line 124: Unless I'm misunderstanding, I think we need to handle the possibility of multiple approvals of the same submission. I've been wondering about that case, and I'm not sure how to select the correct approval when there are more than one. Maybe we could get away with selecting the submission's first approval? I could imagine edge cases where that doesn't work, though they seems pretty unlikely. For example:

- A submission is created for a form that can create entities, but the submission doesn't specify an entity. (That's possible, right?)
- The submission is approved. No entity is created.
- Someone later realizes that there was a mistake and edits the submission. This time, the submission version does specify an entity.
- The submission is reapproved, creating the entity.

If we want to be more robust about cases like that, maybe we select the first approval whose `processed` timestamp is greater than or equal to the `loggedAt` timestamp of the `entity.create` event ? Or maybe we add a column to the entity or entity source that identifies the specific approval that created the entity?

I think the most important thing though is to handle multiple approvals in some way. I think multiple approvals are a real possibility and that we'll want to avoid returning duplicate audit log entries in that case.

lib/model/query/audits.js
```diff
@@ -110,5 +110,56 @@ const getBySubmissionId = (submissionId, options) => ({ all }) =>
...
 
-module.exports = { log, getLatestByAction, get, getBySubmissionId };
+
+const _getByEntityId = (fields, options, entityId) => sql`
+SELECT ${fields} FROM audits
+  LEFT JOIN actors ON actors.id=audits."actorId"
+  LEFT JOIN (
+    SELECT submissions.*, submission_defs."userAgent" FROM submissions
+    JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root -- just for the sake of userAgent which we don't need, we should remove it
+  ) submissions ON (audits.details->'submissionId')::INTEGER = submissions.id AND action IN ('entity.create', 'entity.update.version')
```
matthew-white, line 120: What do you think about joining in `submissions` just for `entity.create`, not `entity.update.version`? At least right now, submissions can't update entities, so I think the only action that's really connected to submissions is `entity.create`.

matthew-white (commented):
Just a couple of preliminary thoughts!

matthew-white:
I'm realizing that Frontend will also need to learn the xmlFormId of the submission so it can link to the submission.

lib/model/query/entities.js
```diff
@@ -96,9 +96,13 @@ const processSubmissionEvent = (event) => (container) =>
     .then((entity) => {
       if (entity != null) {
         return container.Audits.log({ id: event.actorId }, 'entity.create', { acteeId: entity.aux.dataset.acteeId },
-          { entity: { uuid: entity.uuid, label: entity.def.label, dataset: entity.aux.dataset.name },
+          {
+            entityId: entity.id,
```
matthew-white, line 100: Do we need to backfill `entityId`? Or are we OK with the entity activity feed being empty for entities created before .3?

lib/model/query/entities.js
```diff
@@ -96,9 +96,13 @@ const processSubmissionEvent = (event) => (container) =>
...
+            eventId: event.id,
```
matthew-white, line 101: I think we're planning not to backfill this. If that's right, could you leave a comment above this property that notes that it wasn't saved for entities created before v2023.3? I think it'll be nice to have some record of how the details schema has changed over time.

matthew-white (commented):
I'll let @ktuite do the main review, but I thought I'd leave a few comments. (I'm also happy to do the main review if @ktuite prefers.)

@sadiqkhoja and I have discussed the schema of the `details` object of the `entity.create` event. I think that's turned out nicely, and it provides all the information that Frontend needs for the entity activity feed and for "Creating Submission" in the Basic Information section of the entity detail page.

This PR requires storing an additional piece of data in the audit log, `eventId`, which points to the specific approval event that resulted in the entity's creation. I think that's a simple and robust approach to the problem of identifying the correct approval. I don't think we need to backfill `eventId`, since entities are still in preview. For entities created before .3, the entity activity feed won't show the submission approval event, but I think that's fine.

This is only sort of related and would happen later, but I'm coming around to @ktuite's idea of having one set of public audit log details that are returned over the API and another set that is only for internal use:

- Frontend only uses some details, and distinguishing between public details and internal ones might be a way to document which details are more stable (for example, those used by Frontend) and which ones are just to facilitate Backend's implementation.
- Distinguishing between public and internal details would be a way to prevent returning entity and submission numeric ids over the API. We generally don't expose those ids over the API.

lib/model/query/audits.js
```diff
@@ -110,5 +110,75 @@ const getBySubmissionId = (submissionId, options) => ({ all }) =>
...
+  LEFT JOIN audits submission_create_event ON (triggering_event.details->'submissionId')::INTEGER = (submission_create_event.details->'submissionId')::INTEGER AND submission_create_event.action = 'submission.create'
+  LEFT JOIN actors submission_create_event_actor ON submission_create_event_actor.id = submission_create_event."actorId"
+
+  LEFT JOIN (
+    (
+      SELECT submissions.*, submission_defs."userAgent" FROM submissions
+      JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root
+    ) submissions
+      JOIN forms ON forms.id = submissions."formId" AND forms."deletedAt" IS NULL  
+  ) ON (triggering_event.details->'submissionId')::INTEGER = submissions.id
```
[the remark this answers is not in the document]
sadiqkhoja: this is new to me as well. I learnt it from here: https://stackoverflow.com/a/56815807/3633589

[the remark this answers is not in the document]
matthew-white: Fascinating! Even posted by Gajus of Slonik fame.

lib/model/query/entities.js
matthew-white: We've decided not to backfill `entityId`. It feels best to avoid changing the audit log when possible so that it can serve as a historical record.

Similar to my comment below, would you mind leaving a comment in the code that mentions that this property was added in v2023.3?

test/integration/api/entities.js
```diff
@@ -326,26 +327,169 @@ describe('Entities API', () => {
   });
 
   describe('GET /datasets/:name/entities/:uuid/audits', () => {
```
sadiqkhoja, line 329: add tests to verify pagination query options

test/data/xml.js
```diff
@@ -579,6 +579,7 @@ module.exports = {
                 <entities:entity dataset="people" id="uuid:12345678-1234-4123-8234-123456789abc" create="1">
                   <entities:label>Alice (88)</entities:label>
                 </entities:entity>
+                <orx:instanceName>one</orx:instanceName>
```
sadiqkhoja, line 582: add instanceName to other submissions of `simpleEntity`

test/integration/api/entities.js
```diff
@@ -326,26 +327,169 @@ describe('Entities API', () => {
...
+
+          logs[0].details.should.not.have.property('submission');
+
+          logs[0].details.submissionCreate.details.instanceId.should.be.eql('one');
+          logs[0].details.submissionCreate.actor.displayName.should.be.eql('Alice');
+          logs[0].details.submissionCreate.loggedAt.should.be.isoDate();
+        });
+    }));
+
+    it('should return entity audits even when submission and its logs are deleted', testEntities(async (service, container) => {
```
sadiqkhoja, line 415: explain background of this test

test/integration/api/entities.js
```diff
@@ -326,26 +327,169 @@ describe('Entities API', () => {
...
+          .replace('<instanceID>one', '<deprecatedID>one</deprecatedID><instanceID>one2'))
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+
+      await asAlice.patch('/v1/projects/1/forms/simpleEntity/submissions/one')
+        .set('X-Action-Notes', 'create entity')
+        .send({ reviewState: 'approved' })
+        .expect(200);
+
+      await exhaust(container);
```
sadiqkhoja, line 469: approve the submission one more time here

test/integration/api/entities.js
```diff
@@ -326,26 +327,169 @@ describe('Entities API', () => {
   });
 
   describe('GET /datasets/:name/entities/:uuid/audits', () => {
-    it('should return audit logs of the Entity', testService(async (service) => {
+    it('should return audit logs of the Entity', testEntities(async (service) => {
```
sadiqkhoja, line 330: add negative tests in here as well

test/integration/api/entities.js
```diff
@@ -326,26 +327,169 @@ describe('Entities API', () => {
...
+          logs[0].action.should.be.eql('entity.create');
+          logs[0].actor.displayName.should.be.eql('Alice');
+
+          logs[0].details.should.not.have.property('approval');
+          logs[0].details.should.not.have.property('submission');
+          logs[0].details.should.not.have.property('submissionCreate');
         });
     }));
+
+    it('should return right approval details when we have multiple approvals', testService(async (service, container) => {
```
sadiqkhoja, line 439: add a test with POST API

ktuite (approved):
I've had a chance to play around with this a bit on my own, and Sadiq has talked me through it.

matthew-white:
I just had the thought, shouldn't we only return actor objects if extended metadata is requested?

No need to block merging (we could address that question in a follow-up PR), but I thought I'd put the question out there.

ktuite:
> I just had the thought, shouldn't we only return actor objects if extended metadata is requested?
> 
> No need to block merging (we could address that question in a follow-up PR), but I thought I'd put the question out there.

Sadiq could comment more, but I think because of all the work the queries are doing to fetch and fill in the details, it's harder to distinguish between what extended vs. not should really mean. It might be tricky or not particularly useful to just collapse the actors within the different levels of nesting. Maybe extending could mean un-collapsing something in addition to the actors, or maybe not.

There are actors within
- the main audit event
- if the event is an entity create event
   - the approval event if the entity was created from a submission approval
   - the submission creation event if the entity was from a submission
   - the submitter of the submission itself that made the entity
- maybe more when the event is an entity update event with a submission as its source (not an api call)

Anyway... definitely worth thinking about but it shouldn't block this PR.

sadiqkhoja:
My thinking regarding `extended` is that this endpoint is created to fulfil the need of frontend specifically. I don't foresee situation where we would need the response without Actor object extended

ktuite:
Was staring at this again today and thinking... do we need `submissionCreate` in the details there? Is the Creator information on the submission itself enough?

matthew-white:
I think it's reasonable to leave things the way they are. :+1: The endpoint will sort of return extended metadata by default. It doesn't feel like the typical Backend pattern: even the submission audit log can return extended metadata or not. But it also seems fine to have this exception, and it doesn't feel worth spending a lot of time to change how it's working.

> Was staring at this again today and thinking... do we need `submissionCreate` in the details there? Is the Creator information on the submission itself enough?

`submissionCreate` is needed in case the submission was deleted. Frontend can use `submissionCreate` for most things, but it needs the `submission` property for the instance name.

Issue #588, opened 2022-08-29
Intermittent test failures


matthew-white:
Tests often pass locally for me, but there are sometimes one or more failures. The failures aren't consistent: I've seen a few different errors, and even when it's the same error, it's often a different test that fails. It sounds like @ktuite is seeing some of these same errors, so I thought I'd create an issue to track them. Rather than creating an issue for each error, I thought I'd create a single issue to track the different errors we've seen. (We can always spin one off into its own issue if it starts a discussion, but starting things here seems like it might work well.) Feel free to comment if you encounter an error not listed here, to :+1: an error that you've seen, or to add more information to any of the comments below.

matthew-white:
@ktuite [mentioned](https://github.com/getodk/central-backend/pull/583#issuecomment-1230701615):

> I occasionally see this ` Error: end of central directory record signature not found` error on submission zip tests. Not all the time, not always the same test. But the tests all involve `yauzl`.

```
  1) api: /forms/:id/submissions
       [draft] .csv.zip
         should return draft submissions:
     Error: end of central directory record signature not found
      at /Users/ktuite/Desktop/code/odk/central-backend/node_modules/yauzl/index.js:183:14
      at Immediate.<anonymous> (node_modules/yauzl/index.js:615:38)
      at processImmediate (internal/timers.js:464:21)
```

I've seen a similar error:

```
  2) managed encryption
       end-to-end
         should decrypt over cookie auth with passphrases provided via url-encoded POST body:
     Error: end of central directory record signature not found
      at node_modules/yauzl/index.js:183:14
      at node_modules/yauzl/index.js:622:5
      at node_modules/fd-slicer/index.js:32:7
      at FSReqCallback.wrapper [as oncomplete] (fs.js:561:5)
```

Also in these tests:

```
  1) zipPart streamer
       should close the archive only after parts are finalized:

  2) .csv.zip briefcase output @slow
       should output all rows:

  1) api: /forms/:id/submissions
       .csv.zip GET
         versioning
           should return original instanceId and latest attached audit log when instanceId deprecated with new audit log:
```

I see this error maybe 25% of the time that I run tests.

I've seen this error mentioned on the forum: https://forum.getodk.org/t/error-end-of-central-directory-record-signature-not-found/36182

matthew-white:
I see this error maybe 10% of the time that I run tests:

```
     Error: socket hang up
      at connResetException (internal/errors.js:628:14)
[...]

PR #847, opened 2023-04-26
tests: double explicit timeouts

alxndrsn:
As discussed at https://github.com/getodk/central-backend/issues/588#issuecomment-1523024085, this enables `make test` to pass on my machine.

Scope could possibly be narrowed to these tests:

```
  1) datbase migrations: removing default project
       should put old forms into project:
     Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/odk/odk-central-backend/test/integration/other/migrations.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)

  2) datbase migrations: intermediate form schema
       should test migration:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/odk/odk-central-backend/test/integration/other/migrations.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

### DIFF
--- a/test/integration/other/analytics-queries.js
+++ b/test/integration/other/analytics-queries.js
@@ -853,7 +853,7 @@ describe('analytics task queries', () => {
   // eslint-disable-next-line space-before-function-paren, func-names
   describe('combined analytics', function () {
     // increasing timeouts on this set of tests
-    this.timeout(4000);
+    this.timeout(8000);
 
     it('should combine system level queries', testService(async (service, container) => {
       // encrypting a project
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
This looks like a reasonable change to me, and we should definitely get tests working for you locally. I think it's expected that the analytics queries are slower, as are database migrations.

Issue #329, opened 2021-02-02
Misleading response error message in API from authenticating


jkpr:
I was trying out the API, I got a session token, and then I was sending the token in Python as 

```python
session_token = "MY_TOKEN_HERE"
auth_header = {"Authorization": f"Bearer: {session_token}"}
```

with an extra colon `:` after `Bearer`. When I requested project forms at URL `/v1/projects/<project_number>/forms` the response was a 403

```
{"message":"The authenticated actor does not have rights to perform that action.","code":403.1}
```

This web user with the session was confirmed to have access to that project. The error message is misleading because I thought the server understood who I was based on the message. In fact, it did not understand who I was, and when I removed that extra colon to be

```python
session_token = "MY_TOKEN_HERE"
auth_header = {"Authorization": f"Bearer {session_token}"}
```

I got the data that I was expecting. I would have liked the error message to be something like "Authorization header sent, but it does not match any current user or session".

matthew-white:
Hi @jkpr! We generally don't distinguish between "the actor doesn't exist" and "the actor doesn't have rights": both cases result in a 403, specifically with a `code` property of 403.1. I can see the potential confusion though given the message. @issa-tseng, what do you think about changing the message to something that explicitly mentions both cases? Maybe something like "The actor does not exist or does not have rights to perform that action."?

jkpr:
Actually, looking into this a little bit more, a 401 seems more appropriate. From https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

> The HTTP 401 Unauthorized client error status response code indicates that the request has not been applied because it lacks valid authentication credentials for the target resource.

If I incorrectly format the Authorization header, then the server could naturally and appropriately respond "401 the request lacks valid authentication credentials". A 403 tells me that the server knows who I am but I am forbidden from accessing the resource, and that isn't the case here.

issa-tseng:
so, everyone's got some truth here

on one hand, i don't want to information-leak the existence of tokens via the difference between 401/403. this is yes counter to the exact http semantics but it is not an uncommon practice i think. so we tend to log things on the side of  a 403 rather than a 401.

on the other hand, the issue you mention above is a _formatting_ error, not an existence error. i have revised the code to return a 401 in this instance. i hope that sounds alright.

issa-tseng:
(i also revised the message)

PR #849, opened 2023-04-27
Make Authorization header checks tighter

alxndrsn:
Include trailing space in checks for `Basic` and `Bearer` prefixes in Authorization headers.  This space is already assumed to exist in subsequent parsing code.
    
An extra check is added for requests which _provide_ the Authorization header, but in an unrecognised format.  This ensures that a 401 header is still sent in preference to 404, as per #329.  It also changes the response code for unsupported auth-schemes _when they do not start with either `Bearer` or `Basic`_.

### DIFF
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -7,7 +7,7 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-const { isBlank, noop, without } = require('../util/util');
+const { isBlank, isPresent, noop, without } = require('../util/util');
 const { isTrue } = require('../util/http');
 const Problem = require('../util/problem');
 const { QueryOptions } = require('../util/db');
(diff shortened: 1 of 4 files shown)


lib/http/preprocessors.js
```diff
@@ -62,12 +62,12 @@ const sessionHandler = ({ Sessions, Users, Auth, bcrypt }, context) => {
       });
 
   // Standard Bearer token auth:
-  } else if (!isBlank(authHeader) && authHeader.startsWith('Bearer')) {
+  } else if (!isBlank(authHeader) && authHeader.startsWith('Bearer ')) {
```
sadiqkhoja, line 65: what's going to happen if I separate `Bearer` and value by any other character than space? like `Bearer=token_value`

sadiqkhoja (approved):

matthew-white:
It looks like the branch has a conflict. @alxndrsn, are you able to merge this PR sometime this week? We're hoping to begin QA for .3 next week.

PR #852, opened 2023-04-27
Rename authHandler to reflect wider responsibilities

alxndrsn:
The `sessionHandler` function handles more than just sessions, so it seems appropriate to rename it `authHandler`.

It currently handles authentication via:

* session token in Authorization header & cookie
* field key
* HTTP basic auth

For GET/HEAD requests using session-based authentication via Cookie header, this function also checks the CSRF token.


matthew-white (approved):
Good call, you can definitely have auth without a session. :+1:

PR #865, opened 2023-05-09
Prototype of entity_source table connected to entity audits

ktuite:
This PR is heavily inspired/builds directly off of PR #846, introducing an entity audit log endpoint that implicitly reconstructed important information about the context of an entity's creation.

This is a prototype of introducing an explicit entity_source table instead of reconstruction that information.

### What this code does:

1. migration to add `entity_sources` table
   * includes
     * `auditId` (triggering event id)
     * `submissionDefId` (current one kind of triggering object)
     * `details` (a place to store info about the object that we need to persist after object gets deleted)
   * also adds column to `entity_defs` table
   * populates and links new table
   * migration is additive right now - doesn't yet remove `submissionDefId` (essentially our first version of source) from `entity_defs` table
  
3. change entity flow to create and link source when appropriate (when created via submission approval event)

4. new audit query and temp audit resource that pulls information from this source table instead of elsewhere in the audits table. the output is very similar but `submissionCreate` is removed and there are source details instead. 
   * the one new crucial link is all entity events should have a `entityDefId` to fetch the source if such a thing is relevant to an event. 

6. new tests that match the previous entity audit tests. existing tests continue to work especially because previous audit path with implicit sources is still there.

### Why I think it's a good idea

We may be about to enter the Cambrian explosion era of entities, going from one main way they're created to multiple pathways that they are created and updated. Behold!

2022.3 (entities introduced)
- entities are only created though approved submissions

2023.3 (entities expanded -- in progress release)
- create through approved submission
- direct create through submission creation
- create through API
- update through API 
- delete through API 

future release (expanding possibilities even more)
- update through submission?
- create and update and delete through CSV 
- could imagine other things, too, e.g. "make entities out of all non-rejected submissions"? bulk/batch entity actions through API?
- other kinds of actions on entities (archiving, purging)

I think the current audit approach is great for _right now_ (as in, before this work on automatic entity creation from submissions) but having a clearer way to record different kinds of entity sources will be really useful really soon.

### Challenges of implicitly building source:

1. Working through the audit log details to link up entity sources is going to get more complex as the number of ways entities are created and updated increases.
2. Edge cases, like deleted forms/submissions make this even more tricky... many special cases that have to be represented in SQL directly.
3. Events that are linked from other events are also tricky to reason about, involves changing audit detail schema to account for it.
4. Audit details schema is hard to work with - we can't change it after a release if we realize we don't have what we need. It's also under tested.
7. Response data is heavy (e.g. including one audit event about entity creation has two other audit events nested within it: entity approval and submission creation)

### Benefits of explicitly storing entity source in a table:

Main info we want to store:
- what other triggering object made this entity (if exists)
- what triggering event made this entity (if exists)
- what details do we want to keep even if the triggering object is deleted?

With an entity source table, we can
1. Adapt to different kinds of entity sources, both for new entities and updated entities 
2. Maintain an explicit schema that is easier to debug, reason about, test
3. Be able to change the source schema as we go and backfill data if necessary 
4. Could get access to entity source information in other places besides the audit log if/when we need it

----- 

### DIFF
--- a/lib/model/migrations/20230512-01-add-entity-root.js
+++ b/lib/model/migrations/20230512-01-add-entity-root.js
@@ -1,0 +1,20 @@
+// Copyright 2023 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw('ALTER TABLE entity_defs ADD COLUMN "root" BOOLEAN NOT NULL DEFAULT FALSE');
+
+  // This should be all of the entity_defs
+  await db.raw(`update entity_defs set root=true
+  where id in (select min(id) from entity_defs group by "entityId")`);
+};
+
+const down = (db) => db.raw('ALTER TABLE entity_defs DROP COLUMN "root"');
+
+module.exports = { up, down };
--- a/lib/model/migrations/20230512-01-add-entity-source.js
+++ b/lib/model/migrations/20230512-01-add-entity-source.js
@@ -1,0 +1,100 @@
+// Copyright 2023 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+/* Steps of this migration
+  1. add new table entity source
+  2. add source id col to entity def
+  3. populate rows of source table and add id in entity def
+  4. remove sub def id from entity def table
+*/
+
+const up = async (db) => {
+  await db.schema.createTable('entity_def_sources', (es) => {
+    es.increments('id');
+    es.integer('auditId');
+    es.foreign('auditId').references('audits.id').onDelete('set null');
+    es.integer('submissionDefId');
+    es.foreign('submissionDefId').references('submission_defs.id').onDelete('set null');
+    es.jsonb('details'); // any info you'd want to have when unlinked (e.g. after submission getting deleted)
+  });
+
+  await db.schema.table('entity_defs', (fd) => {
+    fd.integer('sourceId');
+    fd.foreign('sourceId').references('entity_def_sources.id');
+
+    // TODO: right now in this prototype, this migration is purely additive:
+    // it adds a new table and adds a new column linking to that table.
+    // if we choose to go along with this, we'd also want to remove the
+    // submission def id column from the entity def table after populating
+    // the new table.
+  });
+
+  /* eslint-disable no-await-in-loop */
+  const allEntityDefs = await db.select('id', 'submissionDefId')
+    .from('entity_defs');
+  for (const entityDef of allEntityDefs) {
+    const [ approveEvent ] = await db.select('id', 'details', 'action')
+      .from('audits')
+      .whereRaw(`(details->'submissionDefId'::text) = '${entityDef.submissionDefId}'`)
+      .where({ action: 'submission.update' });
+
+    let triggeringEventId = null;
+    let details = {};
+
+    if (approveEvent) {
+      const [ submissionCreateEvent ] = await db.select('id', 'details', 'action')
+        .from('audits')
+        .whereRaw(`(details->'submissionId'::text) = '${approveEvent.details.submissionId}'`)
+        .where({ action: 'submission.create' });
+
+      triggeringEventId = approveEvent.id;
+      // looking up the creation event to get the instance id in case the submission
+      // has already been deleted
+      details = { submission: { instanceId: submissionCreateEvent.details.instanceId } };
+    }
+
+    const [ sourceId ] = await db.insert({
+      submissionDefId: entityDef.submissionDefId,
+      auditId: triggeringEventId,
+      details
+    })
+      .into('entity_def_sources').returning('id');
+
+    await db.update({ sourceId })
+      .into('entity_defs')
+      .where({ id: entityDef.id });
+  }
+  // Done backfilling
+  await db.schema.table('entity_defs', (ed) => {
+    ed.dropColumn('submissionDefId');
+  });
+};
+
+const down = async (db) => {
+  await db.schema.table('entity_defs', (ed) => {
+    ed.integer('submissionDefId');
+    ed.foreign('submissionDefId').references('submission_defs.id').onDelete('set null');
+  });
+
+  await db.raw(`
+    UPDATE entity_defs set "submissionDefId" = entity_def_sources."submissionDefId"
+      FROM entity_def_sources
+      WHERE entity_defs."sourceId" = entity_def_sources.id
+  `);
+
+  await db.schema.table('entity_defs', (ed) => {
+    ed.dropForeign('sourceId');
+    ed.dropColumn('sourceId');
+  });
+
+  await db.schema.dropTable('entity_def_sources');
+};
+
+module.exports = { up, down };
+
--- a/lib/model/migrations/20230512-02-backfill-entity-id.js
+++ b/lib/model/migrations/20230512-02-backfill-entity-id.js
@@ -1,0 +1,23 @@
+// Copyright 2023 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+  update audits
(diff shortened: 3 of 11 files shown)


ktuite:
I made the major needed changes to this PR: 
- now the standard `audits` entity endpoint uses the new query
- migration deletes the `entity_defs."submissionDefId"` column (but it can be rolled back)
- the submission creation event is returned as before

Some minor things still to change:
- fix a conflict (i also changed a test that used a worker where it shouldn't have)
- make sure the current version of a submission that is getting returned is really the current one
- probably remove the "details" part of the migration... was planing to put deleted submission info there but we seem to want to much info about a submission and who created it that we'll continue to look up those details in the audit log.

lib/model/query/audits.js
```diff
@@ -115,23 +115,32 @@ const _getByEntityId = (fields, options, entityId) => sql`
...
     (
-      SELECT submissions.*, submission_defs."userAgent" FROM submissions
-      JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND root AND submissions."deletedAt" IS NULL
+      SELECT submissions.*, '' as "userAgent"FROM submissions
     ) submissions
-      JOIN forms ON forms.id = submissions."formId" AND forms."deletedAt" IS NULL  
-  ) ON (triggering_event.details->'submissionId')::INTEGER = submissions.id
+    JOIN forms ON forms.id = submissions."formId" AND forms."deletedAt" IS NULL
+    JOIN submission_defs ON submissions.id = submission_defs."submissionId" AND submissions."deletedAt" IS NULL
+  ) on submission_defs.id = entity_def_sources."submissionDefId"
```
sadiqkhoja, line 136: this will return the submission_def that created the entity, but we have to return latest instanceName

lib/model/query/entities.js
```diff
@@ -110,18 +118,20 @@ const _processSubmissionDef = (submissionDefId, submissionId) => async ({ Datase
   // If no data was returned, e.g. if create flag was not set, then don't continue
   if (!partial)
     return null;
-  return Entities.createNew(partial, submissionDef);
+  const sourceId = await Entities.createSource(submissionDefId, eventId, { submission: { instanceId: submissionDef.instanceId } });
```
sadiqkhoja, line 121: Even though we are saving instanceId in source table, we are still fetching submission.create event. What is the benefit of persisting this additional information?

lib/model/query/entities.js
```diff
@@ -158,13 +168,13 @@ const createVersion = (dataset, entity, data, label, userAgentIn = null) => ({ c
...
   // For now, attach a dummy 'api' entity def source to return for PATCH request.
-  const source = new Entity.Def.Source({
+  const source = new Entity.Def.SourceSummary({
     type: 'api',
     details: null
   });
 
   return one(sql`
-  with def as (${_defInsert(entity.id, null, creatorId, userAgent, label, json)}),
+  with def as (${_defInsert(entity.id, creatorId, userAgent, label, json)}),
```
sadiqkhoja, line 177: shouldn't we create a row in source table when entity is created/updated via API?

lib/model/query/entities.js
```diff
@@ -242,7 +253,8 @@ const getAll = (datasetId, options = QueryOptions.none) => ({ all }) =>
 const _getDef = extender(Entity.Def, Submission, Submission.Def.into('submissionDef'), Form)(Actor.into('creator'))((fields, extend, options) => sql`
   SELECT ${fields} FROM entities
   JOIN entity_defs ON entities.id = entity_defs."entityId"
-  LEFT JOIN submission_defs ON submission_defs.id = entity_defs."submissionDefId"
+  LEFT JOIN entity_def_sources ON entity_defs."sourceId" = entity_def_sources.id
+  LEFT JOIN submission_defs ON submission_defs.id = entity_def_sources."submissionDefId"
```
sadiqkhoja, line 257: should we handle submission deletion scenario here as well?

lib/model/query/entities.js
```diff
@@ -261,7 +273,7 @@ const getAllDefs = (datasetId, uuid, options = QueryOptions.none) => ({ all }) =
       const isSourceSubmission = !!v.submissionDefId;
 
       // TODO: revisit this when working on POST /entities
```
sadiqkhoja, line 275: I think we can now remove TODOs related to source

sadiqkhoja (approved):
Overall I'm satisfied with this PR, I have posted some questions that may require only explanation and no code change.

ktuite:
I just rebased on master (er, it was more manual than that) after @sadiqkhoja's changes to allow entities to be created without approval. There are still some high level changes I intend to make:

- [ ] In the migration file, account for deleted submissions, multiple approval events, and figuring out which details to store.
- [ ] Add `root` column to `entity_defs` table
- [ ]  Make sure submissions from soft-deleted forms are properly excluded in a few more places
- [ ] Combine `Entity.Def.SourceSummary` and `Entity.Def.Source` frames

lib/model/migrations/20230512-01-add-entity-source.js
```diff
@@ -0,0 +1,100 @@
...
+    // if we choose to go along with this, we'd also want to remove the
+    // submission def id column from the entity def table after populating
+    // the new table.
+  });
+
+  /* eslint-disable no-await-in-loop */
+  const allEntityDefs = await db.select('id', 'submissionDefId')
+    .from('entity_defs');
+  for (const entityDef of allEntityDefs) {
+    const [ approveEvent ] = await db.select('id', 'details', 'action')
```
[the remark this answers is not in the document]
ktuite: the submission approval event will have the submission def if in it, and if there happen to be multiple approval events on the same submission def, we can look up the first one.

ktuite:
I've removed the “source” part of the individual entity GET response _...for now..._
- It's not being used in frontend right now (the creating submission is being fetched from the audit log)
- There was a bug that wasn't being tested with every entity def saying it was made by the API
- It’s nested under ‘current version’ of an entity, which partly makes sense because each entity def has its own source, but  maybe someone requesting an entity and its source would expect the root source.
- It’s stressing me out :sweat_smile:  — i think it will take some time (and tests) to get right and i’d like to get the rest of this PR merged and work on other priority things
- Source info is available through the entities/audit endpoint and If we realize we want to access the entity source more directly, we can add it back.

lib/model/query/entities.js
```diff
@@ -118,12 +130,14 @@ const _processSubmissionDef = (event) => async ({ Datasets, Entities, Submission
...
 
-  const entity = await Entities.createNew(dataset, partial, submissionDef);
+  const sourceId = await Entities.createSource(submissionDefId, event.id, { submission: { instanceId: submissionDef.instanceId } });
+  const entity = await Entities.createNew(dataset, partial, submissionDef, null, sourceId);
 
   return Audits.log({ id: event.actorId }, 'entity.create', { acteeId: dataset.acteeId },
     {
       entityId: entity.id, // Added in v2023.3 - not backfilled
       eventId: event.id, // Added in v2023.3 - not backfilled
+      entityDefId: entity.aux.currentVersion.id, // Added in v2023.3 - not backfilled
```
ktuite, line 140: could remove eventId from above?

lib/model/query/entities.js
```diff
@@ -209,7 +216,8 @@ const _get = (includeSource) => {
     LEFT JOIN actors current_version_actors ON current_version_actors.id=entity_defs."creatorId"
   `}
   ${!includeSource ? sql`` : sql`
-    LEFT JOIN submission_defs ON submission_defs.id = entity_defs."submissionDefId"
+    LEFT JOIN entity_def_sources ON entity_defs."sourceId" = entity_def_sources."id"
+    LEFT JOIN submission_defs ON submission_defs.id = entity_def_sources."submissionDefId"
```
ktuite, line 220: this is clunky - i took out the part below that builds the source response but left this part of the query in.

ktuite (commented):
Small review notes to myself

lib/model/migrations/20230512-01-add-entity-root.js
```diff
@@ -0,0 +1,20 @@
...
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw('ALTER TABLE entity_defs ADD COLUMN "root" BOOLEAN NOT NULL DEFAULT FALSE');
+
+  // This should be all of the entity_defs
+  await db.raw(`update entity_defs set root=true
+  where id in (select min(id) from entity_defs group by "entityId")`);
```
sadiqkhoja, line 15: nice to have (PR can be merged with this too): SQL keywords can be capitalized here.

lib/model/migrations/20230512-02-backfill-entity-id.js
```diff
@@ -0,0 +1,23 @@
...
+const up = (db) => db.raw(`
+  update audits
+    set details = details || jsonb_build_object('entityId', entities."id") || jsonb_build_object('entityDefId', entity_defs."id")
+    from entities
+    join datasets on entities."datasetId" = datasets.id
+    join entity_defs on entities.id = entity_defs."entityId" and root
+    where
+      audits.action = 'entity.create' and
+      audits."acteeId" = datasets."acteeId" and
+      entities.uuid::text = audits.details->'entity'->>'uuid'`);
```
sadiqkhoja, line 19: nice to have (PR can be merged with this too): SQL keywords can be capitalized here.

lib/model/migrations/20230512-03-add-entity-source.js
```diff
@@ -0,0 +1,129 @@
...
+
+/* Steps of this migration
+  1. add new table entity source
+  2. add source id col to entity def
+  3. populate rows of source table and add id in entity def
+  4. remove sub def id from entity def table
+*/
+
+const up = async (db) => {
+  await db.schema.createTable('entity_def_sources', (es) => {
```
sadiqkhoja, line 18: I should have noticed this earlier, I think there was a consensus that new migrations should be written without knex API

sadiqkhoja: Do we need any indices here? That can happen in a separate PR

sadiqkhoja (approved):

ktuite:
Will capitalize SQL in another PR!

matthew-white:
I meant to comment on the PR, but I commented on the commit: https://github.com/getodk/central-backend/commit/85806dfc9862fb79ad4f7b25891634efadec0b95#r114151446

Issue #864, opened 2023-05-09
Create entity without approval


sadiqkhoja:
Right now, entities are created when someone approves the submission. But for many workflows, this is an inconvenience. 

As a project manager, I should be able to configure whether entities are created on submission receipt or on approval. 

**Release criteria** can be found [here](https://docs.google.com/document/d/1tOBNJaWZSJjkZjTYxV7vYBOFSkiZWKsnp48uWb_L5Co/edit#heading=h.9la51antd6wm).

**Additional details:**
- Configuration flag will be at the Dataset level
- By default, approval will not be required
- Existing datasets will still require approval, user would be able to turn it off
- The flag will be set via API, it cannot be defined in form definition

## Backend:

- [ ] GET /datasets and /dataset/:name should return the current value of the flag
- [ ] PATCH /datasets/:name should update the value of the flag
- [ ] Update entity worker

## Frontend:
- [ ] display the value on list of dataset page
- [ ] have ability to update the value of the flag

matthew-white:
Closed by #868, #870, and getodk/central-frontend#795.

srujner:
Tested with Success!

dbemke:
**Tested with success!**

PR #868, opened 2023-05-11
Features/entities/approval required flag

sadiqkhoja:
Moves #864 forward

Unit and integration tests

See comments in the PR

Now users can create entities without approving submission.
We should also test that CSV export for multivalued questions is still working because I've changed background worker code.

Yes, API docs has been updated.

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -29,20 +29,13 @@ const odataToColumnMap = new Map([
 ////////////////////////////////////////////////////////////////////////////
 // ENTITY PARSING
 
-// After collecting the entity data from the submission and form fields,
-// we validate the data (properties must include dataset, label, and uuid)
+// After collecting the entity data from the request body or submission,
+// we validate the data (properties must include label, and uuid)
 // and assemble a valid entity data object. System properties like dataset
 // label, uuid are in entity.system while the contents from the bound form
 // fields are in enttiy.data.
-//
-// Since validateEntity is also used on entities parsed from JSON, which
-// don't include dataset or create flag (these are implicit) we have made
-// these specific validation checks optional (but turned on by default).
-const validateEntity = (entity, checkDataset = true, checkCreate = true) => {
-  const { dataset, label, id } = entity.system;
-
-  if (checkDataset && (!dataset || dataset.trim() === ''))
-    throw Problem.user.entityViolation({ reason: 'Dataset empty or missing.' });
+const validateEntity = (entity) => {
+  const { label, id } = entity.system;
 
   if (!label || label.trim() === '')
     throw Problem.user.entityViolation({ reason: 'Label empty or missing.' });
@@ -54,10 +47,6 @@ const validateEntity = (entity, checkDataset = true, checkCreate = true) => {
   const matches = uuidPattern.exec(id);
   if (matches == null) throw Problem.user.entityViolation({ reason: `ID [${id}] is not a valid UUID.` });
   const uuid = matches[2].toLowerCase();
-
-  // If create is not an allowed value of true, don't return entity data
-  if (checkCreate && !(entity.system.create === '1' || entity.system.create === 'true'))
-    return null;
 
   const newEntity = { ...entity };
   newEntity.system.uuid = uuid;
@@ -86,7 +75,7 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject) 
   });
   stream.on('end', () => {
     try {
-      resolve(validateEntity(entity));
+      resolve(entity);
     } catch (err) {
       reject(err);
     }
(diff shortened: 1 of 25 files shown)


lib/data/entity.js
```diff
@@ -86,7 +75,7 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject)
   });
   stream.on('end', () => {
     try {
-      resolve(validateEntity(entity));
+      resolve(entity);
```
sadiqkhoja, line 78: this function now just parse the dataset/entity related information. Validation is done by consumer function

lib/model/query/submissions.js
```diff
@@ -65,7 +65,11 @@ select ins.*, def.id as "submissionDefId" from ins, def;`)
 };
 
 createNew.audit = (submission, _, form) => (log) =>
-  log('submission.create', form, { submissionId: submission.id, instanceId: submission.instanceId });
+  log('submission.create', form, {
+    submissionId: submission.id,
+    instanceId: submission.instanceId,
+    submissionDefId: submission.def.id // added in v2023.3; not backfilled.
```
sadiqkhoja, line 71: we need this for entity creation

lib/model/query/entities.js
```diff
@@ -106,16 +109,36 @@ const _processSubmissionDef = (submissionDefId, submissionId) => async ({ Datase
...
-  if (!partial)
+
+  const entityData = await parseSubmissionXml(fields, submissionDef.xml);
+
+  // If dataset requires approval on submission to create an entity and this event is not
+  // an approval event, then don't create an entity
+  //
+  // We can't simply use submissionDefId to trace back to dataset and find out if approval
+  // is required because in future there can be multiple entities in a single submission.
+  // So we have to rely on dataset name from the xml to fetch the dataset.approvalRequired flag.
```
ktuite, line 120: If we're fetching the dataset from the name in the submission xml here, maybe we could just pass the dataset to `Entities.createNew` and not have to do the lookup in `_getDataset`, thus simplifying that function or not even needing it because we'll always have the dataset id. It just needs to be set in `partial.datasetId`.

test/integration/worker/entity.js
```diff
@@ -457,5 +491,152 @@ describe('worker: entity', () => {
...
+      await exhaust(container);
+
+      const entity = await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc')
+        .expect(200)
+        .then(({ body }) => body);
+
+      entity.should.not.be.null();
+      entity.currentVersion.data.first_name.should.equal('Alice');
+    }));
+
```
ktuite, line 639: I was also imagining cases like:

1. should handle case where approvalRequired is false and submission is updated - should not make ANOTHER entity, should not log an error (though I feel pretty confident in the `if (existingEntity.isDefined())` check and it's tested by another test in a simpler scenario somewhere else)

2. should not make an entity when approval was required, submission was uploaded, submission was rejected, approval required was turned off, and then _rejected_ submission was updated...

lib/model/query/entities.js
```diff
@@ -93,7 +94,9 @@ createNew.audit.withResult = true;
 
 
 // Entrypoint to where submissions (a specific version) become entities
-const _processSubmissionDef = (submissionDefId, submissionId) => async ({ Datasets, Entities, Submissions }) => {
+const _processSubmissionDef = (event) => async ({ Datasets, Entities, Submissions, Forms }) => {
+  const { submissionDefId, submissionId } = event.details;
+  // const approvalRequired = await Datasets.ge>
```
ktuite, line 99: comment looks like it is leftover from something else

ktuite (commented):
the flow around all of this is looking good! i want to poke at the `runSequentially` worker stuff a little bit more, but also wanted to share my comments on the rest of the code so far.

ktuite:
Another thought to test, what if you make an entity, delete the entity, and then update the submission... ooh is that how people could "edit" entities by editing submissions? (not that we want them to do that) or would the fact that the submission was only soft-deleted prevent that?

lib/worker/entity.js
```diff
@@ -7,9 +7,6 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-const createEntityFromSubmission = ({ Entities }, event) => {
-  if (event.details.reviewState === 'approved')
```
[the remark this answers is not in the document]
sadiqkhoja: this function is called only for `submission.create`, `submission.update.version` and `submission.update` events. And all these events can create an entity, will it create one, it is hard to tell by just looking at the `event` data

ktuite (approved):

PR #870, opened 2023-05-12
Features/entities/auto convert sub to entities

sadiqkhoja:
Moves https://github.com/getodk/central-backend/issues/864 forward


lib/model/query/entities.js
```diff
@@ -144,11 +144,24 @@ const processSubmissionEvent = (event) => (container) =>
...
-          submissionDefId: event.details.submissionDefId,
+        { submissionId,
+          submissionDefId,
           errorMessage: err.message,
           problem: (err.isProblem === true) ? err : null }));
 
+const processSubmissionEvent = (event) =>
+  _processionSubmissionEvent(event, event.details.submissionId, event.details.submissionDefId, event.acteeId);
+
+const createEntitiesFromPendingSubmissions = (pendingSubmissions, event) => (container) =>
```
sadiqkhoja, line 155: eventId for logging should be submission create event or latest update event

sadiqkhoja: and add information about this grand event in details column of entity_source

lib/model/query/entities.js
```diff
@@ -163,6 +165,16 @@ const processSubmissionEvent = (event) => (container) =>
           errorMessage: err.message,
           problem: (err.isProblem === true) ? err : null }));
 
+const createEntitiesFromPendingSubmissions = (submissionEvents, parentEvent) => (container) =>
+  // run sequentially because we want to isolate transaction for each submission
+  runSequentially(submissionEvents.map(event =>
+  // what should be eventId here? If we pass event.id then we won't be able to retrieve submission InstanceId when submission is deleted
+  // and if we pass submission create/edit eventId then we are telling a lie. should we store both? how it will look with new source table
```
ktuite, line 172: Is this comment question answered by passing the parent `dataset.update` event id and storing it in the source details?

ktuite (approved):

test/integration/api/datasets.js
```diff
@@ -2218,6 +2219,188 @@ describe('datasets and entities', () => {
...
+          .then(({ body }) => {
+            body.length.should.be.eql(1);
+          });
+
+        const entityErrors = await container.Audits.get(new QueryOptions({ args: { action: 'entity.create.error' } }));
+
+        entityErrors.length.should.be.eql(1);
+        entityErrors[0].details.errorMessage.should.match(/Label empty or missing/);
+
+      }));
```
ktuite, line 2403: Could you also add a test of fetching the audit log for an entity created this way?

lib/model/query/datasets.js
```diff
@@ -391,12 +396,39 @@ const getDiff = (projectId, xmlFormId, forDraft) => ({ all }) => all(sql`
   .then(r => Object.keys(r).map(k => ({ name: k.split(',')[0], isNew: forDraft ? k.split(',')[1] === 'true' : undefined, properties: r[k] })));
 
 const update = (datasets, data) => ({ one }) => one(updater(datasets, data)).then(construct(Dataset));
-update.audit = (datasets, data) => (log) => log('dataset.update', datasets, { data });
+update.audit = (datasets, data, autoConvert) => (log) => log('dataset.update', datasets, { data, autoConvert });
+
+const _unprocessedSubmissions = (datasetId, fields) => sql`
```
matthew-white, line 401: I think draft submissions need to be excluded: see #880. I think the flag `submissions.draft` can be used.

matthew-white, line 401: I think deleted submissions (for forms in the trash) should also be excluded. I don't think something in the trash should cause side effects elsewhere in the project.

lib/model/query/entities.js
matthew-white: Just my two cents, I vote for making the `dataset.update` event the primary source event. That's because the `dataset.update` event is the event in the causal chain that immediately preceded the creation of the entity. Sort of like how with submission approval, the approval is just the last step that caused the creation of the entity. The entity &hellip;/audits endpoint should still return the `submissionCreate` property, but it seems potentially helpful for the `sourceEvent` returned to be the `dataset.update` event. No strong feelings, feel free to ignore!

sadiqkhoja: will you show this event on the frontend?

PR #873, opened 2023-05-17
enhace #793: query param to sanitize property names

### LINKED ISSUE #793: Refactor Datasets.createOrMerge and Datasets.publishIfExists

sadiqkhoja:
Closes #793 

Integration Test

NA

No

Yes 

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -37,6 +37,7 @@ Here major and breaking changes to the API are listed by version.
 
 **Changed**:
 - ETag support has been added for [Download Dataset](#reference/datasets/download-dataset/download-dataset) and [Download Form Attachment](#reference/forms/individual-form/downloading-a-form-attachment).
+- `?odata=true` can be passed to [Dataset Metadata endpoint](#reference/datasets/datasets/dataset-metadata) to sanitize property names to match the way they will be outputted for OData.
 
 ### ODK Central v2023.2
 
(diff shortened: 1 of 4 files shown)


lib/model/query/datasets.js
```diff
@@ -268,6 +269,7 @@ const getMetadata = (dataset) => async ({ all, Datasets }) => {
     if (!propertiesMap.has(property.name)) {
       propertiesMap.set(property.name, {
         ...property,
+        name: sanitizeFieldsForOdata ? sanitizeOdataIdentifier(property.name) : property.name,
```
matthew-white, line 272: Because of the update modal, I think Frontend needs to know both the actual property name and the OData identifier. It needs to know the actual property name so that it can make the correct PATCH request from the update modal in the entities table. Rather than overwriting `name`, what do you think about adding a second property, maybe `odataName`?

sadiqkhoja: how about I return `name` and `odataName` in all situations and remove `odata` query parameter?

matthew-white: That sounds good to me. :+1:

docs/api.md
```diff
@@ -4608,6 +4609,7 @@ These are in alphabetic order, with the exception that the `Extended` versions o
 + isNew: `true` (boolean, required) - Whether or not this Property is new (will be created by publishing the Draft Form).
 
 ## Property Detailed (object)
-+ name: `first_name` (string, required) - The name of the Property.
++ name: `the.age` (string, required) - The name of the Property.
++ odataName: `the_age` (string, required) - OData santized name of the property.
```
matthew-white, line 4613: I'm thinking that a little more explanation might be helpful:

> The name of the property as it will appear in OData. OData property names can only contain alphanumeric characters and underscores.

matthew-white (approved):

PR #875, opened 2023-05-18
fixes #858: Submission frame should be reinstantiated

### LINKED ISSUE #858: Extended submission is merged with submitter
When a submission is requested with extended metadata, a `submitter` property is added to the response (the submission object). However, it looks like that submitter object is also merged with the submission object. The submission and the submitter share a couple of properties (`createdAt` and `updatedAt`), and as a result of the merge, it looks like the submission's properties are overwritten by the submitter's. For example, I added `console.log(body)` to the following test:

https://github.com/getodk/central-backend/blob/0057300c8dec811cd63f2898842e12c346e48b09/test/integration/api/submissions.js#L3169-L3181

The body looked like this:

```js
{
  instanceId: 'one',
  submitterId: 5,
  deviceId: null,
  createdAt: '2023-05-05T12:45:20.261Z',
  updatedAt: null,
  reviewState: null,
  userAgent: 'node-superagent/3.8.3',
  currentVersion: {
    submitterId: 5,
    createdAt: '2023-05-05T12:45:20.886Z',
    instanceName: null,
    instanceId: 'one',
    current: true,
    deviceId: null,
    userAgent: 'node-superagent/3.8.3',
    submitter: {
      id: 5,
      type: 'user',
      displayName: 'Alice',
      createdAt: '2023-05-05T12:45:20.261Z',
      updatedAt: null,
      deletedAt: null
    }
  },
  submitter: {
    id: 5,
    type: 'user',
    displayName: 'Alice',
    createdAt: '2023-05-05T12:45:20.261Z',
    updatedAt: null,
    deletedAt: null
  },
  id: 5,
  type: 'user',
  displayName: 'Alice',
  deletedAt: null
}
```

sadiqkhoja:
Closes #858 

Added integration test

Other option was to change the behaviour of `forApi()` but looks too risky to me

None, I'll create a separate PR to verify/fix same for entities

No

### DIFF
--- a/lib/model/query/submissions.js
+++ b/lib/model/query/submissions.js
@@ -163,7 +163,10 @@ order by path asc, value asc`)
 
 // Helper function to assign submission.currentVersionSubmitter to submission.currentVersion.submitter
 // Current there is no way to create such complex object using `extender` and `unjoiner` framework functions
-const assignCurrentVersionSubmitter = (x) => x.withAux('currentVersion', x.aux.currentVersion.withAux('submitter', x.aux.currentVersionSubmitter));
+//
+// New instance of Submission is created to remove all aux objects from it (except submitter).
+// If those aux's are kept then properties of aux are merged at the root level cb#858
+const assignCurrentVersionSubmitter = (x) => (new Submission(x, { submitter: x.aux.submitter })).withAux('currentVersion', x.aux.currentVersion.withAux('submitter', x.aux.currentVersionSubmitter));
 
 const _get = extender(Submission, Submission.Def.into('currentVersion'))(Actor.into('submitter'), Actor.alias('current_version_actors', 'currentVersionSubmitter'))((fields, extend, options, projectId, xmlFormId, draft) => sql`
   select ${fields} from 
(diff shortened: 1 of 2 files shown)


matthew-white:
In the `ExtendendSubmission` assertion, what do you think about changing `should.containDeep` to `should.eqlInAnyOrder`? I think that'd help prevent this issue and also similar issues in the future.

https://github.com/getodk/central-backend/blob/85806dfc9862fb79ad4f7b25891634efadec0b95/test/assertions.js#L98-L101

sadiqkhoja:
> In the `ExtendendSubmission` assertion, what do you think about changing `should.containDeep` to `should.eqlInAnyOrder`? I think that'd help prevent this issue and also similar issues in the future.
> 

not sure how that will help here? The issue is not related to the order or existence of the properties

matthew-white:
Right now the extended submission has extra top-level properties: `id`, `type`, `displayName`, `deletedAt`. `containDeep` is happy to ignore them, but `eqlInAnyOrder` would have flagged them.

matthew-white:
Sorry, I probably should highlighted that in #858. The issue isn't just that properties are overwritten, it's also that there are extra properties.

sadiqkhoja:
I see. I will replace every `containDeep` with `eqlInAnyOrder` assertion.js in a separate PR

ktuite (approved):

PR #878, opened 2023-05-18
Change name of approval event in entity audits to something more generic

ktuite:
The entity source table contains the id of a source event that created that entity def. Originally, it was just a submission approval event, but now other types of events, namely `submission.create`, can also make an entity. Instead of calling it `approval` in the API response, we wanted to give it a more generic name.

Went with `sourceEvent` but other options could include `triggeringEvent` or `event`.

This will need docs changes in the other docs PR.

### DIFF
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -167,7 +167,6 @@ const getByEntityId = (entityId, options) => ({ all }) => {
     .then(map(_unjoiner))
     .then(map(audit => {
 
-      // this could have a more generic name like 'triggering event' that was linked to any kind of entity audit event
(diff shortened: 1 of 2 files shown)


lib/model/query/audits.js
```diff
@@ -168,7 +168,7 @@ const getByEntityId = (entityId, options) => ({ all }) => {
     .then(map(audit => {
 
       // this could have a more generic name like 'triggering event' that was linked to any kind of entity audit event
```
matthew-white, line 170: This comment could probably be removed now.

matthew-white (approved):

PR #879, opened 2023-05-18
Add index to audits details->entityId

ktuite:
Closes #

### DIFF
--- a/lib/model/migrations/20230518-01-add-entity-index-to-audits.js
+++ b/lib/model/migrations/20230518-01-add-entity-index-to-audits.js
@@ -8,11 +8,11 @@
 // except according to the terms contained in the LICENSE file.
 
 const up = async (db) => {
-  await db.raw("create index audits_details_entity_index on audits using gin ((details -> 'entityId'::text))");
+  await db.raw("CREATE INDEX audits_details_entity_index ON audits USING HASH (((details ->> 'entityId')::INTEGER))");
 };
 
 const down = async (db) => {
-  await db.raw('drop index audits_details_entity_index');
+  await db.raw('DROP INDEX audits_details_entity_index');
 };
 
 module.exports = { up, down };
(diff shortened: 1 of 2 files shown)


lib/model/migrations/20230518-01-add-entity-index-to-audits.js
```diff
@@ -0,0 +1,20 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw("create index audits_details_entity_index on audits using gin ((details -> 'entityId'::text))");
```
matthew-white, line 11: I'm noticing a difference between the audits queries for submissions vs. entities:

https://github.com/getodk/central-backend/blob/46ac81172919880c4101d757ecea100fa8d049c4/lib/model/query/audits.js#L106
https://github.com/getodk/central-backend/blob/46ac81172919880c4101d757ecea100fa8d049c4/lib/model/query/audits.js#L150

I'm not sure, but I think to make use of the index, the query needs to match it exactly. Could the entities query be changed to look more like the submissions query and the index? Or the index changed to look more like the entities query? Specifically, I think the different type casts would be an issue.

ktuite: I wasn't able to change the type casting in the index to something other than text, but I changed the cast in `_getByEntityId` so it matches and follows the same pattern as submissions. 

@sadiqkhoja, since you've looked more at the indices around submission and audit logs, what do you think of this type casting and use of the gin index here?

sadiqkhoja: I think it is possible to create `hash` index here:

```
CREATE INDEX audits_details_entity_index ON audits USING HASH (CAST (details->>'entityId' as INT))
```

`hash` index should meet our needs since we are not going to search a range of `entityId`s. Otherwise we can use `btree`

matthew-white (approved):

PR #885, opened 2023-05-19
fixes 869: pass property path as array instead of dot delimited string

### LINKED ISSUE #869: Entities CSV files don't include property whose name contains period
I've created a project on staging to reproduce this issue: https://staging.getodk.cloud/#/projects/23. I've uploaded the two tree forms that use entities, except that I've changed the dataset property `circumference_cm` to `circumference.cm` with a period. According to #688, periods are allowed in property names. In the dataset, I've created one entity, and you can see that it has a non-empty value for `circumference.cm`: https://staging.getodk.cloud/v1/projects/23/datasets/trees/entities/c3e0d5f0-2af5-44bb-ad3f-26fb9f93288b. However, when I download the entities CSV file in either format, the property has an empty value:

- https://staging.getodk.cloud/v1/projects/23/datasets/trees/entities.csv
- https://staging.getodk.cloud/v1/projects/23/forms/trees_follow_up/attachments/trees.csv

I took just a quick look at the code, and I'm thinking that the following line will need to change so that a property name that contains `.` isn't split:

https://github.com/getodk/central-backend/blob/7147e567c9b4dfaaaa2f2b7dd88d6fee54832105/lib/data/entity.js#L140

Just a thought, how about passing an array of presplit path arrays to `_entityTransformer()` rather than an array of period-delimited path strings?

https://github.com/getodk/central-backend/blob/7147e567c9b4dfaaaa2f2b7dd88d6fee54832105/lib/data/entity.js#L173-L179

sadiqkhoja:
Closes #869 

Integration Tests 

I could have typed array of array as literal but that looked ugly

No

No

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -372,6 +372,37 @@ describe('datasets and entities', () => {
 
       }));
 
+      it('should return csv file for dataset that have dot in its property name', testService(async (service, container) => {
+        const asAlice = await service.login('alice');
+
+        await asAlice.post('/v1/projects/1/forms?publish=true')
+          .send(testData.forms.simpleEntity.replace(/age/g, 'the.age'))
+          .set('Content-Type', 'application/xml')
+          .expect(200);
+
+        await asAlice.post('/v1/projects/1/forms/simpleEntity/submissions')
+          .send(testData.instances.simpleEntity.one.replace(/age/g, 'the.age'))
+          .set('Content-Type', 'application/xml')
+          .expect(200);
+
+        await exhaust(container);
+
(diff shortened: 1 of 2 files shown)


test/integration/api/datasets.js
```diff
@@ -1290,6 +1321,40 @@ describe('datasets and entities', () => {
                 text.should.equal('name,label,first_name,age\n12345678-1234-4123-8234-123456789abc,Alice (88),Alice,88\n');
               })))));
 
+      it('should return entities csv', testService(async (service, container) => {
```
matthew-white, line 1324: Maybe a more specific test title?

matthew-white (approved):

PR #889, opened 2023-05-23
Add usage metrics for counting entity updates

ktuite:
Backend portion of https://github.com/getodk/central/issues/410
Needs the frontend translation change and the central config version change and the new version of the form to be uploaded.

[odk-analytics.xlsx](https://github.com/getodk/central-backend/files/11566705/odk-analytics.xlsx)

Integration Test +

Tried sending local analytics to test server.

image

NA

No

No

### DIFF
--- a/lib/data/analytics.js
+++ b/lib/data/analytics.js
@@ -155,7 +155,7 @@ const metricsTemplate = {
           "recent": 0,
           "total": 0
         },
-        "num_updated_entities": {
+        "num_entity_updates": {
           "recent": 0,
           "total": 0
         }
--- a/lib/data/analytics.js
+++ b/lib/data/analytics.js
@@ -154,6 +154,10 @@ const metricsTemplate = {
         "num_failed_entities": {
           "recent": 0,
           "total": 0
+        },
+        "num_updated_entities": {
+          "recent": 0,
+          "total": 0
         }
       }]
     }
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -370,7 +370,9 @@ SELECT
   COUNT(DISTINCT fd."formId") num_creation_forms,
   COUNT(DISTINCT CASE WHEN f."currentDefId" IS NOT NULL THEN fa."formId" END) num_followup_forms,
   MAX(COALESCE(errors.total, 0)) num_failed_entities_total,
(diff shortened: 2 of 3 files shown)


lib/data/analytics.js
```diff
@@ -154,6 +154,10 @@ const metricsTemplate = {
         "num_failed_entities": {
           "recent": 0,
           "total": 0
+        },
+        "num_updated_entities": {
```
matthew-white, line 158: I think this name has the potential to be slightly confusing. We're measuring the number of entity updates, where each entity can be counted multiple times (once for each update). However, the name of the metric seems to imply that we're measuring a subset of all entities. Put another way, it's possible for `num_updated_entities` > `num_entities`.

@sadiqkhoja, I'll let you decide whether to use the existing name or a different name like `num_entity_updates`. The name isn't shown in Frontend (text is shown instead), so it's just a question of whether we could end up getting confused internally.

sadiqkhoja: Do we want to count "Number of Entities Updated" or "Number of Updates across all Entities"?

matthew-white: I'm not sure actually. getodk/central#410 suggests the latter. "Number of Entities Updated" would never exceed "Number of Entities", but I think it'd be interesting if we saw a very high "Number of Updates across all Entities". It'd be nice to have some sense of how often updates happen.

lib/model/query/analytics.js
```diff
@@ -392,6 +394,16 @@ FROM datasets ds
     WHERE a."action" = 'entity.create.error'
     GROUP BY dfd."datasetId"
   ) AS errors ON ds.id = errors."datasetId"
+  LEFT JOIN (
+    SELECT
+      COUNT (a.details -> 'entityId'::TEXT) total,
```
matthew-white, line 399: Could this just be `COUNT(1)`?

matthew-white (approved):
I think @sadiqkhoja did the primary review. I took a quick look at the new query. I also confirmed that `num_updated_entities` has been renamed to `num_entity_updates`.

matthew-white:
@sadiqkhoja, I see that you updated the XLSForm as well. Thanks for doing that! I've uploaded it to data.getodk.cloud.

PR #892, opened 2023-05-24
Replace cli with commander

alxndrsn:
The current `cli.js` implementation returns a status code 0 if the command is omitted.  This does not seem to be configurable.

```
$node lib/bin/cli.js --email example@example.com; echo $?
Usage:
  cli.js [OPTIONS] <command> [ARGS]

Options: 
  -u, --email EMAIL      For user create and set password commands, supplies 
                         the email. 
  -h, --help             Display help and usage details

Commands: 
  user-create, user-promote, user-set-password
0
```

### Motivation

This is surprising, and hard to notice if it occurs deep in e.g. a script or docker build.

The original ticket for this is https://github.com/node-js-libs/cli/issues/14, which cites `git` as an example, but `git`, like many commands, will print usage _and return a non-zero status code_:
```
$ git >/dev/null; echo $?
1
$ grep >/dev/null 2>/dev/null; echo $?
2
$ sed >/dev/null 2>/dev/null; echo $?
1
$ curl >/dev/null 2>/dev/null; echo $?
2
...etc
```

### Other benefits:

* this removes a dependency, as `commander` is already included via `knex`
* `commander` is maintained
* resolves comment: `i'm not a huge fan of this library`

### New implementation:
```
$ node lib/bin/cli.js --email example@example.com; echo $?
Usage: node lib/bin/cli.js [options] [command]

Options:
  -u, --email <email-address>  
  -h, --help                   display help for command

Commands:
  user-create
  user-promote
  user-set-password
  help [command]               display help for command
1
```


matthew-white:
@alxndrsn, just wanted to check about the status of this PR. I was assuming that it's not ready for review because it still has [DRAFT] in its title, but I thought I'd double-check.

alxndrsn:
@matthew-white `[DRAFT]` removed - this is ready for review :+1:

matthew-white:
Replacing `cli` with `commander` sounds sensible to me if `cli` isn't maintained and has confusing behavior. I think @ktuite or @sadiqkhoja should weigh in though and do the review.

ktuite (approved):

Issue #874, opened 2023-05-18
Use "property" instead of "field" in 400.28 error message


lognaturel:
We should consistently use "property" when referencing members of an Entity.

Should we consider a more specific message if the error is that properties are not nested in `data`?

lognaturel:
And really this is about unexpected **JSON** properties, not even Entity properties yet.

PR #895, opened 2023-05-25
use specific problems instead of entityViolation and invalidEntity

### LINKED ISSUE #854: Update some entity processing errors to 400.28
PR #844 introduces a new Problem `Problem.user.invalidEntity (400.28)` to return when creating an entity from JSON fails because of some formatting or validation problem. New code around JSON entity extraction uses this problem.

Many of the previous validation error cases throw `Problem.user.entityViolation (409.14)`, which is a conflict problem rather than a bad request problem. This problem was made back when we started parsing submission XML to create entities and foresaw entity conflict as one possible problem. 

I think we could go back and update pure validation errors to throw the `400.28` error instead of `409.14`. The names could also be better - `entityConflict` and `invalidEntity` or something.

sadiqkhoja:
Closes #854

Integration tests

I tried to use specific "Problem" wherever I could and used `invalidEntity` as fall back option

No

No

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -41,12 +41,12 @@ const validateEntity = (entity) => {
     throw Problem.user.missingParameter({ field: 'label' });
 
   if (!id || id.trim() === '')
-    throw Problem.user.missingParameter({ field: 'ID' });
+    throw Problem.user.missingParameter({ field: 'uuid' });
 
   const uuidPattern = /^(uuid:)?([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})$/i;
   const matches = uuidPattern.exec(id);
 
-  if (matches == null) throw Problem.user.invalidDataTypeOfParameter({ value: 'id', expected: 'valid UUID' });
+  if (matches == null) throw Problem.user.invalidDataTypeOfParameter({ value: 'uuid', expected: 'valid UUID' });
 
   const uuid = matches[2].toLowerCase();
 
@@ -91,7 +91,7 @@ const extractEntity = (body, propertyNames, existingEntity) => {
 
   const unexpectedKeys = Object.keys(body).filter(k => !bodyKeys.includes(k));
   if (unexpectedKeys.length > 0)
-    throw Problem.user.unexpectedAttributes({ fields: unexpectedKeys });
+    throw Problem.user.unexpectedAttributes({ expected: bodyKeys, actual: Object.keys(body) });
 
   let entity;
   if (existingEntity) {
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -38,14 +38,16 @@ const validateEntity = (entity) => {
   const { label, id } = entity.system;
 
   if (!label || label.trim() === '')
-    throw Problem.user.entityViolation({ reason: 'Label empty or missing.' });
+    throw Problem.user.missingParameter({ field: 'label' });
 
   if (!id || id.trim() === '')
-    throw Problem.user.entityViolation({ reason: 'ID empty or missing.' });
+    throw Problem.user.missingParameter({ field: 'ID' });
 
   const uuidPattern = /^(uuid:)?([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})$/i;
   const matches = uuidPattern.exec(id);
-  if (matches == null) throw Problem.user.entityViolation({ reason: `ID [${id}] is not a valid UUID.` });
+
+  if (matches == null) throw Problem.user.invalidDataTypeOfParameter({ value: 'id', expected: 'valid UUID' });
+
   const uuid = matches[2].toLowerCase();
 
   const newEntity = { ...entity };
@@ -86,8 +88,10 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject) 
 // request data in the form of JSON.
 const extractEntity = (body, propertyNames, existingEntity) => {
   const bodyKeys = ['label', 'uuid', 'data'];
-  if (Object.keys(body).filter(k => !bodyKeys.includes(k)).length > 0)
-    throw Problem.user.invalidEntity({ reason: 'Unrecognized fields included in request.' });
+
+  const unexpectedKeys = Object.keys(body).filter(k => !bodyKeys.includes(k));
+  if (unexpectedKeys.length > 0)
+    throw Problem.user.unexpectedAttributes({ fields: unexpectedKeys });
 
   let entity;
   if (existingEntity) {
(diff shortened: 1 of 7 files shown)


lib/data/entity.js
```diff
@@ -97,13 +101,13 @@ const extractEntity = (body, propertyNames, existingEntity) => {
   } else {
     entity = { system: Object.create(null), data: Object.create(null) };
     if (body.uuid && typeof body.uuid !== 'string')
-      throw Problem.user.invalidEntity({ reason: 'Value for [uuid] is not a string.' });
+      throw Problem.user.invalidDataTypeOfParameter({ value: 'uuid', expected: 'string' });
```
matthew-white, line 104: It looks like `Problem.user.invalidDataTypeOfParameter` is expecting a value, not a field/property name. Should we pass `body.uuid` for `value` instead of `'uuid'`? Or maybe modify the Problem function to take either a value or a field name? Actually, it looks like elsewhere in the code is trying to pass in a field name, so that would probably be something useful to support:

https://github.com/getodk/central-backend/blob/d113ed2111d52411672a770a7a8b31b69bb13632/lib/http/preprocessors.js#L144-L148

lib/util/problem.js
```diff
@@ -181,9 +187,7 @@ const problems = {
 
     editingClosingOrClosed: problem(409.12, () => 'You are trying to edit a submission of a Form that is in Closing or Closed state. In this version of Central, you can only edit the submissions of Open Forms. We will fix this in a future version for Central.'),
 
-    enketoEditRateLimit: problem(409.13, () => 'You must wait one minute before trying to edit a submission a second time.'),
-
-    entityViolation: problem(409.14, ({ reason }) => `There was a problem with entity processing: ${reason}`)
```
matthew-white, line 186: 409.14 is already in production, right? If so, I think we should leave a comment here that 409.14 was previously used, but has been deprecated. That way, we'll know not to reuse that Problem code for a different type of Problem.

lib/util/problem.js
```diff
@@ -62,6 +62,9 @@ const problems = {
     // { field: "the name of the extraneous parameter" }
     unexpectedAttribute: problem(400.4, ({ field }) => `Passed parameter ${field} was not expected.`),
 
+    // { fields: "the name of the extraneous parameters" }
+    unexpectedAttributes: problem(400.5, ({ fields }) => `Passed parameters (${fields.join(', ')}) were not expected.`),
```
matthew-white, line 66: We previously used 400.5 for something else. It was removed in #191. I think it'd be a good idea to add a new Problem code rather than reusing the old one.

test/unit/data/entity.js
```diff
@@ -183,52 +183,34 @@ describe('extracting entities from submissions', () => {
         });
       });
 
-      it('should reject if request does not form valid entity', () => {
```
matthew-white, line 186: Why remove this test? Is it redundant? For what it's worth, I see "Label empty or missing." elsewhere below, but not "ID empty or missing."

matthew-white (commented):
I have some specific questions, but I think this is going to be a nice change. I like the use of existing specific Problem codes.

lib/data/entity.js
```diff
@@ -86,8 +88,10 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject)
 // request data in the form of JSON.
 const extractEntity = (body, propertyNames, existingEntity) => {
   const bodyKeys = ['label', 'uuid', 'data'];
-  if (Object.keys(body).filter(k => !bodyKeys.includes(k)).length > 0)
-    throw Problem.user.invalidEntity({ reason: 'Unrecognized fields included in request.' });
```
matthew-white, line 90: @lognaturel, correct me if I'm wrong, but from a glance, I think this is the only use of `Problem.user.invalidEntity` that mentions "fields" instead of "properties". So I think this PR will close #874.

lib/data/entity.js
```diff
@@ -86,8 +88,10 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject)
...
+
+  const unexpectedKeys = Object.keys(body).filter(k => !bodyKeys.includes(k));
+  if (unexpectedKeys.length > 0)
+    throw Problem.user.unexpectedAttributes({ fields: unexpectedKeys });
```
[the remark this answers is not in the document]
lognaturel: Let's say you have top-level json properties named `age` and `fav_color` because you forgot to nest in `data`, the message would say `Passed parameters age, fav_color were not expected.`? That doesn't feel all that helpful to me. Maybe stating the **expected** json properties would help? Something like "Expected top-level JSON properties: <expected>. Got: <actual>."

lib/data/entity.js
lognaturel: Yes! Ideas for making this more actionable in the other comment thread.

lib/data/entity.js
sadiqkhoja: What should be the message? Right now the message is `Invalid input data type: expected '${value}' to be (${expected})`, I feel the sentence works for both `value` and `field`, however we should pick one for the sake of consistency.

Code snippet above is buggy. If you debug `should throw an error if a non-numeric offset is given` test, you'll see that `error.message` is "Invalid input data type: expected 'undefined' to be (integer)"

test/unit/data/entity.js
sadiqkhoja: There is a test `should throw errors when id is missing` for that, however I've added one for "uuid to be string"

lib/data/entity.js
matthew-white: It seems like most of the time, we want to pass a field name to it. The only time we pass it a value is when there's a Postgres error:

https://github.com/getodk/central-backend/blob/0661a7b8a122063808301f2c03ab09b76364eb8c/lib/util/db.js#L459-L464

I think that's the original use of the Problem. In that case, I don't think a field name is available, only a value (even though the value is named `field` in the code 😅). But everything since this original case has wanted to pass in a field. The field or value appears in the message, and it also appears in the Problem details object.

What do you think about having the Problem function accept either `field` or `value`? They seem like separate but valid cases. That'd be something like:

```js
invalidDataTypeOfParameter: problem(400.11, ({ field, value, expected }) => {
  const fieldOrValue = field ?? `'${value}'`;
  return `Invalid input data type: expected ${fieldOrValue} to be (${expected})`;
},
```

Maybe that's too complicated for a Problem function though? 🤔 I also think it'd be fine to use either `value` or `field` everywhere. If you decide to standardize on `value`, would you mind fixing the code snippet above? Or that could happen in a separate PR.

lib/util/problem.js
```diff
@@ -62,6 +62,8 @@ const problems = {
     // { field: "the name of the extraneous parameter" }
     unexpectedAttribute: problem(400.4, ({ field }) => `Passed parameter ${field} was not expected.`),
 
+    // 400.6 (uniquenessViolation), was used in the past, but has been deprecated
```
matthew-white, line 65:
```diff
-400.6
+400.5
```

lib/util/problem.js
```diff
@@ -117,6 +119,9 @@ const problems = {
     // warning: there are pending submissions
     pendingSubmissions: problem(400.29, ({ count }) => `There are ${count} pending submissions`),
 
+    // { expected: "list of expected parameters", actual: "list of provided parameters" }
+    unexpectedAttributes: problem(400.30, ({ expected, actual }) => `Expected parameters: (${expected.join(', ')}). Got (${actual.join(', ')}).`),
```
matthew-white, line 123: Problem codes can't be divisible by 10 because 400.30 === 400.3. So we can skip here to 400.31.

matthew-white (approved):

PR #897, opened 2023-05-27
Fixes: testing form that consumes dataset

sadiqkhoja:
# Changes
- added dataset download ability to testing endpoint
- moved the common logic to query module

Integration test + verified running it locally

I could have kept the common logic in resource module but that would entail passing 6+ parameters to new function because we container is only injected in query module.

No

No

### DIFF
--- a/lib/model/query/form-attachments.js
+++ b/lib/model/query/form-attachments.js
@@ -17,10 +17,13 @@ const { expectedFormAttachments } = require('../../data/schema');
 const { Frame, into } = require('../frame');
 const { Blob, Form } = require('../frames');
 const { unjoiner, insertMany } = require('../../util/db');
-const { ignoringResult, resolve } = require('../../util/promise');
+const { ignoringResult, resolve, getOrNotFound, reject } = require('../../util/promise');
 const { construct } = require('../../util/util');
 const Option = require('../../util/option');
-
+const { binary, contentDisposition, withEtag } = require('../../util/http');
+const { streamEntityCsvAttachment } = require('../../data/entity');
+const { md5sum } = require('../../util/crypto');
+const Problem = require('../../util/problem');
 
 ////////////////////////////////////////////////////////////////////////////////
 // CREATING ATTACHMENT SLOTS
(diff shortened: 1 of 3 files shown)


lib/model/query/form-attachments.js
```diff
@@ -141,10 +144,32 @@ select ${_unjoinMd5.fields} from form_attachments
   where "formDefId"=${formDefId}`)
   .then(map(_unjoinMd5));
 
+const streamAttachment = (attachment, response) => async ({ Blobs, Datasets, Entities }) => {
+  if (attachment.blobId == null && attachment.datasetId == null) {
+    return reject(Problem.user.notFound());
+  } else if (attachment.blobId != null) {
+    const blob = await Blobs.getById(attachment.blobId).then(getOrNotFound);
+    return withEtag(`"${blob.md5}"`, () => binary(blob.contentType, attachment.name, blob.content));
```
matthew-white, line 152: Doesn't `withEtag()` add its own quotes now? Should these quotes be removed?

matthew-white (approved):

matthew-white:
> I could have kept the common logic in resource module but that would entail passing 6+ parameters to new function

Could you pass the function the container that's passed to the endpoint? Something like:

```js
service.get(`${base}/attachments/:name`, endpoint((container, { params, auth }, request, response) => {
  const { FormAttachments, Forms } = container;
  return getInstance(Forms, params)
    .then((form) => auth.canOrReject('form.read', form))
    .then((form) => FormAttachments.getByFormDefIdAndName(form.def.id, params.name)
      .then(getOrNotFound)
      .then(attachment => streamAttachment(container, attachment, response)));
}));
```

Not a huge deal from my perspective (I'll let you decide what's best), but I do think it's useful in general to keep the resource/query layers distinct.

matthew-white:
@sadiqkhoja, do you recommend that the QA team verify this change? If so, could you add the "needs testing" label to this PR, as well as any notes for the QA team?

sadiqkhoja:
# Steps to test:

- Create a draft Form that consumes a dataset
- Without publishing the form, go to "Testing" tab and click new
- Form should be opened in Enketo without any error
- Test the same form in ODK collect as well

dbemke:
Tested with success!

PR #908, opened 2023-05-30
fixes 433: convert latin1 to utf for filenames return by multer

sadiqkhoja:
Closes https://github.com/getodk/central/issues/433

## Root cause:

Multer is parsing filename with latin1 encoding instead of UTF-8. https://github.com/expressjs/multer/issues/1104

- Tested locally + integration test

Created local middleware to convert latin1 encoding to utf8

This is localized changed, if submission are tested from Enketo and Collect are tested then we should be good

None

### DIFF
--- a/lib/resources/submissions.js
+++ b/lib/resources/submissions.js
@@ -25,6 +25,8 @@ const { resolve } = require('../util/promise');
 const { isBlank, noargs } = require('../util/util');
 const { zipStreamFromParts } = require('../util/zip');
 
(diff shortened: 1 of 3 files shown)


matthew-white (approved):

lognaturel:
Thank you @sadiqkhoja

Not great that a seemingly routine dependency version bump led to data loss because of a subdependency change. A good reminder that we are at the mercy of our dependencies. Not sure there's anything we could have done differently to catch this one.

PR #910, opened 2023-05-31
Support Single Sign-on with OpenID Connect

alxndrsn:
Backend portion of getodk/central#449. Requires frontend changes at https://github.com/getodk/central-frontend/pull/819


lib/http/preprocessors.js
```diff
@@ -68,6 +77,8 @@ const authHandler = ({ Sessions, Users, Auth, bcrypt }, context) => {
 
   // Basic Auth, which is allowed over HTTPS only:
   } else if (isPresent(authHeader) && authHeader.startsWith('Basic ')) {
+    // REVIEW: do web users still have legitimate reasons for accessing using BasicAuth?
```
lognaturel, line 80: Basic auth is needed for generic OData consumers like PowerBI and Excel, unfortunately. I'm not sure there's a way to address that other than something like an API key in URL.

Makefile
```diff
@@ -1,9 +1,28 @@
...
+	# TODO remove shellcheck call - not available/useful/etc. outside dev(?)
+	shellcheck ./oidc-tester/scripts/docker-start.sh
+	cd oidc-tester && \
+	docker compose down && \
+	docker compose build --build-arg CACHEBUST=$$RANDOM$$(date +%s) && \
+	docker compose up --exit-code-from odk-central-oidc-tester
+
+.PHONY: dev-oidc
+dev-oidc: base
+	NODE_CONFIG_ENV=oidc-development npx nodemon --watch lib --watch config lib/bin/run-server.js
```
sadiqkhoja, line 18: Can we add nodemon to dev dependencies and use npm here

lib/bin/cli.js
```diff
@@ -16,23 +16,33 @@
...
 };
 const cliCommands = [ 'user-create', 'user-promote', 'user-set-password' ];
 cli.parse(cliArgs, cliCommands);
 
 // map commands to tasks.
 cli.main((args, options) => {
+  const withPassword = (f) => {
+    if (options['generate-password']) {
+      const password = crypto.randomBytes(32).toString('base64');
+      console.log(password);
```
sadiqkhoja, line 38: Output with `-g` flag is like this:

```
9BonCXVsfn/tQfmp4c3+xrsTlGQzm+8aQ9UCnEeqkKs=
{
  email: 'sadiq2@gmail.com',
  id: 72,
  type: 'user',
  displayName: 'sadiq2@gmail.com',
  createdAt: 2023-07-11T16:09:59.160Z,
  updatedAt: null,
  deletedAt: null
}
```

Can we beautify the output by either moving password inside the output object or by adding a label in front of the password

sadiqkhoja (commented):
Few thoughts:

- We should either implement sign-out flow or add a message on the frontend that says the user is logged out of only Central not from IDP and other applications
- Currently session timeout due to inactivity would be out of sync. If Central's idle time is shorter than IDP's then user would be out of Central but user would be able to login without entering credentials again. 


DEV: To sign out from Auth0, open https://<tenant-name>.auth0.com/oidc/logout

lib/http/preprocessors.js
```diff
@@ -6,13 +6,20 @@
...
+/* eslint-disable */
 
+const config = require('config');
 const { isBlank, isPresent, noop, without } = require('../util/util');
 const { isTrue } = require('../util/http');
 const Problem = require('../util/problem');
 const { QueryOptions } = require('../util/db');
 const { reject, getOrReject } = require('../util/promise');
 
+// TODO use req.protocol?
```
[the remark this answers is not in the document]
alxndrsn: Yes - if ExpressJS's [`trust proxy`](https://expressjs.com/en/4x/api.html#trust.proxy.options.table) setting is configured correctly then [`req.protocol`](https://expressjs.com/en/4x/api.html#req.protocol) will take this into account:

> Contains the request protocol string: either `http` or (for TLS requests) `https`.
>
> When the [`trust proxy` setting](https://expressjs.com/en/4x/api.html#trust.proxy.options.table) does not evaluate to `false`, this property will use the value of the `X-Forwarded-Proto` header field if present. This header can be set by the client or by the proxy.

Makefile
alxndrsn: `nodemon` is already a dev dependency.  How might `npm` be used here?

config/oidc-development.json
```diff
@@ -0,0 +1,42 @@
...
+  "default": {
+    "env": {
+      "oidcProviderName": "OpenID Connect"
+    },
+    "oidc": {
+      "_description": "in testing, move the config you want to last-place",
+      "enabled": false,
+
+      "_description": "broken config: fiddle with this config to test out different init failure modes",
+      "discoveryUrl": "http://example.com",
```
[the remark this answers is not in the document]
alxndrsn: Not just a hostname - also includes protocol, and optionally a path.  `.well-known...` can be optionally included - see https://github.com/panva/node-openid-client/blob/main/docs/README.md#issuerdiscoverissuer

lib/bin/cli.js
alxndrsn: I have moved the password output to stderr, and added `Password: ` prefix to make clear what is being printed.

config/oidc-development.json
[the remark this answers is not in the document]
sadiqkhoja: can we rename this field to `issuer` instead of `discoveryUrl`? and document possible values

sadiqkhoja (commented):
Should we write isolated integration tests for new endpoints defined in `lib/resources/oidc.js`? That would assert 

- Audit logs are created
- Response body and headers are correct

It would be nice to write unit tests for `lib/util/oidc.js`

matthew-white:
Ah sorry, as feared, it looks there's a conflict with the PRs I just merged.

lib/util/sessions.js
```diff
@@ -0,0 +1,52 @@
...
+const Problem = require('../util/problem');
+const { isBlank, noargs } = require('../util/util');
+const { getOrReject, rejectIf } = require('../util/promise');
+const { success } = require('../util/http');
+
+// TODO use req.protocol?
+const HTTPS_ENABLED = config.get('default.env.domain').startsWith('https://');
+
+// TODO add some thoughts about why __Host is important in prod but impossible in dev
+const SESSION_COOKIE = HTTPS_ENABLED ? '__Host-session' : 'session';
```
matthew-white, line 23: Thinking about this after working on #923: will the endpoints that delete the session cookie also need to choose between `__Host-session` and `session`?

lib/resources/users.js
```diff
@@ -31,52 +34,81 @@ module.exports = (service, endpoint) => {
...
-          .then((token) => mail(savedUser.email, 'accountCreated', { token })))
-        .then(always(savedUser)))));
+  if(oidc.isEnabled()) {
+    // Same as non-OIDC, except that password is random & unguessable
+    service.post('/users', endpoint(({ Users, mail }, { body, auth }) =>
+      auth.canOrReject('user.create', User.species)
+        .then(() => User.fromApi(body).forV1OnlyCopyEmailToDisplayName())
+        .then(Users.create)
+        // REVIEW: is this considered an acceptable way of blocking password-based login?
+        .then((savedUser) => Users.updatePassword(savedUser, randomPassword())
```
matthew-white, line 44: @ktuite brought up an interesting case during the meeting yesterday, of what happens if an existing deployment wants to start using SSO. I see that this endpoint would prevent password-based login for users created after the move to SSO, but should we also try to prevent password-based login for users created before the move? Is that already handled somehow?

lib/resources/oidc.js
```diff
@@ -0,0 +1,280 @@
...
+      <h2>TODO</h2>
+      <pre>
+        * stop displaying error details to client
+        * hide set-password UI
+        * don't request password from admin when creating new user account
+      </pre>
+    </div>
+    <div><a href="/">Continue to central</a></div>
+  `));
+}
```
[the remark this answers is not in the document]
matthew-white: Frontend would also provide i18n for any text.

lib/resources/users.js
alxndrsn: At this stage, enabling OIDC is all-in, so it will need to mentioned in documentation rather than handled in code.

lib/resources/oidc.js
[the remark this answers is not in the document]
alxndrsn: :+1: This should be possible for errors if specific codes are defined and then a corresponding view is implemented in frontend.

For the success path, I think the extra backend step is required to maintain the current session/csrf cookie settings.

lib/resources/users.js
matthew-white: I guess a server admin could use the Central CLI to give any existing users random passwords so that they can't use their old passwords. I can't think of a better way to automate the changing of passwords:

- Users could be deleted and recreated, but they wouldn't be linked to their previous activity.
- A password reset could be requested for every user, but that would send an email to everyone. (And maybe that endpoint just shouldn't exist anymore?)

Is using the CLI the strategy you recommend we mention in the docs? What do you mean that enabling OIDC is all-in?

matthew-white: > A password reset could be requested for every user

Ah that endpoint doesn't exist anymore (which makes a lot of sense), so that wouldn't even be an option.

lib/resources/oidc.js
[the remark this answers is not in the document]
matthew-white: > For the success path, I think the extra backend step is required to maintain the current session/csrf cookie settings.

Could you say more about why this is required? If this requirement is due to the setup in Frontend, maybe we could make changes there. Would it help to make the Frontend request to /v1/oidc/login async?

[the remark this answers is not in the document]
alxndrsn: Session ID and CSRF cookies are both `SameSite=Strict`:
https://github.com/getodk/central-backend/blob/facf33916322f4cea6e2b963372b4e7623715320/lib/resources/sessions.js#L39-L42
This means that if backend redirects to frontend at https://github.com/getodk/central-backend/blob/e7da23a3d3ac248f19a179e59a432c756f7c6361/lib/resources/oidc.js#L130-L146, the browser does not send cookies, so the user cannot be authorised.

There are tests for this at:

* https://github.com/getodk/central-backend/blob/e7da23a3d3ac248f19a179e59a432c756f7c6361/oidc-tester/playwright-tests/src/happy.spec.js
* https://github.com/getodk/central-backend/blob/e7da23a3d3ac248f19a179e59a432c756f7c6361/oidc-tester/playwright-tests/src/happy-next.spec.js

With a 302 redirect, the tests fail like this:

```
3) [firefox] › happy-next.spec.js:21:1 › can log in ──────────────────────────────────────────────

  AssertionError: No session cookie found!

     at utils.js:55

    53 |   console.log('requestCookies:', JSON.stringify(requestCookies, null, 2));
    54 |
  > 55 |   assert(requestCookies[SESSION_COOKIE], 'No session cookie found!');
       |   ^   
    56 |   assert(requestCookies['__csrf'],       'No CSRF cookie found!');
    57 |   assert.equal(Object.keys(requestCookies).length, 2, 'Unexpected requestCookie count!');
    58 | } 

      at assertLoginSuccessful (/odk-central-backend/oidc-tester/playwright-tests/src/utils.js:55:3)
      at /odk-central-backend/oidc-tester/playwright-tests/src/happy-next.spec.js:25:3
```

See e.g. https://github.com/alxndrsn/odk-central-backend/actions/runs/5666083718/job/15352099923#step:5:6265

.github/workflows/standard-suite.yml
alxndrsn, line 1: This is for convenience - these are the same tests that run on CircleCI.  This can be removed before merge.

lib/resources/oidc.js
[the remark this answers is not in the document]
matthew-white: OK, that's super helpful, thank you! I think I'm understanding things better now. Just to confirm:

1. An error results in a redirect to Frontend. The redirect specifies an error code as a query parameter, which Frontend uses to show an alert.
2. A success renders an HTML page. The page does a client-side redirect to Frontend and also links to Frontend.

That all sounds good to me. :+1: For the success page, what do you think about leaving that as unstyled as possible and not trying to make it look like Frontend? Since the user will only be on the page for a short time and is automatically redirected, I don't feel like it needs to be styled. I've seen that pattern elsewhere and feel like it wouldn't be overly confusing to users.

lib/resources/oidc.js
```diff
@@ -0,0 +1,292 @@
...
+      if(DEBUG_IDP_RESPONSE) {
+        return debugIdpResponse(res, session, userinfo, user);
+      }
+
+      // This redirect is neat, but breaks SameSite: Secure cookies.
+      //return redirect(302, '/'); // REVIEW: internally, the redirect() function throws... which is odd
+      // Instead, we need to render a page and then "browse" from that page to the normal frontend:
+      // TODO id=cl only set for playwright.. why can't it locate this anchor in any other way?
+
+      const nextPath = safeNextPathFrom(next);
```
matthew-white, line 137: If we add a generic redirect route in Frontend, maybe we could remove `safeNextPathFrom()` and use a path like this:

```js
const nextPath = `/#/redirect?next=${next}`;
```

Though maybe `next` needs to be encoded for use as a query parameter.

matthew-white:
I was just chatting with @alxndrsn about SSO. One thing that he mentioned is that he's added two new sets of tests:

- A GitHub action that runs the integration test suite using OpenID Connect instead of passwords
- A GitHub action that runs the Backend server alongside a fake OIDC server, then runs separate end-to-end tests in a browser. These tests log in, checking the successful case and a couple of failure modes.

One question that came up is whether it's OK that these tests are run as GitHub actions rather than in CircleCI. I personally don't see any issue with that. I think there are questions to think through around how we'll know if tests fail, but I think that's a general question and also something we can keep mulling after merging.

@sadiqkhoja, just to check about the status of review, have you reviewed how these new sets of tests are set up?

lib/resources/users.js
```diff
@@ -101,36 +130,27 @@ module.exports = (service, endpoint) => {
...
+          if (oidc.isEnabled()) {
+            // don't allow modifying email or password for users when using OIDC
+            // REVIEW is there some official, lower-level place to do this, e.g. in lib/model/frames.js?
+            const { email, password, ...filtered } = body;
+            return filtered;
+          } else {
+            return body;
+          }
+        })
+        .then((filteredBody) => User.fromApi(filteredBody))
```
matthew-white, line 143: > is there some official, lower-level place to do this, e.g. in lib/model/frames.js?

Yes, this is usually handled by the frame, specifically its `fromApi()` method. By default, the `fromApi()` method only accepts fields that have been marked `writable` on the frame (or on an aux frame). That means that `password` should be automatically filtered out because it's not a `writable` field.

`email` is a tougher case because we want to filter it out only sometimes. I think it's reasonable to do that here in the endpoint rather than the frame. An additional complication is that the release criteria now call for admins to be able to update emails.

Altogether, I'm thinking this could look something like:

```js
Users.getByActorId(params.id)
  .then(getOrNotFound)
  .then((user) => auth.canOrReject('user.update', user.actor)
    .then(() => (oidc.isEnabled() ? auth.can('user.update', User.species) : resolve(true))
      .then((canUpdateEmail) => (canUpdateEmail ? body : without(['email'], body))))
    .then((filteredBody) => User.fromApi(filteredBody))
```

matthew-white: Also, you can see that some user endpoints use async/await rather than promise chains. You should feel free to convert this endpoint to async/await if that makes things easier!

lib/resources/oidc.js
matthew-white: There's one complication here that I hadn't thought about, which is redirects to Enketo. If you navigate directly to an Enketo form that requires cookie auth, but you aren't logged into Central, then you're redirected to the Frontend login page, with `next` set to the Enketo form. Once you log into Central, Frontend sends you back to Enketo (in the `AccountLogin` component).

Right now, `safeNextPathFrom()` seems to always redirect the user to Frontend, which means that it doesn't work if `next` is to Enketo. If we replaced `safeNextPathFrom()` with a Frontend `/#/redirect` path, then we could have Frontend continue to handle all the logic around redirecting to Enketo. That feels like the easiest thing from a complexity standpoint: it's one less thing for Backend to do, and there'd be no code duplication between Frontend and Backend. Right now, I don't think Backend makes an assumptions about where Enketo is deployed. In contrast, Frontend assumes that Enketo is deployed at `/-`. It seems better for Backend not to be making a similar assumption and to continue to hand off to Frontend the responsibility of handling the `next` query parameter.

That said, I do think there's a performance cost to that choice. If `next` is to an Enketo form, then Backend will redirect to Frontend, which will redirect to Enketo. In other words, there's more redirecting happening than what's strictly necessary. All things considered, I lean toward having Frontend handle all `next` redirects, but I can see it both ways.

sadiqkhoja:
nb: Auth0 logout URL: https://<tenant_name>.auth0.com/oidc/logout

matthew-white:
> I was just chatting with @alxndrsn about SSO. One thing that he mentioned is that he's added two new sets of tests:
> 
> &hellip;
> 
> @sadiqkhoja, just to check about the status of review, have you reviewed how these new sets of tests are set up?

During the meeting today, @sadiqkhoja said that he's reviewed both sets of tests and that he's OK with the approach and tests. @sadiqkhoja has ideas for things we can do to improve testing in general, but those would come after .4.

config/oidc-development.json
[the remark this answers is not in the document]
alxndrsn: Seems like I didn't rename this previously.

For reference, the `discover()` function does not perform discovery: https://github.com/panva/node-openid-client/discussions/614

I have now renamed.

lib/resources/oidc.js
```diff
@@ -0,0 +1,292 @@
...
+      log('validated ID Token claims:', tokenSet.claims())
+
+      const { access_token } = tokenSet;
+
+      const userinfo = await client.userinfo(access_token);
+
+      const { email, email_verified } = userinfo;
+      if(!email) {
+        container.Sentry.captureException(new Error(`Required claim not provided in UserInfo Response: 'email'`));
+        return errorToFrontend(req, res, 'provider-misconfigured');
```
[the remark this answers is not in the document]
alxndrsn: > Is this error always due to provider misconfiguration? Could it be the result of privacy options that the user chose during the login process?

Great question; I can't confirm!

How about `email-claim-not-provided`?

lib/resources/oidc.js
alxndrsn: I've added an e2e test for enketo redirects, and implemented.

lib/util/oidc.js
```diff
@@ -0,0 +1,137 @@
...
+const log = (...args) => console.error('resources/oidc', ...args);
+
+const oidcConfig = config.has('default.oidc') && config.get('default.oidc') || {};
+
+function isEnabled() {
+  // This is AN EXPLICIT SETTING rather than derived from e.g. client init
+  // failing - we don't want to default to a different authN method to that
+  // requested by the system administrator.
+  return oidcConfig.enabled === true;
+}
```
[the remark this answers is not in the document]
alxndrsn: As discussed, we're sticking with an explicit *enabled* setting.

lib/resources/oidc.js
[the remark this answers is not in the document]
matthew-white: `email-claim-not-provided` sounds good to me.

matthew-white:
This may be addressed by other changes you're working on, but I'm noticing that when I log in on dev, a request is sent to /csp-report.

lib/formats/mail.js
```diff
@@ -32,6 +32,9 @@ const messages = {
   // Notifies a user that an account has been created with a predetermined password.
   accountCreatedWithPassword: message('ODK Central account created', '<html>Hello!<p>An account has been provisioned for you on an ODK Central server.</p><p>If this message is unexpected, simply ignore it. Your account was created with an assigned password. Please use that password to <a href="{{{domain}}}">sign in</a>.</p><p>If you have not been given the password, or you cannot remember it, you can reset it at any time at this link:</p><p><a href="{{{domain}}}/#/reset-password">{{{domain}}}/#/reset-password</a></p></html>'),
 
+  // Notifies a user that an account has been created for login exclusively with OIDC.
+  accountCreatedForOidc: message('ODK Central account created', '<html>Hello!<p>An account has been provisioned for you on an ODK Central data collection server.</p><p>If this message is unexpected, simply ignore it. Your account was created for login with {{{oidcProviderName}}}. Please go to <a href="{{{domain}}}">{{{domain}}}</a> to sign in.</p></html>'),
```
alxndrsn, line 36: @matthew-white do we want to drop `{{{oidcProviderName}}}` here?

matthew-white: That makes sense to me. I think we decided at some point not to use the provider name, so I think we should make this a generic email. We can also remove `oidcProviderName` from .json files.

alxndrsn: Done.

lib/resources/oidc.js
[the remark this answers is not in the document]
alxndrsn: Done :+1:

oidc-tester/.gitignore
alxndrsn, line 1: Rename this dir `oidc-dev`.

oidc-tester/odk-central-backend-config.json
alxndrsn, line 1: move to `./config/oidc-tester-docker.json`

lib/resources/users.js
alxndrsn: Comment updated.

lib/http/preprocessors.js
```diff
@@ -67,7 +68,7 @@ const authHandler = ({ Sessions, Users, Auth, bcrypt }, context) => {
     return authBySessionToken(authHeader.substring(7), () => reject(Problem.user.authenticationFailed()));
 
   // Basic Auth, which is allowed over HTTPS only:
```
sadiqkhoja, line 70: can you please update this comment as well

lib/http/preprocessors.js
```diff
@@ -67,7 +68,7 @@ const authHandler = ({ Sessions, Users, Auth, bcrypt }, context) => {
...
-  } else if (isPresent(authHeader) && authHeader.startsWith('Basic ')) {
+  } else if (!oidc.isEnabled() && isPresent(authHeader) && authHeader.startsWith('Basic ')) {
```
sadiqkhoja, line 71: Should we throw a new `Problem.user.ssoEnabled` that should tell user that we don't support Basic authentication when SSO is enabled? or mentioning this in user docs is sufficient

cc: @lognaturel @matthew-white

lib/http/service.js
```diff
@@ -74,6 +86,8 @@ module.exports = (container) => {
   require('../resources/analytics')(service, endpoint);
   require('../resources/datasets')(service, endpoint);
   require('../resources/entities')(service, endpoint);
+  // REVIEW: should these definitely be at /v1/ URLs and using standard preprocessor?
```
sadiqkhoja, line 89: I am fine with it

lib/http/service.js
```diff
@@ -11,6 +11,16 @@
 // defined elsewhere into an actual Express service. The only thing it needs in
 // order to do this is a valid dependency injection context container.
 
+// REVIEW: per CONTRIBUTING.md, perhaps this should be rewritten?  Specifically:
```
sadiqkhoja, line 14: "sparingly" is still true and `cookie-parser` is stock parser of expressjs, right?

lib/resources/oidc.js
```diff
@@ -0,0 +1,184 @@
...
+
+module.exports = (service, endpoint) => {
+  if (!isEnabled()) return;
+
+  service.get('/oidc/login', endpoint.html(async ({ Sentry }, _, req, res) => {
+    try {
+      const client = await getClient();
+      const code_verifier = generators.codeVerifier(); // eslint-disable-line camelcase
+
+      const code_challenge = generators.codeChallenge(code_verifier); // eslint-disable-line camelcase
```
sadiqkhoja, line 60: can we make these camelCase?

lib/resources/oidc.js
```diff
@@ -0,0 +1,184 @@
...
+      Sentry.captureException(err);
+      // hack to override the defaultErrorWriter TODO should be fixed elsewhere
+      err.isProblem = true;
+      throw err;
+    }
+  }));
+
+  service.get('/oidc/callback', endpoint.html(async (container, _, req, res) => {
+    try {
+      const code_verifier = req.cookies[CODE_VERIFIER_COOKIE]; // eslint-disable-line camelcase
```
sadiqkhoja, line 85: camelCase?

lib/resources/oidc.js
```diff
@@ -0,0 +1,184 @@
...
+  return user;
+}
+
+async function initSession(container, req, res, user) {
+  const applySession = await createUserSession(container, req.headers, user);
+  applySession(req, res);
+}
+
+// Logic adapted from `login.vue` in frontend.
+function safeNextPathFrom(next) {
```
sadiqkhoja, line 165: this can be moved to `lib/util/http` or `lib/util/util`

+ can you please add a unit test as well

lib/resources/oidc.js
```diff
@@ -0,0 +1,184 @@
...
+      res.cookie(CODE_VERIFIER_COOKIE, code_verifier, { ...callbackCookieProps, maxAge: ONE_HOUR });
+
+      const { next } = req.query;
+      if (next) res.cookie(NEXT_COOKIE, next, { ...callbackCookieProps, maxAge: ONE_HOUR });
+
+      redirect(307, authUrl);
+    } catch (err) {
+      Sentry.captureException(err);
+      // hack to override the defaultErrorWriter TODO should be fixed elsewhere
+      err.isProblem = true;
```
sadiqkhoja, line 78: I believe only `getClient();` can raise exception, what if we change that to throw `Problem.internal` instead of `new Error`?

sadiqkhoja (commented):
There are bunch of `TODOs`, to me all of them are about explanation of the code, I recommend that we merge this branch and update comments in a separate PR

lib/http/preprocessors.js
matthew-white: A new Problem sounds nice if it's easy to do. As is, if someone tries to use Basic Auth on a server that uses SSO, they won't receive a 401. Instead, they'll be allowed to access endpoints that don't require auth (e.g., /v1/projects), or they'll receive a 403. In contrast, if they try to use Basic Auth without using HTTPS, they'll immediate receive a 401 with a specific explanation. I think we could do something similar for SSO. I think that change could also come in a follow-up PR.

lib/bin/cli.js
matthew-white, line 1: Do you think we should disallow the use of these tools to set a user's password? That's what I suggested in the release criteria. We could use `oidc.isEnabled()` to determine whether SSO is enabled, then disallow password setting if it is. Changes would be needed for `user-create` and `user-set-password`.

Though maybe it doesn't matter that much? Even if a user has a password set, they won't be able to use it anywhere while SSO is enabled.

I think the experience for an administrator of a server using SSO when they create a new user would be improved though. Instead of wondering why they're being asked for a password or having to specify `--null-password`, they just wouldn't be asked for a password.

lib/resources/oidc.js
alxndrsn: We can, but it complicates their use and dissociates the local variables from their corresponding API usage.  Is that preferred?

lib/resources/oidc.js
alxndrsn: We can, but it complicates their use and dissociates the local variables from their corresponding API usage. Is that preferred?

test/util/authenticate-user.js
```diff
@@ -0,0 +1,97 @@
+// Allow main functionality to stay at top of file:
+/* eslint-disable no-use-before-define */
+
+const makeFetchCookie = require('fetch-cookie');
+
+module.exports = async (service, user, includeCsrf) => {
+  if (!user) throw new Error('Did you forget the **service** arg?');
```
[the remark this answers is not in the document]
alxndrsn: When refactoring tests, I found it easy to mistakenly write `authenticateUser('alice')` instead of `authenticateUser(service, 'alice')`.  This check was very helpful in catching those mistakes - without the check here, the eventual error was nebulous.  I think it will be helpful for future test writers to keep this check here.  But this branch is passing now, so I can delete it if it's not helpful.

Alternatively, maybe the function could be refactored to be an instance method of `service`.  Would that be preferred?

test/integration/api/users.js
```diff
@@ -494,20 +546,27 @@ describe('api: /users', () => {
...
-        asAlice.get('/v1/users/current')
-          .expect(200)
-          .then((before) => asAlice.patch(`/v1/users/${before.body.id}`)
-            .send({ email: 'david123@getodk.org' })
+    // eslint-disable-next-line func-names
+    it('should send an email to the user\'s previous email when their email changes', function () {
+      // REVIEW or could exclude _outside_ it() as per larger blocks of tests in this file.
+      // mocha marks skipped tests as "pending", which doesn't seem appropriate in the case
+      // of uname/pword vs OIDC auth.
+      if (process.env.TEST_AUTH === 'oidc') return this.skip();
```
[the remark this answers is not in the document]
alxndrsn: There is a difference:

* skipped tests appear in the test report as explicitly skipped, whereas
* a plain `return` means that nothing is included in the test report at all

@matthew-white did you have an opinion on this?  I thought it was a comment on this PR, but I can't find it.

oidc-dev/scripts/wait-for-it.sh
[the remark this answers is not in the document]
alxndrsn: I'll give this a go.  It might also be a helpful change to `central` - https://github.com/getodk/central/blob/master/service.dockerfile#L28 / https://github.com/getodk/central/blob/master/files/service/scripts/wait-for-it.sh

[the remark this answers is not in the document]
alxndrsn: central PR at https://github.com/getodk/central/pull/475

lib/bin/cli.js
alxndrsn: I've updated `cli.js`; please review.

test/integration/api/users.js
```diff
@@ -472,13 +518,19 @@ describe('api: /users', () => {
...
-              after.body.email.should.equal('newalice@odk.org');
               should.not.exist(after.body.meta);
               before.body.createdAt.should.equal(after.body.createdAt);
               after.body.updatedAt.should.be.a.recentIsoDate();
-              return service.post('/v1/sessions')
-                .send({ email: 'newalice@odk.org', password: 'alice' })
-                .expect(200);
+
+              if (process.env.TEST_AUTH === 'oidc') {
+                after.body.email.should.equal('alice@getodk.org');
```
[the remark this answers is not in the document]
alxndrsn: What's the motivation for this?  To me, the current code seems more explicit and easier to compare to the assertion below:
```js
after.body.email.should.equal('alice@getodk.org');
...
after.body.email.should.equal('newalice@odk.org');
```
vs.

```js
after.body.email.should.equal(before.body.email);
...
after.body.email.should.equal('newalice@odk.org');
```

lib/resources/oidc.js
alxndrsn: See above :slightly_smiling_face:

lib/util/oidc.js
```diff
@@ -0,0 +1,134 @@
...
+function getRedirectUri() {
+  return `${config.get('default.env.domain')}/v1/oidc/callback`;
+}
+
+let clientLoader; // single instance, initialised lazily
+function getClient() {
+  if (!clientLoader) clientLoader = initClient();
+  return clientLoader;
+}
+async function initClient() {
```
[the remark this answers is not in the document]
alxndrsn: Which part should be unit-tested?  Validation of issuer metadata?

lib/util/html.js
```diff
@@ -0,0 +1,42 @@
...
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+//
+// Helper functions that relate to the HTTP/service layer of the application.
+
+// handy dev function for enabling syntax hilighting of html
+const html = ([ first, ...rest ], ...vars) => first + vars.map((v, idx) => [ v, rest[idx] ]).flat().join('');
+
+// Style to look like odk-central-frontend
```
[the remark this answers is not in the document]
alxndrsn: @matthew-white can you share some input here?  How should this page look?

lib/util/html.js
```diff
@@ -0,0 +1,42 @@
...
+const frontendPage = ({ head='', body }) => html`
```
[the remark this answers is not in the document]
alxndrsn: Should it be moved?

test/util/authenticate-user.js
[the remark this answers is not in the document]
sadiqkhoja: > the function could be refactored to be an instance method of service

I like the idea

test/integration/api/users.js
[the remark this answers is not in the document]
sadiqkhoja: it's easier to read that "before email" and "after email" are same.

lib/util/html.js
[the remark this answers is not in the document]
matthew-white: I had an idea about this [above](https://github.com/getodk/central-backend/pull/910#discussion_r1277685758):

> For the success page, what do you think about leaving that as unstyled as possible and not trying to make it look like Frontend? Since the user will only be on the page for a short time and is automatically redirected, I don't feel like it needs to be styled. I've seen that pattern elsewhere and feel like it wouldn't be overly confusing to users.

I think trying to match the style of Frontend would lead to a maintenance burden. Frontend styles will inevitably change over time, and it'd be good not to have to change them in two places.

lib/bin/cli.js
matthew-white: Looks great!

lib/resources/users.js
```diff
@@ -101,36 +130,27 @@ module.exports = (service, endpoint) => {
     Users.getByActorId(params.id)
       .then(getOrNotFound)
       .then((user) => auth.canOrReject('user.update', user.actor)
-        .then(() => User.fromApi(body))
+        .then(() => {
+          if (oidc.isEnabled()) {
+            // Don't allow modifying email or password for users when using OIDC.
+            // This can't be done in /frames/ as it does not allow conditional filtering.
+            const { email, password, ...filtered } = body;
+            return filtered;
```
matthew-white, line 138: The release criteria suggest that admins still need to be able to update users' email addresses. @sadiqkhoja left a comment about that in the release criteria, but if we do decide to run with that behavior, this logic would need to change.

Also, `password` isn't a `writable` field on the frame, so I don't think it needs to be filtered out here: `fromApi()` should always filter it out. This endpoint can't be used to change a password: that's a different endpoint.

alxndrsn: So can this whole block be deleted?

matthew-white: I think there does need to be something here. If SSO is enabled, we don't want non-admin users to use this endpoint to update their own email. Note that in `auth`, users can `user.update` themselves even if they can't `user.update` other users:

https://github.com/getodk/central-backend/blob/cedadf298742dcf4c3c320ac27f2d0fff95b7476/lib/model/query/auth.js#L26-L29

Non-admin users should still be able to use this endpoint to update their own display name.

I think this means that we should filter out `email` if users can't `user.update` users in general. I think you can check for that by calling `auth.can('user.update', User.species)`, which returns a promise.

lognaturel: I personally have no preference. @issa-tseng maybe you have strong feelings?

I think we could say admins can change their own email because they should have a lot more context on the implications. This is also more helpful in contexts in which there’s only one site wide admin. 

It would also be very defensible to say admins need to use the command line or talk to another admin to get their email address changed. 

Whatever it is, it should be in the docs.

lib/util/html.js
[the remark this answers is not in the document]
alxndrsn: > Frontend styles will inevitably change over time

How frequently?

I think a branded forwarding page maintains users' trust that they are inside a legitimate workflow.

test/integration/other/analytics-queries.js
```diff
@@ -101,7 +101,11 @@ const submitToForm = (service, user, projectId, xmlFormId, xml, deviceId = 'abcd
 ////////////////////////////////////////////////////////////////////////////////
 // Tests!
 ////////////////////////////////////////////////////////////////////////////////
-describe('analytics task queries', () => {
+// eslint-disable-next-line func-names
+describe('analytics task queries', function () {
+  // increasing timeouts on this set of tests
+  this.timeout(8000);
+
```
alxndrsn, line 108: This change should be introduced separately.  Separate PR at https://github.com/getodk/central-backend/pull/967.

lib/http/preprocessors.js
alxndrsn: What would you recommend?

lib/util/sessions.js
```diff
@@ -0,0 +1,47 @@
...
+
+const config = require('config');
+
+// TODO use req.protocol?
+const HTTPS_ENABLED = config.get('default.env.domain').startsWith('https://');
+
+// TODO add some thoughts about why __Host is important in prod but impossible in dev
+const SESSION_COOKIE = HTTPS_ENABLED ? '__Host-session' : 'session';
+
+function createUserSession({ Audits, Sessions }, headers, user) {
```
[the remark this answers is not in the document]
alxndrsn: > This doesn't look like a utility function. Where this should reside?

The current location was my best guess.  What qualifies something as a utility function?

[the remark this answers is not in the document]
alxndrsn: This has been converted to an anonymous function 👍

test/integration/other/analytics-queries.js
alxndrsn: #967 has been merged, so this change can be reverted from this PR.

lib/resources/oidc.js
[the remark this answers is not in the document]
alxndrsn: There's a parallel discussion at https://github.com/getodk/central-backend/pull/910/files#r1310995967:

I've also seen the pattern of unstyled forwarding pages elsewhere and I'm not a fan!  I think a branded forwarding page maintains users' trust that they are inside a legitimate workflow.

alxndrsn:
> nb: Auth0 logout URL: https://<tenant_name>.auth0.com/oidc/logout

N.B. per https://auth0.com/docs/authenticate/login/logout/log-users-out-of-auth0, it's `https://{yourDomain}/oidc/logout`, and should be detectable via the standard OpenID Connect mechanism.  Auth0's OIDC logout is currently behind a feature-flag, and enabling it for a tenant requires a request to their support team.

lib/util/html.js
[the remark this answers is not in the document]
matthew-white: > How frequently?

Styles that would affect this page: likely not often, but definitely sometimes.

Already, the styles don't quite match. There's a max width on `.panel-main` that seems to be missing here. You can see that in a wider viewport. Here's a panel as rendered by Frontend, where you can see the max width:



Here the max width is missing:



That change happened in getodk/central-frontend#775. Follow-up work is planned at getodk/central-frontend#785, which is an example of how styles evolve over time.

> I think a branded forwarding page maintains users' trust that they are inside a legitimate workflow.

If it didn't come with any cost, I'd be in favor of it. But given that it requires ongoing maintenance that will be easy to forget, I don't think we should render a page like that. Redirect pages are often styled very simply, so I don't expect that user trust will be broken.

This definitely shouldn't block merging though. I'd be happy to file an issue and continue discussing there.

lib/resources/users.js
matthew-white: Whatever the behavior is, we should also make sure that Frontend matches it. Right now, Frontend will allow an admin to change their own email.

lib/formats/mail.js
```diff
@@ -32,6 +32,9 @@ const messages = {
   // Notifies a user that an account has been created with a predetermined password.
   accountCreatedWithPassword: message('ODK Central account created', '<html>Hello!<p>An account has been provisioned for you on an ODK Central server.</p><p>If this message is unexpected, simply ignore it. Your account was created with an assigned password. Please use that password to <a href="{{{domain}}}">sign in</a>.</p><p>If you have not been given the password, or you cannot remember it, you can reset it at any time at this link:</p><p><a href="{{{domain}}}/#/reset-password">{{{domain}}}/#/reset-password</a></p></html>'),
 
+  // Notifies a user that an account has been created for login exclusively with OIDC.
+  accountCreatedForOidc: message('ODK Central account created', '<html>Hello!<p>An account has been provisioned for you on an ODK Central data collection server.</p><p>If this message is unexpected, simply ignore it. Please go to <a href="{{{domain}}}">{{{domain}}}</a> to sign in.</p></html>'),
```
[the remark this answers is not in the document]
matthew-white: Resolved by #969 I think.

lib/util/html.js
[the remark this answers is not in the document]
matthew-white: I've created an issue for this: #971

lib/resources/oidc.js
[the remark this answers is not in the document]
matthew-white: I've filed a related issue: #971

lib/http/preprocessors.js
matthew-white: How about "Basic Auth, which is allowed only over HTTPS and only if SSO is not enabled"? Or "Basic Auth, which is allowed only if certain conditions are true"? I think right now, the comment seems a little outdated since it mentions HTTPS but not SSO.

lib/resources/oidc.js
matthew-white: Usually, I think camelCase is much preferred over snake_case. However, I think @alxndrsn's explanation makes sense as to why this is an exception to the rule.

test/integration/api/users.js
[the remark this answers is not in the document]
matthew-white: Just to offer my two cents, I can see this both ways. `before.body.email` makes the intention clear, while `'alice@getodk.org'` is more explicit and arguably makes the if/else more parallel. Probably OK to leave it as-is?

lib/util/sessions.js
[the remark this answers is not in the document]
matthew-white: I agree that it's surprising to see `createUserSession()` here, since it contains specific logic related to endpoints/resources/responses. Usually such logic is located in lib/resources. What makes this a particularly interesting case is that this logic is shared across multiple resource files (sessions.js and oidc.js). It's hard to think of a similar case, which is why I'm guessing @sadiqkhoja suggested lib/resources/shared.

That said, I can see why it's nice for `createUserSession()` to be in the same file as `SESSION_COOKIE` and `HTTPS_ENABLED`. `SESSION_COOKIE` is also used in lib/http/preprocessors.js. Since that's in a different directory from lib/resources, that might be a strike against moving this code to inside lib/resources.

lib/http seems like a reasonable location to me. That directory contains endpoint/resource/response logic that's shared across endpoints (in some cases, logic that's shared across all endpoints, and in other cases, logic that's shared across a smaller subset of endpoints). I also understand though that it's sometimes hard to find the exact right place for a function and that sometimes these sorts of distinctions can get blurry or arbitrary. @alxndrsn, I think you should make a decision one way or another for now, then we can continue thinking in general about what goes where.

lib/util/html.js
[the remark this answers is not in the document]
matthew-white: My instinct is also that lib/formats is the right place for `frontendPage()`. Maybe for `html()` as well.

lib/http/preprocessors.js
alxndrsn: PR at https://github.com/getodk/central-backend/pull/982

lib/util/html.js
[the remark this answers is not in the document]
alxndrsn: No longer exists :upside_down_face:

PR #915, opened 2023-06-26
enhances #722: add ability to select parentID

### LINKED ISSUE #722: OData $select: add ability to select __parent-table-id
OData `$metadata` document shows parent table ID like `__submission-id` as a field for the subtables but selecting it returns 400 error.

Excel OData feed uses metadata document to determine the valid fields for the table and then makes request with $select query parameter. Right now request with $select for subtables are failing with 400 error and Excel falls back to requesting without $select

sadiqkhoja:
Closes #722 

Added integration test
Manually verified via Excel and checked logs 

NA

NA

NA

### DIFF
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -544,7 +544,7 @@ const filterFields = (formFields, select, table) => {
       filteredFields.add(fieldTree[path].value);
 
       if (fieldTree[path].value.type === 'repeat' && i < tableSegments.length-1) {
-        parentIdKey = `__Submissions${path.replace(/\//g, '-')}`;
+        parentIdKey = `__Submissions${path.replaceAll('/', '-')}`;
       }
     }
     parentIdKey += '-id';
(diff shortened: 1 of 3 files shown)


lib/formats/odata.js
```diff
@@ -527,42 +527,55 @@ const filterFields = (formFields, select, table) => {
...
     }
+    parentIdKey += '-id';
+    idKeys.add(parentIdKey);
   }
 
   for (const property of select.split(',').map(p => p.trim())) {
     // validate metadata properties. __system/.. properties are only valid for Submission table
-    if (property.startsWith('__id') || property.startsWith('__system')) {
-      if (!(property === '__id' || (table === 'Submissions' && systemFields.has(property))))
-        throw Problem.user.propertyNotFound({ property });
```
matthew-white, line 543: Could you add a test that `$select`-ing a `__system` property for a subtable results in a 404 response? This `if` used to handle it, but now it's handled by different logic after the `continue`.

matthew-white (approved):

matthew-white:
I think @sadiqkhoja would know best, but I feel like it's probably safe to include this in the patch. Only OData should be affected by this change, and OData tests continue to pass. Sadiq also verified things manually in Excel.

sadiqkhoja:
This is safe to release in the patch, I have performed sanity test on staging after the merge as well

PR #919, opened 2023-07-05
Update to node 18

alxndrsn:
Node 16 EOL is September: https://nodejs.dev/en/about/releases/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   build:
     docker:
-      - image: cimg/node:16.19.1
+      - image: cimg/node:18.16
       - image: cimg/postgres:14.6
         environment:
           POSTGRES_PASSWORD: odktest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   build:
     docker:
-      - image: cimg/node:18.16
+      - image: cimg/node:18.17.0
       - image: cimg/postgres:14.6
         environment:
           POSTGRES_PASSWORD: odktest
--- a/.github/workflows/soak-test.yml
+++ b/.github/workflows/soak-test.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.19.1]
+        node-version: [18]
     services:
       # see: https://docs.github.com/en/enterprise-server@3.5/actions/using-containerized-services/creating-postgresql-service-containers
       postgres:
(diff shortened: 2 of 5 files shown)


matthew-white:
Thanks, @alxndrsn! We were wanting to upgrade to Node 18 this release (getodk/central#432). Would you mind updating the central and central-frontend repos as well?

@sadiqkhoja, you might be a good person to do the review, since you worked on a previous Node upgrade for Central.

alxndrsn:
Test is failing:
```js
  1) endpoints                                                                  
       default format (outputter)                                                  
         should fail semigracefully on PartialPipe stream error:                   
     Uncaught TypeError: Cannot read properties of undefined (reading 'should') 
      at /odk/backend/test/unit/http/endpoint.js:503:18                            
      at /odk/backend/node_modules/streamtest/src/index.js:245:11                  
      at Writable.<anonymous> (node_modules/streamtest/src/index.js:238:9)         
      at Writable.emit (node:events:525:35)                                        
      at Writable.emit (node:domain:489:12)                                        
      at emitErrorNT (node:internal/streams/destroy:151:8)                         
      at emitErrorCloseNT (node:internal/streams/destroy:116:3)                    
      at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

This seems to be because the order of error propagation has changed:

### node 16:

1. `test/unit/http/endpoint.js`'s `streamErrorHandler()`
2. `streamtest`'s `error` event listener

### node 18:

1. `streamtest`'s `error` event listener
2. `test/unit/http/endpoint.js`'s `streamErrorHandler()`

alxndrsn:
Possibly related to this:

> `stream.pipeline()` closes all the streams when an error is raised. The `IncomingRequest` usage with `pipeline` could lead to an unexpected behavior once it would destroy the socket without sending the expected response.
> - _https://nodejs.dev/en/api/v18/stream/#streampipelinesource-transforms-destination-callback_
> - _https://nodejs.dev/en/api/v16/stream/#streampipelinesource-transforms-destination-callback_

alxndrsn:
It may be that the approach to error handling at https://github.com/getodk/central-backend/blob/master/lib/http/endpoint.js#L200 > https://github.com/getodk/central-backend/blob/master/lib/util/stream.js#L100-L101 is unsuitable.  Note the docs for the `callback` argument of `pipeline()` say:

>  [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) Called when the pipeline **is fully done**.

sadiqkhoja (approved):
A short summary of what has changed related stream would be really helpful.

alxndrsn:
> A short summary of what has changed related stream would be really helpful.

I think it deserves its own PR.  I've opened it separately at https://github.com/getodk/central-backend/pull/922.  Does that PR need more explanation?  Would you prefer the changes+comments are just included here?

PR #920, opened 2023-07-07
dev: allow using wrong node version deliberately

alxndrsn:


alxndrsn:
This was very useful when working on https://github.com/getodk/central-backend/pull/919 and trying to run the same test against different versions of node.  E.g.

```sh
nvm use 16 && ALLOW_ANY_NODE_VERSION=1 make test-unit &&
nvm use 17 && ALLOW_ANY_NODE_VERSION=1 make test-unit &&
nvm use 18 && ALLOW_ANY_NODE_VERSION=1 make test-unit
```

Issue #657, opened 2022-11-07
Error: Premature close


alxndrsn:
```js
Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at new NodeError (node:internal/errors:387:5)
    at ServerResponse.onclose (node:internal/streams/end-of-stream:145:30)
    at ServerResponse.emit (node:events:525:35)
    at ServerResponse.emit (node:domain:489:12)
    at Socket.onServerResponseClose (node:_http_server:258:23)
    at Socket.emit (node:events:525:35)
    at Socket.emit (node:domain:489:12)
    at TCP.<anonymous> (node:net:757:14)
```

"Premature close" has been mentioned in previous issues:

* https://github.com/getodk/central-backend/issues/485
* https://github.com/getodk/central-backend/pull/480
* https://github.com/getodk/central-backend/issues/521
* https://github.com/getodk/central-backend/pull/637

The error can be recreated fairly reliably (substitute IDs and auth credentials as appropriate):
```bash
for i in {1..20};do \
echo; echo "Starting iteration #$i..."; echo; \
(curl "localhost:8383/------------------------------ITERATION-[$i]-BEGINNING------------------------------"; \
tim=$((3 + $RANDOM % 12)); \
echo; echo "Running for $tim seconds..."; echo; \
timeout $tim siege -v \
http://localhost:8989/v1/projects/1/forms/all-widgets.svc/Submissions \
-H 'Authorization: Basic ...'); \
done
```

yanokwa:
Can you also trigger a true premature close (e.g., cancelling a download halfway or shutting of the connection)? It'd be ideal if we could identify and ignore true premature closes.

alxndrsn:
More consistent reproduction:
```js
const { Socket } = require('node:net');

const s = new Socket(); 

s.on('connect', () => {
  s.write(
    'GET /v1/projects/1/forms/all-widgets.svc/Submissions HTTP/1.1\r\n' +
    'Host: localhost:8989\r\n' +
    'User-Agent: curl/7.68.0\r\n' +
    'Accept: */*\r\n' +
    'Authorization: Basic ...\r\n' +
    '\r\n' +
  '');
});
s.on('data', data => console.log('RECEIVED:', data.toString()));

s.connect('8989', 'localhost');
s.pause();
```

alxndrsn:
Simpler reproduction:
```sh
[...]

PR #922, opened 2023-07-08
streamErrorHandler: write error trailer before closing stream

alxndrsn:
Per nodejs docs:

> `stream.pipeline()` closes all the streams when an error is raised. The `IncomingRequest` usage with `pipeline` could lead to an unexpected behavior once it would destroy the socket without sending the expected response.
> -_https://nodejs.org/docs/latest-v16.x/api/stream.html#streampipelinestreams-callback_
> -_https://nodejs.org/docs/latest-v18.x/api/stream.html#streampipelinestreams-callback_

This change is required for the build to pass on [node 18](https://github.com/getodk/central-backend/pull/919).


lib/util/stream.js
```diff
@@ -98,7 +98,10 @@ class PartialPipe {
     }
   }
   with(stream) { return new PartialPipe(this.streams.concat([ stream ])); }
-  pipeline(errCb) { return pipeline(this.streams, rejectIfError(errCb)); }
+  pipeline(errHandler) {
+    this.streams[this.streams.length-2].on('error', errHandler);
+    return pipeline(this.streams, () => {});
```
alxndrsn, line 103: Maybe this change is specific to `streamErrorHandler()`, and should only be implemented in `lib/http/endpoint.js`.

alxndrsn: Teesting at https://app.circleci.com/pipelines/github/alxndrsn/odk-central-backend/228/workflows/8564a967-914b-4e96-8b00-f6a338fc043d/jobs/240

lib/http/endpoint.js
```diff
@@ -190,16 +190,19 @@ const endpointBase = ({ preprocessor = noop, before = noop, resultWriter, errorW
...
-  response.end();
   next(err);
 };
 const defaultResultWriter = (result, request, response, next) => {
   if (!response.hasHeader('Content-Type')) response.type('json');
 
   if (result instanceof PartialPipe) {
-    result.with(response).pipeline(streamErrorHandler(response, next));
+    const errHandler = streamErrorHandler(response, next);
+    result.streams[result.streams.length-1].on('error', errHandler);
```
sadiqkhoja, line 200: what if there is an error in non-last (length-1) stream? Does it chain to the last stream so listening for error on the last stream is enough

sadiqkhoja (approved):
Couple of clarification questions

lib/http/endpoint.js
```diff
@@ -190,16 +190,19 @@ const endpointBase = ({ preprocessor = noop, before = noop, resultWriter, errorW
...
+    result.with(response).pipeline(() => {});
   } else if (result.pipe != null) {
-    pipeline(result, response, rejectIfError(streamErrorHandler(response, next)));
+    const errHandler = streamErrorHandler(response, next);
+    result.on('error', errHandler);
+    pipeline(result, response, () => {});
```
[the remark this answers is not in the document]
alxndrsn: The docs for `pipeline()` don't state that the argument is optional: https://nodejs.org/docs/latest-v18.x/api/stream.html#streampipelinesource-transforms-destination-callback

It may be that the actual error handler should be `err => next(err)`, and `next` should be removed from the `streamErrorHandler()` function.

lib/http/endpoint.js
alxndrsn: That's my understanding from the [`pipeline()` docs](https://nodejs.org/docs/latest-v18.x/api/stream.html#streampipelinesource-transforms-destination-callback), which is what `PartialPipe` uses internally.

alxndrsn:
Related to https://github.com/getodk/central-backend/issues/657?

PR #923, opened 2023-07-12
Add endpoint to delete current session

matthew-white:
This PR adds an endpoint to delete the current session, making it easy to log out without specifying the token of the current session in the URL. We will probably use this endpoint in Frontend, but I also imagine that some API users would find it convenient.

New tests. I copied some of the tests of DELETE /v1/sessions/:token. I didn't copy all the tests, since the two endpoints call a shared function: I think the tests of DELETE /v1/sessions/:token provide good coverage of that shared behavior.

I created a shared function because I think there's enough logic shared between DELETE /v1/sessions/:token and the new DELETE /v1/sessions/current to warrant one. DELETE /v1/sessions/current ends up being quite short and looks very similar to GET /v1/sessions/restore.

I chose the path &hellip;/current to be similar to another endpoint we have, /v1/users/current.

API users will be able to use the new endpoint. The DELETE /v1/sessions/:token endpoint was modified, but much of its code was copied as-is to the shared function. Existing tests continue to pass.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -30,6 +30,12 @@ Finally, **system information and configuration** is available via a set of spec
 
 Here major and breaking changes to the API are listed by version.
 
+### ODK Central v2023.4
+
+**Added**:
+
+- [DELETE /sessions/current](/reference/authentication/session-authentication/logging-out-current-session) logs out the current session.
+
 ### ODK Central v2023.3
 
 **Added**:
@@ -387,7 +393,7 @@ _(There is not really anything at `/v1/example`; this section only demonstrates 
 
 #### Logging out [DELETE /v1/sessions/{token}]
 
-Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token. It is also the way Public Link and App User access is revoked. To do so, issue a `DELETE` request to that token resource.
+Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token. It is also the way Public Link and App User access are revoked. To do so, issue a `DELETE` request to that token resource.
 
 + Parameters
     + token: `lSpAIeksRu1CNZs7!qjAot2T17dPzkrw9B4iTtpj7OoIJBmXvnHM8z8Ka4QPEjR7` (string, required) - The session bearer token, obtained at login time.
@@ -419,9 +425,6 @@ Only the session that was used to authenticate the request is logged out. If the
 
 + Response 403 (application/json)
     + Attributes (Error 403)
-
-+ Response 404 (application/json)
-    + Attributes (Error 404)
 
 ## HTTPS Basic Authentication [/v1/example]
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -387,7 +387,7 @@ _(There is not really anything at `/v1/example`; this section only demonstrates 
 
 #### Logging out [DELETE /v1/sessions/{token}]
 
-Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token. It is also the way Public Link and App User access are revoked. To do so, issue a `DELETE` request to that token resource.
+Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token. It is also the way Public Link and App User access is revoked. To do so, issue a `DELETE` request to that token resource.
 
 + Parameters
     + token: `lSpAIeksRu1CNZs7!qjAot2T17dPzkrw9B4iTtpj7OoIJBmXvnHM8z8Ka4QPEjR7` (string, required) - The session bearer token, obtained at login time.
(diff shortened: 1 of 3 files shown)


matthew-white:
This new endpoint should help with getodk/central#400.

lib/resources/sessions.js
```diff
@@ -48,24 +48,30 @@ module.exports = (service, endpoint) => {
   service.get('/sessions/restore', endpoint((_, { auth }) =>
     auth.session.orElse(Problem.user.notFound())));
 
+  const logOut = (Sessions, auth, session) =>
+    auth.canOrReject('session.end', session.actor)
+      .then(() => Sessions.terminate(session))
+      .then(() => (_, response) => {
+        // revoke the cookie associated w the session, if the session was used to
+        // terminate itself.
+        // TODO: repetitive w above.
```
alxndrsn, line 57: what does this TODO imply?

test/integration/api/sessions.js
```diff
@@ -269,6 +269,35 @@ describe('api: /sessions', () => {
...
+      const { body } = await service.delete('/v1/sessions/current')
+        .set('Authorization', `Bearer ${token}`)
+        .expect(200);
+      body.should.eql({ success: true });
+      await service.get('/v1/users/current')
+        .set('Authorization', `Bearer ${token}`)
+        .expect(401);
+    }));
+
+    it('should not allow app users to delete their own sessions', testService(async (service) => {
```
alxndrsn, line 291: Why not?  How do app users log out?

docs/api.md
```diff
@@ -403,6 +403,26 @@ Logging out is not strictly necessary for Web Users; all sessions expire 24 hour
...
+#### Logging out current session [DELETE /v1/sessions/current]
+
+This endpoint causes the current session to log itself out. Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token.
+
+Only the session that was used to authenticate the request is logged out. If the Actor associated with the session has other sessions as well, those are not logged out.
+
++ Request
+    + Headers
+
+            Authorization: Bearer lSpAIeksRu1CNZs7!qjAot2T17dPzkrw9B4iTtpj7OoIJBmXvnHM8z8Ka4QPEjR7
```
alxndrsn, line 415: For web users, the token will also be sent as a cookie, and perhaps in more cases.  Would it be helpful to also allow supplying the token-to-delete via cookie?

lib/resources/sessions.js
```diff
@@ -48,24 +48,30 @@ module.exports = (service, endpoint) => {
   service.get('/sessions/restore', endpoint((_, { auth }) =>
     auth.session.orElse(Problem.user.notFound())));
 
+  const logOut = (Sessions, auth, session) =>
+    auth.canOrReject('session.end', session.actor)
+      .then(() => Sessions.terminate(session))
+      .then(() => (_, response) => {
+        // revoke the cookie associated w the session, if the session was used to
+        // terminate itself.
```
alxndrsn, line 56: When do you foresee clients invalidating _other_ sessions?  Is there even any point in authorizing a request to invalidate a session?  If a client knows a session token, then isn't that enough authority to invalidate that session?

docs/api.md
matthew-white: The token could be supplied via a cookie. Frontend might do that. However, we don't encourage API users to use cookie auth.

test/integration/api/sessions.js
matthew-white: App users auth using a session token. When access is revoked for an app user, that session is deleted. Project managers can revoke access for app users, but app users cannot do so themselves. App users are granted very limited rights and are often shared by multiple people.

lib/resources/sessions.js
matthew-white: DELETE /v1/sessions/{token} is used for multiple things. One of them is for logging yourself out, but another is revoking access for app users and public links. See the comment above about app users. Project managers can revoke app users' sessions, but someone with a token for an app user or public link shouldn't be able to revoke its access.

On the other hand, DELETE /v1/sessions/current will probably just be used for logout.

lib/resources/sessions.js
matthew-white: Heh I'm actually not 100% sure, mostly I'm just copying it over. I'm guessing that what's repetitive is the cookie attributes, which are repeated in the session create endpoint and the session delete endpoint.

I'd be happy to try to resolve this comment by creating a function to return these two cookies with the correct attributes. Both endpoints would call that function.

docs/api.md
```diff
@@ -403,6 +403,26 @@ Logging out is not strictly necessary for Web Users; all sessions expire 24 hour
 + Response 403 (application/json)
     + Attributes (Error 403)
 
+#### Logging out current session [DELETE /v1/sessions/current]
+
+This endpoint causes the current session to log itself out. Logging out is not strictly necessary for Web Users; all sessions expire 24 hours after they are created. But it can be a good idea, in case someone else manages to steal your token.
```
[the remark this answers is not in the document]
matthew-white: Mostly I was just copying this from the docs above for DELETE /v1/sessions/{token}. I guess once you're done with your API session, you might as well delete your token if you're not going to use it again. Or yeah, on a shared computer. I'd be happy to remove or reword this.

docs/api.md
```diff
@@ -403,6 +403,26 @@ Logging out is not strictly necessary for Web Users; all sessions expire 24 hour
...
+
+            Authorization: Bearer lSpAIeksRu1CNZs7!qjAot2T17dPzkrw9B4iTtpj7OoIJBmXvnHM8z8Ka4QPEjR7
+
++ Response 200 (application/json)
+    + Attributes (Success)
+
++ Response 403 (application/json)
+    + Attributes (Error 403)
+
++ Response 404 (application/json)
```
matthew-white, line 423: Most endpoints don't document the possibility of a 404, so I'll remove this.

alxndrsn (approved):

PR #925, opened 2023-07-13
Add ability to specify CSRF token in header

matthew-white:
If a non-GET request is sent, and a session cookie is used to authenticate, then the CSRF token associated with the session must also be provided. Right now, the CSRF token is expected to be in the request body. For example, Frontend provides the CSRF token from the iframe form in the [`SubmissionDownload`](https://github.com/getodk/central-frontend/blob/0a3d8d265da634dbc43f0ae7e0c0eff3ec69044a/src/components/submission/download.vue#L213-L258) component. Enketo also provides the CSRF token when using cookie auth.

However, I think the expectation that the CSRF token is in the request body doesn't work for all requests. The main issue that comes to mind is file uploads. For example, I think requiring the CSRF token to be in the request body would be an obstacle to using cookie auth to upload a form definition or form attachment.

This PR makes it possible to use cookie auth in such cases by allowing the CSRF token to be specified in a header instead of the request body.

New tests.

I don't love that there are now two ways to specify the CSRF token. I thought about removing support for CSRF tokens in the request body so that there would be only one way to specify the token (via the header). However, that wouldn't work for HTML form submission like from the iframe form. I also didn't want to change anything about our Enketo integration.

I also don't think this PR adds that much complexity. Only two lines of code were changed.

If a user of cookie auth has a choice between specifying the token in the header vs. in the request body, specifying it in the header is probably slightly better. That's because it doesn't require the token to be removed from the request body.

During regression testing, we will verify that the parts of Central that rely on cookie auth continue to work.

I don't think changes are needed. We discourage the use of cookie auth in the API docs, and we don't document any requirements around the CSRF token.

### DIFF
--- a/test/integration/api/sessions.js
+++ b/test/integration/api/sessions.js
@@ -271,7 +271,19 @@ describe('api: /sessions', () => {
 
   // this isn't exactly the right place for this but i just want to check the
   // whole stack in addition to the unit tests.
-  describe('cookie CSRF auth', () => {
+  describe('CSRF protection for cookie auth used for non-GET requests', () => {
+    it('should reject with a 403 if session token is invalid', testService(async (service) => {
+      const { body: session } = await service.post('/v1/sessions')
+        .send({ email: 'alice@getodk.org', password: 'alice' })
+        .expect(200);
+      await service.post('/v1/projects')
+        .send({ name: 'my project' })
+        .set('X-Forwarded-Proto', 'https')
+        .set('Cookie', '__Host-session=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
+        .set('X-Csrf-Token', session.csrf)
+        .expect(403);
+    }));
+
     it('should reject if the CSRF token is missing', testService((service) =>
       service.post('/v1/sessions')
         .send({ email: 'alice@getodk.org', password: 'alice' })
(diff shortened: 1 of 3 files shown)


test/integration/api/sessions.js
matthew-white, line 1: I wrapped some tests in a `describe` block, so the diff will look nicer if you hide whitespace.

test/unit/http/preprocessors.js
```diff
@@ -298,18 +298,6 @@ describe('preprocessors', () => {
             )
           )).should.be.rejectedWith(Problem, { problemCode: 401.2 }));
 
-        it('should reject cookie auth with incorrect CSRF token for non-GET requests', () =>
```
matthew-white, line 301: This test was moved below.

test/unit/http/preprocessors.js
```diff
@@ -325,17 +313,134 @@ describe('preprocessors', () => {
...
+                  body: { __csrf: 'secretcsrf', other: 'data' }
+                }),
+                { fieldKey: Option.none() }
+              )
+            );
+            context.body.should.eql({ __csrf: 'secretcsrf', other: 'data' });
+          });
+        });
 
         it('should url-decode the CSRF token', () =>
```
matthew-white, line 340: This test ("should url-decode the CSRF token") was moved below.

matthew-white:
I still think this PR is worth merging (after resolving conflicts). That said, the PR's been open for over a year, and I don't think merging it is a priority at the moment. We can reopen the PR when it becomes something we want to address.

PR #927, opened 2023-07-13
Clear __csrf cookie after logout

matthew-white:
After a session is created, Backend returns two cookies: `__Host-session` and `__csrf`. The cookies expire at the same time as the session, so if a session expires on its own, the cookies will expire as well and be cleared. If the user logs out the session before it expires, then the `__Host-session` cookie is cleared — but the `__csrf` cookie is not. I don't think that's a huge deal, but I think it's cleaner to clear both cookies together, which is what this PR does.

Updated tests.

The change is pretty small.

This change isn't really user-facing. It's more about cleaning up after logout. I think the risk of regression is low.

I don't think changes are needed. We discourage the use of cookie auth in the API docs, and we don't document the `__csrf` cookie at all.

### DIFF
--- a/lib/resources/sessions.js
+++ b/lib/resources/sessions.js
@@ -52,12 +52,15 @@ module.exports = (service, endpoint) => {
     auth.canOrReject('session.end', session.actor)
       .then(() => Sessions.terminate(session))
       .then(() => (_, response) => {
-        // revoke the cookie associated w the session, if the session was used to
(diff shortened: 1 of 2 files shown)


alxndrsn (approved):
It's surprising that the cookies are "unset" by setting them to the string null, rather than an empty string.  But with the given expiry, the client should delete them immediately anyway :+1:

PR #928, opened 2023-07-13
Clarify how often usage reports are sent

### LINKED ISSUE #921: Usage data is sent a day later than expected
I think that we intend to send usage data every 30 days. However, it looks like the data is actually being sent every 31 days.

Relatedly, I've noticed that when I set `default.taskSchedule.analytics` to 1 in the config, usage data is sent every 2 days.

During regression testing, I think it'd be nice if usage data were sent every day so that we could more quickly verify a change to the usage reporting config. @getodk/testers, do you think that would be helpful, or does it not matter much?

For production, if we're currently sending data every 31 days, I don't think we should change that now to every 30 days: I think that would complicate longitudinal analysis.

matthew-white:
Closes #921. This PR doesn't change the current behavior, but it tries to make that behavior clear. For example, I used to think that usage reports were sent every 30 days. After this PR, usage reports will continue to be sent every 31 days — but I think that fact will be clearer.

Updated and new tests.

I considered not making any changes at all. I realized that I could send daily usage reports by setting the `default.taskSchedule.analytics` config to 0 instead of 1. However, I think this PR makes things a little clearer.

I'll also leave a comment about an individual line.

The current behavior is intended to stay the same. I've updated tests to help verify that there hasn't been a change in behavior.

I don't think so. We say in the API docs that usage data is sent monthly, which is consistent with the current interval of 31 days.

### DIFF
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -603,13 +603,18 @@ const previewMetrics = () => (({ Analytics }) => Promise.all([
   return metrics;
 }));
 
-// Get most recent 'analytics' audit from the past N days
-const AUDIT_SCHEDULE = config.has('default.taskSchedule.analytics')
+// Usage reports are sent on a fixed interval: a usage report is sent a fixed
+// number of days after the previous report. The default is to send a report
+// every 31 days. However, that number is configurable so that usage reporting
+// can be verified during regression testing.
+const ANALYTICS_SCHEDULE = config.has('default.taskSchedule.analytics')
   ? config.get('default.taskSchedule.analytics')
-  : 30; // Default is 30 days
+  : 31; // Default is 31 days
 
+// Returns the latest recent attempt to send a usage report. If there is no
+// recent attempt, then it is time for a new report to be sent.
 const getLatestAudit = () => ({ maybeOne }) => maybeOne(sql`select * from audits
-  where action='analytics' and "loggedAt" >= current_date - cast(${AUDIT_SCHEDULE} as int)
+  where action='analytics' and current_date - "loggedAt"::date < ${ANALYTICS_SCHEDULE}
   order by "loggedAt" desc limit 1`);
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


lib/model/query/analytics.js
```diff
@@ -603,13 +603,18 @@ const previewMetrics = () => (({ Analytics }) => Promise.all([
...
+const ANALYTICS_SCHEDULE = config.has('default.taskSchedule.analytics')
   ? config.get('default.taskSchedule.analytics')
-  : 30; // Default is 30 days
+  : 31; // Default is 31 days
 
+// Returns the latest recent attempt to send a usage report. If there is no
+// recent attempt, then it is time for a new report to be sent.
 const getLatestAudit = () => ({ maybeOne }) => maybeOne(sql`select * from audits
-  where action='analytics' and "loggedAt" >= current_date - cast(${AUDIT_SCHEDULE} as int)
+  where action='analytics' and current_date - "loggedAt"::date < ${ANALYTICS_SCHEDULE}
```
matthew-white, line 617: Instead of this change, I could have done:

```sql
"loggedAt" >= current_date - cast(${AUDIT_SCHEDULE} as int) + 1
```

However, I think that would have been less clear overall.

I was initially confused about dates vs. timestamps here. Previously, `current_date - cast(${AUDIT_SCHEDULE} as int)` was implicitly coerced to a timestamp with time part equal to midnight. I started wondering about the time part of `loggedAt` and whether it mattered that it was 3 a.m. instead of midnight. Now I'm converting `loggedAt` to a date and calculating the number of days between that date and `current_date`. That way, the time part of `loggedAt` doesn't figure into things: it doesn't matter when the usage report started or how long it took to complete.

This approach also lets us remove the cast of `ANALYTICS_SCHEDULE`.

matthew-white:
@ktuite, I'm thinking this can be reviewed async.

lib/model/query/analytics.js
ktuite: I like this new approach a lot of making both values dates (without timestamps) and doing the subtraction there.

ktuite (approved):

Issue #811, opened 2023-03-28
Refactor Datasets.publishIfExists


ktuite:
There's a code complexity that we should try to get rid if: 

`Datasets.createOrMerge` can be called when creating a new form and updating a form. It can create a new dataset or update an existing dataset. It can also publish a dataset or not publish a dataset as its final action. That feels like a 2x2x2 combo of possibilities. Additionally, Datasets can separately be published with `Datasets.publishIfExists`. 

We could simplify this by moving the dataset publishing out of `createOrMerge` and only using `publishIfExists` to publish. This would 
* make it simpler to reason about the flow
* make the SQL easier to maintain
* make it less likely to make mistakes with dataset/dataset property publishing
* simplify the dataset audit logging code (see future issue: https://github.com/getodk/central-backend/issues/795).

Updating and publishing datasets does not need to be extremely performant.

matthew-white:
Closed by #931 I think?

PR #931, opened 2023-07-18
Enhances #664: add logs only when changes are published

### LINKED ISSUE #664: Makes changes to dataset-related logging in audit log
For after v2022.3 — moving from https://github.com/getodk/central-backend/issues/649#issuecomment-1308327816:

Backend logs `dataset.create` and `dataset.update` in the audit log. However, I think that there are a couple of possibilities for confusion there:

- The events are logged too early.
- The events are logged too frequently.

More specifically:

- If you upload but don't publish a form definition that would create entities for a new dataset, then `dataset.create` is immediately logged, even if the form draft is ultimately abandoned and the dataset never becomes visible on the project.
- If you upload but don't publish a form definition that would add a property to an existing dataset, then `dataset.update` is immediately logged, even if the form draft is ultimately abandoned and the property never becomes visible on the dataset. (Maybe related to https://github.com/getodk/central-backend/issues/649#issuecomment-1309167680?)
- `dataset.update` is logged even when there are no new properties. It is logged when a form definition is uploaded that would create entities for an existing dataset, not when that form definition would add a property to the dataset.

I'd expect `dataset.create` and `dataset.update` to be logged only when the form draft is published. I know that creating a draft can have an immediate effect on the `datasets` table, and I think it makes sense that it works that way under the hood. But I think we're exposing those mechanics over the audit log in a way that could be surprising and that ultimately makes the audit log less actionable. It also results in a mismatch between what's in the audit log and what's returned by &hellip;/datasets and &hellip;/dataset-diff.

How about we don't set `.audit` on `Datasets.createOrMerge`, but instead wait until publishing the form to log `dataset.create` or `dataset.update`? Maybe we could query the dataset diff then to determine whether and what to log.

@sadiqkhoja has made the point that it could be helpful for debugging to log events earlier / more frequently. One idea is that we could add more dataset actions, distinguishing between draft changes to a dataset and published changes to a dataset. It's not clear whether we'd return all those actions over the API or keep some just internal.

sadiqkhoja:
Closes #664 and #811

Integration tests

NA

None

Yes

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1236,9 +1236,9 @@ You will get following workflow warnings while creating a new form or uploading 
 
 ### Creating Datasets with Forms
 
-Starting from Version 2022.3, a Form can also create a Dataset by defining a Dataset schema in the Form definition (XForms XML or XLSForm). When a Form with a Dataset schema is uploaded, a Dataset and its Properties are created and a `dataset.create` event is logged in the Audit logs. The state of the Dataset is dependent on the state of the Form; you will need to publish the Form to publish the Dataset. Datasets in the Draft state are not returned in [Dataset APIs](#reference/datasets), however the [Related Datasets](#reference/forms/related-datasets/draft-form-dataset-diff) API for the Form can be called to get the Dataset and its Properties.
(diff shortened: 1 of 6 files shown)


ktuite (commented):
I think I want to stare at this a little bit longer, but here are some comments.

1. I'm looking forward to this change!
2. I left suggestions to update some code comments because they are out of date or this part of the code is still complex (but less so now)
3. Is this (property list) what we want to log in the audit details for these events? Anything else?

ktuite (approved):

PR #932, opened 2023-07-18
Silence logging from migrations when setting up database in testing

matthew-white:
We have a couple of migrations that log messages. I think that logging is helpful for users and in migration tests. However, those messages are also logged whenever the database is reinitialized during testing (whenever `testServiceFullTrx()` or `testContainerFullTrx()` is used). I don't think the logging is useful in that case and actually makes it harder to read the test output. This PR tries to improve that case by silencing logging from migrations only when the database is (re)initialized during testing.

Before making this change, I ran a subset of tests and saw logging from migrations. After making this change, that logging was gone. However, other logging continued to come through: I called `console.log()` in a test, and I saw its output.

This is a small change and targeted so that only specific logging is silenced. The migrations didn't have to change. I don't love overwriting `console.log`, but I think it's done safely, because `console.log` is restored in a `finally` block.

This change is for testing only.

No changes needed.

### DIFF
--- a/test/integration/setup.js
+++ b/test/integration/setup.js
@@ -4,6 +4,7 @@ const { sql } = require('slonik');
 const { readdirSync } = require('fs');
 const { join } = require('path');
 const request = require('supertest');
+const { noop } = require(appRoot + '/lib/util/util');
 // eslint-disable-next-line import/no-dynamic-require
 const { task } = require(appRoot + '/lib/task/task');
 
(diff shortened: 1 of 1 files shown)


matthew-white:
@ktuite, I think this can reviewed async!

ktuite (approved):
Looks great; seeing those console logs in the migrations was bugging me, too.

PR #933, opened 2023-07-18
Change dataset-diffs to return properties in form question order

### LINKED ISSUE #708: Guarantee dataset properties in order in dataset-diff
Dataset properties are returned and shown in alphabetical order on the frontend, but it might be clearer for form designers if these properties match the order in which they are found in the form.

Screen Shot 2022-11-29 at 11 00 44 AM

This is related to Issue #675 about guaranteeing property order in the CSV export. Whatever ordering we end up using there, we should use in the dataset-diff endpoint as well.

ktuite:
Closes #708 

Ordering the dataset properties is really tricky because of how they're represented in the database. The gist of this change is that when querying the `dataset-diff` endpoint, that is done relative to a specific form, a specific _version_ of that form, even, so the order of the properties should match the order of the fields in that version of the form and be grouped together, and the order of the rest of the properties really isn't as important. (These properties are not even shown on Frontend.) Or if the order if the rest of the properties does seem important, I want that to be determined by the one other chunk of code dedicated to ordering dataset properties for a whole dataset.

Given all that, I split the query from one massive query into 
1) getting just the properties from the specific version of the form in question and 
2) the rest of the properties in the dataset, queried using the other dataset property getter that was finessed in PR #780

Using the `Datasets.getProperties` getter only gets published properties but that is fine because any _unpublished_ properties will only be part of the diff because they are part of that specific form.  

Added a few new tests and changed other tests to have properties expected in the right order.

I don't know, I think the overall approach is sound - this query needed to be split up because I kept getting hung up on it in other contexts - but I don't know if I _love_ the code I've come up with. It's a weird mixture of an async/await and Promise.all and I don't know if it's any easier to understand than what was there before (but what was there before didn't have the right ordering.)

I think the main thing this will impact is having the properties show up in a clearer order on the form draft and form status pages that describe dataset things.

This code works for draft forms and non-draft forms and published/unpublished datasets and properties so there might be MORE tests we want to add, such as:
- making sure if you reorder questions, the diff looks right (i have a test that removes and renames something but not for just moving)
- more tests of when things get deleted?

There are also existential questions like:
1) if the dataset properties have an order, and your form moves things around, what does that even mean for the ordering within the diff, let alone the dataset properties... i think we covered this already with going on the order that matches the order in which the properties were originally added to the dataset.

2) if you delete a bind/remove the property from that form, the dataset info panel on the form doesn't mention deleted properties anymore, but _old versions of the form_ could still fill in those properties? That seems like a (hopefully rare) issue that someone might run into and go a little bit crazy about because they can't see anywhere how that dataset property is getting filled in.

Will check, maybe add a note about the ordering.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -363,7 +363,11 @@ ORDER BY form_fields."order"
 
 // Used when creating or updating a dataset through the uploading of a form
 // to show the status of the dataset and each property in that dataset.
-const getDiff = (projectId, xmlFormId, forDraft) => ({ all }) => all(sql`
+
+// This helper function gets just the properties specificed by the saveto binds
+// in a single form definition. It orders the properties by their question order
+// within the form.
+const _getFormSpecificProperties = (projectId, xmlFormId, forDraft) => sql`
   WITH form AS (
     SELECT f.id, f."xmlFormId", fd.id "formDefId", fd."schemaId"
     FROM forms f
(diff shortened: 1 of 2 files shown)


ktuite:
Just checked: API docs don't mention order of properties and I'm not going to add it.

sadiqkhoja (approved):
I generally don't like multiple database round trips but I agree with your reasoning that in this case it's not that important.

PR #934, opened 2023-07-19
Odata opaque cursor 

### LINKED ISSUE #665: Add opaque cursor-based pagination
There are a couple of ways to fetch submissions in chunks:

- Use the OData query parameters `$top` and `$skip`. However, it is difficult to use those to fetch only new submissions, because submissions are sorted by `createdAt` descending, not ascending.
- As an alternative, each time after fetching submissions, store the latest `createdAt` timestamp. The next time, use the `$filter` OData query parameter to fetch submissions after that timestamp. However, this runs into #459.

As an alternative to both of these, we could implement opaque cursor-based pagination. Quoting from https://graphql.org/learn/pagination/:

> Especially if the cursors are opaque, either offset or ID-based pagination can be implemented using cursor-based pagination (by making the cursor the offset or the ID), and using cursors gives additional flexibility if the pagination model changes in the future. As a reminder that the cursors are opaque and that their format should not be relied upon, we suggest base64 encoding them.

@lognaturel wrote about this concept:

> I think usually this opaque cursor concept is used for pagination.

> Aggregate/Briefcase use it as a “query resume point”: https://docs.getodk.org/briefcase-api/#get-view-submissionlist. That’s pagination but with the page being “whatever is available now”

> [Opacity] prevents clients from trying to build their own cursors and enforces that they just pass them on to the next request.
> So that you have the flexibility to change the actual contents of your cursor.

> I imagine we could use `createdAt` as a string as a basis for a cursor to get around the node precision loss issue [described in #459].

> [This proposal](https://learn.microsoft.com/en-us/odata/webapi/skiptoken-for-server-side-paging) for using `$skiptoken` for server-side paging is similar to what I have in mind.
> (I don’t really understand the context for this but the general description of the problem it solves is helpful I think)

At the moment, the main resource for which cursor-based pagination is of interest is submissions. However, at some point soon, we will face similar questions around entities.

sadiqkhoja:
Closes #665

What has been done to verify that this works as intended?

Integration tests

Why is this the best possible solution? Were any other approaches considered?

- Base64 of a json object is used for skipToken. Fields of the json object depends on the odata table. In case of Submission, it is instanceId, for Submission subtable (repeat) it is repeatId (hash of the data - existing approach) and for Entities it is UUID
- I didn't compress the token, base64 encoding fulfills the requirement
- We are still fetching all the Submissions when data for the subtable is requested because we need to count the total repeating rows and return `nextlink` conditionally i.e. only when there are more rows to be returned. So pagination for subtable is not much of a help, it only saves bandwidth between server and client. I wish we could get rid of those two requirements then odata feed for subtable would be efficient as well.

**Caveat**
Since I am using hash ID of the repeat row and if the data of that repeat item is changed then subsequent `nextlink` will return nothing

How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

None

Does this change require updates to the API documentation? If so, please update docs/api.md as part of this PR.

TODO

Before submitting this PR, please make sure you have:
[ ] run make test-full and confirmed all checks still pass OR confirm CircleCI build passes
[ ] verified that any code from external sources are properly credited in comments or that everything is internally sourced

### DIFF
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -23,6 +23,8 @@ const { sanitizeOdataIdentifier, without } = require('../util/util');
 const { jsonDataFooter, extractOptions, nextUrlFor } = require('../util/odata');
 const { submissionToOData, systemFields } = require('../data/odata');
 const { SchemaStack } = require('../data/schema');
+const { QueryOptions } = require('../util/db');
+
 
 ////////////////////////////////////////////////////////////////////////////////
 // UTIL
@@ -52,9 +54,10 @@ const extractPathContext = (subpath) =>
 const extractPaging = (table, query) => {
   const parsedLimit = parseInt(query.$top, 10);
   const limit = Number.isNaN(parsedLimit) ? Infinity : parsedLimit;
-  const offset = parseInt(query.$skip, 10) || 0;
+  const offset = (!query.$skiptoken && parseInt(query.$skip, 10)) || 0;
   const shouldCount = isTrue(query.$count);
-  const result = { limit: max(0, limit), offset: max(0, offset), shouldCount };
+  const skipToken = query.$skiptoken ? QueryOptions.parseSkiptoken(query.$skiptoken) : null;
+  const result = { limit: max(0, limit), offset: max(0, offset), shouldCount, skipToken };
 
   return Object.assign(result, (table === 'Submissions')
     ? { doLimit: Infinity, doOffset: 0 }
@@ -366,11 +369,12 @@ const edmxForEntities = (datasetName, properties) => {
 // originalUrl: String is the request URL; we need it as well to formulate response URLs.
 // query: Object is the Express Request query object indicating request querystring parameters.
 // inStream: Stream[Row] is the postgres Submissions rowstream.
-const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, tableCount) => {
+const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, tableCount, tableRemaining) => {
   // cache values we'll need repeatedly.
   const serviceRoot = getServiceRoot(originalUrl);
-  const { limit, offset, doLimit, doOffset, shouldCount } = extractPaging(table, query);
+  const { doLimit, doOffset, shouldCount, skipToken } = extractPaging(table, query);
   const options = extractOptions(query);
+  const isSubTable = table !== 'Submissions';
 
   // make sure the target table actually exists.
   // TODO: now that this doesn't require schema computation, should we move it up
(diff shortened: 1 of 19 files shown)


test/integration/api/odata.js
```diff
@@ -1251,6 +1315,36 @@ describe('api: /forms/:id.svc', () => {
...
           }))));
 
+    it.skip('should limit subtable results', testService(async (service) => {
+      const asAlice = await withSubmissions(service, identity);
+
+      await asAlice.get('/v1/projects/1/forms/withrepeat.svc/Submissions.children.child?$top=10')
+        .expect(200)
+        .then(({ body }) => {
+          const tokenData = {
+            instanceId: body.value[body.value.length - 1].instanceId,
```
matthew-white, line 1325: The strategy of using `instanceId` + `__id` makes sense to me. :+1: How long will this token be? Should we try to compress the data like we do with app user QR codes? (See `CollectQr` in Frontend.)

lib/formats/odata.js
```diff
@@ -406,12 +437,20 @@ const rowStreamToOData = (fields, table, domain, originalUrl, query, inStream, t
...
       // @odata.count and nextUrl.
       const totalCount = (tableCount != null) ? tableCount : counted;
-      const nextUrl = nextUrlFor(limit, offset, totalCount, originalUrl);
+
+      // How many items are remaining for the next page?
+      // if there aren't any then we don't need nextUrl
+      const remaining = (tableRemaining != null) ? tableRemaining - added : remainingItems;
+
+      const skipTokenData = { instanceId: lastInstanceId };
+      if (isSubTable) skipTokenData.repeatId = lastRepeatId;
```
sadiqkhoja, line 451: I am thinking to use just `repeatId` for subtable and remove instanceId

matthew-white:
> Since I am using hash ID of the repeat row and if the data of that repeat item is changed then subsequent `nextlink` will return nothing

Could you say more about this? What data contributes to the hash ID? Is it all fields of the repeat group, such that any edit to the repeat group could change its ID? Or is it just the path (including index) of the repeat group?

sadiqkhoja:
>Could you say more about this? What data contributes to the hash ID? Is it all fields of the repeat group, such that any edit to the repeat group could change its ID? Or is it just the path (including index) of the repeat group?

It's an existing function `lib/data/odata.js:hashId`. The hash is calculated based on `instanceId` of the Submission and the fullpath of the repeat item like `Submissions#double%%children#%%child#1`. 

I was wrong when I said "data of the repeat item" can change the hash, it is the index that can affect the hash. 

Here is an example where there will be a problem:

- Submission has 10 repeat items
- User requests `$top=5`
- User deletes 6 repeat items
- Next odata request will return nothing because hashId will not match

Similar problem can arise if user manually deletes Submission because skipToken is based on instanceId and if instanceId is not found then cursor is pointing nowhere. However, probability of this happening for Submissions is quite low because we don't provide any way to delete individual Submissions.

In both cases, I don't see this being a big problem, just wanted to make note of it.

matthew-white:
I ended up prioritizing reviewing getodk/central-frontend#851, and I ran out of time to review this one. Sorry about that! I think the only thought I want to offer is that I think it's OK if `$skiptoken` isn't 100% resilient to concurrent modification. As long as it returns an error when there's a problem and doesn't return the wrong data or incorrectly return no data, I think it's OK.

test/integration/api/odata.js
```diff
@@ -717,6 +718,83 @@ describe('api: /forms/:id.svc', () => {
...
+        .then(({ body }) => {
+          const tokenData = {
+            instanceId: body.value[0].__id,
+          };
+          const token = encodeURIComponent(QueryOptions.getSkiptoken(tokenData));
+          body['@odata.nextLink'].should.be.eql(`http://localhost:8989/v1/projects/1/forms/withrepeat.svc/Submissions?%24top=1&%24skiptoken=${(token)}`);
+        });
+    }));
+
+    it('should ignore $skip when $skipToken is given', testService(async (service) => {
```
sadiqkhoja, line 738: add a test  about adding new submissions in between calls

ktuite (approved):

PR #936, opened 2023-07-24
test/users: remove outdated TODO

alxndrsn:
This feature looks like it's outlasted the "initial release".


matthew-white (approved):
I don't think we're planning to change this anytime soon, so it makes me sense to remove the TODO. At some point, I think we should also remove the "for v1 only" prefix from this related method:

https://github.com/getodk/central-backend/blob/facf33916322f4cea6e2b963372b4e7623715320/lib/model/frames.js#L195-L197

PR #939, opened 2023-07-27
eslint: allow function use before define

alxndrsn:
We don't currently have any instances of this violation.

Per eslint docs:

> Function declarations are hoisted, so it’s safe.

See: https://eslint.org/docs/latest/rules/no-use-before-define

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,6 +23,7 @@
     "no-nested-ternary": "off",
     "no-restricted-syntax": "off",
     "no-underscore-dangle": "off",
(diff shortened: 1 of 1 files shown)


alxndrsn:
It can be simpler to have primary functionality at the top of the file, and smaller helper functions hidden below the fold.

One example would be https://github.com/getodk/central-backend/pull/910/files#diff-08ab778a4b210b6812adb23f527b1b51c131bb7185ffc6658d45f38e03cd718c.

It looks like those functions will be converted to anonymous functions.

PR #946, opened 2023-08-03
Feature: ?datasets=true to get list of datasets in GET /projects API

sadiqkhoja:

Integration tests

NA

None

TODO

### DIFF
--- a/lib/data/project.js
+++ b/lib/data/project.js
@@ -1,0 +1,52 @@
+// Copyright 2023 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const { Frame } = require('../model/frame');
+
+const combineProjectsAndForms = (projects, forms) => {
+  const projectsWithForms = projects.reduce((acc, proj) => {
+    const project = proj instanceof Frame ? proj.forApi() : proj;
+    return { ...acc, [proj.id]: { ...project, formList: [], forms: 0, lastSubmission: null } };
+  }, {});
+
+  for (const form of forms) {
+    if (form.projectId in projectsWithForms) {
+      const formForApi = form.forApi();
+      const proj = projectsWithForms[form.projectId];
+      proj.formList.push(formForApi);
+      // Update "extended" fields on project even if not extended, because
+      // such data can be calculated from formList.
+      proj.forms += 1;
+      if (proj.lastSubmission == null ||
+        (formForApi.lastSubmission != null && formForApi.lastSubmission > proj.lastSubmission))
+        proj.lastSubmission = formForApi.lastSubmission;
+    }
+  }
+  return Object.values(projectsWithForms);
+};
+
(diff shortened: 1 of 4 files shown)


test/integration/api/projects.js
```diff
@@ -1348,13 +1365,61 @@ describe('api: /projects?forms=true', () => {
...
+      await asAlice.post('/v1/projects/1/forms?publish=true')
+        .send(testData.forms.simpleEntity)
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+
+      await asAlice.get('/v1/projects?datasets=true')
+        .expect(200)
+        .then(({ body }) => {
+          body[0].datasetList.length.should.equal(1);
+          body[0].datasetList[0].name.should.equal('people');
```
ktuite, line 1380: This test should include a line like `datasetList[0].should.be.an.ExtendedDataset()` to look more like the forms test above: `should return projects with verbs and nested extended forms`

test/integration/api/projects.js
```diff
@@ -1381,6 +1446,25 @@ describe('api: /projects?forms=true', () => {
...
+      const chelsea = await Users.getByEmail('chelsea@getodk.org').then((o) => o.get());
+      await asAlice.post(`/v1/projects/1/assignments/formfill/${chelsea.actorId}`);
+
+      await asChelsea.get('/v1/projects?datasets=true')
+        .expect(200)
+        .then(({ body }) => {
+          body[0].datasetList.length.should.equal(0);
+        });
+    }));
+
```
ktuite, line 1467: There is a `projects?forms=true` test below called `should set project data from formList even on non-extended projects` that should be duplicated for the dataset case. Does the `projects[i].datasets` and `projects[i].lastEntity` get updated appropriately? In particular, I think there's a bug in how the `lastEntity` value gets set at the whole project level (not within the dataset list within the project).

lib/data/project.js
```diff
@@ -0,0 +1,58 @@
...
+    if (dataset.projectId in projectsWithDatasets) {
+      const datasetForApi = dataset.forApi();
+      const proj = projectsWithDatasets[dataset.projectId];
+      proj.datasetList.push(datasetForApi);
+      // Update "extended" fields on project even if not extended, because
+      // such data can be calculated from datasetList.
+      proj.datasets += 1;
+      if (proj.lastSubmission == null ||
+        (datasetForApi.lastSubmission != null && datasetForApi.lastSubmission > proj.lastSubmission))
+        proj.lastSubmission = datasetForApi.lastSubmission;
```
ktuite, line 49: This needs to use `.lastEntity` instead of `.lastSubmission` for updating the project from the dataset information.

I was thinking that with the introduction of this new data file, it should have a corresponding test file, but since it is just about manipulating data to build a response, it seems awkward to test in isolation/as a unit test.

ktuite (commented):
I follow the changes here, including adding the `lastEntity` timestamp at the project level and refactoring the code that builds the form list under each project to also work for building a dataset list. 

I left comments about adding a few more tests and fixing a bug with how the non-extended project data related to datasets gets filled in when the datasets are requested.

lib/resources/projects.js
```diff
@@ -9,16 +9,32 @@
...
+      const datasets = await Datasets.getAllByAuth(auth, QueryOptions.extended);
+      projects = combineProjectsAndDatasets(projects, datasets);
+    }
+
+    return projects;
+  }));
+
+  // (isTrue(query.forms)
+  //   ? Projects.getAllByAuthWithForms(auth, queryOptions)
+  //   : Projects.getAllByAuth(auth, queryOptions))));
```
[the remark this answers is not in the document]
sadiqkhoja: yes, this is dead code

ktuite (approved):

PR #947, opened 2023-08-04
Do not export individual client audit attachments in .csv.zip

matthew-white:
The submissions.csv.zip exports client audit logs in two ways:

1. It exports each submission's individual client audit log along with other submission attachments (via `SubmissionAttachments.streamForExport()`).
2. It combines all client audit logs into a single CSV file (via `ClientAudits.streamForExport()`).

If we're doing (2), then I think we don't really need to do (1). Exporting both (1) and (2) effectively exports each client audit log twice. For some forms, there can be more client audit log data than actual submission data, in which case exporting both ends up adding quite a bit of redundant data.

Note that the individual client audit logs aren't easily accessible from the .zip. Typically, each client audit log has the same name: audit.csv. That means that each log is written to the same path in the .zip: media/audit.csv. Apparently the .zip spec allows multiple files to be written to the same path, but the end result is that most users won't know how to access these files (or even know that there are multiple files to access). And again, there shouldn't be the need to access the individual files, because they're already combined elsewhere.

This PR excludes individual client audit log attachments from the submissions.csv.zip.

I had to change tests in ways that were expected. After making those changes, tests continue to pass.

Just one line changed in lib/, but I'll leave a comment about that line.

Hopefully this PR just reduces the size of the submissions.csv.zip. `SubmissionAttachments.streamForExport()` looks to be used in only one place, the submissions.csv.zip endpoint.

I sort of doubt that anyone is accessing the individual client audit logs in the submissions.csv.zip, in large part because they're not easy to access. If someone really needs to access an individual client audit log, they can still do so using the individual submission attachment endpoint (at least for an unencrypted form).

If this does end up being an issue for someone, we could come back and add a flag to control whether to include individual client audit log attachments. But I think we should start with the simplest approach, which I'm hopeful will be enough.

I don't see client audit logs mentioned in the API docs.

### DIFF
--- a/lib/model/query/submission-attachments.js
+++ b/lib/model/query/submission-attachments.js
@@ -204,6 +204,7 @@ where submission_defs.current=true
   and "deletedAt" is null
   and ${odataFilter(options.filter, odataToColumnMap)}
   and submission_attachments.name is distinct from submission_defs."encDataAttachmentName"
+  and submission_attachments."isClientAudit" is not true
   and (form_defs."keyId" is null or form_defs."keyId" in (${keyIdCondition(keyIds)}))`);
 
 module.exports = {
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -1859,10 +1859,8 @@ describe('api: /forms/:id/submissions', () => {
           .then(() => exhaust(container))
           .then(() => pZipStreamToFiles(asAlice.get('/v1/projects/1/forms/audits/submissions.csv.zip'))
             .then((result) => {
-              result.filenames.should.containDeep([
+              result.filenames.should.eql([
                 'audits.csv',
-                'media/audit.csv',
-                'media/log.csv',
                 'audits - audit.csv'
               ]);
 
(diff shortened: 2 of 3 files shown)


lib/model/query/submission-attachments.js
```diff
@@ -204,6 +204,7 @@ where submission_defs.current=true
   and "deletedAt" is null
   and ${odataFilter(options.filter, odataToColumnMap)}
   and submission_attachments.name is distinct from submission_defs."encDataAttachmentName"
+  and submission_attachments."isClientAudit" is not true
```
matthew-white, line 207: `isClientAudit` can be `TRUE`, `FALSE`, or `NULL`. After a quick look, the only place where I see it being set to `NULL` is for the encrypted data attachment:

https://github.com/getodk/central-backend/blob/2d8d20dd823cd478bb062d5d5b7f13070abf39f0/lib/model/query/submission-attachments.js#L28
https://github.com/getodk/central-backend/blob/2d8d20dd823cd478bb062d5d5b7f13070abf39f0/lib/model/query/submission-attachments.js#L67-L68

That attachment should be excluded by the line above this one. However, I see other submission attachments on the QA server where `isClientAudit` is `NULL`, not `FALSE`. Maybe that's just junk data, but given that the column is nullable, I think the safest thing is to include submission attachments for which `isClientAudit` is either `FALSE` or `NULL`.

That's what `IS NOT TRUE` does: https://www.postgresql.org/docs/current/functions-comparison.html. `= FALSE` or `<> TRUE` would exclude submission attachments for which `isClientAudit` is `NULL`.

sadiqkhoja (approved):
looks good

matthew-white:
@getodk/testers, this is ready to be verified on staging:

- Create a form that uses client audit logs.
- Create submissions to the form.
- Download and unzip the submissions .csv.zip.
- You should continue to see all the client audit logs consolidated in a single CSV file whose filename ends with "- audit.csv". However, in the media folder, you should not see any individual client audit logs. You should continue to see other files in the media folder.

srujner:
**Tested with Success!**

Issue #832, opened 2023-04-18
Track entity versionNumbers


ktuite:
We need some way of tracking versions within an entity as it gets updated. Submissions have a unique instance id for each version and a deprecated id pointing to the previous version. We probably won't do the exact same thing for entities. We do want some way of uniquely identifying versions within an entity. We also need to be able to manage updates from multiple sources including partly offline updates.

matthew-white:
Now that we're doing this in .4, I'll move notes here from the draft release criteria for .5:

**Allow an API update to an Entity to specify a base version**

- Right now, I can make an API update to an Entity by specifying the ?force=true query parameter.
- As an alternative, I should be able to make an API update by specifying the Entity version on top of which my changes should apply.
- If my changes conflict with another update to the Entity, then I should receive a 409 error.
  - Backend might reject for “soft conflicts” (server version has moved forward) too.

PR #951, opened 2023-08-16
Enhance: add entity version number and modify update entity API to work with version

sadiqkhoja:
### Changes:

- Added version column in the entity_defs table
- OData Feed returns version number
- `forms/attachment/dataset.csv` also returns version number
- PATCH /entities/:uuid checks conflict if force flag is not provided

Integration tests and manual testing with frontend

Other option was to add version in query parameter or body, but version in ETag and If-Match is a cleaner approach

None

Will do it in a separate PR


matthew-white:
It sounds like this PR closes #832. Does that sound right?

test/integration/api/entities.js
```diff
@@ -829,13 +838,24 @@ describe('Entities API', () => {
...
-        .expect(501)
+        .expect(409)
         .then(({ body }) => {
-          body.code.should.equal(501.1);
+          body.code.should.equal(409.15);
+        });
+    }));
+
+    it('should reject if version does not match', testEntities(async (service) => {
+      // TODO: change logic around force flag and enforcing certain kinds of updates
```
ktuite, line 852: can remove this and above TODO comment now?

ktuite (approved):

ktuite:
I was trying to use this branch and had some problem loading the dataset properties in frontend. Null source forms?? It may have nothing to do with this PR, but reverting it made the problem go away.

Screen Shot 2023-08-29 at 10 03 14 AM

test/integration/api/datasets.js
```diff
@@ -1317,7 +1317,7 @@ describe('datasets and entities', () => {
               .then(({ headers, text }) => {
                 headers['content-disposition'].should.equal('attachment; filename="goodone.csv"; filename*=UTF-8\'\'goodone.csv');
                 headers['content-type'].should.equal('text/csv; charset=utf-8');
-                text.should.equal('name,label,first_name,age\n12345678-1234-4123-8234-123456789abc,Alice (88),Alice,88\n');
+                text.should.equal('name,label,version,first_name,age\n12345678-1234-4123-8234-123456789abc,Alice (88),1,Alice,88\n');
```
ktuite, line 1320: One thing I noticed when trying to make a form that looked up the version, and based on what we're talking about in the team meeting on Aug 29, `name` and `label` have to be what they are, but should `version` be `__version`?

sadiqkhoja: I like `version` for the attachment csv

PR #952, opened 2023-08-16
Validate savetos on structural fields

### LINKED ISSUE #670: Validate fields saved to dataset properties
Comment from [Matt](https://github.com/getodk/central-backend/issues/649#issuecomment-1301597288)

> Should we validate the type of fields that can be used as entity properties? My understanding is that groups and repeat groups are fields that can't be used as entity properties.
> 
> Less sure about this one: for now, should we prohibit any fields within a repeat group from being used as entity properties?
> 
> <hr>
> 
> - check what happens when a structural field is used in a bind that gets picked up by entities 
> - validate and don't allow binds on structural fields if they are for some reason allowed currently
> - don't allow binds on fields within repeat groups
> - keep all this in mind for future versions of entities that use repeats and can make multiple entities per form 

Check what happens when we try to add binds about entity properties to structural fields or fields within repeat groups and don't allow if they don't make sense. 

We'll come back to an updated version of the entity XForm spec that allows for a form to create multiple entities in the future.

ktuite:
Closes #670 

We had a number of questions from the issue
- What happens if a `saveto` is on a bind for a structural field, like a group or repeat group?

    * Answer: pyxform is more strict than Central when it comes to dataset/entity schemas in form definitions. If you try to put a `saveto` on a structural field, it will return an error like `Groups and repeats can't be saved as entity properties.`

- What does Central currently allow/disallow, given that pyxform is more strict?

    * Answer: Central is looking for binds with `type` attributes and ignoring other kinds of binds. It doesn't disallow them, though, because sometimes a bind might be used on a structural field for handling relevance (see https://github.com/getodk/central-backend/issues/147). 

- What happens when a `saveto` is on a field within a repeat group?

    * Answer: pxyform doesn't allow this and gives the error `Currently, you can't create entities from repeats. You may only specify save_to values for form fields outside of repeats.'`. In Central through v2023.3, the XML parsing allows this and creates a new dataset property. When parsing a corresponding submission, it saves the field value of the _last_ repeat instance into the entity. This PR adds a check that doesn't even let forms that do this be uploaded.

Tests.

This PR is mostly formally checking behavior we informally tested earlier when releasing entities. We have also been assuming most everyone working with entities is making forms in a spreadsheet and has had pyxform, which has been stricter than Central, stop them from making these mistakes. 

The only change to the code I made was adding a check to be more like pyxform where savetos on fields in repeats are no longer allowed.

It makes XML parsing around datasets a little stricter, but they probably weren't running into these things anyway. 

No.

### DIFF
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -82,8 +82,11 @@ const _recurseFormFields = (instance, bindings, repeats, selectMultiples, envelo
       if (binding != null) {
         field.type = binding.type || 'unknown'; // binding should have a type.
         const prop = saveToAttr(binding);
-        if (prop)
+        if (prop) {
+          if (Array.from(repeats).find((repeat) => bindingPath.startsWith(repeat)))
+            throw Problem.user.invalidEntityForm({ reason: 'Currently, entities cannot be populated from fields in repeat groups.' });
           field.propertyName = binding[prop];
+        }
       } else if (tag.children != null) {
         // if we have no binding node but we have children, assume this is a
         // structural node with no repeat or direct data binding; recurse.
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -2080,98 +2080,6 @@ describe('datasets and entities', () => {
           });
       }));
 
-      it.skip('should not break when a saveto/bind is on a field inside a repeat group', testService(async (service, container) => {
-        // Entities made from repeat groups are not yet supported. pyxform throws an error but central does not.
-        // When parsing the corresponding submission, it will pick the last item in the repeat group.
-        const form = `<?xml version="1.0"?>
-        <h:html xmlns="http://www.w3.org/2002/xforms" xmlns:entities="http://www.opendatakit.org/xforms/entities" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
-          <h:head>
-            <h:title>Repeat Children Entities</h:title>
-            <model entities:entities-version="2022.1.0" odk:xforms-version="1.0.0">
-              <instance>
-                <data id="repeat_entity" version="2">
-                  <num_children/>
-                  <child jr:template="">
-                    <child_name/>
-                  </child>
-                  <child>
-                    <child_name/>
-                  </child>
-                  <meta>
-                    <instanceID/>
-                    <instanceName/>
-                    <entity create="1" dataset="children" id="">
-                      <label/>
-                    </entity>
-                  </meta>
-                </data>
-              </instance>
-              <bind entities:saveto="num_children" nodeset="/data/num_children" type="int"/>
-              <bind entities:saveto="child_name" nodeset="/data/child/child_name" type="string"/>
-              <bind jr:preload="uid" nodeset="/data/meta/instanceID" readonly="true()" type="string"/>
-              <bind calculate=" /data/num_children " nodeset="/data/meta/instanceName" type="string"/>
-              <bind calculate="1" nodeset="/data/meta/entity/@create" readonly="true()" type="string"/>
-              <bind nodeset="/data/meta/entity/@id" readonly="true()" type="string"/>
-              <setvalue event="odk-instance-first-load" readonly="true()" ref="/data/meta/entity/@id" type="string" value="uuid()"/>
-              <bind calculate="concat(&quot;Num children:&quot;,  /data/num_children )" nodeset="/data/meta/entity/label" readonly="true()" type="string"/>
-            </model>
-          </h:head>
-          <h:body>
-            <input ref="/data/num_children">
-              <label>Num Children</label>
-            </input>
-            <group ref="/data/child">
-              <label>Child</label>
-              <repeat nodeset="/data/child">
-                <input ref="/data/child/child_name">
-                  <label>Child Name</label>
-                </input>
-              </repeat>
-            </group>
-          </h:body>
-        </h:html>
-        `;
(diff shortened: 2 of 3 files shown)


lib/data/schema.js
```diff
@@ -82,8 +82,11 @@ const _recurseFormFields = (instance, bindings, repeats, selectMultiples, envelo
       if (binding != null) {
         field.type = binding.type || 'unknown'; // binding should have a type.
         const prop = saveToAttr(binding);
-        if (prop)
+        if (prop) {
+          if (Array.from(repeats).find((repeat) => bindingPath.startsWith(repeat)))
```
sadiqkhoja, line 86: Does this work when repeat is inside a group? What would be the value of `bindingPath` and `repeat` when saveto on /group1/repeat1/somefield

test/integration/api/datasets.js
```diff
@@ -2030,6 +2030,209 @@ describe('datasets and entities', () => {
...
+          .then(({ body }) => {
+            body.name.should.be.eql('things');
+            const { properties } = body;
+            properties.length.should.equal(2);
+            properties[0].name.should.equal('prop1');
+            properties[1].name.should.equal('prop3');
+          });
+      }));
+
+      it.skip('should not break when a saveto/bind is on a field inside a repeat group', testService(async (service, container) => {
```
sadiqkhoja, line 2083: forgot to remove `skip`?

this is about Submission parsing, right?

test/integration/api/datasets.js
```diff
@@ -2030,6 +2030,209 @@ describe('datasets and entities', () => {
             }));
       }));
 
+      it('should ignore a saveto incorrrectly placed on a bind on a structural field', testService(async (service) => {
```
sadiqkhoja, line 2033: shouldn't we reject such forms?

test/integration/api/datasets.js
```diff
@@ -2030,6 +2030,209 @@ describe('datasets and entities', () => {
...
+            <model entities:entities-version="2022.1.0" odk:xforms-version="1.0.0">
+              <instance>
+                <data id="repeat_entity" version="2">
+                  <num_children/>
+                  <child jr:template="">
+                    <child_name/>
+                  </child>
+                  <child>
+                    <child_name/>
+                  </child>
```
sadiqkhoja, line 2190: duplicate nodes

test/unit/data/schema.js
```diff
@@ -524,6 +524,158 @@ describe('form schema', () => {
...
+            { name: 'title', path: '/occupation/title', type: 'string', order: 2 },
+            { name: 'dates', path: '/occupation/dates', type: 'structure', order: 3 },
+            { name: 'joined', path: '/occupation/dates/joined', type: 'date', order: 4 },
+            { name: 'departed', path: '/occupation/dates/departed', type: 'date', order: 5 },
+            { name: 'salary', path: '/occupation/salary', type: 'decimal', order: 6 }
+          ]);
+        });
+      });
+
+      it.skip('should allow binds on fields in repeats even if functionality is not correct', () => { // gh cb#670
```
sadiqkhoja, line 572: I guess this can be removed?

test/unit/data/schema.js
```diff
@@ -524,6 +524,158 @@ describe('form schema', () => {
...
+            { name: 'children', path: '/children', type: 'structure', order: 1 },
+            { name: 'child', path: '/children/child', type: 'repeat', order: 2 },
+            { name: 'name', path: '/children/child/name', type: 'string', order: 3, propertyName: 'child_name' },
+            { name: 'toy', path: '/children/child/toy', type: 'repeat', order: 4 },
+            { name: 'name', path: '/children/child/toy/name', type: 'string', order: 5 }
+          ]);
+        });
+      });
+
+      it('should allow binds on fields in repeats even if functionality is not correct', () => { // gh cb#670
```
sadiqkhoja, line 630: caption needs to be changed

test/integration/api/datasets.js
ktuite: this [line](https://github.com/getodk/central-backend/blob/83e453326c9aeded6297c1ef5af63aa0efc782b4/lib/data/schema.js#L146C8-L146C8) of the code looks for binds that explicitly include `type`, so a field that is explicitly not a structural field. 

i believe the parsing code works like that (filtering for just binds with a type) because of this [test](https://github.com/getodk/central-backend/blob/de41d8fa09d8c6d3cd6ed652703974f1cce645d2/test/unit/data/schema.js#L139C36-L179) for this [issue](https://github.com/getodk/central-backend/issues/147) where sometimes there _is_ a bind on a structural field, but for something that collect/enketo does, not central.

it would be more work (and might introduce other bugs) to find these other binds and check they don't have a saveto and reject if they do... 

- it doesn't seem like a big enough problem to justify the work/risk: if a person adds a saveto on a structural field and it gets totally ignored, nothing terrible happens. hopefully they notice the problem when the property isn't in the list of properties for the dataset. they'd also have to have done it manually, because pyxform _does_ throw an error about this.
- if a person adds a type and a saveto on a bind for a structural field, many other things about the rendering of the form itself breaks.

lib/data/schema.js
ktuite: excellent question

test/integration/api/datasets.js
ktuite: this is me leaving a trail of my thoughts in a confusing manner
- towards the beginning of my work on this issue, i wrote this test checking that you could have binds on a field in a repeat, and the submission parsing would pick up the last value in the repeat.
- i decided i could get more aggressive about checking the form definition up front. with that change, this test no longer passes. i wasn't ready to delete it, though.
- i'll delete it unless for some reason we _do_ want to let people upload form definitions with binds on fields inside of repeats.

lib/data/schema.js
ktuite: in a test i am about to add, the binding path might look like
`/data/children/child/possessions/toys/toy`
and the set of repeats like
`'/data/children/child',  '/data/children/child/possessions/toys'`
so it does find it and reject, even when it is inside a complicated nesting.

sadiqkhoja (approved):

PR #958, opened 2023-08-23
fixes #673: dataset named itemsets and external choices form is uploaded

### LINKED ISSUE #673: Attempting to use a form with "fast external itemsets" in a project that has a dataset called itemsets leads to 500
Quote from [LN](https://github.com/getodk/central-backend/issues/649#issuecomment-1305057297):

> It feels to me like `itemsets` would be a really strange dataset name so I doubt there would be collisions. `itemsets` is plural because it essentially lets you define multiple datasets that are processed in a custom/magic way by Collect. As @issa-tseng says, there's nothing special about the `itemsets.csv` filename other than IF you're using "fast external itemsets" then that's the name of the CSV that's produced. I don't think it needs to be treated in any special way but it's a good thing to bring up!
> 
> ⚠️ Edit: we should verify the case of a dataset named itemsets.csv and a form that uses “fast external itemsets.” The unusual thing about that feature is that pyxform produces the csv.

sadiqkhoja:
Closes #673

Integration tests and manually verified it

NA

NA

NA

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -1,3 +1,4 @@
+const { readFileSync } = require('fs');
 const appRoot = require('app-root-path');
 const { testService } = require('../setup');
 const testData = require('../../data/xml');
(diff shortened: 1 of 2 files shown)


test/integration/api/datasets.js
```diff
@@ -1276,6 +1277,58 @@ describe('datasets and entities', () => {
                 .catch(error => {
                   error.constraint.should.be.equal('check_datasetId_is_null_for_non_file');
                 }))))));
+
+      // cb#673
+      it('should not throw problem for fast external itemsets when there is existing "itemset" dataset', testService(async (service) => {
```
lognaturel, line 1282: Nit: "itemset" -> "itemsets"!

ktuite (approved):

PR #959, opened 2023-08-23
Remove entity label from audit log and look up info in database

### LINKED ISSUE #822: Remove entity label from audit log
I don't think we need to make any changes right away, but as we start thinking about deleting and purging entities, I think we should think about what information about an entity should remain in the audit log even after the entity is deleted and purged. We should store the entity UUID and dataset name, but I think we should remove any entity data, including the entity label, because that information could be sensitive. Ideally, we don't log any of that information in the audit log to begin with.

We currently log the entity label as part of the details of the `entity.create` event. Frontend uses the entity label on the submission detail page, so if we remove it from the details of the audit log entry, we'll need to figure out how to get that information to Frontend.

Update from Kathleen:
Backend PR: https://github.com/getodk/central-backend/pull/959
Frontend PR: https://github.com/getodk/central-frontend/pull/850

ktuite:
Backend part of issue #822 
Frontend PR [here](https://github.com/getodk/central-frontend/pull/850) (Makes small change on frontend to look up `label` from a new place.)

In the same way we look up current information about a submission when we get audit log events for a specific entity (so we can appropriately handle deleted submissions), this does the same thing for entity information.

Previously, the details for an event like `entity.create` contained an `entity` object with `uuid`, `dataset`, _and_ `label`. 

`label` has been removed, so the audit details only contain `uuid` and `dataset` but default.

But through the `getBySubmissionId` query, for events about entities, `details.entity` contains entity data populated from the database. This object has the same structure you'd see elsewhere in the API for getting details about an `entity`, but with the additional `dataset` property that was already in the audit event details. 

```
...
    'action': 'entity.create',
    'details': {'entity': {'createdAt': '2023-08-30T19:16:38.812Z',
                         'creatorId': 5,
                         'currentVersion': {'createdAt': '2023-08-30T19:16:38.812Z',
                                            'creatorId': 5,
                                            'current': True,
                                            'data': {'circumference_cm': '333',
                                                     'geometry': '44.296333 '
                                                                 '-75.364095 0 '
                                                                 '0',
                                                     'species': 'purpleheart'},
                                            'label': '333cm purpleheart',
                                            'userAgent': 'Enketo/6.2.2 '
                                                         'Mozilla/5.0 '
                                                         '(Macintosh; Intel '
                                                         'Mac OS X 10_15_7) '
                                                         'AppleWebKit/537.36 '
                                                         '(KHTML, like Gecko) '
                                                         'Chrome/116.0.0.0 '
                                                         'Safari/537.36'},
                         'dataset': 'trees',
                         'deletedAt': None,
                         'updatedAt': None,
                         'uuid': 'f08a4912-1c45-4dc5-a27a-36a6bb5d2638'},
              'entityDefId': 2,
              'entityId': 2,
              'submissionDefId': 2,
              'submissionId': 2},
...
```

If the entity has been soft deleted or purged, OR if the request is NOT EXTENDED, the details will look like this (closer to how they looked before but now without `label`):
```
'details': {'entity': {'dataset': 'trees',
                     'uuid': 'f08a4912-1c45-4dc5-a27a-36a6bb5d2638'},
          'entityDefId': 2,
          'entityId': 2,
          'submissionDefId': 2,
          'submissionId': 2},
```

Tests.

Follows a similar pattern as submission events for entities. Adds more useful info.

Going to make a change on frontend so deleted entities are displayed differently and dont have broken links.

I think this kind of detail about audit log event details isn't documented.

### DIFF
--- a/lib/model/migrations/20230830-01-remove-entity-label-from-audits.js
+++ b/lib/model/migrations/20230830-01-remove-entity-label-from-audits.js
@@ -1,0 +1,17 @@
+// Copyright 2023 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+update audits set details=(details #- '{entity,label}')
+  where action='entity.create'`);
+
+const down = () => {};
+
+module.exports = { up, down };
+
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -9,7 +9,7 @@
 
 const { sql } = require('slonik');
 const { map, mergeLeft } = require('ramda');
-const { Actee, Actor, Audit, Dataset, Form, Project, Submission } = require('../frames');
+const { Actee, Actor, Audit, Dataset, Entity, Form, Project, Submission } = require('../frames');
 const { extender, equals, page, QueryOptions, unjoiner } = require('../../util/db');
 const Option = require('../../util/option');
 const { construct } = require('../../util/util');
@@ -99,7 +99,7 @@ const get = (options = QueryOptions.none) => ({ all }) =>
       return new Audit(row, { actor: row.aux.actor, actee: Option.firstDefined(actees) });
     });
   });
-
+/*
 const _getBySubmissionId = extender(Audit)(Option.of(Actor))((fields, extend, options, submissionId) => sql`
 select ${fields} from audits
   ${extend|| sql`left outer join actors on actors.id=audits."actorId"`}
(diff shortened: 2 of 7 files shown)


matthew-white:
> Need to make small change on frontend to look up label from a new place.

If the entity is deleted, how about we show its UUID, similar to how we show the instance ID if a submission is deleted? Ideally, we also wouldn't render a link if the entity is deleted. I think those are nice-to-haves though: we could probably back to this later once we're thinking more about entity deletion and purging.

lib/model/migrations/20230830-01-remove-entity-label-from-audits.js
```diff
@@ -0,0 +1,17 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+update audits set details=(details #- '{entity,label}')
```
sadiqkhoja, line 11: not super important but uppercase for SQL keywords would be nice

lib/model/query/audits.js
```diff
@@ -100,15 +100,35 @@ const get = (options = QueryOptions.none) => ({ all }) =>
...
 
-const _getBySubmissionId = extender(Audit)(Option.of(Actor))((fields, extend, options, submissionId) => sql`
-select ${fields} from audits
-  ${extend|| sql`left outer join actors on actors.id=audits."actorId"`}
-  where (details->'submissionId'::text)=${submissionId}
-  order by "loggedAt" desc, audits.id desc
+const _getBySubmissionId = extender(Audit)(Option.of(Actor), Option.of(Entity), Option.of(Entity.Def.alias('current_entity_def', 'currentVersion')))((fields, extend, options, submissionId) => sql`
+SELECT ${fields} FROM audits
+${extend|| sql`
+  left outer join actors on actors.id=audits."actorId"
```
sadiqkhoja, line 106: uppercase for SQL keywords

test/integration/api/submissions.js
```diff
@@ -3012,6 +3012,22 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff
                 body[2].details.should.eql({ instanceId: 'one', submissionId, submissionDefId: submissionDefIds[1] });
               }))))));
 
+    it.skip('should not expand actor when not extended', testService((service) =>
```
sadiqkhoja, line 3015: skipped test?

sadiqkhoja (approved):
I like the response of this API now

Issue #949, opened 2023-08-11
Update API docs for 2023.4


ktuite:
**The goal to write these API docs in the OpenAPI format.**

- [x] If SSO is enabled, then the API won't be accessible. #910
- [x] API doc changes for https://github.com/getodk/central-backend/pull/946
    * add sentence to `### Listing Projects [GET]` about also getting count of dataset and last entity timestamp when requesting extended metadata
    * add section paralleling `### Listing Projects with nested Forms [GET /v1/projects?forms=true]` about nested Datasets

PR #960, opened 2023-08-24
fixes #464: always return source forms

sadiqkhoja:
Closes getodk/central#464

Integration tests and manual verification

This is the simplest solution, other thing that could do was return properties without forms and a separate field to specify relation between sourceForm and properties

None

Should be done in #949 

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -891,7 +891,7 @@ describe('datasets and entities', () => {
           });
       }));
 
-      // bug #464
+      // bug central#464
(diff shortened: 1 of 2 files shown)


test/integration/api/datasets.js
```diff
@@ -887,6 +891,23 @@ describe('datasets and entities', () => {
           });
       }));
 
+      // bug #464
```
matthew-white, line 894: Maybe indicate that it's central#464, not central-backend#464?

matthew-white (approved):

PR #962, opened 2023-08-28
tests: always use real bcrypt lib

alxndrsn:
* remove dependency injection of `bcrypt`
* remove `mock-bcrypt.js`
* remove multiple paths to `password.verify()`/`verifyPassword()` and `password.hash()`/`hashPassword()`
* cut CI build times from ~8.5 mins to ~2.5 mins


matthew-white:
How does bcrypt with a salt of 1 compare to the bcrypt mock in terms of speed? That is, how do times compare when testing locally (where `BCRYPT=no` previously)?

alxndrsn:
I've just run `time make test` locally, once for each:

### `master`
```
real	3m20.993s
user	1m39.687s
sys	0m10.963s
```

### this branch
```
real	3m21.863s
user	1m41.048s
sys	0m10.810s
```

alxndrsn:
It's a surprising result though, as both `bcrypt.hashSync()` and `bcrypt.validateSync()` take about 1 second on my machine with a cost factor of 4.

sadiqkhoja:
I like this change, and want this to be merged soon

matthew-white:
I haven't taken a close look, but no objection on my end! Maybe @alxndrsn should rebase, then @sadiqkhoja should review.

One thing that comes to mind is that this PR seems thematically related to #990 in that both use environment variables rather than dependency injection, which we've tended to use in the past. #990 is partly meant to facilitate testing of the CLI, for which we don't have dependency injection set up: see #937. I don't think there's an issue with a greater use of environment variables, but I thought I'd note the pattern.

alxndrsn:
> I haven't taken a close look, but no objection on my end! Maybe @alxndrsn should rebase, then @sadiqkhoja should review.

This has been rebased :+1:

Makefile
```diff
@@ -105,4 +101,4 @@ check-file-headers:
 .PHONY: api-docs
 api-docs:
 	(test "$(docker images -q odk-docs)" || docker build --file odk-docs.dockerfile -t odk-docs .) && \
-	docker run --rm -it -v ./docs/api.yaml:/docs/docs/_static/api-spec/central.yaml -p 8000:8000 odk-docs
\ No newline at end of file
+	docker run --rm -it -v ./docs/api.yaml:/docs/docs/_static/api-spec/central.yaml -p 8000:8000 odk-docs
```
alxndrsn, line 104: revert this change

matthew-white:
@sadiqkhoja, I know you're focused on web forms at the moment, but would you be up for reviewing this PR? I can also take a look if you don't have time.

sadiqkhoja (approved):
this is a nice enhancement

PR #965, opened 2023-08-30
Api entity lists

lognaturel:
Progress towards https://github.com/getodk/central/issues/466

Ran it with https://github.com/getodk/docs/pull/1601/files

I tried a few different ways of presenting this information. I landed on a warning box because it's eye-catching and can be used in different contexts. It also feels appropriate: we want to warn users that they might have seen another word for this concept.

Docs-only.

That's what this is!

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -18,13 +18,17 @@ info:
 
     The `/projects/:id/app-users` subresource can be used to create, manage, and delete **App Users**.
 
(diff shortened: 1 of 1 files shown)


docs/api.yaml
```diff
@@ -18,14 +18,18 @@ info:
...
+    The `/projects/:id/forms` resource and its subresource `/projects/:id/forms/…/submissions` provide full access to create, manage, and delete **`Form`s and `Submission`s to those Forms**. Each Form is a single [ODK XForms form](https://getodk.github.io/xforms-spec/), and many Submissions (filled-out forms, also sometimes called `Instance`s) may be attached to each Form. These resources are are created by sending XML in the ODK XForms format instead of JSON. One can also retrieve all the multimedia attachments associated with any submission through the `/projects/:id/forms/…/submissions/…/attachments` subresource.
```
matthew-white, line 21: Second "are" after "These resources are"

matthew-white (approved):

docs/api.yaml
lognaturel: Sorry, thank you!

PR #968, opened 2023-09-09
Enumerator should be able to access only open Form

sadiqkhoja:
Closes [sec#2](https://github.com/getodk/security/issues/2)

**Impact:**

Following endpoints were accessible for "App User", "Data Collector" and "Public link" when Form was closed, after this change they return 403:

- /v1/projects/1/forms/:name
- /v1/projects/1/forms/:name.xls
- /v1/projects/1/forms/:name.xml
- /v1/projects/1/forms/:name/versions
- /v1/projects/1/forms/:name/fields
- /v1/projects/1/forms/:name/manifest
- /v1/projects/1/forms/:name/attachments
- /v1/projects/1/forms/:name/attachments/:name.csv

Integration tests + Manual testing

Discussed the solution over Slack

None

No

### DIFF
--- a/lib/model/query/auth.js
+++ b/lib/model/query/auth.js
@@ -20,17 +20,21 @@ with recursive implied(id) as (
   (select unnest(ARRAY[ parent, species ]) from actees
     join implied on implied.id=actees.id))`;
 
-const can = (actor, verb, actee) => ({ oneFirst }) => {
+const _verbClause = (verbs) => sql.join(Array.from(verbs).map(v => sql`verbs ? ${v}`), sql` or `);
+
+const can = (actor, verbs, actee) => ({ oneFirst }) => {
   const acteeId = actee.acteeId || actee;
+
+  const verbsSet = new Set(typeof verbs === 'string' ? [verbs] : verbs);
 
   if (actor.acteeId === acteeId) {
     // special privileges actors always get on themselves.
-    if ((verb === 'user.read') || (verb === 'user.update'))
+    if ((verbsSet.has('user.read')) || (verbsSet.has('user.update')))
       return resolve(true);
 
     // all actors except app users can always log themselves out.
     // (public links cannot reach this route.)
-    if ((verb === 'session.end') && (actor.type !== 'field_key'))
+    if ((verbsSet.has('session.end')) && (actor.type !== 'field_key'))
       return resolve(true);
   }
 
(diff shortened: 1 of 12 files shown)


lib/model/query/auth.js
```diff
@@ -47,6 +51,10 @@ limit 1`)
 const canAssignRole = (actor, role, actee) => ({ Auth }) =>
   Auth.verbsOn(actor.id, actee).then((hasArray) => {
     const has = new Set(hasArray);
+    // `open_form` is subset of `form` so if someone has grant access on `form`
+    // they should be able do it on `open_form` as well
+    if (has.has('form.list')) has.add('open_form.list');
+    if (has.has('form.read')) has.add('open_form.read');
```
matthew-white, line 57: Rather than inferring `open_form.list` and `open_form.read` from `form.list` and `form.read`, could we add `open_form.list` and `open_form.read` to the list of verbs for admins and project managers? I think that would simplify some of the SQL below as well.

ktuite (approved):
Talked to Sadiq interactively just now and I'm convinced that this is a sensible tradeoff of code complexity vs. cleanly reasoning about the verbs. 

With this approach, the meaning of `form.list` and `form.read` doesn't change for viewers/managers/admins, it continues to mean "access to all forms regardless of their state". 

The lower access roles have their `form` verbs removed and replaced with `open_form` verbs, which also makes sense -- we're limiting the access for these roles.

If we were instead to add `open_form.list/read` to the higher access roles and try to use it, we would need to forever maintain this double verb (`form` + `open_form`) and `form` on its own might take on a kind of unintentional `closed_form` meaning. Anyone who happened to create custom roles (through the DB) would also need to know about this double verb and need to remember to add the lower level `open_form` verb to make things work.

(One example where we already have to check a double verb is in the `_getSql` query of `Projects` where users must have `project.read` and `form.list`. This check makes sense to me, though, and isn't nearly as complicated as the `Forms` queries.)

Caveats I remember from code review:
- `canAssignRole()` has to change because higher access roles like admins/managers are expected to have all the verbs they assign to lower access roles, but admins don't have `open_form` verb so the function fakes it if that role has the superset `form` verb.

matthew-white:
Very helpful explanation, that sounds reasonable to me. I can see how there are tradeoffs here. It's a tricky problem!

matthew-white:
Just to make sure I'm understanding, Data Collectors will still be able to access closing forms over the API, right? So they'll see a closing form in the form list, but then Frontend will restrict them from actually navigating to the form page? If that's the case, I'm thinking of filing a separate Frontend issue about that. Frontend could hide closing forms from Data Collectors in the forms table.

sadiqkhoja:
>  I'm thinking of filing a separate Frontend issue about that. Frontend could hide closing forms from Data Collectors in the forms table.

I am not sure what's the ideal here. We allow "App Users" and "Public links" to submit against "Closing Forms", by that logic "Data Collector" should also be able to submit. But if they can't see the form in the form list then how are they gonna submit, by saving Enketo URL somewhere offline?

PR #969, opened 2023-09-12
Improve flow of SSO account creation email

lognaturel:
Adds "Otherwise" to more closely match other similar emails.

Only visual verification.

I considered moving the "please sign in" up to the first paragraph but I decided to match the structure of other existing emails instead.


matthew-white (approved):
Just to link things, @sadiqkhoja suggested something similar here: https://github.com/getodk/central-backend/pull/910#discussion_r1311007114

PR #973, opened 2023-09-12
Add Postgres version requirement

lognaturel:
Making this implicit requirement explicit!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![License](https://img.shields.io/badge/license-Apache_2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 [![Build status](https://circleci.com/gh/getodk/central-backend.svg?style=shield)](https://circleci.com/gh/getodk/central-backend)
 
-ODK Central Backend is a minimal [ODK](https://getodk.org/) server based on Node.js and Postgres. It is currently under development.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 [![License](https://img.shields.io/badge/license-Apache_2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 [![Build status](https://circleci.com/gh/getodk/central-backend.svg?style=shield)](https://circleci.com/gh/getodk/central-backend)
 
-ODK Central Backend is a minimal [ODK](https://getodk.org/) server based on Node.js and Postgres. It is currently under development.
+ODK Central Backend is the API server for [ODK](https://getodk.org/). It's built with Node.js and Postgres.
```
lognaturel, line 7: Making a couple of quick cosmetic changes while here, hopefully that's ok! For better or worse, I think Central stopped being minimal a little while ago!

We used to use the "It is currently under development" language when we had some dead repos but now those have been removed.

matthew-white:
Just to document one reason why Postgres 14+ is required, @lognaturel wrote on Slack:

> From @eyelidlessness about postgres 14 now being required: it’s about keywords that now can be used as column labels without `AS`

<blockquote>

```
-- Syntax error in 12, 13
SELECT '' action;

-- No syntax error in 14+
SELECT '' as action;
SELECT '' "action";
```
</blockquote>

matthew-white (approved):

PR #974, opened 2023-09-12
Return 404 if accessing unpublished dataset aka entity list

ktuite:
Backend part of https://github.com/getodk/central/issues/467

Tested it with frontend and added a specific test.

When getting metadata about a dataset/entity list, it doesn't even seem to include the `publishedAt` timestamp, so it makes sense to only return published datasets at the `/datasets/:name` url and return 404 resource not found otherwise.

Limits access to unpublished datasets but makes some parts of frontend more consistent.

Writing this out in the API would be a good idea. Tagging https://github.com/getodk/central-backend/issues/949

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -621,6 +621,15 @@ describe('datasets and entities', () => {
 
       }));
 
+      it('should reject if dataset is not published', testService((service) =>
+        service.login('alice', (asAlice) =>
+          asAlice.post('/v1/projects/1/forms')
+            .send(testData.forms.simpleEntity)
+            .set('Content-Type', 'application/xml')
+            .expect(200)
+            .then(() => asAlice.get('/v1/projects/1/datasets/people')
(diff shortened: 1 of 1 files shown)


lib/resources/datasets.js
```diff
@@ -26,6 +27,7 @@ module.exports = (service, endpoint) => {
     Datasets.get(params.projectId, params.name)
       .then(getOrNotFound)
       .then((dataset) => auth.canOrReject('dataset.read', dataset)
+        .then(rejectIf((() => dataset.publishedAt == null), noargs(Problem.user.notFound)))
```
matthew-white, line 30: Rather than adding a separate check, I think you could use the optional third parameter of `Datasets.get()`. That's how some other endpoints are returning a 404 (for example, lib/resources/entities.js).

matthew-white (approved):

ktuite:
This issue was actually fixed by #934 and now this PR is just a new test :D

Issue #971, opened 2023-09-12
SSO: render simple redirect page rather than copying Frontend styles


matthew-white:
Continuing the discussion in #910:

- https://github.com/getodk/central-backend/pull/910#discussion_r1310995967
- https://github.com/getodk/central-backend/pull/910#discussion_r1277685758

I suggested there that it'd be better for Backend to render a simple redirect page rather than trying to match Frontend's styles. I think trying to match styles would lead to a maintenance burden as styles change over time. On wider viewports, the current styles don't match on the max width of the panel to an extent that I think would be noticeable to users.

@alxndrsn, let's continue discussing here! Let me know what you think about my most recent comment.

matthew-white:
I brought up this question in a meeting earlier today, and @issa-tseng and @lognaturel offered some thoughts from a design perspective. (I'll try to summarize here, but please correct anything I get wrong.)

- @issa-tseng said that either way would be fine from a design perspective. A bare page wouldn't be a problem. If we do opt for a branded redirect page, the styles should exactly match Frontend.
- @lognaturel also thought that a bare page would be fine.
- Specific idea: Render a bare page with centered text. Use the following text: "Redirecting. Click here if you are not automatically redirected to Central within 5 seconds."
  - @lognaturel said that the text "to Central" could provide branding/continuity for anyone who actually ends up seeing that page for any length of time.
  - One goal is for this text to be short so that users don't feel like they've missed something when the page flashes by.
- Another idea from @issa-tseng: Don't show anything at all on that page at first. After 5 seconds, show some text (using a CSS animation).

@alxndrsn, do any of these ideas seem reasonable to you? There may be other options as well that we haven't considered.

lognaturel:
A quick note that @ktuite brought up and I agree with -- the current page goes by so fast it's hard not to think that something is broken. The advantage of having something really simple like "Redirecting" or "Loading" or a spinner is that it would be understandable in the very short time it's up.

alxndrsn:
Thanks for all the thoughts on this.  

I've opened a PR at https://github.com/getodk/central-backend/pull/981 with the following behaviour:

1. initially: centred animated spinner is displayed

![Screenshot_2023-09-13_11-49-13](https://github.com/getodk/central-backend/assets/191496/c8e85154-a9cc-4cd5-993f-232692857cf1)

2. after 10 seconds: "continue" link fades in below

[...]

PR #981, opened 2023-09-13
oidc: show unbranded spinner on forwarding page

alxndrsn:
Spinner source: https://loading.io/css/
Licence: CC0 (public domain)

Closes https://github.com/getodk/central-backend/issues/971

-----

# Behaviour

1. initially: centred animated spinner is displayed

Screenshot_2023-09-13_11-49-13

2. after 10 seconds: "continue" link fades in below

Screenshot_2023-09-13_11-49-22

### DIFF
--- a/lib/http/endpoint.js
+++ b/lib/http/endpoint.js
@@ -241,7 +241,11 @@ const defaultEndpoint = endpointBase({
 const htmlEndpoint = endpointBase({
   resultWriter: (result, request, response) => {
     response.type('text/html');
-    response.send(frontendPage(result));
+    if (typeof result === 'string') {
+      response.send(result);
+    } else {
+      response.send(frontendPage(result));
+    }
   },
   errorWriter: (error, request, response) => {
     response.type('text/html');
--- a/lib/resources/oidc.js
+++ b/lib/resources/oidc.js
@@ -118,13 +118,47 @@ module.exports = (service, endpoint) => {
       // Instead, we need to render a page and then "browse" from that page to the normal frontend:
 
       // id=cl only set for playwright. Why can't it locate this anchor in any other way?
-      return {
-        head: html`<meta http-equiv="refresh" content="0; url=${nextPath}">`,
-        body: html`
-          <h1>Authentication Successful</h1>
-          <div><a href="${nextPath}" id="cl">Continue to ODK Central</a></div>
-        `,
-      };
+      return html`
+        <!doctype html>
+        <html>
+          <head>
+            <title>Loading... | ODK Central</title>
+            <meta http-equiv="refresh" content="0; url=${nextPath}">
+            <style>
+              #container { text-align:center }
+
+              .lds-spinner { display:inline-block; position:relative; width:80px; height:80px }
+              .lds-spinner div { transform-origin:40px 40px; animation:lds-spinner 1.2s linear infinite }
+              .lds-spinner div:after {
+                content:" "; display:block; position:absolute; top:3px; left:37px;
+                width:6px; height:18px; border-radius:20%; background:#000;
+              }
+              .lds-spinner div:nth-child(1) { transform:rotate(0deg); animation-delay:-1.1s }
+              .lds-spinner div:nth-child(2) { transform:rotate(30deg); animation-delay:-1s }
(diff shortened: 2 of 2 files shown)


matthew-white:
@sadiqkhoja, if you have time to review this before release, we could try to get this change into .4. We should only do so if you evaluate that the risk is low and if someone is able to verify the change on staging. I'm excited for this change! I also think it'd be OK to wait until .5 to ship it.

lib/resources/oidc.js
```diff
@@ -118,13 +118,47 @@ module.exports = (service, endpoint) => {
...
-          <h1>Authentication Successful</h1>
-          <div><a href="${nextPath}" id="cl">Continue to ODK Central</a></div>
-        `,
-      };
+      return html`
+        <!doctype html>
+        <html>
+          <head>
+            <title>Loading... | ODK Central</title>
+            <meta http-equiv="refresh" content="0; url=${nextPath}">
```
matthew-white, line 126: Looking at the `${nextPath}` interpolation, is the `content` attribute being escaped for HTML? Is there any risk of a maliciously constructed `next` query parameter that includes a quote, thereby escaping the attribute? If so, I'd start worrying about XSS and such, e.g., somehow breaking out of the `<meta>` tag and inserting a `<script>` tag. Or does `safeNextPathFrom()` ensure against that? It'd be great if we had tests related to that. (Maybe we do, and I've just missed them.)

alxndrsn: @matthew-white Great questions, and I think they apply equally to `master` as this PR.

alxndrsn: @matthew-white PR at https://github.com/getodk/central-backend/pull/1003 for more test cases for `safeNextPathFrom()`

matthew-white: Awesome, I just approved that PR. I think there are still issues here though related to ampersands. It looks like `new URL()` doesn't encode `&` and `;`, thereby opening the door to HTML entities. I'm pretty sure that HTML entities are valid in HTML attributes.

For example, `safeNextPathFrom()` prevents `next=/login`, but I'm wondering if HTML entities could be used to bypass that check along the lines of `next=/login/foo&sol;&period;&period;`. As a side note, I actually don't think it's important for `safeNextPathFrom()` to prevent `next=/login`: it's important in Frontend just to prevent the `AccountLogin` component from navigating to itself, which would cause a Vue Router error. But I do think that ampersands open up possibilities that we haven't thought through.

Is there a function we could call to encode `nextPath` for HTML? Could we do that rather than trying to prevent HTML special characters in `nextPath`? I feel like there should be something in Express or another package to do that. I think that would result in a neater separation of concerns, making it clear that `safeNextPathFrom()` is mostly meant to prevent a redirect away from Central, not prevent XSS.

lib/http/endpoint.js
```diff
@@ -241,7 +241,11 @@ const defaultEndpoint = endpointBase({
 const htmlEndpoint = endpointBase({
   resultWriter: (result, request, response) => {
     response.type('text/html');
-    response.send(frontendPage(result));
+    if (typeof result === 'string') {
+      response.send(result);
+    } else {
+      response.send(frontendPage(result));
```
matthew-white, line 247: How else is `frontendPage()` used? I think the goal should be to remove `frontendPage()` entirely.

lib/resources/oidc.js
alxndrsn: > Is there a function we could call to encode `nextPath` for HTML?

I think that's what `safeNextPath()` should be.  If there is a standard approach to this (especially with a path to supporting search/fragment switching & enketo URLs), I'd be happy to use that instead.

> For example, `safeNextPathFrom()` prevents `next=/login`

I'll remove this exception.

> but I'm wondering if HTML entities could be used to bypass that check along the lines of `next=/login/foo&sol;&period;&period;`.

They can.  `next=/login/foo&sol;&period;&period;` will be treated the same as `next=/login/foo/..`, which there isn't a specific problem with.  Perhaps an issue with escaping enketo URLs?  I'll add test cases for that as well.

ktuite (approved):

PR #985, opened 2023-09-14
fixes #949: Updated API doc for the changes done in 2023.4

### LINKED ISSUE #949: Update API docs for 2023.4

sadiqkhoja:
- Basic Auth not available when SSO is enable
- ?datasets=true for project listing
- Version number for Entities

Closes #949

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -42,6 +42,12 @@ info:
 
     - [DELETE /sessions/current](/central-api-authentication/#logging-out-current-session) logs out the current session.
 
+    **Changed**:
+
+    - If single sign-on (SSO) with Open ID Connect (OIDC) is enabled then HTTP Basic Authentication and login with `POST /v1/session` are disabled.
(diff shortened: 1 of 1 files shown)


ktuite (approved):
Can you add `sourceForms` around line 10459 to the example dataset metadata response to go with https://github.com/getodk/central-backend/pull/960? I'm seeing `sourceForms` defined somewhere in the code but I'm not seeing it come through -- i'm looking at the output of the yaml file in a swagger editor but maybe i should be inspecting it a different way. 

Can you add a note to `/projects/{projectId}/datasets/{name}:` that it will only return metadata for published datasets for https://github.com/getodk/central-backend/pull/974?

PR #994, opened 2023-09-18
GitHub Actions: set job timeouts

alxndrsn:
Without these, a runaway job can continue for the default job timeout of **[6 hours](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)**.

I've set the timeouts as approximately double the time I've seen for recent ✅ green builds.


lognaturel (approved):
This seems helpful and very low risk. Worst case, some of these start timing out and we can adjust the configuration.

PR #989, opened 2023-09-15
Request Enketo IDs during request when form is created or published

matthew-white:
This PR addresses three cases related to getodk/central#385:

- When a form is created for the first time, request an `enketoId` for the draft.
- When a form is published for the first time, request an `enketoId` and `enketoOnceId` for the form.
- When a form is created for the first time and immediately published (via the `?publish=true` flag of the API, thereby skipping a draft), request an `enketoId` and `enketoOnceId` for the form.

These are the last cases to address, so this PR closes getodk/central#385.

#### Does this change require updates to the API documentation? If so, please update docs/api.md as part of this PR.

I quickly searched the API docs for "Enketo", but I didn't see anything that needed to case.

#### Before submitting this PR, please make sure you have:

- [x] run `make test-full` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1001, opened 2023-09-22
Test that Enketo IDs are requested during request to create form

matthew-white:
Follow-up PR to #989, adding more tests. Only test/ is changed, not lib/.

#989 included tests of requests to publish an existing form. This PR also tests requests to create a form (either as a draft form or as an immediately published form).

### DIFF
--- a/test/integration/api/forms/draft.js
+++ b/test/integration/api/forms/draft.js
@@ -1251,7 +1251,7 @@ describe('api: /projects/:id/forms (drafts)', () => {
       it('should request Enketo IDs from worker if request from endpoint fails', testService(async (service, container) => {
         const asAlice = await service.login('alice');
 
-        // First request to Enketo, from endpoint
+        // First request to Enketo, from the endpoint
         await asAlice.post('/v1/projects/1/forms')
           .send(testData.forms.simple2)
           .set('Content-Type', 'application/xml')
@@ -1264,7 +1264,7 @@ describe('api: /projects/:id/forms (drafts)', () => {
         should.not.exist(beforeWorker.enketoId);
         should.not.exist(beforeWorker.enketoOnceId);
 
-        // Second request, from worker
+        // Second request, from the worker
         global.enketo.callCount.should.equal(2);
         await exhaust(container);
         global.enketo.callCount.should.equal(3);
--- a/test/integration/api/forms/draft.js
+++ b/test/integration/api/forms/draft.js
@@ -123,9 +123,9 @@ describe('api: /projects/:id/forms (drafts)', () => {
         should.not.exist(body.enketoId);
       }));
 
-      it('should stop waiting for Enketo after 0.5 seconds @slow', testService(async (service) => {
+      it('should wait for Enketo only briefly @slow', testService(async (service) => {
         const asAlice = await service.login('alice');
-        global.enketo.wait = (f) => { setTimeout(f, 501); };
+        global.enketo.wait = (done) => { setTimeout(done, 600); };
         await asAlice.post('/v1/projects/1/forms/simple/draft').expect(200);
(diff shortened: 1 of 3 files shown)


test/integration/api/forms/draft.js
```diff
@@ -123,9 +123,9 @@ describe('api: /projects/:id/forms (drafts)', () => {
         should.not.exist(body.enketoId);
       }));
 
-      it('should stop waiting for Enketo after 0.5 seconds @slow', testService(async (service) => {
+      it('should wait for Enketo only briefly @slow', testService(async (service) => {
         const asAlice = await service.login('alice');
-        global.enketo.wait = (f) => { setTimeout(f, 501); };
+        global.enketo.wait = (done) => { setTimeout(done, 600); };
```
matthew-white, line 128: The change from 501 to 600 is meant to address this intermittent test failure: https://github.com/getodk/central-backend/issues/588#issuecomment-1722828303. Note that this test was not introduced in #989, but rather #782. I don't think I did anything in #989 that would have changed the POST/GET &hellip;/draft endpoints. However, if this test is failing, it certainly seems possible that new tests in #989 could fail, so I think it's worth addressing now.

Here's one possible explanation for the failure:

- A request to Enketo is sent. After 501 ms, the Enketo mock will respond.
- A moment later, the endpoint time-bounds the request. After 500 ms, the endpoint will stop waiting for the request.
- If "a moment later" is greater than 1 ms, then the test will fail.

Another possible explanation:

- `setTimeout()` doesn't run callbacks after an exact number of milliseconds, but after _at least_ the specified number of milliseconds. So the Enketo mock could respond after 502 ms instead of after 501 ms.
- Similarly, the endpoint could stop waiting after 502 ms instead of after 500 ms.
- In other words, both timers could exceed the number of milliseconds specified and maybe be processed together in a single batch. In that case, the order in which the callbacks are run might be indeterminate.

I'm not sure what the exact explanation is, but the fact that the test is failing intermittently doesn't make me think that there's anything wrong with the functionality here. I think we just need to provide a little more wiggle room for the test.

test/integration/api/forms/forms.js
```diff
@@ -324,6 +297,156 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
...
+
+    describe('Enketo IDs for published form', () => {
+      it('should request Enketo IDs', testService(async (service, { env }) => {
+        const asAlice = await service.login('alice');
+        const { body } = await asAlice.post('/v1/projects/1/forms?publish=true')
+          .set('Content-Type', 'application/xml')
+          .send(testData.forms.simple2)
+          .expect(200);
+        global.enketo.callCount.should.equal(1);
+        without(['token'], global.enketo.createData).should.eql({
```
matthew-white, line 384: I don't think we need to test the `token` argument at this time. Unlike in the draft case above, `token` won't be `undefined`. However, I'm not sure that the token is even needed for this flow to function: see #997. We don't test the token in existing tests.

test/util/enketo.js
```diff
@@ -24,6 +24,8 @@ const defaults = {
   // The number of times that the mock has been called during the test, that is,
   // the number of requests that would be sent to Enketo
   callCount: 0,
+  // An object with a property for each argument passed to the create() method
+  createData: undefined,
```
matthew-white, line 28: I added `createData` instead of using `receivedUrl` in the new tests here, for two reasons:

- This allows us to test the `xmlFormId` argument of `enketo.create()` (the second argument), not just the first argument (`openRosaUrl`/`receivedUrl`).
- Consistency with `editData` elsewhere in this file, which we use to test `enketo.edit()`.

matthew-white:
@ktuite, I think this can be reviewed async, but let me know if you'd like to review interactively!

ktuite (approved):

PR #1002, opened 2023-09-22
Update systeminformation dependency

matthew-white:
I updated the `systeminformation` package by running `npm audit fix`. `systeminformation` is a subdependency of `pm2`:

```
$ npm ls systeminformation
odk-central@0.1.0
└─┬ pm2@5.3.0
  └─┬ pm2-sysmonit@1.2.8
    └── systeminformation@5.21.8
```

Just running tests.

Given that the package is a subdependency within the allowed semver range, I'm hopeful that any risk of regression is low.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9446,9 +9446,9 @@
       }
     },
     "node_modules/systeminformation": {
-      "version": "5.17.12",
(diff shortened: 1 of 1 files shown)


ktuite (approved):
I think this should be ok to merge before release, risk seems low.

PR #1003, opened 2023-09-23
safeNextPathFrom(): add tests for bad content

alxndrsn:


test/unit/util/html.js
```diff
@@ -30,6 +30,10 @@ describe('util/html', () => {
       [ '/-/xyz',                    'http://localhost:8989/-/xyz' ], // eslint-disable-line no-multi-spaces
       [ '/account/edit',             '/#/account/edit' ],             // eslint-disable-line no-multi-spaces
       [ '/users',                    '/#/users' ],                    // eslint-disable-line no-multi-spaces
+      [ '/users"><badTag ',          '/#/users%22%3E%3CbadTag' ],     // eslint-disable-line no-multi-spaces
```
matthew-white, line 33: Just to be extra sure, could we add tests for a quote in a query string and for a quote after a hash? I think there are subtle differences in the encoding requirements for those parts of the URL, so it'd be reassuring to test that those don't provide a way for a quote to get through.

```
/users?foo="
/users#"
```

matthew-white (approved):

test/unit/util/html.js
alxndrsn: Added - please rereview.

PR #1005, opened 2023-09-26
Exploratory PR for starting to update entities from submissions

ktuite:
Closes https://github.com/getodk/central/issues/509

This uses the existing submission parsing and existing update entity via PATCH infrastructure to update entities from submissions. 

There's more to do but I thought it might make sense to merge something basic in early.

Related issues that could be finished while other work starts in parallel
- https://github.com/getodk/central/issues/510
- https://github.com/getodk/central/issues/514

I'm trying to get something in place so other improvements (conflicts, merging strategies) can be worked on asap.

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -70,6 +70,8 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject) 
       entity.system.dataset = field.attrs.dataset;
       entity.system.id = field.attrs.id;
       entity.system.create = field.attrs.create;
+      entity.system.update = field.attrs.update; // TODO: add tests
+      entity.system.baseVersion = field.attrs.baseVersion; // TODO: add tests
     } else if (field.path.indexOf('/meta/entity') === 0)
       entity.system[field.name] = text;
     else if (field.propertyName != null)
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -110,7 +110,7 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
     });
 };
 
-const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent) => async ({ Entities }) => {
+const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event) => async ({ Entities }) => {
 
   // Get version of entity on the server
   const serverEntity = await Entities.getById(dataset.id, entityData.system.id).then(o => o.get());
@@ -121,11 +121,7 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
 
   // make some kind of source object
   const sourceDetails = {
-    submission: { instanceId: submissionDef.instanceId },
-    dataParsedFromSubmission: entityData.data, // or some other info
-    labelParsedFromSubmission: entityData.label,
-    baseVersionParsedFromSubmission: 0,
-    parentEventId: parentEvent ? parentEvent.id : undefined,
+    submission: { instanceId: submissionDef.instanceId }
   };
   const sourceId = await Entities.createSource(sourceDetails, submissionDefId, event.id);
 
@@ -193,7 +189,7 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Datasets, Entit
     return Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent);
 
   if (entityData.system.update === '1' || entityData.system.update === 'true')
-    return Entities._updateEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent);
+    return Entities._updateEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event);
 
   return null;
 };
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -10,7 +10,7 @@
 const { sql } = require('slonik');
 const { Actor, Entity, Submission, Form } = require('../frames');
 const { equals, extender, unjoiner, page, markDeleted } = require('../../util/db');
-const { map } = require('ramda');
+const { map, mergeRight } = require('ramda');
 const { blankStringToNull, construct } = require('../../util/util');
 const { QueryOptions } = require('../../util/db');
 const { odataFilter } = require('../../data/odata-filter');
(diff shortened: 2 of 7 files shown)


lib/model/query/entities.js
```diff
@@ -195,9 +229,9 @@ const createEntitiesFromPendingSubmissions = (submissionEvents, parentEvent) =>
 ////////////////////////////////////////////////////////////////////////////////
 // UPDATING ENTITIES
 
-const createVersion = (dataset, entity, data, label, version, sourceId, userAgentIn = null) => ({ context, one }) => {
+const createVersion = (dataset, entity, data, label, version, sourceId, userAgentIn = null, submitterId = null) => ({ context, one }) => {
   // dataset is passed in so the audit log can use its actee id
-  const creatorId = context.auth.actor.map((actor) => actor.id).orNull();
+  const creatorId = (context.auth ? context.auth.actor.map((actor) => actor.id).orNull() : submitterId);
```
ktuite, line 234: this whole inferring the `creatorId` from the context is awkward and came from when entities were originally updated only via the PATCH api. it doesn't work with other update pathways.

test/data/xml.js
```diff
@@ -649,6 +671,52 @@ module.exports = {
...
+              <orx:instanceName>one</orx:instanceName>
+              <entity dataset="people" id="12345678-1234-4123-8234-123456789aaa" baseVersion="1" update="1">
+                <label>Julie (30)</label>
+              </entity>
+            </meta>
+            <name>Julie</name>
+            <age>30</age>
+          </data>`,
+      three: `<data xmlns:jr="http://openrosa.org/javarosa" xmlns:entities="http://www.opendatakit.org/xforms" id="updateEntity" version="1.0">
+          <meta>
```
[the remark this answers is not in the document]
matthew-white: Also, I only see `one` used, so maybe we just commit that one for now?

test/data/xml.js
```diff
@@ -649,6 +671,52 @@ module.exports = {
...
+                <label>Alicia (85)</label>
+              </entity>
+            </meta>
+            <name>Alicia</name>
+            <age>85</age>
+          </data>`,
+      two: `<data xmlns:jr="http://openrosa.org/javarosa" xmlns:entities="http://www.opendatakit.org/xforms" id="updateEntity" version="1.0">
+            <meta>
+              <instanceID>one</instanceID>
+              <orx:instanceName>one</orx:instanceName>
```
matthew-white, line 689:
```suggestion
              <instanceID>two</instanceID>
              <orx:instanceName>two</orx:instanceName>
```

lib/model/query/entities.js
```diff
@@ -89,9 +89,53 @@ const createSource = (details = null, subDefId = null, eventId = null) => ({ one
...
+  const sourceId = await Entities.createSource(sourceDetails, submissionDefId, event.id);
+
+  // TODO: createVersion infers creator id from context.auth, shoving the submitter id in as another arg
+  // should be refactored
+  return Entities.createVersion(dataset, serverEntity, mergedData, mergedLabel, serverEntity.aux.currentVersion.version + 1, sourceId, null, submissionDef.submitterId);
+};
 
 // Entrypoint to where submissions (a specific version) become entities
-const _processSubmissionEvent = (event, parentEvent) => async ({ Datasets, Entities, Submissions, Forms, Audits }) => {
+const _processSubmissionEvent = (event, parentEvent) => async ({ Datasets, Entities, Submissions, Forms }) => {
```
matthew-white, line 138: (Not sure exactly where to put this comment.) An entity update should be processed even if the entity list requires approval for entity _creation_. So I think this PR should update this `if`:

https://github.com/getodk/central-backend/blob/6f7750b19f95e34b6b20f5f13891f6fb027cd393/lib/model/query/entities.js#L143-L145

Maybe that `if` should be moved to `_createEntity()`.

lib/model/query/entities.js
```diff
@@ -89,9 +89,53 @@ const createSource = (details = null, subDefId = null, eventId = null) => ({ one
...
+    {
+      entityId: entity.id, // Added in v2023.3 and backfilled
+      entityDefId: entity.aux.currentVersion.id, // Added in v2023.3 and backfilled
+      entity: { uuid: entity.uuid, dataset: dataset.name },
+      submissionId,
+      submissionDefId
+    });
+};
+
+const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent) => async ({ Entities }) => {
```
matthew-white, line 113: It doesn't look like `submissionId` is used in this function.

lib/model/query/entities.js
```diff
@@ -89,9 +89,53 @@ const createSource = (details = null, subDefId = null, eventId = null) => ({ one
...
+
+  // Get version of entity on the server
+  const serverEntity = await Entities.getById(dataset.id, entityData.system.id).then(o => o.get());
+
+  // merge data
+  const mergedData = mergeRight(serverEntity.aux.currentVersion.data, entityData.data);
+  const mergedLabel = entityData.system.label || serverEntity.aux.currentVersion.label;
```
matthew-white, line 120: The updates should be applied on top of the specified base version, not on top of the current version. But we could come back to that.

lib/model/query/entities.js
```diff
@@ -89,9 +89,53 @@ const createSource = (details = null, subDefId = null, eventId = null) => ({ one
...
+
+  // make some kind of source object
+  const sourceDetails = {
+    submission: { instanceId: submissionDef.instanceId },
+    dataParsedFromSubmission: entityData.data, // or some other info
+    labelParsedFromSubmission: entityData.label,
+    baseVersionParsedFromSubmission: 0,
+    parentEventId: parentEvent ? parentEvent.id : undefined,
```
matthew-white, line 128: I don't feel like `parentEvent.id` needs to be stored here. For entity creation, changing `approvalRequired` to `false` can cause entities to be created from pending submissions. Without the change to `false`, the entity wouldn't have been created. However, that's not the case for entity updates: the `approvalRequired` flag shouldn't affect whether an entity update is completed. So I don't think we should log a change to the flag as a `parentEvent`.

Changing `approvalRequired` to `false` might happen to cause the entity update to be processed sooner than it would have otherwise, if there's a backlog in the worker queue. That's because `Entities.createEntitiesFromPendingSubmissions()` processes submission events. Yet even in that case, I don't think the change to `approvalRequired` qualifies as a `parentEvent`. The entity update would have been processed anyway, just later.

I also think we could circle back to these questions later. In that case, it'd be great if you could track this question in an issue (new or existing).

lib/model/query/entities.js
```diff
@@ -89,9 +89,53 @@ const createSource = (details = null, subDefId = null, eventId = null) => ({ one
...
+  };
+  const sourceId = await Entities.createSource(sourceDetails, submissionDefId, event.id);
+
+  // TODO: createVersion infers creator id from context.auth, shoving the submitter id in as another arg
+  // should be refactored
+  return Entities.createVersion(dataset, serverEntity, mergedData, mergedLabel, serverEntity.aux.currentVersion.version + 1, sourceId, null, submissionDef.submitterId);
```
matthew-white, line 134: Above, `submissionId` and `submissionDefId` are logged in the details of the `entity.create` event. However, `Entities.createVersion.audit` won't log those things. We could circle back to that in getodk/central#513, but I thought I'd mention it.

matthew-white (commented):
I left a number of thoughts, but I think many of these we could come back to in future PRs. I think it'd be good to make a few changes now though.

lib/model/query/entities.js
matthew-white: It looks like this is addressed in #1013.

lib/model/query/entities.js
matthew-white: I feel like `creatorId` should be a required parameter of `createVersion()`, then the endpoint should call `createVersion()` with `context.auth.actor...`. Then `context` wouldn't need to appear here at all.

lib/model/query/entities.js
```diff
@@ -89,9 +89,53 @@ const createSource = (details = null, subDefId = null, eventId = null) => ({ one
...
+  // merge data
+  const mergedData = mergeRight(serverEntity.aux.currentVersion.data, entityData.data);
+  const mergedLabel = entityData.system.label || serverEntity.aux.currentVersion.label;
+
+  // make some kind of source object
+  const sourceDetails = {
+    submission: { instanceId: submissionDef.instanceId },
+    dataParsedFromSubmission: entityData.data, // or some other info
+    labelParsedFromSubmission: entityData.label,
+    baseVersionParsedFromSubmission: 0,
```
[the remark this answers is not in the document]
ktuite: I was looking at this again, thinking about entity deletion, and yeah, this entity data shouldn't go here

lib/model/query/entities.js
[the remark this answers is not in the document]
ktuite: i'll come back to this, maybe even in the PR that refactors createVersion

ktuite: i'll make sure this is cleaned up in the refactor in #1013

lib/model/query/entities.js
ktuite: I'll remove `parentEvent.id`. I think the rest of your comments about how `approvalRequired` plays into all this, and the order in which things get processed, is one of the themes behind https://github.com/getodk/central/issues/517 so we can come back to it then.

lib/model/query/entities.js
ktuite: that sounds like it would go against "simultaneous updates"

lib/model/query/entities.js
ktuite: Will do in #1013

lib/model/query/entities.js
ktuite: will do in #1013! because of the way i've split things up, it's hard for me to think about where to put these changes. maybe i shouldn't have split it into two PRs...

lib/data/entity.js
```diff
@@ -70,6 +70,8 @@ const parseSubmissionXml = (entityFields, xml) => new Promise((resolve, reject)
       entity.system.dataset = field.attrs.dataset;
       entity.system.id = field.attrs.id;
       entity.system.create = field.attrs.create;
+      entity.system.update = field.attrs.update; // TODO: add tests
+      entity.system.baseVersion = field.attrs.baseVersion; // TODO: add tests
```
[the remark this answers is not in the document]
ktuite: will do in https://github.com/getodk/central/issues/510

lib/worker/jobs.js
```diff
@@ -14,10 +14,10 @@ const jobs = {
     require('./submission.attachment.update')
   ],
 
-  'submission.create': [ require('./submission').submissionCreate, require('./entity').createEntityFromSubmission ],
-  'submission.update.version': [ require('./submission').submissionUpdateVersion, require('./entity').createEntityFromSubmission ],
+  'submission.create': [ require('./submission').submissionCreate, require('./entity').createOrUpdateEntityFromSubmission ],
+  'submission.update.version': [ require('./submission').submissionUpdateVersion, require('./entity').createOrUpdateEntityFromSubmission ],
 
-  'submission.update': [ require('./entity').createEntityFromSubmission ],
+  'submission.update': [ require('./entity').createOrUpdateEntityFromSubmission ],
```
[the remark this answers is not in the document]
ktuite: mentioned in https://github.com/getodk/central/issues/517

matthew-white (approved):
For things that this PR doesn't do, it sounds like there are concrete plans to do them elsewhere. #1013 is up next, and I expect that we'll continue some of these conversations there.

PR #1008, opened 2023-09-27
oidc: encode HTML entities in next URL

alxndrsn:
Alternative approach to https://github.com/getodk/central-backend/pull/1011

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "express": "~4.18",
     "express-graceful-exit": "~0.5",
     "fast-myers-diff": "~3",
+    "html-entities": "^2.4.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     "express": "~4.18",
     "express-graceful-exit": "~0.5",
     "fast-myers-diff": "~3",
+    "html-entities": "^2.4.0",
```
matthew-white, line 27: We already use the `entities` package for decoding. Could we use it for encoding as well?

alxndrsn:
Resolved by https://github.com/getodk/central-backend/pull/1014.

PR #1011, opened 2023-09-29
html``: encode HTML entities in all vars

alxndrsn:
Alternative approach to #1008

### DIFF
--- a/lib/util/html.js
+++ b/lib/util/html.js
@@ -9,12 +9,24 @@
 //
 // Helper functions that relate to the HTML/frontend layer of the application.
 
+const { encodeNonAsciiHTML } = require('entities');
 const config = require('config');
 
 const envDomain = config.get('default.env.domain');
 
-// handy dev function for enabling syntax hilighting of html
-const html = ([ first, ...rest ], ...vars) => first + vars.map((v, idx) => [ v, rest[idx] ]).flat().join('');
+class Html {
+  constructor(...content) { this.content = content; }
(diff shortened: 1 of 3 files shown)


lib/util/html.js
```diff
@@ -9,12 +9,24 @@
 //
 // Helper functions that relate to the HTML/frontend layer of the application.
 
+const { encodeNonAsciiHTML } = require('entities');
 const config = require('config');
 
 const envDomain = config.get('default.env.domain');
 
-// handy dev function for enabling syntax hilighting of html
-const html = ([ first, ...rest ], ...vars) => first + vars.map((v, idx) => [ v, rest[idx] ]).flat().join('');
```
matthew-white, line 17: I don't 100% understand what this did before. It just concatenates everything, right? And the purpose is that your text editor or IDE recognizes <code>html``</code> and does syntax-highlighting?

matthew-white (commented):
This feels like the beginnings of building our own HTML templating, but I feel like it'd be better to use an existing package. We already use mustache in several places in the codebase. Now that I think of it, that's probably why we haven't used the `entities` package to encode before: we've used mustache for that. What do you think about using mustache to encode `nextPath` in /v1/oidc/callback? I know that'd be a third PR after this PR and #1008, but that's feeling like the right approach to me.

alxndrsn:
Resolved by https://github.com/getodk/central-backend/pull/1014.

PR #1013, opened 2023-09-29
Refactor Entities.createVersion

ktuite:
Addresses some clunkiness in https://github.com/getodk/central-backend/pull/1005 but I wanted to put it in a separate PR so the meaning of #1005 would be clearer.

There's a 2x2 square of situations and I tried to make the via submission vs. API cases more similar. 

[creating entities, updating entities] x [via submission, via API]

Via API:
* parse data from request
* entity def source is basic, practically empty
* creator is inferred from request auth
* audit logging happens through `createNew.audit` and `createVersion.audit`, which checks that submission def was not provided

Via submission:
* data parsed from submission
* entity def source is complex, based on submission, def, event, etc.
* creator is the submission's creator
* audit logging happens in a submission event processing method because different audit logging kicks in in case of a processing error

### DIFF
--- a/test/integration/worker/entity.js
+++ b/test/integration/worker/entity.js
@@ -845,6 +845,14 @@ describe('worker: entity', () => {
           person.currentVersion.version.should.equal(2);
         });
 
+      await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc/audits')
+        .expect(200)
+        .then(({ body: logs }) => {
+          logs[0].should.be.an.Audit();
+          logs[0].action.should.be.eql('entity.update.version');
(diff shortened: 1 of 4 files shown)


test/integration/worker/entity.js
```diff
@@ -845,6 +845,14 @@ describe('worker: entity', () => {
           person.currentVersion.version.should.equal(2);
         });
 
+      await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc/audits')
```
ktuite, line 848: This test of audit log content deserves its own test/set of tests.

ktuite (commented):

matthew-white (approved):

PR #1014, opened 2023-10-02
oidc: use mustache for templating

alxndrsn:
Alternative approach to https://github.com/getodk/central-backend/pull/1008 & https://github.com/getodk/central-backend/pull/1011.

### DIFF
--- a/lib/resources/oidc.js
+++ b/lib/resources/oidc.js
@@ -15,8 +15,9 @@
 
 const { generators } = require('openid-client');
 const config = require('config');
+const { parse, render } = require('mustache');
 
-const { html, safeNextPathFrom } = require('../util/html');
+const { safeNextPathFrom } = require('../util/html');
 const { redirect } = require('../util/http');
 const { createUserSession } = require('../util/sessions');
 const { // eslint-disable-line object-curly-newline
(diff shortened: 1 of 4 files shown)


lib/resources/oidc.js
```diff
@@ -118,47 +162,7 @@ module.exports = (service, endpoint) => {
       // Instead, we need to render a page and then "browse" from that page to the normal frontend:
 
       // id=cl only set for playwright. Why can't it locate this anchor in any other way?
```
matthew-white, line 164: Maybe move this comment closer to `loaderTemplate` above?

lib/util/html.js
```diff
@@ -37,11 +34,14 @@ const frontendPage = ({ head='', body }) => html`
...
       <div id="content">
-        ${body}
+        {{{body}}}
       </div>
     </body>
   </html>
 `;
+parse(template); // caches template for future perf.
+
+const frontendPage = ({ head='', body }) => render(template, { head, body });
```
matthew-white, line 44: I [commented](https://github.com/getodk/central-backend/pull/981#discussion_r1335236657) in #981 that I think we should try to remove `frontendPage()` entirely. But that's probably a task for another PR.

matthew-white (approved):

PR #1016, opened 2023-10-02
added test-fast in Makefile

sadiqkhoja:
Added a rule in Makefile to skip @slow test, this is helpful for me during development because most of the time I'm not changing the code which is tested by slow tests like encryption

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,9 +61,14 @@ debug: base
 .PHONY: test
 test: lint
 	BCRYPT=no npx mocha --recursive --exit
+
 .PHONY: test-full
 test-full: lint
 	npx mocha --recursive --exit
+
+.PHONY: test-fast
+test-fast: node_version
+	BCRYPT=no npx mocha --recursive --exit --fgrep @slow --invert
 
 .PHONY: test-integration
 test-integration: node_version


Makefile
```diff
@@ -61,10 +61,15 @@ debug: base
...
 test: lint
 	BCRYPT=no npx mocha --recursive --exit
+
 .PHONY: test-full
 test-full: lint
 	npx mocha --recursive --exit
 
+.PHONY: test-fast
+test-fast:
+	BCRYPT=no npx mocha --recursive --exit --invert="--fgrep @slow"
```
matthew-white, line 71: The Mocha docs show:

```
-i, --invert  Inverts --grep and --fgrep matches                     [boolean]
```

Which makes me think it should be `--fgrep @slow --invert`. But if this is working, then maybe it doesn't matter.

matthew-white (approved):
I really like this idea!

Makefile
```diff
@@ -61,10 +61,15 @@ debug: base
 .PHONY: test
 test: lint
 	BCRYPT=no npx mocha --recursive --exit
+
 .PHONY: test-full
 test-full: lint
 	npx mocha --recursive --exit
 
+.PHONY: test-fast
+test-fast:
```
matthew-white, line 70: Maybe this should run `node_version` like `test-integration` and `test-unit`? Or it could run `lint` like `test` and `test-full`.

sadiqkhoja: I tend to fix lint at the very end that's why I didn't add `lint` or `node_version`.

matthew-white: I also like linting and testing separately. :+1: Note that `node_version` doesn't run `lint`, it just runs `npm install` and checks the Node version.

PR #1017, opened 2023-10-02
Add columns to track entity conflict

sadiqkhoja:
Adds new columns to record conflict information on Entity

### DIFF
--- a/lib/model/migrations/20231002-01-add-conflict-details.js
+++ b/lib/model/migrations/20231002-01-add-conflict-details.js
@@ -1,0 +1,37 @@
+// Copyright 2023 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw(`CREATE TYPE "conflictType" AS ENUM ('soft', 'hard')`); // eslint-disable-line quotes
+
+  await db.raw('ALTER TABLE entities ADD COLUMN conflict "conflictType" NULL');
+
+  await db.raw(`ALTER TABLE entity_defs 
+    ADD COLUMN "dataReceived" JSONB NOT NULL,
+
+    -- null means, it's a first version
+    ADD COLUMN "baseVersion" INT4,  
+
+    -- array of conflicting properties
+    ADD COLUMN "conflictingProperties" JSONB NULL 
+
+    -- Not adding explicit 'conflict' column: version created conflict if "baseVersion" < "version" - 1
+  `);
+
+  // Sets the value for "dataReceived" and "baseVersion" for existing row
+  await db.raw('UPDATE entity_defs SET "dataReceived" = data, "baseVersion" = CASE WHEN version > 1 THEN version - 1 ELSE NULL END');
+
+};
+
+const down = async (db) => {
+  await db.raw('ALTER TABLE entities DROP COLUMN conflict');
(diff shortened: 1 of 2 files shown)


lib/model/migrations/20231002-01-add-conflict-details.js
```diff
@@ -0,0 +1,30 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw('ALTER TABLE entities ADD COLUMN conflict VARCHAR(10) NULL');
```
sadiqkhoja, line 11: `varchar` because we should store type of conflict like:

- `NULL`: no conflict
- `soft`: version discrepancy
- `hard`: two properties modified simultaneously 

maybe in future we need to show type of conflict on home/entity-list page

matthew-white: What do you think about using an `ENUM`?

lib/model/migrations/20231002-01-add-conflict-details.js
```diff
@@ -0,0 +1,30 @@
...
+
+  await db.raw(`ALTER TABLE entity_defs 
+    ADD COLUMN "dataReceived" JSONB NOT NULL,
+    ADD COLUMN "baseVersion" INT4 NOT NULL,
+    ADD COLUMN "conflictingProperties" JSONB NULL -- array of conflicting properties
+    -- if version created conflict if "baseVersion" < "version" - 1
+  `);
+
+  // Sets the value for "dataReceived" for existing row
+  await db.raw('UPDATE entity_defs SET "dataReceived" = data, "baseVersion" = version - 1');
```
sadiqkhoja, line 21: set Null for the first version

lib/model/migrations/20231002-01-add-conflict-details.js
```diff
@@ -0,0 +1,37 @@
...
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw(`CREATE TYPE "conflictType" AS ENUM ('soft', 'hard')`); // eslint-disable-line quotes
+
+  await db.raw('ALTER TABLE entities ADD COLUMN conflict "conflictType" NULL');
+
+  await db.raw(`ALTER TABLE entity_defs 
+    ADD COLUMN "dataReceived" JSONB NOT NULL,
```
ktuite, line 16: When I tried to run the migration, I got this error
```
column "dataReceived" of relation "entity_defs" contains null values
```

I think this column should have a default like ` DEFAULT '{}'::jsonb`

lib/model/query/entities.js
```diff
@@ -23,8 +23,8 @@ const { runSequentially } = require('../../util/promise');
 // ENTITY CREATE
 
 const _defInsert = (id, root, creatorId, userAgent, label, json, version, sourceId = null) => sql`
-  insert into entity_defs ("entityId", "root", "sourceId", "creatorId", "userAgent", "label", "data", "current", "createdAt", "version")
-  values (${id}, ${root}, ${sourceId}, ${creatorId}, ${userAgent}, ${label}, ${json}, true, clock_timestamp(), ${version})
+  insert into entity_defs ("entityId", "root", "sourceId", "creatorId", "userAgent", "label", "data", "current", "createdAt", "version", "baseVersion", "dataReceived")
+  values (${id}, ${root}, ${sourceId}, ${creatorId}, ${userAgent}, ${label}, ${json}, true, clock_timestamp(), ${version}, ${version - 1}, ${json})
```
ktuite, line 27: will this insert baseVersion=0 for new entities? it could use root (or version, but root seems more explicit) to conditionally set baseVersion to null for new entities.

ktuite (approved):
See comments about default column value and consistently setting baseVersion to null for new entities. Otherwise looking good and start building on top of!

PR #1019, opened 2023-10-03
Clarify that OData operator for "not equals" is ne, not neq

matthew-white:
For getodk/central#503, I'm currently planning to use the OData "not equals" operator in the `$filter` query parameter, something like `__system/conflict ne null`. While looking into that, I noticed that the API docs say that `neq` is the OData operator. However, the operator is actually `ne`: https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_NotEquals. I updated a test to confirm that.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -7745,7 +7745,7 @@ paths:
         in: query
         description: If provided, will filter responses to those matching the given
           OData query. Only [certain fields](/central-api-odata-endpoints/#data-document)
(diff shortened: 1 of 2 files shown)


sadiqkhoja:
ha! I have used not equal like this: `$filter=not __system/reviewState eq \'rejected\'` in a test :)

sadiqkhoja (approved):

PR #1020, opened 2023-10-06
Feature/entities/return conflicts in versions api

sadiqkhoja:
Part of getodk/central#501

Fills in conflict information in the database.
Returns conflict information in the /versions API

#### Before submitting this PR, please make sure you have:

- [x] run `make test-full` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code from external sources are properly credited in comments or that everything is internally sourced

Issue #1009, opened 2023-09-27
Refactor worker/entity tests


ktuite:
Addressed in PR https://github.com/getodk/central-backend/pull/1021

There are a number of tests in `tests/integration/worker/entity.js` that were made before we had more of the API in place that we might be able to remove. This file could be made to focus just on the entity worker? 

I find myself either forgetting about this file, or over-relying on this file as a kind of catch-all of entity workflow test scenarios. For example, I'm writing entity update tests here because I'm not sure where else they should go... in the GET entities/uuid endpoint? Sometimes I want to test one scenario, like entity updates, and maybe check that multiple endpoints look as expected instead of duplicating the scenario across the tests for different endpoints.

Do other people have issues with not knowing where to put tests? Suggestions?

matthew-white:
For getodk/central#385, existing tests of the form worker were in test/integration/api/forms/forms.js and test/integration/api/forms/draft.js. I think that's because the worker is triggered by the endpoints to create a form or draft or to publish a draft. By that logic though, we should test the entity worker in the tests of the submission creation endpoints, and that doesn't feel right. 🤔 I feel like there are multiple reasonable options here. If this feels like a special case since it touches so many endpoints and is about a larger workflow, one option might be that we move these tests to their own file or set of files.

PR #1021, opened 2023-10-09
Refactoring entity tests and adding new entity update tests

ktuite:
Closes the following issues:
* https://github.com/getodk/central-backend/issues/1009
* https://github.com/getodk/central/issues/510 
    * except for offshoot issue: https://github.com/getodk/central/issues/520)

In particular:

**commit 1**
- updated entity-parsing unit tests 

**commit 2**
- moved extra tests of streaming CSVs of entities from `worker/entity` test to `api/datasets` test (multi form and updated entities in particular)
- removed redundant checks and tests from `worker/entity` test file
- added entity update tests to relevant sections of `api/entities` API test file
	* returning current version and data of updated entity
	* audit log events for an entity

**commit 3**
- added tests of entity processing errors flagged by worker

**commit 4**
- added tests about when submissions about entity updates do or do not get processed including tests for comments:
    - submisssion.update/approval should never update entity https://github.com/getodk/central-backend/pull/1005#discussion_r1341794108
    - update should be processed even if approval required for creation https://github.com/getodk/central-backend/pull/1005#discussion_r1341833278

**commit 5**
- tests of basic update logic (can update label and data) 
    * does not test or even consider conflicts
    * has some issues (skipped tests) where setting a field to blank doesn't work. waiting for https://github.com/getodk/central-backend/pull/1020 first to not get in the way (or maybe that PR can directly address this issue)

it's mostly all tests!

Some of the tests got refactored. I moved tests that started in worker/entity but weren't really about the worker into other files, e.g. csv streaming of entities, which was one of the first tests about entities to exist. 

Many tests about entity updates have an API endpoint that is most relevant, and it seems reasonable to move tests/parts of tests to those places.

I added one special section to the end of `integration/api/entities` for capturing update logic. There are some special sections at the bottom of `integration/api/datasets` specifically for dataset scenarios.

More reliable entity update behavior 

No

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -157,6 +157,9 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
 };
 
 const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event) => async ({ Audits, Entities }) => {
+  if (!(event.action === 'submission.create')) // only update on submission.create
+    return null;
+
   // Get client version of entity
   const clientEntity = await Entity.fromParseEntityData(entityData); // validation happens here
 
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -13,6 +13,7 @@ const { equals, extender, unjoiner, page, markDeleted } = require('../../util/db
 const { map, mergeRight } = require('ramda');
 const { blankStringToNull, construct } = require('../../util/util');
 const { QueryOptions } = require('../../util/db');
+const { getOrNotFound } = require('../../util/promise');
 const { odataFilter } = require('../../data/odata-filter');
 const { odataToColumnMap, parseSubmissionXml } = require('../../data/entity');
 const { isTrue } = require('../../util/http');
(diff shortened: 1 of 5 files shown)


lib/model/query/entities.js
```diff
@@ -156,11 +157,14 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
 };
 
 const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event) => async ({ Audits, Entities }) => {
+  if (!(event.action === 'submission.create')) // only update on submission.create
```
sadiqkhoja, line 160: should this condition be in `_processSubmissionEvent` function?

sadiqkhoja (approved):
Liked this reorganization of tests

lib/model/query/entities.js
ktuite: Because I moved these lines into `_createEntity`

```
if ((dataset.approvalRequired && event.details.reviewState !== 'approved') ||
      (!dataset.approvalRequired && event.action === 'submission.update'))
```

it felt reasonable to move this event action check into `_updateEntity`

PR #1023, opened 2023-10-11
Feature/entities/resolve conflict

sadiqkhoja:
Closes #


test/integration/api/entities.js
```diff
@@ -1340,6 +1341,107 @@ describe('Entities API', () => {
...
+        await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?resolve=true')
+          .expect(200);
+
+        await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc')
+          .set('X-Extended-Metadata', true)
+          .expect(200)
+          .then(({ body: person }) => {
+            should(person.conflict).be.null();
+          });
+      }));
```
ktuite, line 1394: Can you add a test checking that the right person/actor is associated with the resolve conflict audit event?

docs/api.yaml
```diff
@@ -10950,6 +10950,10 @@ paths:
         Use this API to update one or all properties of an Entity. It will throw `400 - Bad Request` if any of the updating properties doesn't exist in the dataset.
 
         To unset value of any property, you can set it to empty string (""). Setting it to `null` will throw an error.
+
```
ktuite, line 10953: Since you're editing the API docs here, would it make sense to also change the docs to show the new `conflict` and `conflictingProperties` flags on entities and entity versions? 

It's more to ask, but you could even start the changelog part of the docs?

Changed
* Entities accessed via API now have `conflict` and version-specific `conflictingProperties` fields
    * Conflicts are either hard (baseVersion conflicts and multiple versions update the same property), soft (baseVersion conflicts but data updates are independent), or null

Added
* Conflicts can be resolved using the entity PATCH endpoint with or without new data

lib/model/query/entities.js
```diff
@@ -132,7 +132,16 @@ createVersion.audit = (updatedEntity, dataset, partial, subDef) => (log) => {
 };
 createVersion.audit.withResult = true;
 
+////////////////////////////////////////////////////////////////////////////////
+// RESOLVE CONFLICT
 
+const resolveConflict = (entity, dataset) => ({ run }) => // eslint-disable-line no-unused-vars
+  run(sql`update entities set conflict=null where "id"=${entity.id}`);
+resolveConflict.audit = (entity, dataset) => (log) => log('entity.update.resolve', dataset, {
```
ktuite, line 140: If this new action is getting added, there are a couple other things to do
1. add it to the nonverbose and entity action catetory in model/query/audits
2. test that it comes through in the entity category filtering in the audit tests

docs/api.yaml
```diff
@@ -36,7 +36,19 @@ info:
 
     Here major and breaking changes to the API are listed by version.
 
-    ## ODK Central v2023.4
+    ### ODK Central v2023.5
```
ktuite, line 39: Should this just be `##`? Can you put backticks around the appropriate words below?

test/integration/api/audits.js
```diff
@@ -382,6 +382,23 @@ describe('/audits', () => {
...
+
+      // all properties changed
+      await asAlice.post('/v1/projects/1/forms/updateEntity/submissions')
+        .send(testData.instances.updateEntity.one)
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+
+      await exhaust(container);
+
+      await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?resolve=true')
```
ktuite, line 399: Is this suggesting that you can resolve a conflict even if there is no conflict? It seems like it shouldn't log anything if you try to resolve a conflict that doesn't exist.

ktuite (approved):
Thanks for adding the doc changes and the audit action tests.

One of the additions to a test made me curious about calling 'resolve' on an entity without any conflict. It seems like it shouldn't resolve a conflict that doesn't exist. Mainly it shouldn't log it in the audit log, because that might be confusing later.

PR #1026, opened 2023-10-12
Return conflicts count in Project and Dataset GET APIs

sadiqkhoja:
Part of getodk/central#500

Test

Yes, updated

### DIFF
--- a/test/integration/setup.js
+++ b/test/integration/setup.js
@@ -9,6 +9,7 @@ const { noop } = require(appRoot + '/lib/util/util');
 const { task } = require(appRoot + '/lib/task/task');
 const authenticateUser = require('../util/authenticate-user');
 const testData = require('../data/xml');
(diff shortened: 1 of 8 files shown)


test/integration/setup.js
```diff
@@ -219,5 +220,36 @@ const withClosedForm = (f) => async (service) => {
   return f(service);
 };
 
-module.exports = { testService, testServiceFullTrx, testContainer, testContainerFullTrx, testTask, withClosedForm };
+const createConflict = async (user, container) => {
```
ktuite, line 223: It feels weird to me to put this here in setup.js, which I think of as super generic test infrastructure of how tests are run. (I suppose `withClosedForm` is in there....) 

Is there some other place we can put complex scenarios like this for use in multiple tests across multiple files? Could that be a separate issue or PR where we factor out the other places we have duplicate complicated entity logic?

ktuite (approved):
Looks great!

PR #1027, opened 2023-10-13
Migrate entity.create.error audit action to entity.error

ktuite:
Closes https://github.com/getodk/central/issues/520

Migrate `entity.create.error` to `entity.error` and add new `404.8` Problem with descriptive error about entity UUID not found. Example:

`The entity with UUID (12345678-1234-4123-8234-123456789abc) specified in the submission does not exist in the dataset (people).`

Tests, trying it in frontend.

It makes the most sense to have a generic entity processing error, so all previous `entity.create.error` error events, which were the only possible processing errors before updates, have been migrated to `entity.error`. 

Makes clearer errors. I don't think there are regression risks.

Did make a small change the one place this error was mentioned in the documentation.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -814,7 +814,7 @@ tags:
     * `dataset.update` when a Dataset is updated.
     * `dataset.update.publish` when a Dataset is published.
     * `entity.create` when an Entity is created.
-    * `entity.create.error` when there is an error during entity creation process.
+    * `entity.error` when there is an error during entity creation process.
     * `config.set` when a system configuration is set.
     * `analytics` when a Usage Report is attempted.
     * Deprecated: `backup` when a backup operation is attempted for Google Drive backups.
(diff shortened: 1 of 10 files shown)


docs/api.yaml
```diff
@@ -814,7 +814,7 @@ tags:
     * `dataset.update` when a Dataset is updated.
     * `dataset.update.publish` when a Dataset is published.
     * `entity.create` when an Entity is created.
-    * `entity.create.error` when there is an error during entity creation process.
+    * `entity.error` when there is an error during entity creation process.
```
ktuite, line 817: list other entity actions in API (`entity.update.version`, `entity.update.resolve`, `entity.delete`) not super relevant to this PR

test/integration/worker/entity.js
```diff
@@ -575,11 +575,11 @@ describe('worker: entity', () => {
         subEvent.failures.should.equal(0);
 
         // the entity creation error should be logged
-        const event = await container.Audits.getLatestByAction('entity.create.error').then((o) => o.get());
```
ktuite, line 578: change comment above - it's not a creation error

ktuite (commented):
interactive review

matthew-white (approved):

PR #1028, opened 2023-10-16
Features/entities/relevant to conflict flag

sadiqkhoja:
Part of getodk/central#504

Added tests

The name of querystring parameter could be shorter but we came to consensus on `relevantToConflict` in Slack

None

Yes

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -355,8 +355,6 @@ const ConflictType = {
   HARD: 'hard' // discrepancy in version number and one/more properties updated simultaneously
 };
 const getWithConflictDetails = (defs, audits, relevantToConflict) => {
-  const defMap = new Map(defs.map(d => [d.version, d]));
-
   const result = [];
 
   const lastResolveEvent = audits.findLast(a => a.action === 'entity.update.resolve');
@@ -364,8 +362,6 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
   const auditMap = new Map(audits
     .filter(a => a.action === 'entity.create' || a.action === 'entity.update.version')
     .map(a => [a.details.entityDefId, a.details]));
-
-  const versionMap = new Map();
 
   let lastGoodVersion = 0; // Entity Version is never 0
 
@@ -387,23 +383,20 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
     if (v.version > 1) { // v.root is false here - can use either
       const conflict = v.version !== (v.baseVersion + 1);
 
-      v.baseDiff = getDiffProp(v.dataReceived, defMap.get(v.baseVersion).data);
-      if ('label' in v.dataReceived && v.dataReceived.label !== defMap.get(v.baseVersion).label) v.baseDiff.push('label');
+      v.baseDiff = getDiffProp(v.dataReceived, defs[v.baseVersion - 1].data);
+      if ('label' in v.dataReceived && v.dataReceived.label !== defs[v.baseVersion - 1].label) v.baseDiff.push('label');
 
-      v.serverDiff = getDiffProp(v.dataReceived, defMap.get(v.version - 1).data);
-      if ('label' in v.dataReceived && v.dataReceived.label !== defMap.get(v.version - 1).label) v.serverDiff.push('label');
+      v.serverDiff = getDiffProp(v.dataReceived, defs[v.version - 2].data);
+      if ('label' in v.dataReceived && v.dataReceived.label !== defs[v.version - 2].label) v.serverDiff.push('label');
 
       if (conflict) {
         v.conflict = v.conflictingProperties && v.conflictingProperties.length > 0 ? ConflictType.HARD : ConflictType.SOFT;
 
         v.resolved = lastResolveEvent && lastResolveEvent.details.entityDefId >= def.id;
 
-        // lastGoodVersion is the baseVersion of first unresolved conflict
         if (!v.resolved && lastGoodVersion === 0) {
-          lastGoodVersion = v.baseVersion;
-
-          // baseVersion will be always in versionMap; if not then something terrible has happened - 500 error is fine
-          versionMap.get(v.baseVersion).lastGoodVersion = true;
+          lastGoodVersion = v.version - 1;
+          result[v.version - 2].lastGoodVersion = true;
         }
       }
 
@@ -412,8 +405,6 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
         relevantBaseVersions.add(v.baseVersion);
       }
     }
-
-    versionMap.set(v.version, v);
 
     result.push(v);
   }
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -361,7 +361,9 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
 
   const lastResolveEvent = audits.findLast(a => a.action === 'entity.update.resolve');
 
-  const auditMap = new Map(audits.map(a => [a.details.entityDefId, a.details]));
+  const auditMap = new Map(audits
+    .filter(a => a.action === 'entity.create' || a.action === 'entity.update.version')
+    .map(a => [a.details.entityDefId, a.details]));
 
   const versionMap = new Map();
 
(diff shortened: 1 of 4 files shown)


lib/data/entity.js
```diff
@@ -354,20 +354,32 @@ const ConflictType = {
...
+const getWithConflictDetails = (defs, audits, relevantToConflict) => {
   const defMap = new Map(defs.map(d => [d.version, d]));
 
   const result = [];
 
+  const lastResolveEvent = audits.findLast(a => a.action === 'entity.update.resolve');
+
+  const auditMap = new Map(audits.map(a => [a.details.entityDefId, a.details]));
+
+  const versionMap = new Map();
```
matthew-white, line 366: Is `defs` guaranteed to be ordered by version number? If so, instead of creating a separate map, could we look up each version in the array by index using the version's version number? Or are there cases where that might not work?

lib/data/entity.js
```diff
@@ -380,13 +392,35 @@ const getWithConflictDetails = (defs) => {
...
 
       if (conflict) {
-        v.conflict = v.conflictingProp && v.conflictingProp.length > 0 ? ConflictType.HARD : ConflictType.SOFT;
+        v.conflict = v.conflictingProperties && v.conflictingProperties.length > 0 ? ConflictType.HARD : ConflictType.SOFT;
+
+        v.resolved = lastResolveEvent && lastResolveEvent.details.entityDefId >= def.id;
+
+        // lastGoodVersion is the baseVersion of first unresolved conflict
+        if (!v.resolved && lastGoodVersion === 0) {
+          lastGoodVersion = v.baseVersion;
```
matthew-white, line 401: The release criteria define it as:

> The update immediately prior to the earliest parallel update (hereafter the “known good” record)

So I think:

```diff
-lastGoodVersion = v.baseVersion;
+lastGoodVersion = v.version - 1;
```

lib/data/entity.js
```diff
@@ -380,13 +392,35 @@ const getWithConflictDetails = (defs) => {
...
     }
+
+    versionMap.set(v.version, v);
+
     result.push(v);
   }
+
+  // We return all the versions after last good version and the base versions of those.
+  if (relevantToConflict) {
+    return result.filter(v => lastGoodVersion > 0 && (v.version >= lastGoodVersion || relevantBaseVersions.has(v.version)));
```
matthew-white, line 421: Could we move the `lastGoodVersion > 0` condition out of the loop?

```suggestion
  if (relevantToConflict && lastGoodVersion > 0) {
    return result.filter(v => v.version >= lastGoodVersion || relevantBaseVersions.has(v.version));
```

lib/data/entity.js
```diff
@@ -354,20 +354,32 @@ const ConflictType = {
...
 };
-const getWithConflictDetails = (defs) => {
+const getWithConflictDetails = (defs, audits, relevantToConflict) => {
   const defMap = new Map(defs.map(d => [d.version, d]));
 
   const result = [];
 
+  const lastResolveEvent = audits.findLast(a => a.action === 'entity.update.resolve');
+
+  const auditMap = new Map(audits.map(a => [a.details.entityDefId, a.details]));
```
matthew-white, line 364: There can be multiple audit log events with the same `entityDefId`, right? For example, if you cause a conflict, then you resolve the conflict. But I guess all the relevant details would be the same among those multiple events, right? Just making sure I understand how this works.

lib/data/entity.js
matthew-white: Sort of a different question, but relevant if we decide we need to keep `versionMap`: what's the difference between `versionMap` and `defMap`? Could we just use one or the other?

matthew-white (commented):
I haven't looked at tests yet, but I thought I'd start with these comments. It's looking good so far!

lib/data/entity.js
matthew-white: Oh wait, I guess that doesn't really work, because we want to return `[]` if `lastGoodVersion === 0`. I think it'd be nice to move the static `lastGoodVersion > 0` condition out of the loop, but maybe it's not worth an extra if/ternary. 🤔

lib/data/entity.js
sadiqkhoja: `defMap` values are incoming Entity.Def frames and `versionMap` values are results with all additional fields. 

Regarding `versionMap`, I created it in a fear of missing versions. Now after thinking a bit, I feel it's an unfounded fear and I should use array indexing in both places instead. Can you think of any situation where we would have missing version number?

lib/data/entity.js
sadiqkhoja: `v.version - 1` is parallel to `v` on which we have marked the conflict. Although, `v.version - 1` has no conflict in the database because it arrived before `v`, it may still not be the truth. Both versions are in contention.

Consider following scenario, I think v30 is the last known good version, because (31, 32) and (33) are two parallel update branches

```

                            ┌───────────┐            ┌──────────┐
           ┌───────────────►│           │            │          │
           │                │    v31    ├───────────►│   v32    │
           │                │           │            │          │
           │                └───────────┘            └──────────┘
           │
   ┌───────┴──────┐
   │              │
   │  v30         │
   │              │
   └────────┬─────┘
            │
            │
            │
            │
            │                                                     ┌───────────┐
            │                                                     │           │
            └─────────────────────────────────────────────────────►   v33     │
                                                                  │           │
                                                                  └───────────┘

```

lib/data/entity.js
sadiqkhoja: For the safe side, I have filtered the audits for only `create` and `update` events

lib/data/entity.js
sadiqkhoja: Added inline ternary.

lib/data/entity.js
matthew-white: I can't think of a case where a version number would be outright missing. The version number is incremented one by one, and versions can't be deleted, so I don't see how a number could get skipped.

Another issue would be if versions are out-of-order (e.g., somehow 1, 3, 2, not 1, 2, 3). But I again don't see that happening, since I think we sort on `entity_defs."createdAt", entity_defs.id`, and version numbers are incremented one by one. In order to increment the version number of version 2, version 2 has to have been created previously.

In conditions of high concurrency, is there any chance that there could be duplicate version numbers? That would definitely be a problem. What do you think about adding a uniqueness constraint on `entityId` + `version` to the `entity_defs` table? (Could be a separate PR.) Or is it already guaranteed that duplicate version numbers are not possible?

lib/data/entity.js
matthew-white: If v32 were considered the last good version, then we would show v32, v33 (the conflict), and v30 (the base version of the conflict) in the summary table. We'd be skipping v31, but I'm not sure that the user needs to compare v31 and v33, as v32 includes the changes from v31. Maybe in some cases the user would want to thoroughly review the history, but I'm not sure that we need to include v31 in the summary table.

I do see what you're saying about the two branches. I think taking that approach would have design implications though. For example, what status icon would be shown in the summary table for v31? I don't think the current icons and text would work. What do you think about starting with `v.version - 1`, then we can ask issa what she thinks? If you don't like the language of "last good", we could use a more specific term ("last resolved"? "last before conflict"?).

By the way, if we were to account for branching, I think we'd need to adjust the current approach slightly. In that case, I think the last good version wouldn't be the base version of the first unresolved conflict; it would be the minimum base version among unresolved conflicts. The second conflict might have an earlier base version than the first conflict.

lib/data/entity.js
sadiqkhoja: Agree, we need to ensure duplicate that there's no duplicate version. I'll create separate PR for that.

matthew-white (approved):
I think there are a few places where we need more test coverage, but besides that, I only have small suggestions to make. ✅ Entity updates via submission are a complex feature, so I really think we'll benefit from having excellent test coverage.

lib/data/entity.js
```diff
@@ -354,39 +354,66 @@ const ConflictType = {
...
+
+    const v = mergeLeft(def.forApi(),
       {
         conflict: null,
         resolved: false,
         baseDiff: [],
         serverDiff: [],
-        conflictingProp: [],
-        source: {} // TODO
+        source: { sourceEvent, submissionCreate, submission },
```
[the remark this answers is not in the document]
sadiqkhoja: created an issue for this https://github.com/getodk/central-backend/issues/1036

PR #1030, opened 2023-10-24
Fixed: missing label in dataReceived

sadiqkhoja:

### DIFF
--- a/lib/model/frames/entity.js
+++ b/lib/model/frames/entity.js
@@ -49,8 +49,10 @@ class Entity extends Frame.define(
       });
     const { uuid, label } = entityData.system;
     const { data } = entityData;
+    const dataReceived = { ...body.data };
+    if (body.label != null) dataReceived.label = body.label;
     return new Entity.Partial({ uuid, datasetId: dataset.id, id: oldEntity?.id }, {
-      def: new Entity.Def({ data, label, dataReceived: body.data || {} }),
+      def: new Entity.Def({ data, label, dataReceived }),
       dataset: dataset.name
     });
   }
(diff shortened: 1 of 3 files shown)


matthew-white (commented):
The change here looks good to me. Should we also change `Entity.fromJson()`? I think right now, it doesn't include the label in `dataReceived`:

https://github.com/getodk/central-backend/blob/f682891eb985b69116c49d0fdc1c988a376fecf3/lib/model/frames/entity.js#L39-L56

So far, this PR tests that an entity update via submission includes the label in `dataReceived`. I'm thinking that we should test that for an API update as well.

lib/model/frames/entity.js
```diff
@@ -50,7 +50,7 @@ class Entity extends Frame.define(
     const { uuid, label } = entityData.system;
     const { data } = entityData;
     return new Entity.Partial({ uuid, datasetId: dataset.id, id: oldEntity?.id }, {
-      def: new Entity.Def({ data, label, dataReceived: body.data || {} }),
+      def: new Entity.Def({ data, label, dataReceived: { ...body.data, label } || {} }),
```
matthew-white, line 53: In the case of an update, `oldEntity` is specified, which means that even if no label was specified in the request body, `label` will exist (it will be taken from `oldEntity`). But in that case, I don't think we should include label in `dataReceived` (I don't think we should add a `label` property at all). How about something like:

```js
const dataReceived = { ...body.data };
if (body.label != null) dataReceived.label = body.label;
```

matthew-white (approved):
As I mentioned on Slack, I think there's still an issue with `Entity.fromParseEntityData()`:

https://github.com/getodk/central-backend/blob/643fd63a92fe056a0d3e9889e58b45d94ac2c3c1/lib/model/frames/entity.js#L28-L37

Right now, `validateEntity()` requires `entityData` to include a label. But I think we should allow `entityData` not to include a label. In that case, we shouldn't add a `label` property to `dataReceived`. However, I think that change could come in a follow-up PR.

PR #1033, opened 2023-10-26
Lock the Entity while it's being updated

sadiqkhoja:
Closes getodk/central#541.

Added a comprehensive test

With this approach Postgresql is responsible for handling concurrency and locking mechanism, and application code remains clean.

Other option was to create an unique constraint on uuid+version and reprocess the submission when there is a violation. I was worried about reprocessing because in that case we would need to refetch the latest Entity version and merge it again that can cause unique constraint violation at the time of insertion.

None

NA

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -176,7 +176,7 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
   const clientEntity = await Entity.fromParseEntityData(entityData); // validation happens here
 
   // Get version of entity on the server
-  const serverEntity = (await Entities.getById(dataset.id, clientEntity.uuid))
+  const serverEntity = (await Entities.getById(dataset.id, clientEntity.uuid, QueryOptions.forUpdate))
     .orThrow(Problem.user.entityNotFound({ entityUuid: clientEntity.uuid, datasetName: dataset.name }));
 
   let { conflict } = serverEntity;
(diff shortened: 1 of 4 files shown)


matthew-white (commented):
This approach makes a lot of sense to me.

Advantages of this approach:

- Simplicity. The source didn't have to change much.
- We don't need to implement retry logic for entity processing. That would almost certainly be needed with a uniqueness constraint.

Potential risks:

- Is it not possible to read an entity while it's locked? Does that mean that it will be hard to download the entities CSV file (including to OpenRosa clients) while entities are being updated en masse?
- Every path that updates entities needs to lock the entity. It's not automatic like a uniqueness constraint. When we add another path to update entities (bulk upload), we'll need to remember to lock the entity.
- The test needs to use query module functions directly rather than via the API. I think we'll need to maintain the test as query module functions change.
- Locking the entity in one update could lead another entity update to time out. If that happened, we'd need to implement retry logic anyway, removing one of the advantages. But hopefully entity updates will be fast enough that there's no timeout. We could see how it does in practice.

Altogether, I lean toward merging the PR and seeing how it works for us. We can always change it later. However, it'd be good to get more information about the first risk above.

sadiqkhoja:
> Is it not possible to read an entity while it's locked? Does that mean that it will be hard to download the entities CSV file (including to OpenRosa clients) while entities are being updated en masse?

This is not an issue at all. `pg_advisory_xact_lock` doesn't lock any row, it locks an arbitrary integer and because we don't use it during CSV download or OData feed there won't be any issue.

> Locking the entity in one update could lead another entity update to time out. If that happened, we'd need to implement retry logic anyway, removing one of the advantages. But hopefully entity updates will be fast enough that there's no timeout. We could see how it does in practice.

This definitely needs to be tested with some load. I feel that the retry logic in the Worker process should be enough to handle those cases.

matthew-white:
> `pg_advisory_xact_lock` doesn't lock any row, it locks an arbitrary integer and because we don't use it during CSV download or OData feed there won't be any issue.

Ooh interesting, I get it now. Sounds great.

> > Locking the entity in one update could lead another entity update to time out.
> 
> I feel that the retry logic in the Worker process should be enough to handle those cases.

I don't think entity processing uses that standard retry logic: entity processing does its own error handling, logging `entity.error` events. However, I think it'd be reasonable for us to merge this and see whether we run into any issues.

matthew-white (approved):

PR #1034, opened 2023-10-27
Tests for entity-updating forms modifying dataset properties

ktuite:
Closes getodk/central#516

Forms that update entities vs. creating entities don't get parsed any differently at upload time. The effects these forms have on dataset properties are the same as entity-creation forms. This PR adds a few tests of this behavior. 

I think these tests will be more important to have in place when we get back to this issue: https://github.com/getodk/central/issues/518 (Restrict submissions to the entity actions listed in the form) because that will involve code changes around dataset parsing and form uploading.

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -2609,6 +2609,119 @@ describe('datasets and entities', () => {
 
       }));
 
+      describe('uploading forms that UPDATE but do not create entities', () => {
+        it('should be able to add properties to datasets', testService(async (service) => {
+          const asAlice = await service.login('alice');
+          await asAlice.post('/v1/projects/1/forms?publish=true')
+            .send(testData.forms.simpleEntity)
+            .set('Content-Type', 'text/xml')
+            .expect(200);
+
+          await asAlice.post('/v1/projects/1/forms')
+            .send(testData.forms.updateEntity
+              .replace('entities:saveto="first_name"', 'entities:saveto="nickname"'))
+            .set('Content-Type', 'text/xml')
+            .expect(200);
+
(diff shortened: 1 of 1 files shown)


test/integration/api/datasets.js
```diff
@@ -2609,6 +2609,119 @@ describe('datasets and entities', () => {
...
+            .expect(200);
+
+          await asAlice.get('/v1/projects/1/datasets/people')
+            .expect(200)
+            .then(({ body }) => {
+              body.name.should.be.eql('people');
+              body.properties.map(p => p.name).should.eql([ 'first_name', 'age' ]);
+            });
+
+          await Audits.getLatestByAction('dataset.create')
```
matthew-white, line 2720: Instead of querying the database, how about using the API to request /v1/audits?action=dataset.create ?

test/integration/api/datasets.js
```diff
@@ -2609,6 +2609,119 @@ describe('datasets and entities', () => {
...
+              ]);
+            });
+        }));
+
+        it('should create dataset if it does not exist and and log it', testService(async (service, { Audits }) => {
+          const asAlice = await service.login('alice');
+          await asAlice.post('/v1/projects/1/forms?publish=true')
+            .send(testData.forms.updateEntity)
+            .set('Content-Type', 'text/xml')
+            .expect(200);
```
matthew-white, line 2711: Just a thought, before publishing the form, would it be worth it to test that &hellip;/draft/dataset-diff returns the dataset with `isNew: true`? No worries if that doesn't feel useful.

matthew-white (approved):

test/integration/api/datasets.js
ktuite: I can do that!

PR #1037, opened 2023-10-31
Minor fixes for Resolve Conflict

sadiqkhoja:
- Fixes: it should not resolve if resolve flag is not true
- Fixes: set updatedAt on resolve conflict
- Check for If-Match before resolve conflict

### DIFF
--- a/test/integration/api/entities.js
+++ b/test/integration/api/entities.js
@@ -614,6 +614,7 @@ describe('Entities API', () => {
         await createConflictOnV2(asAlice, container);
 
         await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?resolve=true')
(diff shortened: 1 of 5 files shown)


test/integration/api/entities.js
```diff
@@ -1576,6 +1611,31 @@ describe('Entities API', () => {
...
+
+        await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?resolve=true')
+          .set('If-Match', '"0"')
+          .expect(409)
+          .then(({ body }) => {
+            body.code.should.equal(409.15);
+          });
+      }));
+
+      it('should reject if version does not match', testService(async (service, container) => {
```
matthew-white, line 1627: I think this test needs a different title.

matthew-white (approved):

PR #1038, opened 2023-11-02
Add relevantToConflict field in the GET /versions response

sadiqkhoja:
### Changes:

- Added `relevantToConflict` field in GET /versions response; this should help in creation of conflict summary table on Entity feed page
- Return the fields that are stored in `entity_defs` table for all Entity related endpoints, except `data` and `dataReceived`
- Updated assertions and API docs to reflect the changes

### DIFF
--- a/lib/model/frames/entity.js
+++ b/lib/model/frames/entity.js
@@ -81,7 +81,7 @@ Entity.Def.Metadata = class extends Entity.Def {
 
   // we don't want `data` to be selected from database, hence return in forApi()
   static get fields() {
-    return super.fields.filter(f => f !== 'data');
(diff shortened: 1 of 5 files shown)


lib/model/frames/entity.js
```diff
@@ -81,7 +81,7 @@ Entity.Def.Metadata = class extends Entity.Def {
 
   // we don't want `data` to be selected from database, hence return in forApi()
   static get fields() {
-    return super.fields.filter(f => f !== 'data');
+    return super.fields.filter(f => f !== 'data' && f !== 'dataReceived');
```
matthew-white, line 84: Would you mind adding a test for this change?

matthew-white (approved):

PR #1039, opened 2023-11-02
API-docs container: use env variable for spec path

sadiqkhoja:
API-docs container: use env variable for spec path, this fixes issue where git is unable to operate on mounted file.


lognaturel (approved):
I'm not able to reproduce the problem that this addresses. But it works and seems fine so I'd rather merge it now and make further changes if it causes other problems.

PR #1043, opened 2023-11-10
Support updating AND creating entities from a single submission/form

ktuite:
Closes https://github.com/getodk/central/issues/514

A submission describing an entity can now set both `create` and `update` flags to true. The submission processing will try to update first, and if it can't find the entity with the specified uuid, it will try to create the entity.

If there is any error besides 404.8 (entityNotFound) when performing the update, it will log the error and stop before trying to create. For example, if the base version is wrong in the update+create, the error will be about not finding the correct base version.

Lots of tests.

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -41,14 +41,13 @@ const normalizeUuid = (id) => {
 };
 
 // Input: object representing entity, parsed from submission XML
-const extractLabelFromSubmission = (entity) => {
-  const { label, create, update } = entity.system;
-  if ((create === '1' || create === 'true') && (!label || label.trim() === ''))
-    throw Problem.user.missingParameter({ field: 'label' });
-
-  if ((update === '1' || update === 'true') && (!label || label.trim() === ''))
+const extractLabelFromSubmission = (entity, options = { create: true }) => {
+  const { create, update } = options;
+  const { label } = entity.system;
+  if (update && (!label || label.trim() === ''))
     return null;
-
+  if (create && (!label || label.trim() === ''))
+    throw Problem.user.missingParameter({ field: 'label' });
   return label;
 };
 
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -278,13 +278,19 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Datasets, Entit
   const dataset = (await Datasets.get(form.get().projectId, entityData.system.dataset, true))
     .orThrow(Problem.user.datasetNotFound({ datasetName: entityData.system.dataset }));
 
-  // Create entity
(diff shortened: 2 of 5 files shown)


lib/model/query/entities.js
```diff
@@ -278,13 +278,19 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Datasets, Entit
   const dataset = (await Datasets.get(form.get().projectId, entityData.system.dataset, true))
     .orThrow(Problem.user.datasetNotFound({ datasetName: entityData.system.dataset }));
 
-  // Create entity
-  if (entityData.system.create === '1' || entityData.system.create === 'true')
-    return Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent);
-
   // Or update entity
```
matthew-white, line 281: Update or remove comment?

test/integration/api/entities.js
```diff
@@ -2054,5 +2054,139 @@ describe('Entities API', () => {
...
+
+        await exhaust(container);
+
+        await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789aaa')
+          .expect(404);
+
+        await asAlice.get('/v1/projects/1/forms/updateEntity/submissions/one/audits')
+          .expect(200)
+          .then(({ body: logs }) => {
+            logs[0].action.should.be.eql('entity.error');
```
matthew-white, line 2104: What do you think about asserting something about the audit log details to verify that the `entity.error` is the one that `_createEntity()` threw, not the 404.8 that `_updateEntity()` threw?

matthew-white (approved):
Looks great. 💯

I'll make sure to review #1035 before Monday so that you can merge that one before this one.

lib/model/query/entities.js
matthew-white: Specifically, I think the "Or" seems out-of-place now that "Create entity" above is gone.

lib/data/entity.js
```diff
@@ -41,14 +41,13 @@ const normalizeUuid = (id) => {
...
 
 // Input: object representing entity, parsed from submission XML
-const extractLabelFromSubmission = (entity) => {
-  const { label, create, update } = entity.system;
-  if ((create === '1' || create === 'true') && (!label || label.trim() === ''))
-    throw Problem.user.missingParameter({ field: 'label' });
-
-  if ((update === '1' || update === 'true') && (!label || label.trim() === ''))
+const extractLabelFromSubmission = (entity, options = { create: true }) => {
+  const { create, update } = options;
```
ktuite, line 45: @matthew-white I added this options arg because I was running into a problem where a submission specified both `create` and `update`, but the validation behavior needed to be different depending on if we were updating or creating.

PR #1044, opened 2023-11-10
Trigger entity update on submission edit

ktuite:
Closes https://github.com/getodk/central/issues/517 by letting `submission.update.version` events also be used to update entities. 

All the scenarios listed in the issue are tested.

### processing pending submissions
I also tested what happens when some submissions that are meant to _update_ entities get picked up in the "pending submissions" when a dataset that requires approval for entity creation gets toggled to no longer require approval. Since we don't know ahead of time which submissions create vs. update entities, we can't really treat them differently here. 

However, entity updates should get processed right away (regardless of if approval is required) so this probably wont come up often. The main situation this comes up is if a entity update submission comes in, the worker hasn't come by to process it yet, and the user toggles the dataset flag in the time between submission and worker processing. Or if a submission comes in after they toggle the dataset, but the worker still hasn't run yet.

Within this flow, it could also happen that dataset update event gets queued after the submission events, then the worker processes the individual submissions, and by the time it gets to the bulk processing event, there is nothing left to process.

There's a lot going on but ultimately none of it seems like a problem.

Also because of how entity updates get processed, there is no link from the update event back to the parent bulk processing `dataset.update` event.

### frontend changes 

I don't actually think anything needs to change here on the front end.

This shows two updates, one which came from a `submission.update.version` event and one from a `submission.create` event.  We just show the submission and don't mention the specific kind of submission event for an entity update. 
Screenshot 2023-11-15 at 3 50 47 PM

Here's the submission feed
Screenshot 2023-11-15 at 3 55 35 PM

I made a [conditional update form](https://staging.getodk.cloud/#/projects/49/forms/tree_conditional_update) on staging that works locally and will work once this PR is merged. 

This PR is mostly tests. 

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -171,7 +171,7 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
 };
 
 const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event) => async ({ Audits, Entities, maybeOne }) => {
-  if (!(event.action === 'submission.create')) // only update on submission.create
+  if (!(event.action === 'submission.create' || event.action === 'submission.update.version'))
     return null;
 
   // Get client version of entity
--- a/test/data/xml.js
+++ b/test/data/xml.js
@@ -691,6 +691,15 @@ module.exports = {
             <label>Alicia - 85</label>
           </entity>
         </meta>
+      </data>`,
+      three: `<data xmlns:jr="http://openrosa.org/javarosa" xmlns:entities="http://www.opendatakit.org/xforms" id="updateEntity" version="1.0">
+        <meta>
+          <instanceID>three</instanceID>
+          <orx:instanceName>three</orx:instanceName>
+          <entity dataset="people" id="12345678-1234-4123-8234-123456789abc" baseVersion="1" update="1">
+          </entity>
+        </meta>
+        <age>55</age>
       </data>`
     },
     groupRepeat: {
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -3324,6 +3324,89 @@ describe('datasets and entities', () => {
 
       }));
 
+      it('should use latest version of submission in pending submissions', testService(async (service, container) => {
+        const asAlice = await service.login('alice');
+
+        await asAlice.post('/v1/projects/1/forms?publish=true')
+          .send(testData.forms.simpleEntity)
+          .set('Content-Type', 'application/xml')
+          .expect(200);
+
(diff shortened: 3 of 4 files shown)


matthew-white:
One other thing I thought to test: an `entity.error` should not preclude a later submission edit from creating or updating an entity. Example:

- Submit a submission that defines a new entity: `<entity create="true" ...>`
- However, there's something wrong (the UUID is invalid, say). Processing the submission results in an `entity.error` audit log event, not a new entity.
- Edit the submission to correct the error.
- The submission edit should be reprocessed and should create an entity.

matthew-white:
I'm thinking that this PR won't fully close getodk/central#517. I think that issue needs some Frontend changes as well.

matthew-white:
I've filed getodk/central#547 related to processing pending submissions that (failed to) update entities. I think any change for that should come in a follow-up PR: we should merge this one first. I'll finish taking a look at this PR tomorrow.

matthew-white:
> processing pending submissions

> There's a lot going on but ultimately none of it seems like a problem.

I agree with this. If there are unprocessed entity updates at the time of the `dataset.update`, then they will get processed either before or during when the `dataset.update` is processed — and either possibility seems fine.

As I mentioned in getodk/central#547, part of me would like a `dataset.update` to only trigger entity creation, not entity updates. However, I don't think that works given that `update="1" create="1"` is a possibility, when update is attempted before create. There's not necessarily a neat division between submissions that create entities and submissions that update them.

I'm wondering wonder whether we should update the text in `DatasetPendingSubmissions` in Frontend to make it clear that pending submissions will be processed for any entity action, not just creation. Right now, the text says:

> Convert all pending Submissions to Entities now.
> Change the setting and create Entities out of {count} Submissions not yet Approved or Rejected.

I would lean toward no change. Like you said, entity updates should be processed quickly, and it doesn't really matter when they are processed relative to when the `dataset.update` is processed. Also, I don't know how common it will be for users to opt into requiring submission approval for entity creation, yet at the same time use entity updates, which never require approval. What do you think about `DatasetPendingSubmissions`? Any change needed?

matthew-white (approved):

ktuite:
> I would lean toward no change. Like you said, entity updates should be processed quickly, and it doesn't really matter when they are processed relative to when the `dataset.update` is processed. Also, I don't know how common it will be for users to opt into requiring submission approval for entity creation, yet at the same time use entity updates, which never require approval. What do you think about `DatasetPendingSubmissions`? Any change needed?

I don't think we need to change any wording, I hope it's a scenario that doesn't come up often! But if we are eventually able to identify submissions that do not create entities (because we track form actions), we should probably skip them here.

PR #1045, opened 2023-11-10
Add ETag header to all blobs

alxndrsn:
### Benefits

* improved server performance in situations where a client re-accesses the same data multiple times
* near-zero implementation cost, as blob hashes have already been fetched from db
* centralises code for serving blobs, which will simplify implementation of:
  * support for external storage options
  * streaming blob data from db (not currently scheduled AFAIK)
  
### Costs

* additional tests
* additional documentation

### DIFF
--- a/test/integration/api/forms/forms.js
+++ b/test/integration/api/forms/forms.js
@@ -772,25 +772,6 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
               text.should.equal(testData.forms.simple);
             }))));
 
-      it('should return standard body if incorrect ETag provided', testService((service) =>
-        service.login('alice', (asAlice) =>
-          asAlice.get('/v1/projects/1/forms/simple.xml')
-            .set('If-None-Match', 'a-random-string')
-            .expect(200)
-            .then(({ text }) => {
-              text.should.equal(testData.forms.simple);
-            }))));
-
-      it('should return 304 if correct ETag provided', testService((service) =>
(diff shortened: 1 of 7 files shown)


test/integration/api/forms/forms.js
```diff
@@ -772,6 +772,25 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
...
+          asAlice.get('/v1/projects/1/forms/simple.xml')
+            .set('If-None-Match', 'a-random-string')
+            .expect(200)
+            .then(({ text }) => {
+              text.should.equal(testData.forms.simple);
+            }))));
+
+      it('should return 304 if correct ETag provided', testService((service) =>
+        service.login('alice', async (asAlice) => {
+          const res = await asAlice.get('/v1/projects/1/forms/simple.xml').expect(200);
```
matthew-white, line 786: Form XML isn't stored as a blob, so I don't think this test is exercising the new functionality. We do store XLSForms as blobs though.

matthew-white (commented):
I'm convinced that this is a change worth making. Especially once we support API keys, I could imagine a client requesting the same submission attachment multiple times.

One thing I'm not seeing though is CircleCI / GitHub Actions for some reason.

matthew-white (approved):

matthew-white:
I've added this PR to the GitHub project and assigned to @alxndrsn, since it looks like the PR will be merged in time for the v2024.1 release.

PR #1046, opened 2023-11-16
Allow 2023.1 entity xform spec version (entity updates)

ktuite:
Makes Central support the latests version of the entities xform spec (v2023.1.0). This version is all about submissions that allow (online) entity updates. We wanted a way to alert users that if their form used the new entity update xform spec and they uploaded it into an old version of Central, it would let them know that it couldn't do the updating. 

The xform spec changes between versions was additive, adding the `update=true`, `baseVersion` so this is mostly allowing the new spec version, but it doesn't change how the forms are processed under the hood. 

Tests

Is it too simple? Is there a better "check if version is in acceptable list of versions" snippet of code?

Lets us try entity update forms converted with the latest pyxform! 

### DIFF
--- a/lib/data/dataset.js
+++ b/lib/data/dataset.js
@@ -68,7 +68,7 @@ const getDataset = (xml) => {
       return Option.none();
     if (version.isEmpty())
       throw Problem.user.invalidEntityForm({ reason: 'Entities specification version is missing.' });
-    else if (!version.get().startsWith('2022.1.'))
+    else if (!(version.get().startsWith('2022.1.') || version.get().startsWith('2023.1.')))
       throw Problem.user.invalidEntityForm({ reason: `Entities specification version [${version.get()}] is not supported.` });
     // check that dataset name is valid
     if (datasetName.isEmpty())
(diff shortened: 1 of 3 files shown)


lib/data/dataset.js
```diff
@@ -68,7 +68,7 @@ const getDataset = (xml) => {
       return Option.none();
     if (version.isEmpty())
       throw Problem.user.invalidEntityForm({ reason: 'Entities specification version is missing.' });
-    else if (!version.get().startsWith('2022.1.'))
+    else if (!(version.get().startsWith('2022.1.') || version.get().startsWith('2023.1.')))
```
matthew-white, line 71: You could use the `semver` package if you wanted. It's already a dependency. (Though it's a dev dependency, so you'd probably need to reinstall it as a prod dependency.) This also looks fine though.

matthew-white (approved):

lib/data/dataset.js
ktuite: maybe *next* version bump?

PR #1048, opened 2023-11-21
Update backfill migration to also set label in dataReceived column

ktuite:
Closes https://github.com/getodk/central/issues/540

I debated between adding a new migration that would fill in this data, or changing the old migration file. I decided to change the old migration file, the one that added `baseVersion` and that introduced the column `dataReceived`, to also populate the label field in `dataReceived`. 

I was nervous about getting that change right so I wrote a test, but I've committed the code with the test skipped. 

A test.

I think this is tidier. The change is simple, it'll mean fewer migrations have to run on upgrade. The only downside is some of our test data may not have the label populated. 

no 

no

### DIFF
--- a/test/integration/other/migrations.js
+++ b/test/integration/other/migrations.js
@@ -866,3 +866,46 @@ describe('database migrations from 20230802: delete orphan submissions', functio
 
   }));
 });
+
+describe.skip('database migration: 20231002-01-add-conflict-details.js', function test() {
+  this.timeout(20000);
+
+  it('should update dataReceived and baseVersion of existing entity defs', testServiceFullTrx(async (service, container) => {
+    await upToMigration('20231002-01-add-conflict-details.js', false);
+    await populateUsers(container);
+    await populateForms(container);
+
+    const asAlice = await service.login('alice');
+
+    await asAlice.post('/v1/projects/1/forms?publish=true')
+      .set('Content-Type', 'application/xml')
+      .send(testData.forms.simpleEntity)
+      .expect(200);
+
+    const dsId = await container.oneFirst(sql`select id from datasets limit 1`);
+
+    // Create the entity in the DB directly because application code has changed.
+    const newEntity = await container.one(sql`
+    INSERT INTO entities (uuid, "datasetId", "creatorId", "createdAt")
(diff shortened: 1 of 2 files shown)


test/integration/other/migrations.js
```diff
@@ -866,3 +866,46 @@ describe('database migrations from 20230802: delete orphan submissions', functio
...
+    // Create the entity in the DB directly because application code has changed.
+    const newEntity = await container.one(sql`
+    INSERT INTO entities (uuid, "datasetId", "creatorId", "createdAt")
+    VALUES (${uuid()}, ${dsId}, 5, now())
+    RETURNING "id"`);
+
+    // Create two entity defs
+    await container.run(sql`
+      INSERT INTO entity_defs ("entityId", "createdAt", "creatorId", "current", "label", "data")
+      VALUES (${newEntity.id}, now(), 5, true, 'some label', '{"foo": "bar"}')`);
```
matthew-white, line 896: Seems like `current` should be `false` if another def is about to be created.

test/integration/other/migrations.js
```diff
@@ -866,3 +866,46 @@ describe('database migrations from 20230802: delete orphan submissions', functio
...
+
+    await container.run(sql`
+      INSERT INTO entity_defs ("entityId", "createdAt", "creatorId", "current", "label", "data", "version")
```
matthew-white, line 899: Should `version` be specified for both defs? The `baseVersion` of the first def will be `null` after the migration, but I think its `version` should be 1.

matthew-white (approved):

test/integration/other/migrations.js
ktuite: The column has a default value of 1, which I visually saw, and is how the `baseVersion` gets set to null.

test/integration/other/migrations.js
ktuite: it's not the best/most coherent entity def 😅 but at least `current` is not a relevant part of the migration.

ktuite: I will make these improvements in case anyone (future me?) decides to "borrow" this entity-insertion code in the future...

PR #1049, opened 2023-11-21
Tests of reprocessing submissions that yielded entity.error earlier

ktuite:
Related to https://github.com/getodk/central/issues/547

The main code change here is that in the list of submissions to reprocess when toggling the dataset `approvalRequired` flag, submissions that were already processed for some reason and led to an entity processing `entity.error` are now filtered out and not reprocessed.

Two main scenarios: 
1. A submission failed to create an entity, and then the status was changed so there was a new `submission.update` event. Because the first attempt created an error, this will no longer be re-processed.
2. A submission failed to update an entity, and then something with the dataset changed (maybe a non-existent entity came into existence) but that submission wont be reprocessed in the "reprocess all unprocessed submissions" flow. (It was awkward it got caught up in this process anyway, but we don't have a good way of knowing ahead of time if a submission is about creating or updating so we have to re-examine it.)
    * In this entity update case, if you _edit_ the submission, that can re-trigger the desired entity update.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -456,9 +456,14 @@ LEFT JOIN (
   JOIN entity_def_sources es ON es.id = ed."sourceId"
   JOIN entities e ON e.id =  ed."entityId" AND e."datasetId" = ${datasetId}
 ) ON es."submissionDefId" = sd.id
+LEFT JOIN (
+  SELECT DISTINCT ON ((details->'submissionId')::INTEGER) * FROM audits
+  WHERE action = 'entity.error'
+) errors ON (errors.details->'submissionId')::INTEGER = s.id
 WHERE sd.current AND dfd."datasetId" = ${datasetId}
 AND (s."reviewState" IS NULL OR s."reviewState" = 'edited' OR s."reviewState" = 'hasIssues')
 AND ed.id IS NULL
+AND errors.action IS NULL
 `;
 
 const countUnprocessedSubmissions = (datasetId) => ({ oneFirst }) => oneFirst(_unprocessedSubmissions(datasetId, sql`COUNT(1)`));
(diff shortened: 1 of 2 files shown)


lib/model/query/datasets.js
```diff
@@ -456,9 +456,14 @@ LEFT JOIN (
   JOIN entity_def_sources es ON es.id = ed."sourceId"
   JOIN entities e ON e.id =  ed."entityId" AND e."datasetId" = ${datasetId}
 ) ON es."submissionDefId" = sd.id
+LEFT JOIN (
+  SELECT DISTINCT ON ((details->'submissionId')::INTEGER) * FROM audits
```
matthew-white, line 460: Should we use `submissionDefId` instead of `submissionId`? If the submission had an error, then an edit, and no error specifically associated with the edit, it should still be processed I think.

ktuite: I thought about this, but the way this would happen is
- submission
- approval
- error
- edit (new sub def)
- DON'T approve yet
- change dataset flag

... and that seems so specialized to have the auto processing happen then. 

Someone could approve earlier or could do the edit after changing the dataset flag and it should work.

It was easier for me to think about how if the submission had ever caused an error with an entity, it wouldn't get caught up in this reprocessing scenario.

test/integration/api/datasets.js
```diff
@@ -3546,5 +3546,181 @@ describe('datasets and entities', () => {
...
+          .send(testData.instances.simpleEntity.two)
+          .set('Content-Type', 'application/xml')
+          .expect(200);
+
+        // Approve the submission.
+        await asAlice.patch('/v1/projects/1/forms/simpleEntity/submissions/two')
+          .send({ reviewState: 'approved' })
+          .expect(200);
+
+        // Create a submission isn't yet approved and DOES need to be reprocessed
```
ktuite, line 3588: processed for the first time

matthew-white (approved):

lib/model/query/datasets.js
ktuite: Changing the code so it _will_ process any new version of a submission that hasn't specifically generated an error.

PR #1052, opened 2023-11-29
Upgrade to node 20.10 and postgres 14.10

lognaturel:
Ran locally, clicked around and downloaded some submission zips.

Latest minor versions available.

There could be some new bugs introduced in these versions.

No.


lognaturel:
In CI I see 

```
  1) api: /forms/:id/submissions
       .csv.zip GET
         should include all repeat rows @slow:
     Error: read ECONNRESET
      at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
```

I can't seem to repro locally. It's a timeout, right?

matthew-white:
For posterity: Previously, in the one failing test, 50 concurrent requests were sent in a `Promise.all()`. The test passed for @lognaturel locally, but not in CircleCI. When she reduced the number of requests from 50 to 30, the test passed in CircleCI. I came across ladjs/supertest#709, which mentions issues with SuperTest and Node 20 when there are many concurrent requests. I deparallelized the one test, and then it passed. I assume that the failing test was just due to that GitHub issue and not a performance degradation in Backend. The soak tester continues to pass.

matthew-white:
@lognaturel, I think we also need to bump Node in oidc-dev/playwright-tests/package.json.

matthew-white:
Should we bump Postgres for `run-docker-postgres` in the makefile? I'm actually not sure what the effect would be for someone who's already created that database locally. We could look into that later, but if you happen to know that it's OK, feel free to bump the version there!

matthew-white (approved):
To review, I searched the codebase for the strings "18.17" and "14.6". I left a couple of comments above, but this PR is looking great. Thank you, @lognaturel!

matthew-white:
Do we need to wait until we have a PR to the `central` repo to merge this PR? I'm wondering what the effect on staging would be to merge this PR without also making a change to `central`.

lognaturel:
Since no code changes were needed, we can just go ahead and get the Central PR in first whether or not it's necessary. Give me a sec.

lognaturel:
Infrastructure PR at https://github.com/getodk/central/pull/562

matthew-white:
@ktuite, you use `run-docker-postgres` locally, right? You might be in a good position to check whether it's easy to bump the Postgres version. Also probably not a big deal to leave it as-is.

PR #1053, opened 2023-11-29
Enforce that entity field is always structure type

ktuite:
In response to https://github.com/getodk/central/issues/551, https://github.com/getodk/central/issues/552, and https://github.com/getodk/central/issues/552 

It was the case that the whether or not the `<entity>` block had a child like `<label>` would change the type of field that entity was from `structure` (if it had children) to `unknown`. This lead to each of the errors above. 

Example 1: This is possible for an update form that doesn't update an entity's label, but wasn't really possible before because a label was needed to create an entity.
```
<entity dataset="people" id="" update="" baseVersion="" />
```

Example 2: This was the only kind of form we saw before, where `entity` was always structural.
```
<entity dataset="people" id="" create="">
	<label/>
</entity>
```

The main change in this PR is forcing the entity field, which so far is always at `/meta/entity` to be structural, even if it is empty. 

Tests.

This is the simplest approach. The code change is small, if solves a bunch of issues at once.
The other option we considered was not changing the field type and adding workarounds for when `entity` was `unknown`. But these were getting cumbersome and likely hard/annoying to maintain or work around in the future. 

**Risk 1:** The fact that `entity` can be an empty structural node might cause some problems somewhere. There are places in the code that use the structure type of the field to recurse over children. That could lead to a problem that we haven't realized yet. The entity field is the only one that gets this treatment, though, so the problem would be contained to those fields and those label-less update forms. 

We want to check exports and OData because that is where, if there were a problem, it would probably come up.

**Risk 2:** It's unlikely that people got forms into the system that had `<entity>` blocks without labels/children that got saved as type `unknown`... but maybe? We could add a migration to make sure all meta/entity fields are structural. Or we could not, because it's probably not common. It would have to be a super old form that someone is oddly committed to upgrading rather than replacing to run into this problem at all. 

Luckily no.


test/integration/api/datasets.js
```diff
@@ -2864,6 +2864,243 @@ describe('datasets and entities', () => {
...
+        .expect(200);
+
+      await asAlice.post('/v1/projects/1/forms/brokenForm/draft')
+        .send(form2)
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+    }));
+
+    // cb#553 issue, forms with and without entity label show different fields
+    // (because entity has type 'unknown' instead of 'structure')
```
ktuite, line 2985: update comment -- they now both have the same field type, and `structure` types wont be shown on frontend.
https://github.com/getodk/central/issues/553

lib/data/schema.js
```diff
@@ -91,6 +91,10 @@ const _recurseFormFields = (instance, bindings, repeats, selectMultiples, envelo
         // if we have no binding node but we have children, assume this is a
         // structural node with no repeat or direct data binding; recurse.
         field.type = 'structure';
+      } else if (tag.name === 'entity' && joinedPath === '/meta/entity') {
```
ktuite, line 94: This condition might be overkill, i just wanted to be very clear which tag is targeted here!

test/integration/api/datasets.js
```diff
@@ -2864,6 +2864,243 @@ describe('datasets and entities', () => {
...
+
+      await asAlice.get('/v1/projects/1/forms/brokenForm/submissions/one/audits')
+        .expect(200)
+        .then(({ body: logs }) => {
+          logs[0].action.should.equal('entity.update.version');
+        });
+    }));
+
+    // c#552 issue, can't add label to entity update form that previously didnt have label
+    it('should allow update where no label or no properties are updated', testService(async (service) => {
```
ktuite, line 2934: change name of test

test/integration/api/datasets.js
```diff
@@ -2864,6 +2864,243 @@ describe('datasets and entities', () => {
...
+
+      await asAlice.post('/v1/projects/1/forms/brokenForm/draft')
+        .send(form2)
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+    }));
+
+    // c#553 issue, forms with and without entity label show different fields
+    // (because entity was previously type 'unknown' instead of 'structure')
+    it('should allow update where no label or no properties are updated', testService(async (service) => {
```
ktuite, line 2986: change name of test

test/integration/api/datasets.js
```diff
@@ -2864,6 +2864,243 @@ describe('datasets and entities', () => {
...
+      await asAlice.get('/v1/projects/1/forms/updateWithoutLabel/fields?odata=true')
+        .then(({ body }) => {
+          body[2].path.should.equal('/meta/entity');
+          body[2].type.should.equal('structure');
+        });
+
+      await asAlice.get('/v1/projects/1/forms/updateWithLabel/fields?odata=true')
+        .then(({ body }) => {
+          body[2].path.should.equal('/meta/entity');
+          body[2].type.should.equal('structure');
```
ktuite, line 3046: test that label field is there, too

test/integration/other/empty-entity-structure.js
```diff
@@ -0,0 +1,276 @@
...
+          body.one2.should.eql([
+            { old: 'one', new: 'one2', path: [ 'meta', 'instanceID' ] },
+            { new: 'one', path: [ 'meta', 'deprecatedID' ] },
+            { old: '88', new: '', path: [ 'age' ] },
+            { old: '', new: 'seattle', path: [ 'location', 'hometown' ] }
+          ]);
+        });
+    }));
+
+    it('should check diff when form version changes to add label', testService(async (service) => {
```
ktuite, line 81: File an issue with this messy diff

<details>
<summary>Example of details block in markdown</summary>

Look, there's cool stuff hidden inside here!
</details>

ktuite (commented):

matthew-white (approved):

Issue #1000, opened 2023-09-22
Different port numbers for pyxform-http


sahilcodes-com:
I encountered this issue when i was setting up central-frontend and central-backend locally. In the [README.md](https://github.com/getodk/pyxform-http#run-in-docker) file of pyxform-http, it specifies port number 5001. However, central-backend is actually utilizing port 5000 to establish a connection with pyxform-http.

yanokwa:
Will be fixed when https://github.com/getodk/pyxform-http/pull/41 is merged.

PR #1055, opened 2023-12-05
Change pyxform-http port because macOS uses port 5000 for ControlCenter

yanokwa:
I updated pyxform-http's port to 5000 to match Central's docs because of https://github.com/getodk/central-backend/issues/1000.

But that broke running pyxform-http on macOS because port 5000 is used by [ControlCenter](https://nono.ma/port-5000-used-by-control-center-in-macos-ventura-controlce).

So, I'll be reverting the pyxform-http change, but I think we should also make this change here. I say think because I have no idea the implications for this change in Central.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -15,7 +15,7 @@
     },
     "xlsform": {
       "host": "localhost",
-      "port": 5000
+      "port": 5001
     },
     "enketo": {
       "url": "http://localhost:8005/-",


matthew-white (approved):
Looks good to me! I don't think anything else needs to change. We specify a different port in production, so everything should be fine there: https://github.com/getodk/central/blob/774b491961578bd6e03ceaa6cf8032dd27b828a7/files/service/config.json.template#L26. Feel free to merge!

PR #1059, opened 2023-12-08
Upgrade postgres to 14.10 in dev container setup

lognaturel:
Follow on to https://github.com/getodk/central-backend/pull/1052

I wanted the changes that affect production to go on staging as quickly as possible so left this behind. At the time I was having trouble running the make task because I already had a postgres server running on my machine. The make task would start the container but then would fail when trying to create the `jubilant` role because it already existed in my local db.

I've also switched to the alpine image. I don't think we need any particular utilities to be available and this saves ~40mb in download.

Discarded my existing container and image, ran the make task, verified I could use the db with Central and that the postgres version was 14.10.

This is not really a complete solution because it won't upgrade an existing container. It seems like a reasonable first step but I think we should give some more thought to what we want this task to do:
- If the goal is to quickly spin up a clean database, then this is probably fine. Folks who use it will periodically throw away their existing containers and images and create new ones which will have the latest postgres version. 
- If the goal is to have a stable local database for development, I think we should use a named volume and make it so containers can be upgraded

None.

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ lint: node_version
 
 .PHONY: run-docker-postgres
 run-docker-postgres: stop-docker-postgres
-	docker start odk-postgres14 || (docker run -d --name odk-postgres14 -p 5432:5432 -e POSTGRES_PASSWORD=odktest postgres:14.10 && sleep 5 && node lib/bin/create-docker-databases.js)
+	docker start odk-postgres14 || (docker run -d --name odk-postgres14 -p 5432:5432 -e POSTGRES_PASSWORD=odktest postgres:14.10-alpine && sleep 5 && node lib/bin/create-docker-databases.js)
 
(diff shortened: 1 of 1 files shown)


ktuite (approved):
Looks good! I tried it (following the path of just deleting my existing dev DB) and now I'm onto this fresh new database.

I like the simplicity of the first goal and that's how I seem to use it: 
> If the goal is to quickly spin up a clean database, then this is probably fine. Folks who use it will periodically throw away their existing containers and images and create new ones which will have the latest postgres version.

It's just a helpful little utility! And the less is baked into it, the less extra stuff someone needs to load into their brain. If someone needs their data to persist, they can choose their favorite option 
- run postgres some other way
- don't run this command to remake the postgres dev container
- use other database (or central) backup/restore techniques

Issue #1006, opened 2023-09-26
Update dependencies for v2023.5


matthew-white:

lognaturel:
Make sure to upgrade `bcrypt` as described by @swoodcock at https://forum.getodk.org/t/bcrypt-dependency-install-fails-on-arm64-architectures/43992

I think we still want to keep control over the minor version to ensure consistency but could be convinced otherwise.

matthew-white:
> I think we still want to keep control over the minor version to ensure consistency but could be convinced otherwise.

Just to clarify, are you thinking of the semver range specified in package.json? package-lock.json is what ensures consistency across installs. The semver range constrains what updates are possible when a dev runs `npm update`.

Relatedly, I noticed this statement in the forum post:

> It's currently pinned with a tilde `~5`, meaning the max version is `5.0.1`.

However, I'm pretty sure that `~5` includes any `5.x`.

ktuite:
Ah yes I see what you mean about `5.x` ([npm docs](https://docs.npmjs.com/cli/v6/using-npm/semver#tilde-ranges-123-12-1)). I just updated bcypt within the `package-lock.json` file.

PR #1060, opened 2023-12-08
Upgrade backend dependencies

ktuite:
Closes https://github.com/getodk/central-backend/issues/1006

This updates `bcrypt` from `5.0.1` to `5.1.1`.

This PR also resolves the **critical** vulnerability:
```
@babel/traverse  <7.23.2
Severity: critical
Babel vulnerable to arbitrary code execution when compiling specifically crafted malicious code - https://github.com/advisories/GHSA-67hx-6x53-jw92
fix available via `npm audit fix`
node_modules/@babel/traverse
```

This PR doesn't change this other set of issues with `axios`. I think we should wait for `pm2` or something else to change, because otherwise this fix looks like it rolls all the way back to an old pm2 (or pm2/js-api) when we are currently on `pm2@5.3.0`.

```
axios  0.8.1 - 1.5.1
Severity: moderate
Axios Cross-Site Request Forgery Vulnerability - https://github.com/advisories/GHSA-wf5p-g6vw-rhxx
fix available via `npm audit fix --force`
Will install pm2@2.10.4, which is a breaking change
node_modules/axios
  @pm2/js-api  *
  Depends on vulnerable versions of axios
  node_modules/@pm2/js-api
    pm2  >=3.0.0
    Depends on vulnerable versions of @pm2/js-api
    node_modules/pm2
```

`npm audit` also reports the usual knex vulnerability that doesn't affect is but that we're not able to deal with right now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -89,15 +89,87 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.18.6",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.18.6.tgz",
-      "integrity": "sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==",
+      "version": "7.23.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.23.5.tgz",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "@sentry/node": "~7",
     "archiver": "~5",
-    "bcrypt": "~5",
+    "bcrypt": "~5.1",
```
matthew-white, line 14: See my comment on the issue: https://github.com/getodk/central-backend/issues/1006#issuecomment-1850884855. I don't think we really need to pin the minor version in package.json. It's enough that package-lock.json specifies the exact version to install. If we specify a minor version here, it'll be slightly harder to update to a later version in the future. Not a big deal though, we can always change it in package.json later if you prefer to merge this.

matthew-white (approved):

PR #1061, opened 2023-12-08
Check entity actions in submission against those permitted by form def

matthew-white:
Closes getodk/central#518.

New tests.

The permitted actions from the form def are persisted on the `dataset_form_defs` record. It seems like a good idea to persist the actions rather than parsing the form XML each time a submission is processed for entities. `dataset_form_defs` also seems like the right table to store the actions. In the future, a form def will be able to take actions on multiple datasets, and in theory, it could take different actions on the different datasets. So conceptually, the permitted actions aren't per form def: they're per form def + dataset.

Note also that a single form may have different versions/defs that permit different actions. When a submission is processed, the form version specified in the submission will be used.

All existing forms that make changes to entities will permit entity creation and only entity creation.

One edge case that was tolerated before now results in an error. See test/unit/data/dataset.js for details about that change. I think it's likely good that we no longer tolerate that edge case.

No API responses have changed (other than some Problems). Maybe one day we'll want to return the list of permitted actions over the API.

### DIFF
--- a/lib/data/dataset.js
+++ b/lib/data/dataset.js
@@ -15,7 +15,7 @@
 
 const Option = require('../util/option');
 const Problem = require('../util/problem');
-const { traverseXml, findOne, and, root, node, hasAttr, tree, attr } = require('../util/xml');
+const { traverseXml, findOne, root, node, attr, stripNamespacesFromPath } = require('../util/xml');
 
 const validateDatasetName = (name) => {
   // Regex explanation:
@@ -61,21 +61,36 @@ const getDataset = (xml) => {
   const metaNode = findOne(root('html'), node('head'), node('model'), node('instance'), node(), node('meta'));
   return traverseXml(xml, [
     findOne(root('html'), node('head'), node('model'))(attr('entities-version')),
-    metaNode(findOne(root(), node('entity'))(tree())),
-    metaNode(findOne(root(), and(node('entity'), hasAttr('dataset')))(attr('dataset')))
-  ]).then(([ version, entityTag, datasetName ]) => {
-    if (entityTag.isEmpty() && datasetName.isEmpty())
+    metaNode(findOne(root(), node('entity'))(attr()))
+  ]).then(([ version, entityAttrs ]) => {
+    if (entityAttrs.isEmpty())
       return Option.none();
+
     if (version.isEmpty())
       throw Problem.user.invalidEntityForm({ reason: 'Entities specification version is missing.' });
     else if (!(version.get().startsWith('2022.1.') || version.get().startsWith('2023.1.')))
       throw Problem.user.invalidEntityForm({ reason: `Entities specification version [${version.get()}] is not supported.` });
+
+    const strippedAttrs = Object.create(null);
+    for (const [name, value] of Object.entries(entityAttrs.get()))
+      strippedAttrs[stripNamespacesFromPath(name)] = value;
+
     // check that dataset name is valid
-    if (datasetName.isEmpty())
+    const datasetName = strippedAttrs.dataset;
+    if (datasetName == null)
       throw Problem.user.invalidEntityForm({ reason: 'Dataset name is missing.' });
-    if (!validateDatasetName(datasetName.get()))
+    if (!validateDatasetName(datasetName))
       throw Problem.user.invalidEntityForm({ reason: 'Invalid dataset name.' });
-    return Option.of(datasetName.get().trim());
+
+    // Entity actions permitted by the form def
+    const actions = [];
+    const { create, update } = strippedAttrs;
+    if (create != null) actions.push('create');
+    if (update != null) actions.push('update');
+    if (actions.length === 0)
+      throw Problem.user.invalidEntityForm({ reason: 'The form must specify at least one entity action, for example, create or update.' });
+
+    return Option.of({ name: datasetName.trim(), actions });
   });
 };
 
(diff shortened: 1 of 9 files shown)


lib/util/problem.js
```diff
@@ -136,6 +136,9 @@ const problems = {
     // TODO: should have details but not sure what yet.
     insufficientRights: problem(403.1, () => 'The authentication you provided does not have rights to perform that action.'),
 
+    entityActionNotAllowed: problem(403.2, ({ action, permitted }) =>
+      `The submission attempts an entity ${action}, but the form does not permit that action. The form permits the following actions: ${permitted.join(',')}.`),
```
matthew-white, line 140: `', '`

lib/data/dataset.js
matthew-white, line 1: Update the comments above `getDataset()`.

lib/data/dataset.js
```diff
@@ -61,21 +61,36 @@ const getDataset = (xml) => {
   const metaNode = findOne(root('html'), node('head'), node('model'), node('instance'), node(), node('meta'));
   return traverseXml(xml, [
     findOne(root('html'), node('head'), node('model'))(attr('entities-version')),
-    metaNode(findOne(root(), node('entity'))(tree())),
-    metaNode(findOne(root(), and(node('entity'), hasAttr('dataset')))(attr('dataset')))
-  ]).then(([ version, entityTag, datasetName ]) => {
-    if (entityTag.isEmpty() && datasetName.isEmpty())
+    metaNode(findOne(root(), node('entity'))(attr()))
```
matthew-white, line 64: Move the `metaNode` here since it's no longer referenced twice.

lib/data/dataset.js
```diff
@@ -61,21 +61,36 @@ const getDataset = (xml) => {
...
     else if (!(version.get().startsWith('2022.1.') || version.get().startsWith('2023.1.')))
       throw Problem.user.invalidEntityForm({ reason: `Entities specification version [${version.get()}] is not supported.` });
+
+    const strippedAttrs = Object.create(null);
+    for (const [name, value] of Object.entries(entityAttrs.get()))
+      strippedAttrs[stripNamespacesFromPath(name)] = value;
+
     // check that dataset name is valid
-    if (datasetName.isEmpty())
+    const datasetName = strippedAttrs.dataset;
```
matthew-white, line 79:
```suggestion
    const datasetName = strippedAttrs.dataset?.trim();
```

lib/data/dataset.js
```diff
@@ -61,21 +61,36 @@ const getDataset = (xml) => {
...
+    if (!validateDatasetName(datasetName))
       throw Problem.user.invalidEntityForm({ reason: 'Invalid dataset name.' });
-    return Option.of(datasetName.get().trim());
+
+    // Entity actions permitted by the form def
+    const actions = [];
+    const { create, update } = strippedAttrs;
+    if (create != null) actions.push('create');
+    if (update != null) actions.push('update');
+    if (actions.length === 0)
```
matthew-white, line 90: Allow an empty array. As part of this, maybe update the Problem message.

lib/model/query/entities.js
```diff
@@ -148,6 +148,14 @@ resolveConflict.audit = (entity, dataset) => (log) => log('entity.update.resolve
 /////////////////////////////////////////////////////////////////////////
 // Processing submission events to create and update entities
 
+const _getFormDefActions = async (oneFirst, datasetId, submissionDefId) => oneFirst(sql`
```
matthew-white, line 151: Pass this the `formDefId` from the `submissionDef`, then remove the `JOIN`.

test/unit/data/dataset.js
```diff
@@ -53,7 +53,7 @@ describe('parsing dataset from entity block', () => {
                 <name/>
                 <age/>
                 <meta>
-                  <entities:entity dataset="foo">
+                  <entities:entity dataset="foo" create="">
```
matthew-white, line 56: Revert.

test/unit/data/dataset.js
```diff
@@ -162,40 +165,71 @@ describe('parsing dataset from entity block', () => {
...
-              </model>
+          <model entities:entities-version="2022.1.0">
+            <instance>
+              <data id="NoName">
+                <name/>
+                <age/>
+                <meta>
+                  <entities:entity create="">
+                    <label/>
+                  <entities:entity>
```
matthew-white, line 188:
```suggestion
                  </entities:entity>
```

test/unit/data/dataset.js
```diff
@@ -162,40 +165,71 @@ describe('parsing dataset from entity block', () => {
...
+      result.get().actions.should.eql(['create', 'update']);
+    });
+
+    it('should strip namespaces from action attributes', async () => {
+      const result = await getDataset(testData.forms.updateEntity
+        .replace('update=""', 'entities:create="" entities:update=""'));
+      result.get().actions.should.eql(['create', 'update']);
+    });
+
+    it('should reject for an entity form without an action', () => {
```
matthew-white, line 225: Instead of rejecting, `actions` should be an empty array.

Let's also add an integration test that a submission is rejected if it attempts a create for a form that doesn't permit any entity actions.

ktuite (approved):

matthew-white (commented):
Notes from interactive code review

lib/data/dataset.js
matthew-white: It actually turns out that the XForm spec requires at least one action attribute, so we'll continue to reject in that case.

PR #1066, opened 2023-12-14
bug #930: added email property check, ensuring the correct response.

### LINKED ISSUE #930: Return better error message when email is not specified
It looks like /v1/users/reset/initiate will return a 500 error if the request body does not have an `email` property. It'd be better if Central returned a 400 error with a more informative message.

We saw such a request in a deployment of Central v2023.3.1. As to why the request body didn't have an `email` property, it looks like the `Content-Type` header of the request was `multipart/form-data`, not `application/json`. The following error was thrown:

```
InvalidInputError: SQL tag cannot be bound an undefined value.
  File "/usr/odk/node_modules/slonik/dist/src/factories/createSqlTag.js", line 27, col 19, in sql
    throw new errors_1.InvalidInputError('SQL tag cannot be bound an undefined value.');
  File "/usr/odk/lib/util/db.js", line 218, col 12, in equals
    : sql`${sql.identifier(k.split('.'))}=${obj[k]}`;
  File "/usr/odk/lib/model/query/users.js", line 71, col 9, in _getSql
    where ${equals(options.condition)} and actors."deletedAt" is null
  File "/usr/odk/lib/model/query/users.js", line 79, col 12, in <anonymous>
    maybeOne(_getSql(options.withCondition({ email }))).then(map(_unjoin));
  File "/usr/odk/lib/model/container.js", line 30, col 33, in Object.module.<computed> [as getByEmail]
    const result = fn(...args)(container);
  File "/usr/odk/lib/resources/users.js", line 47, col 11, in <anonymous>
    Users.getByEmail(body.email)
  File "/usr/odk/lib/http/endpoint.js", line 159, col 26, in <anonymous>
    .then((context) => resource(localContainer.with({ context }), context, request, response))
  File "/usr/odk/node_modules/slonik/dist/src/factories/createSqlTag.js", in runMicrotasks
  File "node:internal/process/task_queues", line 96, col 5, in processTicksAndRejections
  File "/usr/odk/node_modules/slonik/dist/src/connectionMethods/transaction.js", line 22, col 24, in Object.transaction
    const result = await handler(binders_1.bindTransactionConnection(log, connection, clientConfiguration, connection.connection.slonik. {snip}
  File "/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js", line 97, col 18, in Object.createConnection
    result = await connectionHandler(connectionLog, connection, boundConnection, clientConfiguration);
```

dev-rahulbhadoriya:
Closes #930

Added a check for the email field. If the property is missing or not included in the request body, an error 400.23 ( propertyNotFound) will be returned. 

```javascript
{
    "message": "Could not find a property named 'email'",
    "code": 400.23,
    "details": {
        "property": "email"
    }
}
``` 

 Tested with postman.


matthew-white:
Thanks, @dev-rahulbhadoriya! I have a couple of small suggestions. First, I think we should use `Problem.user.missingParameter` instead of `propertyNotFound`. Currently, we only use `propertyNotFound` when validating the `$select` OData query parameter, where the Problem means "you specified a property that doesn't exist." In other cases, it's more common that we use `missingParameter`, which has more of a meaning of "you didn't specify a property that we expected."

It'd be great if you could also add a test to test/integration/api/users.js for the new behavior. The test could send a request without `email`, then assert that the resulting response body has correct values for `code` and `details`.

Also, just FYI, the project is going to be quieter for the next couple of weeks. You may find that additional review is delayed until 2024. Thanks again for submitting this PR!

dev-rahulbhadoriya:
@matthew-white , thanks for the suggestions and for reviewing my PR. I have corrected the code and added a test case for it. Kindly check and review it

test/integration/api/users.js
```diff
@@ -402,6 +402,12 @@ describe('api: /users', () => {
             asAlice.post('/v1/users/reset/verify')
               .send({ new: 'coolpassword' })
               .expect(403))));
+
+        it('should fail the request if email field is sent blank in request body', testService((service) =>
+          service.login('alice', (asAlice) =>
+            asAlice.post('/v1/users/reset/initiate')
+              .send({ email: '' })
+              .expect(400))));
```
matthew-white, line 410: Would you mind also adding assertions about the response body? `code` should equal 400.2, and `details` should equal `{ field: 'email' }`.

dev-rahulbhadoriya: can do something like this ?
`it('should fail the request if email field is sent blank in request body', testService((service) =>
            asAlice.post('/v1/users/reset/initiate')
              .send({ email: '' })
              .expect(400)
              .then(({ body: { code, details } }) => {
                details.should.eql({ field: 'email' });
                code.should.eql(400.2);
              }))));`

matthew-white (approved):
Sorry for the long delay on review! This is looking great. I'll go ahead and merge now. Thank you for the PR, @dev-rahulbhadoriya!

PR #1070, opened 2024-01-11
Bulk entity creation via API

ktuite:
Closes https://github.com/getodk/central/issues/573

Uses POSTing to existing `/v1/projects/1/datasets/people/entities` endpoint so it can handle
* single entities with just `{uuid, label, data}`
* multiple entities with `{entities: [{}, {}], source: {name, size}}`

For the source object,
* `source` is required
* `name` within source is required (not sure if it should be but it will help with displaying on the front end...)
* `size` (meant to represent file size) 

I was thinking about how to handle the bulk SQL insert and decided it was easiest to think/build two separate queries, one for the `entities` table and another for the `entity_defs` table. Elsewhere the code, we have complex queries for inserting a single entity's data into both tables at once, but it didn't seem reasonable to try to do that with multiple entities. 

Tests, trying it out.

Documentation is included in this PR!!

### DIFF
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -36,7 +36,7 @@ const actionCondition = (action) => {
     // The backup action was logged by a backup script that has been removed.
     // Even though the script has been removed, the audit log entries it logged
     // have not, so we should continue to exclude those.
-    return sql`action not in ('entity.create', 'entity.error', 'entity.update.version', 'entity.update.resolve', 'entity.delete', 'submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'analytics')`;
+    return sql`action not in ('entity.create', 'entity.bulk.create', 'entity.error', 'entity.update.version', 'entity.update.resolve', 'entity.delete', 'submission.create', 'submission.update', 'submission.update.version', 'submission.attachment.update', 'backup', 'analytics')`;
   else if (action === 'user')
     return sql`action in ('user.create', 'user.update', 'user.delete', 'user.assignment.create', 'user.assignment.delete', 'user.session.create')`;
   else if (action === 'field_key')
@@ -52,7 +52,7 @@ const actionCondition = (action) => {
   else if (action === 'dataset')
     return sql`action in ('dataset.create', 'dataset.update')`;
   else if (action === 'entity')
-    return sql`action in ('entity.create', 'entity.error', 'entity.update.version', 'entity.update.resolve', 'entity.delete')`;
+    return sql`action in ('entity.create', 'entity.bulk.create', 'entity.error', 'entity.update.version', 'entity.update.resolve', 'entity.delete')`;
 
   return sql`action=${action}`;
 };
@@ -136,7 +136,7 @@ const _getByEntityId = (fields, options, entityId) => sql`
 SELECT ${fields} FROM entity_defs
 
   LEFT JOIN entity_def_sources on entity_def_sources.id = entity_defs."sourceId"
-  LEFT JOIN audits ON ((audits.details->'entityDefId')::INTEGER = entity_defs.id OR (audits.details->'sourceId')::INTEGER = entity_def_sources.id)
+  INNER JOIN audits ON ((audits.details->'entityDefId')::INTEGER = entity_defs.id OR (audits.details->'sourceId')::INTEGER = entity_def_sources.id)
 
   LEFT JOIN actors ON actors.id=audits."actorId"
 
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -133,11 +133,12 @@ const getBySubmissionId = (submissionId, options) => ({ all }) => _getBySubmissi
 
 
 const _getByEntityId = (fields, options, entityId) => sql`
-SELECT ${fields} FROM audits
-  LEFT JOIN actors ON actors.id=audits."actorId"
+SELECT ${fields} FROM entity_defs
 
-  LEFT JOIN entity_defs ON (audits.details->'entityDefId')::INTEGER = entity_defs.id
   LEFT JOIN entity_def_sources on entity_def_sources.id = entity_defs."sourceId"
+  LEFT JOIN audits ON ((audits.details->'entityDefId')::INTEGER = entity_defs.id OR (audits.details->'sourceId')::INTEGER = entity_def_sources.id)
+
+  LEFT JOIN actors ON actors.id=audits."actorId"
 
   LEFT JOIN audits triggering_event ON entity_def_sources."auditId" = triggering_event.id
   LEFT JOIN actors triggering_event_actor ON triggering_event_actor.id = triggering_event."actorId"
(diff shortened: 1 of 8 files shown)


lib/model/query/entities.js
```diff
@@ -86,6 +87,49 @@ createNew.audit = (newEntity, dataset, partial, subDef) => (log) => {
...
+  return defs;
+};
+
+const createMany = (dataset, entities, sourceId, userAgentIn) => (container) =>
+  container.db.transaction((trxn) =>
+    container.with({ db: trxn }).Entities._createMany(dataset, entities, sourceId, userAgentIn))
+    .catch((err) => reject(err));
+
+createMany.audit = (dataset, entities, sourceId) => (log) =>
+  log('entity.bulk.create', dataset, { sourceId, count: entities.length });
```
ktuite, line 130: I think this should be about an entity (rather than a dataset) because while it does change the overall dataset by growing the dataset, it is more about entities. Also, it is the dataset actee ID, so there is already that link. 

Maybe it should be `entity.create.bulk` instead of `entity.bulk.create`?

lib/resources/entities.js
```diff
@@ -78,21 +78,34 @@ module.exports = (service, endpoint) => {
...
+    if (!entities) {
+      // not a bulk operation
+      const partial = await Entity.fromJson(body, properties, dataset);
+      const sourceId = await Entities.createSource();
+      const entity = await Entities.createNew(dataset, partial, null, sourceId, userAgent);
+      // Entities.createNew doesn't return enough information for a full response so re-fetch.
+      return Entities.getById(dataset.id, entity.uuid).then(getOrNotFound);
+    } else {
+      // bulk operation
+      // TODO: maybe validate source fields
```
ktuite, line 102: I think we may need to revisit this (how the source of a bulk action is represented, if/how it should be validated) when we get to this issue: https://github.com/getodk/central/issues/583

lib/model/query/audits.js
```diff
@@ -214,7 +216,8 @@ const getByEntityId = (entityId, options) => ({ all }) => {
 
       const details = mergeLeft(audit.details, sourceEvent
         .map(event => ({ source: { event, submission } }))
-        .orElse({ source: {} })); // Add default empty source to all other entity audit events
+        .orElse({ source: entitySourceDetails ?? {} })); // Add details from entity def source, or add default empty source to all other entity audit events
```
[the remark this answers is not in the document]
ktuite: This source I think always represents the version source, rather than the audit event source.

test/integration/api/entities.js
```diff
@@ -2081,6 +2081,528 @@ describe('Entities API', () => {
...
+    // functionality of an existing endpoint:
+    // - should return notfound if the dataset does not exist
+    // - should reject if the user cannot write
+    // - should reject creating new entity if dataset not yet published
+
+    it('should reject malformed json', testDataset(async (service) => {
+      const asAlice = await service.login('alice');
+
+      await asAlice.post('/v1/projects/1/datasets/people/entities')
+        .send([{ broken: 'json' }])
```
ktuite, line 2096: Change to test something like `entities: [{ broken: 'json' }]`

test/integration/api/entities.js
```diff
@@ -2081,6 +2081,528 @@ describe('Entities API', () => {
...
+      await asAlice.get('/v1/projects/1/datasets/people/entities')
+        .then(({ body }) => {
+          body.length.should.equal(2);
+        });
+
+      // Most recent event IS bulk create
+      await asAlice.get('/v1/audits')
+        .then(({ body }) => {
+          body[0].action.should.equal('entity.bulk.create');
+          body[0].details.count.should.equal(2);
```
ktuite, line 2151: test source details more thoroughly

test/integration/api/entities.js
```diff
@@ -2081,6 +2081,528 @@ describe('Entities API', () => {
...
+
+        // resolve conflict without making new version
+        await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?resolve=true&force=true')
+          .expect(200);
+
+        await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc/versions')
+          .expect(200)
+          .then(({ body: versions }) => {
+            versions.length.should.equal(3);
+            //console.log('versions', versions);
```
ktuite, line 2526: rm

test/integration/api/entities.js
```diff
@@ -2081,6 +2081,528 @@ describe('Entities API', () => {
...
+
+        // third version: hard conflict - all properties are changed
+        await asAlice.post('/v1/projects/1/forms/updateEntity/submissions')
+          .send(testData.instances.updateEntity.one)
+          .set('Content-Type', 'application/xml')
+          .expect(200);
+
+        await exhaust(container);
+
+        // resolve conflict without making new version
```
ktuite, line 2566: comment should be we are making a new version

test/integration/api/entities.js
ktuite: what if you dont pass an array to entities 
what if the data within one of these things is not an object

test/integration/api/entities.js
```diff
@@ -2081,6 +2081,528 @@ describe('Entities API', () => {
...
+        })
+        .expect(200)
+        .then(({ body }) => {
+          body.success.should.be.true();
+        });
+
+      // Entity list has more entities!
+      await asAlice.get('/v1/projects/1/datasets/people/entities')
+        .then(({ body }) => {
+          body.length.should.equal(2);
```
ktuite, line 2144: check contents of entities?
check data and data received (different path that database isn't enforcing)

matthew-white (approved):

ktuite (commented):

lib/model/query/audits.js
```diff
@@ -133,11 +133,12 @@ const getBySubmissionId = (submissionId, options) => ({ all }) => _getBySubmissi
 
 
 const _getByEntityId = (fields, options, entityId) => sql`
-SELECT ${fields} FROM audits
-  LEFT JOIN actors ON actors.id=audits."actorId"
+SELECT ${fields} FROM entity_defs
 
-  LEFT JOIN entity_defs ON (audits.details->'entityDefId')::INTEGER = entity_defs.id
   LEFT JOIN entity_def_sources on entity_def_sources.id = entity_defs."sourceId"
+  LEFT JOIN audits ON ((audits.details->'entityDefId')::INTEGER = entity_defs.id OR (audits.details->'sourceId')::INTEGER = entity_def_sources.id)
```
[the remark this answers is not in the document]
ktuite: This entity source id is only used in the query for audits for a single entity and i don't know if that justifies adding another index to this whole audit table (even though it is what we did for the entity id). I want to wait and add this index later if it seems necessary.

lib/model/query/entities.js
```diff
@@ -86,6 +86,49 @@ createNew.audit = (newEntity, dataset, partial, subDef) => (log) => {
...
 createNew.audit.withResult = true;
 
+const _createMany = (dataset, entities, sourceId, userAgentIn) => async ({ all, context }) => {
+  const creatorId = context.auth.actor.map((actor) => actor.id).orNull();
+  const userAgent = blankStringToNull(userAgentIn);
+
+  const entityInsert = sql.join(entities.map(e => sql`(${sql.join([dataset.id, e.uuid, creatorId, sql`clock_timestamp()`], sql`,`)} )`), sql`,`);
+  const newEntities = await all(sql`
+  INSERT INTO entities ("datasetId", "uuid", "creatorId", "createdAt")
+  VALUES ${entityInsert}
```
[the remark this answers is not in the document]
ktuite: Felt blocked on looking at this in depth so I made a separate issue: https://github.com/getodk/central-backend/issues/1095 

That issue can be used to track any other performance issues/limitations we think of regarding bulk entity uploads.

PR #1071, opened 2024-01-13
Support S3-compatible blob storage

alxndrsn:
Closes https://github.com/getodk/central/issues/585

Adds:

* `s3_status` column on `blobs` table
* CLI tool for:
  * uploading blobs to s3
  * resetting failed blobs -> pending

Changes:

* `blob.content` uses must now be accompanied by a conditional fetch of the data from s3

Queries:

* would it be preferable to make `blob.content` `async` everywhere, and moving the s3 check/fetch inside the getter?
  * pro: neater code, less likely to have bugs where `blob.content` used when data is not available
  * con: seems intrusive of frames
* should "purging" forms in central also delete content from S3? *yes*


.github/workflows/soak-test.yml
```diff
@@ -26,10 +23,10 @@ jobs:
           --health-retries 5
     steps:
     - uses: actions/checkout@v3
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js 20
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: 20.10.0
```
alxndrsn, line 29: See https://github.com/getodk/central-backend/pull/1080

alxndrsn:
Tests which touch blobs:

- should not be able access closed forms and its sub-resources
- should filter out submission and backup events given action=nonverbose
- should override blob and link dataset
- should not link dataset if previous version has blob
- should not autolink if attachment already filled in
- should throw problem if blobId and datasetId are being set
- should not throw problem for fast external itemsets when there is existing "itemsets" dataset
- should not throw problem for new version of "fast external itemsets" when there is existing "itemsets" dataset
- should accept xlsx drafts
- should ignore xlsx warnings if asked
- should deal with xlsx itemsets
- should pick up xlsx itemsets when newly required
- should pick up new xlsx itemsets replacing the old one
- should carry forward matching published attachments
- should carry forward matching draft attachments
- should provide attachments as expected
- should return a manifest
- should accept the file with a success result
- should accept xml type files
- should replace an extant file with another
- should allow the same file in different slots
- should log the action in the audit log
- should reject unless the user may update the form
- should delete the attachment contents
- should log the action in the audit log
- should accept and convert valid xlsx files
- should fail on warnings even for valid xlsx files
- should create the form for xlsx files with warnings given ignoreWarnings
- should apply itemsets.csv if it is returned and expected
- should ignore itemsets.csv if it is returned but not expected
- should return an appropriate response upon conversion error
- should accept xls forms and return them over that endpoint
- should reject with xls and deleted form exists warnings
- should reject with xls and structure changed warnings
- should create the form for xml files with warnings given ignoreWarnings
- should reject if the user cannot read
- should return xls notfound given xlsx file
- should return the xlsx file originally provided
- should return the xlsx file originally provided
- should continue to offer the xlsx file after a copy-draft
- should not continue to offer the xlsx file after a noncopy-draft
- should allow xls file download only
- should return xls content type with extended form details
- should include attachments that have been uploaded
- should flag exists: true for extant files
- should return upload updatedAt for extended metadata
- should return deletion exists and updatedAt for extended metadata
- should reject unless the user may read the form
- should return file contents with appropriate headers
- should return 304 content not changed if ETag matches
- should return latest content and correct ETag
- should reject if the draft has been published
- should reject if the draft has been deleted
- should reject if the key is wrong
- should return a manifest with present attachments
- should reject if the draft has been published
- should reject if the draft has been deleted
- should reject if the key is wrong
- should return the attachment
- should return xls content type with extended form details
- should return the xlsx file originally provided
- should return a manifest
- should return a list of attachments
- should return an attachment
- should not make additional form schemas for a form with unchanged fields
- should make additional form schema when fields change
- should return a single row result
- should return success if "uuid:" prefix is encoded
- should return an accurate edit count
- should return a single encrypted frame (no formdata)
- should return a single encrypted frame (has formdata)
- should return subtable results
- should limit and offset subtable results
- should return just a count if asked
- should gracefully degrade on encrypted subtables
- should return encoded URLs
- should return a single row result with all properties
- should return a single row result with selected properties
- should return a single row result with all system properties
- should return subtable results with selected properties
- should not return parent IDs if __id is selected
- should return only parent ID
- should return toplevel rows
- should return subtable results
- should limit and offset toplevel rows
- should limit Submissions
- should ignore $skip when $skipToken is given
- should have no impact on skipToken when a new submission is created
- should limit and filter Submissions
- should limit and return selected fields of Submissions
- should provide toplevel row count if requested
- should return submitter-filtered toplevel rows if requested
- should return submissionDate-filtered toplevel rows if requested
- should only return submissions with submissionDate gt provided timestamp
- should return submissionDate-filtered toplevel rows with a function
- should return updatedAt-filtered toplevel rows if requested
- should return reviewState-filtered toplevel rows if requested
- should filter toplevel rows by $root expression
- should count correctly while windowing
- should count correctly while filtering
- should count correctly while filtering and windowing
- should return encrypted frames (no formdata)
- should return encrypted frames (has formdata)
- should limit and offset subtable results
- should filter subtable results
- should filter and paginate subtable results
- should count correctly while filtering and windowing subtable
- should limit subtable results
- should limit and filter subtable
- should gracefully degrade on encrypted subtables
- should return toplevel rows with selected properties
- should return toplevel rows with group properties
- should return toplevel row with nested group properties
- should return subtable results with selected properties
- should return only parent ID
- should return subtable results with group properties
- should return results even when repeat name is not a valid OData name 
- should return submissions in specified order
- should combine orderby and other things like filtering
- should return null values at the correct end of list with orderby
- should reject if both orderby and skiptoken are used together
- should return a single row result
- should return toplevel rows
- should not be able access closed forms and its sub-resources
- should save given attachments
- should save attachments with unicode / non-english char
- should not fail given identical attachments
- should not fail given identical attachment references
- should create audit log entries for saved attachments
- should save client audit log attachments
- should take in additional attachments via additional POSTs
- should successfully save attachment binary data
- should successfully save additionally POSTed attachment binary data
- should accept encrypted submissions, with attachments
- should accept the new submission
- should copy forward missing attachments
- should replace extant attachments
- should upsert attachments
- should save client audit log attachments
- should copy forward matching attachments
- should pass the appropriate parameters to the enketo module
- should return a zipfile with the relevant data
- should include all repeat rows @slow
- should not include data from other forms
- should return a submitter-filtered zipfile with the relevant data
- should return a review state-filtered zipfile with the relevant data
- should return a submissionDate-filtered zipfile with the relevant data
- should return an updatedAt-filtered zipfile with the relevant data
- should return a zipfile with the relevant attachments
- should filter attachments by the query
- should list the original submitted form version per submission
- should split select multiple values if ?splitSelectMultiples=true
- should omit multiples it does not know about
- should split select multiples and filter given both options
- should export deleted fields and values if ?deletedFields=true
- should skip attachments if ?attachments=false is given
- should give the appropriate filename if ?attachments=false is given
- should omit group paths if ?groupPaths=false is given
- should split select AND omit group paths given both options
- should properly count present attachments
- should return worker-processed consolidated client audit log attachments
- should return adhoc-processed consolidated client audit log attachments
- should return consolidated client audit log filtered by user
- should return consolidated client audit log filtered by review state
- should return the latest attached audit log after openrosa replace
- should return the latest attached audit log after REST replace
- should tolerate differences in line lengths
- should tolerate quote inside unquoted field of client audit log
- should return original instanceId and latest attached audit log when instanceId deprecated with new audit log
- should return the root csv table
- should return only the root csv table given repeats
- should split select multiple values if ?splitSelectMultiples=true
- should export deleted fields and values if ?deletedFields=true
- should split select multiple values if ?splitSelectMultiples=true
- should return draft submissions
- should not include draft submissions in nondraft csvzip
- should split select multiple values submitted over /test/ if ?splitSelectMultiples=true
- should return attachment information
- should return notfound if the attachment does not exist
- should reject if the user cannot read
- should return attachment data
- should return a draft attachment
- should not return a draft attachment nondraft
- should return notfound if the form does not exist
- should return notfound if the submission does not exist
- should reject if the user cannot update a submission
- should reject if the attachment does not exist
- should attach the given file
- should log an audit entry about initial attachment
- should log an audit entry about reattachment
- should clear the given attachment
- should log an audit entry about the deletion
- should delete a draft attachment
- should not delete a draft attachment nondraft
- should count failed audits
- should count the total number of client audit submission attachments
- should count client audit attachments that failed processing
- should count the number of rows extracted from client audit attachments
- should calculate forms with audits
- should combine system level queries
- should fill in all project.forms queries
- should not purge xls blob that is still referenced
- should handle blob collisions and not purge still attached blobs
- should show submissions in odata
- should show submissions in odata when label added
- should reject with a reasonable message given incorrect passphrase
- should decrypt to CSV successfully
- should decrypt to CSV successfully as a direct root table
- should decrypt with passphrases provided via url-encoded POST body
- should decrypt over cookie auth with passphrases provided via url-encoded POST body
- should decrypt with passphrases provide via JSON POST body
- should decrypt attached files successfully
- should strip .enc suffix from decrypted attachments
- should decrypt client audit log attachments
- should decrypt client audit log attachments
- should handle mixed [plaintext/encrypted] attachments (not decrypting)
- should handle mixed [plaintext/encrypted] attachments (decrypting)
- should handle mixed[plaintext/encrypted] formdata (decrypting)
- should handle mixed[plaintext/encrypted] formdata (not decrypting)
- should handle mixed[managedA/managedB] formdata (decrypting)
- should handle mixed [plaintext/missing-encrypted-xml] formdata (decrypting)
- should handle mixed [plaintext/missing-encrypted-xml] formdata (not decrypting)
- should purge attachments (and blobs) of a form
- should purge attachments and blobs associated with the submission
- should purge client audit log attachments
- should purge xls blob of a form
- should copy forward matching attachments
- should not copy forward nonmatching attachments
- should not be able access closed forms and its sub-resources
- should be able to list form submissions
- should be able to download submissions
- should be able to get submission detail
- should be able to fetch a submission attachment
- should handle when field changes from select1 to select (selectMultiple)
- should do nothing if the attachment is not a client audit
- should process the given logs
- should not reprocess already-processed logs

alxndrsn:
### Identified uses

- [ ] /home/user/workspaces/odk/backend/lib/model/query/form-attachments.js:36:16
- [ ] /home/user/workspaces/odk/backend/lib/worker/submission.attachment.update.js:30:36
- [ ] Transform.transform [as _transform] (/home/user/workspaces/odk/backend/lib/data/briefcase.js:328:44)
- [ ] _makeAttachment (/home/user/workspaces/odk/backend/lib/model/query/submission-attachments.js:31:7)
- [ ] /home/user/workspaces/odk/backend/lib/util/blob-response.js:26:21

### Unidentified uses

```
get content [as content] (/home/user/workspaces/odk/backend/lib/model/frames/blob.js:49:11)
    at /home/user/workspaces/odk/backend/lib/model/query/blobs.js:25:50
    at module.<computed> [as ensure] (/home/user/workspaces/odk/backend/lib/model/container.js:30:33)
    at /home/user/workspaces/odk/backend/lib/model/query/forms.js:40:53
```

```
get content [as content] (/home/user/workspaces/odk/backend/lib/model/frames/blob.js:49:11)
    at /home/user/workspaces/odk/backend/lib/model/query/blobs.js:25:50
    at module.<computed> (/home/user/workspaces/odk/backend/lib/model/container.js:30:33)
    at async Promise.all (index 1)
```

```
get content [as content] (/home/user/workspaces/odk/backend/lib/model/frames/blob.js:49:11)
    at /home/user/workspaces/odk/backend/lib/model/query/blobs.js:25:50
    at module.<computed> (/home/user/workspaces/odk/backend/lib/model/container.js:30:33)
    at async Promise.all (index 2)
    at async Object.transaction (/home/user/workspaces/odk/backend/node_modules/slonik/dist/src/connectionMethods/transaction.js:22:24)
    at async Object.createConnection (/home/user/workspaces/odk/backend/node_modules/slonik/dist/src/factories/createConnection.js:97:18)
```

alxndrsn:
### Identified Uses

- [ ] ./lib/model/query/form-attachments.js:36:16
- [ ] ./lib/worker/submission.attachment.update.js:30:36
- [ ] Transform.transform [as _transform] (./lib/data/briefcase.js:328:44)
- [ ] _makeAttachment (./lib/model/query/submission-attachments.js:31:7)
- [ ] ./lib/util/blob-response.js:26:21

### Unidentified Uses

```
get content [as content] (./lib/model/frames/blob.js:49:11)
    at ./lib/model/query/blobs.js:25:50
    at module.<computed> [as ensure] (./lib/model/container.js:30:33)
    at ./lib/model/query/forms.js:40:53
```

```
get content [as content] (./lib/model/frames/blob.js:49:11)
    at ./lib/model/query/blobs.js:25:50
    at module.<computed> (./lib/model/container.js:30:33)
    at async Promise.all (index 1)
```

```
get content [as content] (./lib/model/frames/blob.js:49:11)
    at ./lib/model/query/blobs.js:25:50
    at module.<computed> (./lib/model/container.js:30:33)
    at async Promise.all (index 2)
    at async Object.transaction (./node_modules/slonik/dist/src/connectionMethods/transaction.js:22:24)
    at async Object.createConnection (./node_modules/slonik/dist/src/factories/createConnection.js:97:18)
```

ktuite (commented):
I have a lot to look at still, but I want to summarize the shape I see:

Blob touchpoints (with a little bit of code for each case because the generic blob table is joined with other types of tables):
- submission attachments
- encrypted submissions
- form xls files
- client audits (subset of submission attachments) which are processed by a worker
- that briefcase file

The function `blobResponse` that serves blobs modified to be able to serve from s3 if not in the database.

`s3_status` added to blobs table (pending, in_progress, uploaded, failed)

Simple worker defined in `util/s3.js` that is started in `runServer.js`
This is where the blob uploading happens!
Worker doesn't retry any failures, that is handled by an external cli tool. 

`exhaustBlobs` function for tests and for cli script to use.

Am I missing anything big?


I haven't yet looked at:
- the details of the worker
- the different blob scenarios/touchpoints
- tests (changes to existing tests, or e2e tests)

lib/util/http.js
```diff
@@ -144,28 +138,23 @@ const url = (strings, ...parts) => {
 
 // Checks Etag in the request if it matches serverEtag then returns 304
 // Otherwise executes given function 'fn'
-const withEtag = (serverEtag, fn) => (request, response) => {
-
-  response.set('ETag', `"${serverEtag}"`);
+const withEtag = (serverEtag, fn, always=true) => (request, response) => {
+  if (always) response.set('ETag', `"${serverEtag}"`);
```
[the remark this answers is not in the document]
alxndrsn: I've tried to clarify by:

1. adding a comment about `always`, and
2. splitting the exported `withEtag()` function to add an additional `withEtagOnMatch()`

Does this help?  Perhaps `withEtagOrRedirect()` would be a clearer name for the new function?

lib/bin/s3-create-bucket.js
```diff
@@ -0,0 +1,39 @@
+// Copyright 2024 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
```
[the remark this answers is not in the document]
alxndrsn: It's used indirectly by developers in the `dev-s3` and `test-s3-integration` make targets.

For me the ideal would be to put scripts which are not for end-user use somewhere else which communicates this, e.g. in `test/scripts`, or `lib/bin/private`.

There are currently a few of these, e.g.

* `check-file-headers.js`
* `create-docker-databases.js`
* `enforce-node-version.js`
* `check-open-db-queries.js`

lib/util/http.js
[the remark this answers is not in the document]
ktuite: I mostly follow but now I'm confused by the `!always` line below.  I saw online that a 304 not modified request should always be sending back the etag in the header anyway. 

But also in the s3 redirect case, `always` will be false, so:
- dont set etag
- check client etag
- if it matches, DO set etag and return 304 not modified

Does s3 make its own etag? Will these ever match and would we ever return a 304 in place of a 307?

lib/data/attachments.js
```diff
@@ -33,14 +34,24 @@ const streamAttachments = (inStream, decryptor) => {
...
+          : decryptor(rawContent, att.keyId, att.localKey, att.instanceId, att.index);
 
-      archive.append(content, { name: join('media', processName(att.name)) }, done);
+        archive.append(content, { name: join('media', processName(att.name)) }, done);
+      };
+
+      if (att.s3_status === 'uploaded') {
+        s3.getContentFor(att)
+          .then(appendContent)
+          .catch(err => { this.destroy(err); });
```
ktuite, line 51: I see what you mean about this query: 
> would it be preferable to make blob.content async everywhere, and moving the s3 check/fetch inside the getter?
> * pro: neater code, less likely to have bugs where blob.content used when data is not available
> * con: seems intrusive of frames

(Just highlighting this one because it seemed like the smallest example.)

Maybe it's okay to leave it like this since there are only about 4 places blobs are used (unless you have an idea for some clever alternative):
1. exporting attachments (here)
2. exporting encrypting submissions that were stored as blobs
3. processing client audit attachments
4. (different mechanism, `urlForBlob` instead of `getContentFor` line the ones above) the http blob response 

As long as each of these 4 paths is tested, since each might have a different behavior if there is an error fetching from S3.

Some errors i'm thinking of:
- auth problem
- data doesn't exist where it should
- response from s3 is taking too long
- data is broken in some other way?

ktuite (commented):
I'm going through this bit by bit (while a bigger discussions also take place). 

I think I now have a good understanding of the part of the system that *uses* blobs that may or may not be on S3.

Next up, I'll probably take a closer look s3.js and how the blobs actually get to and from s3.

lib/util/http.js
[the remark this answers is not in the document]
alxndrsn: > Does s3 make its own etag? Will these ever match and would we ever return a 304 in place of a 307?

Great question.  If a client sees an odk-central URL for an s3-backed blob, odk-central needs to ensure that the final ETag provided to the user is the same regardless of the binary data source.

This means either both systems need to generate the same ETag for the same data, or odk-central needs to substitute its ETag in place of S3's.

Currently:

* odk-central's nginx should be transparently following 307 redirects for s3-backed content
* both S3 and odk-central-backend use MD5(blob-content) as a strong ETag

This means that the _current_ implementation should work fine.  As there are no end-to-end central tests running both odk-central-backend and nginx, there's a risk this could break in future.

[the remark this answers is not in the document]
alxndrsn: > What's this `always` flag about?

I've reversed this flag, renaming it as `onlyOnMatch`, and updated the comments.  Is the intention clearer now?

Makefile
```diff
@@ -16,6 +16,14 @@ test-oidc-e2e: node_version
...
 	NODE_CONFIG_ENV=oidc-development npx nodemon --watch lib --watch config lib/bin/run-server.js
 
+.PHONY: fake-s3-accounts
+fake-s3-accounts: node_version
+	NODE_CONFIG_ENV=s3-dev node lib/bin/s3-create-bucket.js
+
+.PHONY: dev-s3
+dev-s3: fake-s3-accounts base
+	NODE_CONFIG_ENV=s3-dev npx nodemon --watch lib --watch config lib/bin/run-server.js
+
```
alxndrsn, line 26: Group all s3 targets together.

.github/workflows/s3-e2e.yml
```diff
@@ -0,0 +1,54 @@
...
+        node-version: 20.10.0
+        cache: 'npm'
+    - run: npm ci --legacy-peer-deps
+    - run: node lib/bin/create-docker-databases.js
+    - name: E2E Test
+      timeout-minutes: 10
+      run: ./test/e2e/s3/ci
+    - name: Backend Logs
+      if: always()
+      run: "! [[ -f ./backend.log ]] || cat ./backend.log"
```
alxndrsn, line 54: Maybe just send all the backend logs to stdout/stderr?

lib/bin/s3-create-bucket.js
```diff
@@ -0,0 +1,39 @@
...
+const Minio = require('minio');
+
+const { server, bucketName, accessKey, secretKey } = require('config').get('default').s3blobStore;
+
```
alxndrsn, line 13: Can this all be moved back to the `Makefile`?  This might make it less tempting for end-users, and should make it clearer that an ideal S3 production credential would not have authority to create buckets.

lib/model/migrations/20240311-01-add-blob-s3.js
```diff
@@ -0,0 +1,29 @@
...
+  `);
+};
+
+const down = async (db) => {
+  await db.raw(`
+    ALTER TABLE blobs
+      DROP COLUMN s3_status,
+      ALTER COLUMN content SET NOT NULL
+  `);
+  await db.raw('DELETE TYPE S3_UPLOAD_STATUS');
```
ktuite, line 26: I think this should be DROP TYPE (I tried to roll back this migration once and it didn't work)

ktuite (commented):
Comments from review today (more review tomorrow!)

lib/bin/s3-create-bucket.js
alxndrsn: There's not obviously a simpler way to do this which supports both CI & local, and doesn't require more dependencies.

Discussed at:

* https://github.com/minio/minio/issues/4769
* https://stackoverflow.com/questions/71827591/how-to-create-default-bucket-for-minio-within-github-actions-using-minioedge-ci
etc

.github/workflows/s3-e2e.yml
```diff
@@ -0,0 +1,51 @@
+name: S3 E2E Tests
+
+on: push
+
+env:
+  LOG_LEVEL: DEBUG
```
alxndrsn, line 6: This can go

test/unit/data/attachments.js
```diff
@@ -5,74 +5,80 @@ const { streamAttachments } = require(appRoot + '/lib/data/attachments');
...
-    const inStream = streamTest.fromObjects([
-      { row: { instanceId: 'subone', name: 'firstfile.ext', content: 'this is my first file' } },
-      { row: { instanceId: 'subone', name: 'secondfile.ext', content: 'this is my second file' } },
-      { row: { instanceId: 'subtwo', name: 'thirdfile.ext', content: 'this is my third file' } }
-    ]);
-    zipStreamToFiles(zipStreamFromParts(streamAttachments(inStream)), (err, result) => {
-      // eslint-disable-next-line keyword-spacing
-      if(err) return done(err);
+  [false, true].forEach(s3enabled => {
+    // REVIEW s3 enabled/disabled doesn't make any difference to these tests
```
alxndrsn, line 9: Revert this change, and add a new test which includes uploaded blobs

test/unit/data/briefcase.js
```diff
@@ -7,7 +7,6 @@ const { fieldsFor } = require(appRoot + '/test/util/schema');
 const { streamBriefcaseCsvs } = require(appRoot + '/lib/data/briefcase');
 const { zipStreamFromParts } = require(appRoot + '/lib/util/zip');
 
-
```
alxndrsn, line 10: revert this change

test/unit/data/briefcase.js
```diff
@@ -24,1126 +23,1131 @@ const withSubmitter = (id, displayName, row) => ({ ...row, aux: { ...row.aux, su
...
-    ]);
-
-    callAndParse(inStream, formXml, 'mytestform', (err, result) => {
-      // eslint-disable-next-line keyword-spacing
-      if(err) return done(err);
-
-      result.filenames.should.eql([ 'mytestform.csv' ]);
-      result['mytestform.csv'].should.equal(
-        `SubmissionDate,name,age,hometown,KEY,SubmitterID,SubmitterName,AttachmentsPresent,AttachmentsExpected,Status,ReviewState,DeviceID,Edits,FormVersion
+  [false, true].forEach(s3enabled => {
```
alxndrsn, line 27: Revert this change, and add a new test which includes uploaded blobs

test/util/s3.js
```diff
@@ -0,0 +1,64 @@
...
+  resetMock() {
+    delete this.enabled;
+    delete this.s3bucket;
+    delete this.error;
+    delete this.uploads;
+  }
+
+  enableMock() {
+    this.enabled = true;
+    this.s3bucket = {};
```
alxndrsn, line 11: use `new Map()`

ktuite (commented):
Comments from review

TODO.md
```diff
@@ -0,0 +1,23 @@
...
+* make sure that a failed/killed process can't result in a blob stuck as "uploading", or if that is possible then there is a way to clean that up from the CLI
+* check what happens with S3 when trying to re-upload identical content - https://groups.google.com/g/jets3t-users/c/i1gusIu5mTw
+* consider NOT deleting blob.content on upload - instead add CLI script to EXPLICITLY delete all uploaded blobs
+* if e.g. 1 blob is missing from s3, and user is trying to export large amount of data, how should we handle?
+  * simpler: just fail completely
+  * more helpful: e.g. inline error message in exported data... but if done badly this could be misleading and the user may never notice that they've got a partial export
+* e2e test is currently not very interesting - could the upload-pending CLI call be async, and we test what happens if we try to fetch attachments while upload is in progress?
+* should config key be default.external.s3blobStore?
+* remove this file
+
```
ktuite, line 14: Additional questions:
- Should S3 config be set in central .env and passed through? 

- What should be documented about setting up S3?
  - Minimal recommended setup of S3
  - S3 retention strategy? Stop things from being deleted manually from S3?
  - CLI usage

- Should upload-blobs cli be part automatically part of cron job or should it be fully under admin's control? Manual control (e.g. running upload script or adding to a cron job themselves) can allow uploading to take place during a service window or using a custom schedule appropriate to specific blob usage.

- More e2e tests could also include multiple overlapping upload processes (e.g. one starts before the other has finished, two processes try to upload the same blob (shouldn't happen))

test/integration/api/submissions.js
```diff
@@ -485,6 +485,50 @@ describe('api: /submission', () => {
                 .set('If-None-Match', '"25bdb03b7942881c279788575997efba"')
                 .expect(304)))))));
 
+    it('should successfully save additionally POSTed attachment binary data with s3 enabled', testService((service, { Blobs }) => {
```
ktuite, line 488: I remembered some other integration tests:
- downloading form attachments rather than submission attachments (technically tested in e2e test)
- downloading the xlsx form if it was provided (and converted to xml behind the scenes)
- checking what happens if you update or delete a form or submission attachment (old blob could be in S3, new blob could be in DB)

lib/resources/submissions.js
```diff
@@ -297,14 +298,14 @@ module.exports = (service, endpoint) => {
               response.append('Content-Type', 'application/zip');
               return zipStreamFromParts(
                 // TODO: not 100% sure that these streams close right on crash.
-                streamBriefcaseCsvs(rows, fields, form.xmlFormId, selectValues, decryptor, false, options),
-                streamAttachments(attachments, decryptor),
-                streamClientAudits(clientAudits, form, decryptor)
+                streamBriefcaseCsvs(s3, rows, fields, form.xmlFormId, selectValues, decryptor, false, options),
+                streamAttachments(s3, attachments, decryptor),
+                streamClientAudits(s3, clientAudits, form, decryptor)
```
alxndrsn, line 303: Don't pass `s3` to these functions - `X.streamForExport()` functions should fetch from s3 transparently.

Makefile
```diff
@@ -26,6 +34,18 @@ fake-oidc-server-ci:
 	cd test/e2e/oidc/fake-oidc-server && \
 	node index.mjs
 
+.PHONY: fake-s3-server
+fake-s3-server:
```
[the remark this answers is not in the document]
alxndrsn: I've split this target into a persistent and an ephemeral version - devs can use whichever suits.

test/integration/api/submissions.js
alxndrsn: Added to TODO.md :+1:

TODO.md
alxndrsn: Added to this file.

Makefile
```diff
@@ -83,15 +103,15 @@ lint: node_version
...
 
 .PHONY: stop-docker-postgres
 stop-docker-postgres:
-	docker stop odk-postgres14 || true
+	docker stop odk-postgres15 || true
 
 .PHONY: rm-docker-postgres
 rm-docker-postgres: stop-docker-postgres
-	docker rm odk-postgres14 || true
+	docker rm odk-postgres15 || true
```
alxndrsn, line 114: Revert all changes relating to postgres docker image.

alxndrsn:
Slonik is issuing an extra SQL query every time the s3_status column is referenced(?)

```
2024-08-24 09:19:28.760 UTC [502] LOG:  execute <unnamed>: SELECT oid, typarray, typname FROM pg_type WHERE typname = ANY($1::text[])
2024-08-24 09:19:28.760 UTC [502] DETAIL:  parameters: $1 = '{date,int8,interval,numeric,timestamp,timestamptz}'
2024-08-24 09:19:28.771 UTC [502] LOG:  execute <unnamed>: SELECT COUNT(*) FROM blobs WHERE s3_status=$1
2024-08-24 09:19:28.771 UTC [502] DETAIL:  parameters: $1 = 'failed'
```

In slonik:

https://github.com/gajus/slonik/blob/d43cfa33eea6a8ac82277c8a72f3e436522b1602/src/routines/createTypeOverrides.ts#L21-L25

alxndrsn:
> Slonik is issuing an extra SQL query every time the s3_status column is referenced(?)

This is not a concern, as it looks like the query is made only once _per connection pool_.  It happens frequently in the integration tests on this branch, as each call to `node lib/bin/s3.js` creates a new connection pool.

lib/external/s3.js
```diff
@@ -0,0 +1,175 @@
...
+
+const init = (config) => {
+  if (!config) return disabled;
+
+  const { server, accessKey, secretKey, bucketName, requestTimeout } = config;
+  if (!(server && accessKey && secretKey && bucketName)) return disabled;
+
+  let destroyed = false;
+
+  const inflight = new Set();
```
alxndrsn, line 23: * Could use weakRefs in this set if there is concern about memory leaks

lib/external/s3.js
```diff
@@ -0,0 +1,175 @@
...
+    // Include blob SHA sum to prevent name collision in case multiple odk-central-
+    // backend instances point to the same bucket.  There are a few scenarios where
+    // this could happen, e.g.
+    //
+    //   * instance reset after testing/training
+    //   * staging & prod instances pointed to the same bucket
+    //   * temporary loss of access to postgres data on upgrade error
+    if (typeof id !== 'number') throw new Error('Invalid id: ' + id);
+    if (!sha) throw new Error('Missing sha sum for blob: ' + id);
+    return `blob-${id}-${sha}`;
```
alxndrsn, line 103: Also include optional bucket prefix?

lib/external/s3.js
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  async function uploadFromBlob(blob) {
+    const objectName = objectNameFor(blob);
+
+    const { md5, sha } = blob;
+    const { length } = blob.content;
+    console.log('Uploading blob:', JSON.stringify({ md5, sha, length })); // eslint-disable-line no-console
+
+    const { Readable } = require('stream');
```
alxndrsn, line 151: Move `require()` to top

ktuite (approved):
Left a couple small comments about the database migration and package lock, but otherwise it looks good to go!!!

lib/model/migrations/20240619-01-add-blob-s3.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+const down = async (db) => {
+  await db.raw(`
+    ALTER TABLE blobs
+      DROP COLUMN s3_status,
+      ALTER COLUMN content SET NOT NULL
+  `);
+  await db.raw('DROP TYPE S3_UPLOAD_STATUS');
+  await db.raw('DROP EXTENSION pgrowlocks');
```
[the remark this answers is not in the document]
alxndrsn: What's your recommendation?  Can I just delete the `down` migration?

package-lock.json
```diff
@@ -1470,6 +1471,12 @@
         "undici-types": "~5.26.4"
       }
     },
+    "node_modules/@zxing/text-encoding": {
```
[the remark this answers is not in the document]
alxndrsn: I've reset to `package-lock.json` from `master` and then re-installed dependencies.  Let's see if that improves it.

lib/model/migrations/20240619-01-add-blob-s3.js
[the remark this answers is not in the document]
alxndrsn: OK, it looks like there's precedent for this:

https://github.com/getodk/central-backend/blob/33f2213bd8271059bdbbd9e2a4e53516c5b29098/lib/model/migrations/20171030-01-add-default-authz-records.js#L42-L44

PR #1074, opened 2024-01-25
Return empty source object for all entity audit events

ktuite:
This was meant to be part of backend PR https://github.com/getodk/central-backend/pull/1072 but i forgot to push it!! 

Frontend PR https://github.com/getodk/central-frontend/pull/932 expects a source for every entity audit event, even API events, where the source is currently just `{}`.

### DIFF
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -214,7 +214,7 @@ const getByEntityId = (entityId, options) => ({ all }) => {
 
       const details = mergeLeft(audit.details, sourceEvent
         .map(event => ({ source: { event, submission } }))
-        .orElse(undefined));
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
It seems a little funny to have a `source` object that's completely empty, but I guess it'd also be funny not to have a `source` property at all. I bet it simplifies Frontend to be able to assume that there is always a `source` property.

ktuite:
> It seems a little funny to have a `source` object that's completely empty, but I guess it'd also be funny not to have a `source` property at all. I bet it simplifies Frontend to be able to assume that there is always a `source` property.

We will have to put something in there soon to distinguish between "API" bulk upload and API single operation when the source is not a submission.

PR #1075, opened 2024-01-26
eslint: relax enforcement of string quotes

alxndrsn:
Keeps the expectation of single quotes for strings, but makes enforcement less strict when:

* a single quote is included inside the string
* template literals (`backticks`) are used when not strictly required

This allows removing all existing inline exceptions for `quotes` rule, except where JSON(?) is directly included in `lib/data/analytics.js`.


matthew-white (approved):
Definitely an improvement to the linting rule! 👍

PR #1078, opened 2024-01-30
Orderby support in entity odata endpoint

ktuite:
Closes #

Adds $orderby support to OData endpoints for Entities and Submissions. This will be useful overall but it is also part of Frontend showing existing entities in the bulk entity creation flow. 

Tests

New feature for users! We've covered a lot of the OData spec at this point!

Documentation has been updated in this PR. 

### DIFF
--- a/lib/data/odata-filter.js
+++ b/lib/data/odata-filter.js
@@ -88,23 +88,29 @@ const odataFilter = (expr, odataToColumnMap) => {
 };
 
 const odataOrderBy = (expr, odataToColumnMap, stableOrderColumn = null) => {
-  if (expr == null) return sql`true`;
-
   let initialOrder = null;
   const clauses = expr.split(',').map((exp) => {
     const [col, order] = exp.trim().split(/\s+/);
+
+    // validate field
     if (!odataToColumnMap.has(col))
-      throw Problem.internal.unsupportedODataField({ at: 0, text: col });
+      throw Problem.internal.unsupportedODataField({ text: col });
+
+    // validate order (asc or desc)
+    if (order && !order?.toLowerCase().match(/^(asc|desc)$/))
+      throw Problem.internal.unsupportedODataField({ text: order });
+
+    const sqlOrder = (order?.toLowerCase() === 'desc') ? sql`DESC` : sql`ASC`;
 
     // Save the order of the initial property to use for the stable sort column order
     if (initialOrder == null)
-      initialOrder = order;
+      initialOrder = sqlOrder;
 
-    return sql`${sql.identifier(odataToColumnMap.get(col).split('.'))} ${((order?.toLowerCase() === 'asc') ? sql`ASC` : sql`DESC`)}`;
+    return sql`${sql.identifier(odataToColumnMap.get(col).split('.'))} ${sqlOrder}`;
   });
 
   if (stableOrderColumn != null)
-    clauses.push(sql`${sql.identifier(stableOrderColumn.split('.'))} ${((initialOrder?.toLowerCase() === 'asc') ? sql`ASC` : sql`DESC`)}`);
+    clauses.push(sql`${sql.identifier(stableOrderColumn.split('.'))} ${initialOrder}`);
 
   return sql`ORDER BY ${sql.join(clauses, sql`,`)}`;
 };
--- a/lib/data/odata-filter.js
+++ b/lib/data/odata-filter.js
@@ -87,19 +87,27 @@ const odataFilter = (expr, odataToColumnMap) => {
   return op(ast);
 };
 
-const odataSort = (expr, odataToColumnMap) => {
+const odataOrderBy = (expr, odataToColumnMap, stableOrderColumn = null) => {
   if (expr == null) return sql`true`;
 
+  let initialOrder = null;
   const clauses = expr.split(',').map((exp) => {
     const [col, order] = exp.trim().split(/\s+/);
     if (!odataToColumnMap.has(col))
       throw Problem.internal.unsupportedODataField({ at: 0, text: col });
 
+    // Save the order of the initial property to use for the stable sort column order
+    if (initialOrder == null)
+      initialOrder = order;
+
     return sql`${sql.identifier(odataToColumnMap.get(col).split('.'))} ${((order?.toLowerCase() === 'asc') ? sql`ASC` : sql`DESC`)}`;
   });
+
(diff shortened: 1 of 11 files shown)


matthew-white:
Is there a way to ensure that the sort is stable? Currently, the `ORDER BY` clause ends with `entities.id DESC`, and I think it'd be useful to continue doing that. Though with `$orderby`, maybe the `DESC`/`ASC` after `entities.id` should match the `DESC`/`ASC` that the user specifies for the first `$orderby` property. 🤔

matthew-white:
I'm going to add this PR to the project even though getodk/central#589 was the main impetus behind it. It's a nice improvement on its own that we'll want to mention in release notes.

lib/data/odata-filter.js
```diff
@@ -87,5 +87,27 @@ const odataFilter = (expr, odataToColumnMap) => {
   return op(ast);
 };
 
-module.exports = { odataFilter };
+const odataOrderBy = (expr, odataToColumnMap, stableOrderColumn = null) => {
+  if (expr == null) return sql`true`;
```
ktuite, line 91: 'true' doesn't make sense in context

lib/data/odata-filter.js
```diff
@@ -87,5 +87,27 @@ const odataFilter = (expr, odataToColumnMap) => {
...
+
+  let initialOrder = null;
+  const clauses = expr.split(',').map((exp) => {
+    const [col, order] = exp.trim().split(/\s+/);
+    if (!odataToColumnMap.has(col))
+      throw Problem.internal.unsupportedODataField({ at: 0, text: col });
```
ktuite, line 97: make "at" optional

lib/data/odata-filter.js
```diff
@@ -87,5 +87,27 @@ const odataFilter = (expr, odataToColumnMap) => {
   return op(ast);
 };
 
-module.exports = { odataFilter };
+const odataOrderBy = (expr, odataToColumnMap, stableOrderColumn = null) => {
+  if (expr == null) return sql`true`;
+
+  let initialOrder = null;
+  const clauses = expr.split(',').map((exp) => {
+    const [col, order] = exp.trim().split(/\s+/);
```
ktuite, line 95: have order default to ascending here

lib/data/odata-filter.js
```diff
@@ -87,5 +87,27 @@ const odataFilter = (expr, odataToColumnMap) => {
...
+    if (!odataToColumnMap.has(col))
+      throw Problem.internal.unsupportedODataField({ at: 0, text: col });
+
+    // Save the order of the initial property to use for the stable sort column order
+    if (initialOrder == null)
+      initialOrder = order;
+
+    return sql`${sql.identifier(odataToColumnMap.get(col).split('.'))} ${((order?.toLowerCase() === 'asc') ? sql`ASC` : sql`DESC`)}`;
```
ktuite, line 103: odata spec (https://learn.microsoft.com/en-us/azure/search/search-query-odata-orderby) mentions " If there are null values in the field, null values appear first if the sort is asc and last if the sort is desc."

Add `NULLS FIRST` to end of each sort clause.

test/unit/data/odata-filter.js
```diff
@@ -73,3 +74,48 @@ describe('OData filter query transformer', () => {
...
+    odataOrderBy('__system/updatedAt   asc').should.eql(sql`ORDER BY ${sql.identifier([ 'submissions', 'updatedAt' ])} ASC`);
+    odataOrderBy('  __system/updatedAt   AsC ').should.eql(sql`ORDER BY ${sql.identifier([ 'submissions', 'updatedAt' ])} ASC`);
+  });
+
+  it('should combine multiple sort operators', () => {
+    odataOrderBy('__system/updatedAt desc, __system/submissionDate ASC').should.eql(sql`ORDER BY ${sql.identifier([ 'submissions', 'updatedAt' ])} DESC,${sql.identifier([ 'submissions', 'createdAt' ])} ASC`);
+  });
+
+  it('should handle no asc/desc provided', () => {
+    odataOrderBy('__system/updatedAt').should.eql(sql`ORDER BY ${sql.identifier([ 'submissions', 'updatedAt' ])} DESC`);
```
ktuite, line 90: talked about defaulting to ASC

lib/data/odata-filter.js
ktuite: let's also validate the order

test/unit/http/endpoint.js
```diff
@@ -656,12 +656,6 @@ describe('endpoints', () => {
           .should.be.rejectedWith(Problem, { problemCode: 404.1 });
       });
 
-      it('should reject requests for unsupported OData features', () => {
-        const request = createRequest({ url: '/odata.svc?$orderby=magic' });
-        return odataPreprocessor('json')(null, new Context(request), request)
-          .should.be.rejectedWith(Problem, { problemCode: 501.1 });
```
ktuite, line 662: find a different unsupported feature, e.g. $search?? $inlineCount?

test/unit/data/odata-filter.js
```diff
@@ -73,3 +74,48 @@ describe('OData filter query transformer', () => {
...
+      err.message.should.equal('The given OData filter expression references fields not supported by this server: myfield at 0');
+      return true;
+    });
+  });
+
+  it('should ignore unsupported operators and default to DESC', () => {
+    odataOrderBy('__system/updatedAt UP').should.eql(sql`ORDER BY ${sql.identifier([ 'submissions', 'updatedAt' ])} DESC`);
+  });
+
+  it('should add last sort clause to insure stable sort order', () => {
```
ktuite, line 118: also add test of stable id in reverse direction

lib/model/query/entities.js
```diff
@@ -449,7 +449,9 @@ WHERE
   AND entities."deletedAt" IS NULL
   AND entity_defs.current=true
   AND  ${odataFilter(options.filter, odataToColumnMap)}
-ORDER BY entities."createdAt" DESC, entities.id DESC
+${options.orderby ? sql`
```
ktuite, line 452: Hmmmmm, how does it interact with skiptoken?

ktuite: maybe throw a problem if both are used together?

ktuite (commented):
notes from interactive review with Matt

matthew-white (approved):

PR #1079, opened 2024-01-31
docs: upload form attachment does not support etag

alxndrsn:
I don't think uploading a form attachment API supports `ETag` header.


sadiqkhoja (approved):
that's right

PR #1083, opened 2024-02-06
endpoint: for pipes, only call next() on error

### LINKED ISSUE #1057: "Not found" Problem after request to submissions.csv.zip
In testing, I see a Problem logged after each request to &hellip;/submissions.csv.zip. For example:

```
::ffff:127.0.0.1 - - [07/Dec/2023:21:11:05 +0000] "GET /v1/projects/1/forms/simple/submissions.csv.zip HTTP/1.1" 200 -
      ✔ should return a zipfile with the relevant headers (267ms)
Error: Could not find the resource you were looking for.
    at Object.result [as notFound] (/home/circleci/repo/lib/util/problem.js:45:31)
    at /home/circleci/repo/lib/http/service.js:88:64
    at Layer.handle [as handle_request] (/home/circleci/repo/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/home/circleci/repo/node_modules/express/lib/router/index.js:328:13)
    at /home/circleci/repo/node_modules/express/lib/router/index.js:286:9
    at Function.process_params (/home/circleci/repo/node_modules/express/lib/router/index.js:346:12)
    at next (/home/circleci/repo/node_modules/express/lib/router/index.js:280:10)
    at next (/home/circleci/repo/node_modules/express/lib/router/route.js:136:14)
    at /home/circleci/repo/lib/http/endpoint.js:203:47
    at node:internal/util:464:12
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

The error seems to come from here:

https://github.com/getodk/central-backend/blob/a0cb6068b5fe3af78f5112f9005a5d9777920350/lib/http/service.js#L87-L88

I think I'm only seeing the error after &hellip;/submissions.csv.zip, not other responses that stream.

alxndrsn:
Closes #1057

From express docs:

> If the current middleware function does not end the request-response cycle, it must call next() to pass control to the next middleware function.
> - https://expressjs.com/en/guide/using-middleware.html

`pipeline()` should be assumed to have ended the "request-response cycle" unless an error is passed to its callback function.

### DIFF
--- a/lib/http/endpoint.js
+++ b/lib/http/endpoint.js
@@ -200,7 +200,7 @@ const defaultResultWriter = (result, request, response, next) => {
     result.with(response).pipeline((err) => next?.(err));
   } else if (result.pipe != null) {
     result.on('error', streamErrorHandler(response));
-    pipeline(result, response, (err) => next?.(err));
+    pipeline(result, response, (err) => err && next?.(err));
   } else {
     response.send(serialize(result));
   }


lib/http/endpoint.js
```diff
@@ -200,7 +200,7 @@ const defaultResultWriter = (result, request, response, next) => {
     result.with(response).pipeline((err) => next?.(err));
```
alxndrsn, line 200: I would expect the check to also be required here, but adding it doesn't make any different to errors logged by the integration test suite.

lib/http/endpoint.js
```diff
@@ -200,7 +200,7 @@ const defaultResultWriter = (result, request, response, next) => {
...
   } else if (result.pipe != null) {
     result.on('error', streamErrorHandler(response));
-    pipeline(result, response, (err) => next?.(err));
```
alxndrsn, line 203: It's unclear to me if `next` would _ever_ be `null`/`undefined` :thinking:

alxndrsn:
```sh
git checkout master >/dev/null 2>&1 &&
echo "master: $(make test-integration 2>&1 | grep -F 'Error: Could not find the resource you were looking for.' | wc -l)" &&
git checkout zip-notfound-error >/dev/null 2>&1 &&
echo "fix:    $(make test-integration 2>&1 | grep -F 'Error: Could not find the resource you were looking for.' | wc -l)"
```

Output:

```
master: 64
fix:    0
```

matthew-white (approved):

PR #1085, opened 2024-02-06
migrations: convert some from knex to raw SQL

alxndrsn:
Ref #744


alxndrsn:
@matthew-white is there any interest in moving _old_ migrations off knex?

matthew-white:
I don't think so, at least not at this moment. Before changing any migrations, I think we should decide on our strategy for moving off of Knex. That's definitely something we want to do, but it's not clear how. Here's the issue for it: #744. Here's one idea from that issue:

> One idea that we've had is to take a snapshot of the schema of a particular past version of Central (version X), then remove all migrations from before that version.

(More details in the issue.) If we took that approach, then we wouldn't need to modify old migrations.

Maybe there's a case for modifying a handful of more recent Knex migrations, if that enables us to push version X further back. But even in that case, I don't see us wanting to modify migrations from 2017. We don't have tests for a lot of those migrations.

For now, I think we should close this PR. We can always reopen if it we want to use these modified migrations. In the meantime, if you have any ideas for the overall strategy for moving off Knex, it'd be awesome to hear (you could drop them in #744 or in Slack).

PR #1094, opened 2024-02-22
WIP: http: strip etags from error responses

alxndrsn:
This PR currently demonstrates that strong ETags are set on error responses, and that if these are stripped, express still sets weak ETags on those responses.

Setting strong ETags which correspond with OK responses on error responses with non-matching content is not useful, and may cause caching issues.

Adding a test for this without making invasive changes to production code seems tricky.

Running the test locally:

* `STRIP_ERR_ETAGS=yes make test-integration`
* `STRIP_ERR_ETAGS=no  make test-integration`

### DIFF
--- a/lib/http/service.js
+++ b/lib/http/service.js
@@ -87,6 +87,13 @@ module.exports = (container) => {
   // first, translate routing fallthroughs to 404:
   service.use((request, response, next) => { next(Problem.user.notFound()); });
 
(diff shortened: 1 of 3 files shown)


alxndrsn:
Closing as this PR just demonstrates the issue.  Fix at https://github.com/getodk/central-backend/pull/1103

Issue #1095, opened 2024-02-23
Benchmark bulk entity create API


ktuite:
### Questions about bulk upload limits:
* What is the upper limit of an entity CSV that backend can handle?
* How does that change with # of rows vs. # of columns?
* What limits will be imposed from other parts of the system (e.g. 100MB nginx request body limit ) or limits of what frontend can handle? 
* Frontend might be able to parse a larger file than backend?

### Optimization idea:
There was an optimization using `sql.unnest` made to another part of backend for inserting many form fields at once. Should that approach be used here, too?  https://github.com/gajus/slonik?tab=readme-ov-file#inserting-large-number-of-rows

Commit where this was added to `insertMany`: https://github.com/getodk/central-backend/commit/222b2a85501f68e614c587d2081a28dc2c8a938e

What is the upper limit?

> Postgresql because database supports only 65535 parameters

10 parameters per entity def --> maybe ~6K entities able to be inserted at once without the unnest optimization? Things to try!!

### Related idea:
Should `createMany` and `createNew` (multiple vs. single entity creation) be combined? Does it make sense to use the multi case for a single entity or is `createNew` better optimized for the single-entity scenario?

matthew-white:
I have some basic CSV parsing in Frontend in place, and I've been benchmarking that parsing. I also tried sending the parsed entities to Backend, and so far, I'm seeing two limits.

I'm able to send 6553 entities, but there's a Postgres error for 6554 entities. That's consistent with the limit mentioned above. Happy to share my JSON data if that'd be helpful!

Secondly, even though `client_max_body_size` is `100m` in nginx ([here](https://github.com/getodk/central/blob/20dcbf46703ded595b82bda713c624673a3faff9/files/nginx/odk.conf.template#L19)), Backend itself seems to set a much smaller limit for JSON request bodies, of 250 kB. I had to increase that limit in order for larger requests to go through.

https://github.com/getodk/central-backend/blob/f98187d76307dafe76482a882beba16854008c9f/lib/http/service.js#L33

ktuite:
I just ran into the 6553 max entity and 250kb JSON limit myself. 

Removing that limit, I've so far been able to upload 6553 entities with 1000 properties, but not 1500 properties.

matthew-white:
Based on discussion yesterday, we're hoping to be able to bulk-create 50,000 entities with 20 properties each, where each value is 100 characters.

lognaturel:
> 250kb JSON limit 

[...]

PR #1098, opened 2024-03-06
Use sql unnest to optimize entity insertion

ktuite:
Related to https://github.com/getodk/central-backend/issues/1095 

Happy to report that I was able to insert 500,000 entities w/ 10 properties in about ~30 seconds (on my M3 mac) with this `sql.unnest` change. 

That payload was about `76,000KB` (76MB) and this limit was increased to `250000kb` (250MB)
[service.use(bodyParser.json({ type: 'application/json', limit: '250kb' }));](https://github.com/getodk/central-backend/blob/a04caadcb9052105b03e6be4e431c8f1923b1f4c/lib/http/service.js#L33-L34)

Tests of small insertions still work. 
I tried it on my own with those larger numbers.

There is still more to do to create many entities at once (streaming JSON?), but I think it's worth merging this particular change now so the bottleneck is no longer the database insert.

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -97,7 +97,7 @@ const createMany = (dataset, rawEntities, sourceId, userAgentIn) => async ({ all
   // Augment parsed entity data with dataset and creator IDs
   const entitiesForInsert = rawEntities.map(e => new Entity({ datasetId: dataset.id, creatorId, ...e }));
 
-  const entities = await all(insertMany(entitiesForInsert));
+  const entities = await all(sql`${insertMany(entitiesForInsert)} RETURNING id`);
 
   // Augment defs with IDs of freshly inserted entities and
   // other default values
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -109,8 +109,7 @@ const createMany = (dataset, rawEntities, sourceId, userAgentIn) => async ({ all
     sourceId,
     version: 1,
     userAgent,
-    ...e.def,
-    data: { ...e.def.data }, // was [Object: null prototype]
+    ...e.def
   }));
   return all(insertMany(defsForInsert));
 };
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -94,13 +94,14 @@ const createMany = (dataset, entities, sourceId, userAgentIn) => async ({ all, c
   const creatorId = context.auth.actor.map((actor) => actor.id).orNull();
   const userAgent = blankStringToNull(userAgentIn);
 
-  const entityInsert = sql.join(entities.map(e => sql`(${sql.join([dataset.id, e.uuid, creatorId, sql`clock_timestamp()`], sql`,`)} )`), sql`,`);
+  const entityRows = entities.map(e => [dataset.id, e.uuid, creatorId]);
+  const entityColumnTypes = ['int4', 'uuid', 'int4'];
   const newEntities = await all(sql`
-  INSERT INTO entities ("datasetId", "uuid", "creatorId", "createdAt")
-  VALUES ${entityInsert}
+  INSERT INTO entities ("createdAt", "datasetId", "uuid", "creatorId")
+  SELECT clock_timestamp(), * FROM ${sql.unnest(entityRows, entityColumnTypes)} as t
   RETURNING id`);
 
-  const defInsert = sql.join(entities.map((e, i) => sql`(${sql.join([
+  const defRows = entities.map((e, i) => [
     newEntities[i].id,
     e.def.label,
     JSON.stringify(e.def.data),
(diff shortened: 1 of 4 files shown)


lib/model/query/entities.js
```diff
@@ -94,13 +94,14 @@ const createMany = (dataset, entities, sourceId, userAgentIn) => async ({ all, c
   const creatorId = context.auth.actor.map((actor) => actor.id).orNull();
   const userAgent = blankStringToNull(userAgentIn);
 
-  const entityInsert = sql.join(entities.map(e => sql`(${sql.join([dataset.id, e.uuid, creatorId, sql`clock_timestamp()`], sql`,`)} )`), sql`,`);
+  const entityRows = entities.map(e => [dataset.id, e.uuid, creatorId]);
+  const entityColumnTypes = ['int4', 'uuid', 'int4'];
```
matthew-white, line 98: Postgres does have a `uuid` data type, but I don't think we store `entities.uuid` that way. It looks like the column is `character varying(255)`.

lib/model/query/entities.js
```diff
@@ -110,16 +111,26 @@ const createMany = (dataset, entities, sourceId, userAgentIn) => async ({ all, c
...
     'true',
     'true',
-    sql`clock_timestamp()`,
     '1'
-  ], sql`,`)} )`), sql`,`);
+  ]);
+  const defColumnTypes = [
+    'int4',
+    'text',
+    'json',
```
matthew-white, line 119: Maybe it'd be better for this to be `jsonb`, since that's what the column ultimately is.

lib/model/query/entities.js
matthew-white, line 1: We use `insertMany()` for several tables, so I'm curious about whether it'd be possible to use it here. Did you run into any obstacles with trying `insertMany()`? I see that it would probably insert values for columns that it doesn't need to, e.g., `entities."updatedAt"`. Or is there an issue utilizing column defaults, e.g., for `entities.root`?

lib/model/query/entities.js
```diff
@@ -94,13 +94,14 @@ const createMany = (dataset, entities, sourceId, userAgentIn) => async ({ all, c
...
   const userAgent = blankStringToNull(userAgentIn);
 
-  const entityInsert = sql.join(entities.map(e => sql`(${sql.join([dataset.id, e.uuid, creatorId, sql`clock_timestamp()`], sql`,`)} )`), sql`,`);
+  const entityRows = entities.map(e => [dataset.id, e.uuid, creatorId]);
+  const entityColumnTypes = ['int4', 'uuid', 'int4'];
   const newEntities = await all(sql`
-  INSERT INTO entities ("datasetId", "uuid", "creatorId", "createdAt")
-  VALUES ${entityInsert}
+  INSERT INTO entities ("createdAt", "datasetId", "uuid", "creatorId")
+  SELECT clock_timestamp(), * FROM ${sql.unnest(entityRows, entityColumnTypes)} as t
```
matthew-white, line 101:
```suggestion
  SELECT clock_timestamp(), * FROM ${sql.unnest(entityRows, entityColumnTypes)} AS t
```

matthew-white (commented):
Very exciting that the database is no longer the bottleneck! 🎉 I have a question about `insertMany()`, but this is looking like a nice change to me.

lib/model/query/entities.js
ktuite: I was worried (without a good reason) about using `insertMany` for this two-step case (and using the new ids for the followup query), but after doing it manually, I saw it was just fine to do with `insertMany`.

lib/model/query/entities.js
```diff
@@ -90,38 +90,29 @@ createNew.audit.withResult = true;
...
-  const defs = await all(sql`
-  INSERT INTO entity_defs ("entityId", "label", "data", "dataReceived",
-    "sourceId", "creatorId", "userAgent", "root", "current", "createdAt", "version")
-  VALUES ${defInsert}
-  RETURNING *
-  `);
-
-  return defs;
+    ...e.def,
+    data: { ...e.def.data }, // was [Object: null prototype]
```
matthew-white, line 113: Could you say more about this line? Were you encountering an error involving `data`? I think it's a good thing that we create `e.def.data` with a `null` prototype. If we could avoid copying the data, that seems ideal.

ktuite: I was getting this error:
```
'InvalidInputError: Invalid unnest tuple member type. Must be a primitive value expression.',
'at Object.createUnnestSqlFragment (/Users/ktuite/Desktop/code/odk/central-backend/node_modules/slonik/dist/src/sqlFragmentFactories/createUnnestSqlFragment.js:35:23)',
...
```

matthew-white: Hmm it sort of sounds like `JSON.stringify()` isn't getting called. 🤔 I'm guessing that this function of ours is at fault:

https://github.com/getodk/central-backend/blob/7839e07b88facbbf10336cfdd38ce3d9242405f7/lib/util/db.js#L135-L143

If that's the cause, I think we could call `JSON.stringify()` there if `typeof v === 'object' && Object.getPrototypeOf(v) == null`. I'm not sure how best to modify that ternary, but I'm thinking of something like:

```js
const _assign = (obj) => (k) => {
  if (k === 'createdAt') return sql`clock_timestamp()`;
  const v = obj[k];
  if (v === null || v === undefined) return null;
  if (typeof v === 'object') {
    if (v instanceof Date) return v.toISOString();
    if (v.constructor === Object || Object.getPrototypeOf(v) == null) return JSON.stringify(v);
  }
  return v;
};
```

matthew-white (approved):

lib/util/db.js
```diff
@@ -179,7 +180,8 @@ const insertMany = (objs) => {
 
   return sql`
   INSERT INTO ${raw(Type.table)} (${columns})
-  SELECT ${selectExp} FROM ${sql.unnest(rows, columnTypes)} AS t`;
+  SELECT ${selectExp} FROM ${sql.unnest(rows, columnTypes)} AS t
+  RETURNING *`;
```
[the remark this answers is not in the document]
matthew-white: Alternatively, maybe `insertMany()` could take an optional argument for `RETURNING` that defaults to no `RETURNING` clause?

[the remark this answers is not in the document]
ktuite: I like the idea of just adding `RETURNING id` to the sql in `Entities.createMany()` instead of changing `insertMany`. Changing it to that.

PR #1099, opened 2024-03-11
Bug Fix for `istrue` & `isfalse` utility function

### LINKED ISSUE #806: isFalse() and non-string query parameters
`isFalse()` throws an error if a query parameter is not blank or string:

```
TypeError: x.toLowerCase is not a function
  File "/usr/odk/lib/util/http.js", line 22, col 43, in isFalse
    const isFalse = (x) => (!isBlank(x) && (x.toLowerCase() === 'false'));
  File "/usr/odk/lib/util/db.js", line 391, col 9, in Function.fromSubmissionCsvRequest
    if (isFalse(query.groupPaths))
  File "/usr/odk/lib/resources/submissions.js", line 270, col 42, in <anonymous>
    const options = QueryOptions.fromSubmissionCsvRequest(query);
  File "/usr/odk/lib/util/http.js", in runMicrotasks
  File "node:internal/process/task_queues", line 96, col 5, in processTicksAndRejections
```

For example, we've seen this error for the (invalid) query parameter `deletedFields=%5B%22false%22%2C%22false%22%5D`. `isFalse()` seems to work fine for valid user input, but it'd be nice if it were more resilient to user error. For example, Express allows a query parameter to be an array, but `isFalse()` doesn't account for that possibility.

`isTrue()` should probably be updated in a similar way as `isFalse()`.

JosephVoid:
Closes #806

I have ran `make test-unit` and all the test passed
It was the most straight forward solution. No other approaches were needed, just letting the function consider string only
It doesn't affect much, the function now is doing what is was supposed to do, which is check for string types only.
No

### DIFF
--- a/lib/util/http.js
+++ b/lib/util/http.js
@@ -18,8 +18,8 @@ const sanitize = require('sanitize-filename');
 // REQUEST PARSING
 
 // Returns t/f only if the input is a string of any upper/lowercase combination.
-const isTrue = (x) => (!isBlank(x) && (x.toLowerCase() === 'true'));
-const isFalse = (x) => (!isBlank(x) && (x.toLowerCase() === 'false'));
+const isTrue = (x) => (!isBlank(x) && typeof x === 'string' && (x.toLowerCase() === 'true'));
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
Thank you, @JosephVoid!

PR #1101, opened 2024-03-13
Specify that entity uuids need to be version 4

lognaturel:
I randomly generated some UUIDs but they were v1. I was confused for a bit when I got an invalid UUID error.

Tests

It gives a hint as to what's going wrong without making invasive changes.

None, text only.

No.


matthew-white (approved):

matthew-white:
The different versions of UUID may be one more reason why it's nice that starting with .1, we won't require users to specify UUIDs via the API (getodk/central#574).

lognaturel:
Yes but having Central generate UUIDs introduces the risk of duplicates. Users will need to be very careful not to upload the same file multiple times because we’ll have no way of detecting duplicate entities.

PR #1103, opened 2024-03-13
http: strip ETag headers from error responses

alxndrsn:
Setting strong ETags which correspond with OK responses on error responses with non-matching content is not useful, and may cause caching issues.

Adding a test for this without making invasive changes to production code seems tricky.

For invasive tests, see: https://github.com/getodk/central-backend/pull/1094


sadiqkhoja (approved):
LGTM

matthew-white:
I'm going to add this one to the project, since it's user-facing / affects API responses, and I want to remember to add it to release notes.

PR #1107, opened 2024-03-17
Don't leak error details to client

### LINKED ISSUE #549: Prod stacktraces are returned to client for unhandled 500 errors
E.g.

```
$ curl 2>/dev/null https://dev.getodk.cloud/v1/sessions \
    --header 'content-type: application/json' \
    --data '{"email":{},"password":{}}' | jq .
{
  "message": "Completely unhandled exception: Unexpected value expression.",
  "details": {
    "stack": [
      "TypeError: Unexpected value expression.",
      "at sql (/usr/odk/node_modules/slonik/dist/src/factories/createSqlTag.js:45:23)",
      "at equals (/usr/odk/lib/util/db.js:197:12)",
      "at _getSql (/usr/odk/lib/model/query/users.js:66:9)",
      "at /usr/odk/lib/model/query/users.js:74:12",
      "at Object.module.<computed> [as getByEmail] (/usr/odk/lib/model/container.js:30:33)",
      "at /usr/odk/lib/resources/sessions.js:24:18",
      "at /usr/odk/lib/http/endpoint.js:141:26",
      "at processTicksAndRejections (internal/process/task_queues.js:95:5)",
      "at async Object.transaction (/usr/odk/node_modules/slonik/dist/src/connectionMethods/transaction.js:22:24)",
      "at async Object.createConnection (/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js:96:18)"
    ]
  }
}
```

Possibly written at: https://github.com/getodk/central-backend/blob/master/lib/http/endpoint.js#L202-L211

alxndrsn:
Closes #549

### DIFF
--- a/lib/http/endpoint.js
+++ b/lib/http/endpoint.js
@@ -218,15 +218,10 @@ const defaultErrorWriter = (error, request, response) => {
       details: error.problemDetails
     });
   } else {
-    const details = {};
-    if (error?.stack != null)
-      details.stack = error.stack.split('\n').map((x) => x.trim());
-
     debugger; // trip debugger if attached.
     process.stderr.write(inspect(error));
     response.status(500).type('application/json').send({
-      message: `Completely unhandled exception: ${error?.message}`,
-      details
+      message: 'Internal Server Error',
     });
   }
 };
--- a/test/unit/http/endpoint.js
+++ b/test/unit/http/endpoint.js
@@ -60,13 +60,12 @@ describe('endpoints', () => {
       defaultErrorWriter(new Problem(409.1138, 'test message', { x: 1 }), null, response);
     });
 
-    it('should turn remaining errors into unknown Problems', (done) => {
+    it('should turn remaining errors into internal server errors', (done) => {
       const response = createModernResponse();
       const error = new Error('oops');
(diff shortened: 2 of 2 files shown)


lib/http/endpoint.js
```diff
@@ -218,15 +218,10 @@ const defaultErrorWriter = (error, request, response) => {
...
     });
   } else {
-    const details = {};
-    if (error?.stack != null)
-      details.stack = error.stack.split('\n').map((x) => x.trim());
-
     debugger; // trip debugger if attached.
     process.stderr.write(inspect(error));
     response.status(500).type('application/json').send({
-      message: `Completely unhandled exception: ${error?.message}`,
```
matthew-white, line 228: The discussion in the issue was about stack traces, but here the error message is also not being returned. That seems OK to me: the user will still be able to see it in their logs. But could you say more about why you made that change? Do you think that as with stack traces, secrets could end up in error messages?

matthew-white (approved):

lib/http/endpoint.js
alxndrsn: > Do you think that as with stack traces, secrets could end up in error messages?

Yes, this :+1:

PR #1111, opened 2024-03-21
Remove X-Powered-By header

lognaturel:
Follows the OWASP recommendation: https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#x-powered-by

This is not super important because there are many other ways to learn about the stack but it does get flagged by some security scanners.

#### What has been done to verify that this works as intended?
Ran it, looked at headers in console

#### Why is this the best possible solution? Were any other approaches considered?
I think this is the only approach!

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
Should have no risk of negative effect.

#### Does this change require updates to the API documentation? If so, please update docs/api.md as part of this PR.
No.

#### Before submitting this PR, please make sure you have:

- [x] run `make test` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1112, opened 2024-03-21
Add test for X-Powered-By header

alxndrsn:
Test for https://github.com/getodk/central-backend/pull/1111


lognaturel:
Cherry picked into https://github.com/getodk/central-backend/pull/1111 so we could merge together. Thanks!

I went back and forth on whether or not to test. I agree I should have -- it's something we could easily regress on and it's intentional so we want to document that.

PR #1113, opened 2024-03-21
http/endpoint: extract common pipe code

alxndrsn:
defaultResultWriter and odataJsonWriter use identical code for piping results to response.  This commit extracts the shared code into its own function.

Follow-up to #1110


lib/http/endpoint.js
```diff
@@ -192,20 +192,28 @@ const endpointBase = ({ preprocessor = noop, before = noop, resultWriter, errorW
...
   if (result instanceof PartialPipe) {
     result.streams.at(-1).on('error', streamErrorHandler(response));
     result.with(response).pipeline((err) => next?.(err));
+    return true;
   } else if (result.pipe != null) {
     result.on('error', streamErrorHandler(response));
     pipeline(result, response, (err) => err && next?.(err));
+    return true;
+  }
+};
```
matthew-white, line 209: I know `undefined` is falsy, but I think it'd be nice if the function returned `false` if it doesn't return `true`.

lib/http/endpoint.js
```diff
@@ -192,20 +192,28 @@ const endpointBase = ({ preprocessor = noop, before = noop, resultWriter, errorW
 // ENDPOINT FORMAT SPECIALIZATIONS
 
 ////////////////////////////////////////
```
matthew-white, line 194: I think this line comment goes together with the line comment on the next line, so the two should be moved together. I think the two lines should come before `defaultResultWriter()`/`defaultErrorWriter()`/`defaultEndpoint()`, similar to how there are line comments above `odataJsonWriter()`/`odataXmlErrorWriter()`/`odataJsonEndpoint()`/etc.

matthew-white (approved):
I know it's been a while since this PR was pushed, but I just took a look, and the changes make sense to me. 💯

lib/http/endpoint.js
alxndrsn: Could we just remove the comment completely?

lib/http/endpoint.js
alxndrsn: Added.

lib/http/endpoint.js
matthew-white: Sounds fine to me, I'm not attached to it. :+1:

PR #1116, opened 2024-03-22
Update db index on entity audit query to use entityDefId and sourceId

ktuite:
Entity details page was getting really slow. `/audits` and `/versions` endpoints in particular were really slow (order of seconds on test server). Both endpoints use `Audits.getByEntityId()`.

Looking into it, when I updated the entity audit query to use the new `sourceId` for bulk entity uploads, this definitely impacted the query's performance.
https://github.com/getodk/central-backend/commit/f98187d76307dafe76482a882beba16854008c9f?diff=split&w=0#diff-754724053d58d67c63f58d13d8708e4d185a11bb2d172614693cc3f611bb5361R171

The code used to filter on `audits.details->>'entityId'`, which we had an index for. Now, not every audit log will reference an entity directly. Instead, we look up audits by `audits.details->'entityDefId'` or `audits.details->'sourceId'` in a different part of the query. 

This PR removes the index on `details->entityId` and adds two new indexes on `entityDefId` and `sourceId` instead.

I'm trying it locally. I think it's helping a bit but it's hard to tell. I haven't tried it on the test server.

### DIFF
--- a/lib/model/migrations/20240322-01-add-entity-source-index-to-audits.js
+++ b/lib/model/migrations/20240322-01-add-entity-source-index-to-audits.js
@@ -1,0 +1,24 @@
+// Copyright 2024 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw('DROP INDEX audits_details_entity_index');
+  await db.raw("CREATE INDEX audits_details_entity_def_index ON audits USING HASH (((details ->> 'entityDefId')::INTEGER))");
+  await db.raw("CREATE INDEX audits_details_entity_source_index ON audits USING HASH (((details ->> 'sourceId')::INTEGER))");
+};
+
+const down = async (db) => {
+  await db.raw("CREATE INDEX audits_details_entity_index ON audits USING HASH (((details ->> 'entityId')::INTEGER))");
+  await db.raw('DROP INDEX audits_details_entity_def_index');
+  await db.raw('DROP INDEX audits_details_entity_source_index');
+};
+
+module.exports = { up, down };
+
+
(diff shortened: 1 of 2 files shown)


lib/model/migrations/20240322-01-add-entity-source-index-to-audits.js
```diff
@@ -0,0 +1,24 @@
...
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw('DROP INDEX audits_details_entity_index');
+  await db.raw("CREATE INDEX audits_details_entity_def_index ON audits USING HASH (((details ->> 'entityDefId')::INTEGER))");
+  await db.raw("CREATE INDEX audits_details_entity_source_index ON audits USING HASH (((details ->> 'sourceId')::INTEGER))");
```
matthew-white, line 13: It looks like the query uses `->`, not `->>`:

https://github.com/getodk/central-backend/blob/428ebdd06143684759f1507a708db69fd2f0b76c/lib/model/query/audits.js#L139

It might be a good idea for the index to exactly match the query. We could either change the index to match the query or change the query to match the index.

lib/model/migrations/20240322-01-add-entity-source-index-to-audits.js
```diff
@@ -0,0 +1,24 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw('DROP INDEX audits_details_entity_index');
```
matthew-white, line 11: It looks like this query might still use this index:

https://github.com/getodk/central-backend/blob/428ebdd06143684759f1507a708db69fd2f0b76c/lib/model/query/audits.js#L110

Though again, there seems to be a mismatch between the query, which uses `->`, and the index, which uses `->>`.

matthew-white (approved):

ktuite:
Because the `_getBySubmissionId` query also uses `details->>'entityId'` I will keep that index. Though it feels like there are a lot of indexes on this table...

I think we decided the [hash index](https://github.com/getodk/central-backend/pull/879/commits/0aafd1739788485f9cd21bd398def74aedc312fa) worked well for `entityId` so I'm using the same thing for `entityDefId` and `sourceId` and I've updated the queries (with `->>`) themselves to match.

PR #1117, opened 2024-03-22
Fix API docs for adding property

lognaturel:
Small mistake.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -9992,7 +9992,7 @@ paths:
         example: "16"
       - name: name
         in: path
-        description: Name of the Property
+        description: Name of the Dataset
         required: true
         schema:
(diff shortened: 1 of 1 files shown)


lognaturel:
@ktuite is the content type for adding properties and datasets really supposed to be */* and not application/json? https://github.com/getodk/central-backend/blob/master/docs/api.yaml#L10002

ktuite (approved):
Many of the example `requestBody -> content` examples in our docs specify `*/*` rather than `application/json`. 

One of the few places I saw json specified in the requestBody (vs. the response) was here: https://github.com/getodk/central-backend/blob/master/docs/api.yaml#L9919-L9921 but it seems like it's `*/*` most other places.

PR #1122, opened 2024-04-11
Add example content to OData `/datasets/{name}.svc` 200 response

spwoodcock:
#### The problem this PR solves

The `Service Document` endpoint for datasets was missing an example 200 response:
https://docs.getodk.org/central-api-odata-endpoints/#id1

image

#### What was done to fix it

- Previously we had: `content: {}`.
- Updated to include actual content.

Loaded the yaml into https://editor.swagger.io/

image

image

- I assume this will be copied to getodk/docs, so is user facing.

- I think this PR template needs an update to reference `docs/api.yaml` ! 😄 

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -11834,7 +11834,76 @@ paths:
       responses:
         200:
           description: OK
-          content: {}
+          content:
+            application/json; charset=utf-8; odata.metadata=minimal:
+              schema:
+                type: object
+                properties:
+                  '@odata.context':
+                    type: string
+                  value:
+                    type: array
+                    items:
+                      required:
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Looks good, thank you!

Do you have a use for the service document? We have so far assumed it's only of value to clients that already use the OData standard (Excel, etc). It could be somewhat useful for arbitrary clients using submissions because of repeats but it feels less clear to me for entities.

spwoodcock:
No I don't use the service document, I just saw this when I was going through the docs 😄

I only use the data endpoint with filter and select params.

PR #1123, opened 2024-04-11
Apply increased json body limit to entity post endpoint

ktuite:
Part of https://github.com/getodk/central-backend/issues/1095

Adds custom middleware to use a higher JSON size limit for `bodyParser` middleware for the specific entity bulk creation route.

Uses `path-to-regex` to match the route, because I think that is what express is using internally, so this dependency should technically already be included.

### DIFF
--- a/lib/http/service.js
+++ b/lib/http/service.js
@@ -10,6 +10,7 @@
 // This file glues together all the middleware and the HTTP REST resources we have
 // defined elsewhere into an actual Express service. The only thing it needs in
 // order to do this is a valid dependency injection context container.
+const { match } = require('path-to-regexp');
 
 module.exports = (container) => {
   const service = require('express')();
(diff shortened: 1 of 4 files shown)


lib/http/service.js
```diff
@@ -31,7 +32,17 @@ module.exports = (container) => {
...
   const bodyParser = require('body-parser');
-  service.use(bodyParser.json({ type: 'application/json', limit: '250kb' }));
+
+  // use a default json limit except for URLs that explicitly need a larger limit.
+  const defaultJsonLimit = bodyParser.json({ type: 'application/json', limit: '250kb' });
+  const largeJsonLimit = bodyParser.json({ type: 'application/json', limit: '100mb' });
+  const largeJsonUrlMatch = match('/:apiVersion/projects/:id/datasets/:name/entities');
+
+  service.use((req, res, next) => {
+    if (largeJsonUrlMatch(req.url))
```
matthew-white, line 42: I'm noticing that `req.url` will include the query string if there is one, but then `largeJsonUrlMatch()` won't match it:

```js
// Returns an object.
largeJsonUrlMatch('/v1/projects/1/datasets/foo/entities')
// Returns false.
largeJsonUrlMatch('/v1/projects/1/datasets/foo/entities?bar=baz')
```

However, it doesn't really matter now, because the endpoint doesn't use query parameters. I thought I'd mention it, because I could see the endpoint using query parameters in the future. But we could always circle back to it then.

ktuite: I was able to use `req.path` instead of `req.url` to handle this better.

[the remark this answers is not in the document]
ktuite: I can check for the POST method!

matthew-white (approved):

Issue #1096, opened 2024-02-27
Update API docs for v2024.1


ktuite:
- [x] https://github.com/getodk/central-backend/pull/1045
- [x] https://github.com/getodk/central-backend/pull/1078
- [x] https://github.com/getodk/central-backend/pull/1086  
- [x] https://github.com/getodk/central-backend/pull/1070
- [ ] https://github.com/getodk/central/issues/580
- [ ] https://github.com/getodk/central/issues/575
- [ ] https://github.com/getodk/central/issues/576

PR #1127, opened 2024-04-19
Prevent Forms from introducing dataset name case conflicts

ktuite:
Closes https://github.com/getodk/central/issues/627

Prevents Forms and Dataset API from introducing published datasets with conflicting names, specifically the same name with different capitalization. 

* Error on _Form Upload_ if dataset name conflicts with a published dataset
* Error on _Form Publish_ if publishing will introduce a conflict, which can happen if there were multiple draft forms with different name variations, one was published, and then a second is attempted to be published.
* Error on "New Entity List" / API if dataset name conflicts with a published dataset
   * Now allowed to create/publish a dataset that directly collides with an unpublished dataset

Tests, trying with frontend.

I basically added another database query for a similarly named dataset. In the same area of code, we're also querying for an existing exactly matching dataset so we can reuse its ID and Actee ID for various things. I wasn't comfortable trying to combine these queries at this time, given the complicated logic of draft datasets, similar but not exact matches, multiple similar draft datasets, etc. 

This could potentially be added to the API docs. (https://github.com/getodk/central-backend/issues/1096)

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -324,6 +324,7 @@ const getPublishedBySimilarName = (projectId, name) => ({ maybeOne }) => {
     WHERE LOWER(name) = LOWER(${name})
       AND name != ${name}
       AND "publishedAt" IS NOT NULL
+      AND "projectId" = ${projectId}
     LIMIT 1`)
     .then(map(_unjoiner));
 };
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { sql } = require('slonik');
-const { extender, insert, QueryOptions, equals, updater } = require('../../util/db');
+const { extender, insert, QueryOptions, equals, unjoiner, updater } = require('../../util/db');
 const { Dataset, Form, Audit } = require('../frames');
 const { validatePropertyName } = require('../../data/dataset');
 const { isEmpty, isNil, either, reduceBy, groupBy, uniqWith, equals: rEquals, map } = require('ramda');
@@ -90,6 +90,14 @@ const createOrMerge = (parsedDataset, form, fields) => async ({ one, Actees, Dat
   // Need to check for existing dataset, and if not found, need to also
   // fetch the project since the dataset will be an actee child of the project.
   const { name } = parsedDataset;
+
+  // Check for a published dataset with a conflicting name (same name, different capitalization).
+  // There can be multiple draft datasets with conflicting names but we won't throw an error
+  // until publishing.
+  const conflictingName = await Datasets.getPublishedBySimilarName(projectId, name);
+  if (conflictingName.isDefined())
+    throw Problem.user.datasetNameConflict({ current: conflictingName.get().name, provided: name });
+
   const existingDataset = await Datasets.get(projectId, name, false);
   const acteeId = existingDataset.isDefined()
     ? existingDataset.get().acteeId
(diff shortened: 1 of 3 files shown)


lib/model/query/datasets.js
```diff
@@ -153,41 +179,63 @@ createPublishedProperty.audit = (property, dataset) => (log) =>
...
+
+  const _datasetNameConflict = () => sql`
+    SELECT existing_ds.name as current, ds_to_publish.name as provided FROM dataset_form_defs dfd
+    JOIN datasets ds_to_publish ON dfd."datasetId" = ds_to_publish.id
+    JOIN datasets existing_ds ON LOWER(ds_to_publish.name) = LOWER(existing_ds.name)
+      AND ds_to_publish.id != existing_ds.id
+      AND existing_ds."publishedAt" IS NOT NULL
+      AND ds_to_publish.name != existing_ds.name
+    WHERE dfd."formDefId" = ${formDefId}
+      AND ds_to_publish."publishedAt" IS NULL
```
ktuite, line 229: Join on `existing_ds` needs to filter by `projectId`

lib/model/query/datasets.js
```diff
@@ -153,41 +179,63 @@ createPublishedProperty.audit = (property, dataset) => (log) =>
...
+  ORDER BY "propertyId"
+  `;
+
+  const _datasetNameConflict = () => sql`
+    SELECT existing_ds.name as current, ds_to_publish.name as provided FROM dataset_form_defs dfd
+    JOIN datasets ds_to_publish ON dfd."datasetId" = ds_to_publish.id
+    JOIN datasets existing_ds ON LOWER(ds_to_publish.name) = LOWER(existing_ds.name)
+      AND ds_to_publish.id != existing_ds.id
+      AND existing_ds."publishedAt" IS NOT NULL
+      AND ds_to_publish.name != existing_ds.name
```
ktuite, line 227: Condition not necessary, dataset already exists at this point so it wont be possible for two with the same name to already exist.

lib/model/query/datasets.js
```diff
@@ -153,41 +179,63 @@ createPublishedProperty.audit = (property, dataset) => (log) =>
...
+    WHERE dfd."formDefId" = ${formDefId}
+      AND ds_to_publish."publishedAt" IS NULL
+  `;
+
+  const d = await maybeOne(_datasetNameConflict(formDefId));
```
ktuite, line 232:
```suggestion
  const d = await maybeOne(_datasetNameConflict());
```

lib/model/query/datasets.js
```diff
@@ -153,41 +179,63 @@ createPublishedProperty.audit = (property, dataset) => (log) =>
...
+  if (d.isDefined()) {
+    const { current, provided } = d.get();
+    throw Problem.user.datasetNameConflict({ current, provided });
+  }
+  return all(_publishIfExists(formDefId, publishedAt));
```
ktuite, line 237:
```suggestion
  return all(_publishIfExists());
```

lib/model/query/datasets.js
```diff
@@ -123,7 +131,25 @@ const createOrMerge = (parsedDataset, form, fields) => async ({ one, Actees, Dat
...
 // Insert a Dataset when there is no associated form and immediately publish
-const createPublishedDataset = (dataset, project) => async ({ one, Actees }) => {
+const createPublishedDataset = (dataset, project) => async ({ one, Actees, Datasets }) => {
+
+  // Check for existing dataset with the same name
+  const existingDataset = await Datasets.get(project.id, dataset.name, false);
+  if (existingDataset.isDefined()) {
+    const { publishedAt, id: datasetId } = existingDataset.get();
+    if (publishedAt)
+      throw Problem.user.uniquenessViolation({ table: 'datasets', fields: [ 'name', 'projectId' ], values: [ dataset.name, project.id ], });
```
ktuite, line 141:
```suggestion
      throw Problem.user.uniquenessViolation({ table: 'datasets', fields: [ 'name', 'projectId' ], values: [ dataset.name, project.id ] });
```

lib/model/query/datasets.js
```diff
@@ -123,7 +131,25 @@ const createOrMerge = (parsedDataset, form, fields) => async ({ one, Actees, Dat
...
+
+    // If existing dataset is only a draft, allow it to be published
+    return new Dataset(await one(sql`UPDATE datasets SET "publishedAt" = "createdAt" where id = ${datasetId} RETURNING *`));
```
ktuite, line 144: Set `publishedAt` to current clock timestamp rather than `createdAt`

lib/model/query/datasets.js
```diff
@@ -123,7 +131,25 @@ const createOrMerge = (parsedDataset, form, fields) => async ({ one, Actees, Dat
...
+  }
+
+  // Check for a published dataset with a similar name but different capitalization
+  const conflictingName = await Datasets.getPublishedBySimilarName(project.id, dataset.name);
+  if (conflictingName.isDefined())
+    throw Problem.user.datasetNameConflict({ current: conflictingName.get().name, provided: dataset.name });
```
ktuite, line 150: Move this above. Situation:

draft "people"
draft "PEOPLE"
publish "people"
API to create "PEOPLE" (should not be allowed)

lib/util/problem.js
```diff
@@ -206,7 +206,10 @@ const problems = {
 
     // 409.14 (entityViolation) was used in the past, but has been deprecated
 
-    entityVersionConflict: problem(409.15, ({ current, provided }) => `Current version of the Entity is '${current}' and you provided '${provided}'. Please correct the version number or pass '?force=true' in the URL to forcefully update the Entity.`)
+    entityVersionConflict: problem(409.15, ({ current, provided }) => `Current version of the Entity is '${current}' and you provided '${provided}'. Please correct the version number or pass '?force=true' in the URL to forcefully update the Entity.`),
+
+    datasetNameConflict: problem(409.16, ({ current, provided }) => `A dataset named '${current}' exists and you provided '${provided}' with the same name but different capitalization. Please change the capitalization of '${provided}' to match '${current}'.`)
```
ktuite, line 211: Remove second part of sentence to make it make more sense for the new Entity List via API modal.

test/integration/api/datasets.js
```diff
@@ -98,22 +98,81 @@ describe('datasets and entities', () => {
...
-      }));
+          // Second time
+          await asAlice.post('/v1/projects/1/datasets')
+            .send({
+              name: 'trees'
+            })
+            .expect(409)
+            .then(({ body }) => {
+              body.code.should.equal(409.3);
+              body.message.should.startWith('A resource already exists with name,projectId value(s) of trees,');
```
ktuite, line 119: Can add projectID 1 back

test/integration/api/datasets.js
```diff
@@ -3464,6 +3523,180 @@ describe('datasets and entities', () => {
...
+          await alice.post('/v1/projects/1/forms/simpleEntity/draft/publish')
+            .expect(200);
+
+          await alice.post('/v1/projects/1/forms/simpleEntity2/draft/publish')
+            .expect(409)
+            .then(({ body }) => {
+              body.code.should.equal(409.16);
+              body.message.should.startWith("A dataset named 'people' exists and you provided 'PEOPLE'");
+            });
+        }));
```
ktuite, line 3611: Check path of publishing a form while also specifying a new version

test/integration/api/datasets.js
```diff
@@ -3464,6 +3523,180 @@ describe('datasets and entities', () => {
...
+            .expect(200);
+
+          // publish "people"
+          await alice.post('/v1/projects/1/forms/simpleEntity/draft/publish')
+            .expect(200);
+
+          // dataset "people" (same as first/published)
+          await alice.post('/v1/projects/1/forms')
+            .send(testData.forms.simpleEntity
+              .replace('dataset="people"', 'dataset="people"')
```
ktuite, line 3693: dont need to replace people

ktuite (commented):

matthew-white (approved):

PR #1133, opened 2024-04-29
Use schema template for 403 errors

lognaturel:
I noticed that https://docs.getodk.org/central-api-entity-management/#updating-an-entity has a 403 example with code and message both set to `pencil`. I imagine that's some fallback when no examples are specified.

We have a nice schema template for 403 errors so I went through and used that everywhere.

Some endpoints like the one above had 403s defined with json and json/extended response types. In those cases, I removed the `json/extended` one.

Some endpoints redundantly specified an example along with the schema reference which itself specifies the same examples. I removed the redundant ones.

Some endpoints have 403 responses with other response types (e.g. https://docs.getodk.org/central-api-form-management/#draft-form). I think that's also a mistake but didn't want to look too deeply into it for now.

Rendered docs with `make api-docs`, spot checked a few 403s.

Yay for reduction of duplication.

Docs-only.

This does update it.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -1043,9 +1043,6 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error403'
-              example:
-                code: "403.1"
-                message: The authenticated actor does not have rights to perform that action.
   /v1/sessions/current:
     delete:
       tags:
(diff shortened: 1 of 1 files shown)


matthew-white (approved):

matthew-white:
This is only sort of related, but one thing I'm wondering about is the `code` property of the `Error403` schema. It looks to be a string, but Problem codes are actually numbers.

lognaturel:
Great, we now have a single place to fix it! Let's roll it into this change.

matthew-white (approved):

PR #1135, opened 2024-05-01
improve: Also pass along a state (nonce) paramater.

### LINKED ISSUE #1134: OIDC login fails due to Authentik sending state QS and node-openid-client failing a check
It's all here.  https://forum.getodk.org/t/oidc-issue-when-logging-in/46619

openbrian:
bug: #1134

It's not needed for CSRF protection because code_challenge does this, but the Authentik IDP will return &state= in the query string.  This will trigger node-openid-client to fail a check.

Closes #1134 

Testing the oidc way failed for me.  I mean the tests would not kick off.


lib/resources/oidc.js
```diff
@@ -102,15 +103,19 @@ module.exports = (service, endpoint) => {
       const client = await getClient();
       const code_verifier = generators.codeVerifier(); // eslint-disable-line camelcase
 
+      // State is not strictly required because PKCS (code_challenge) protects us from CSRF attacks.
```
lognaturel, line 106: It feels important to then mention why it's there! Maybe just add "but some IDPs always send it"

lognaturel:
Thanks for digging into this and figuring out a solution, @openbrian! I've left one small suggestion. @alxndrsn could you please do a deeper review?

lib/resources/oidc.js
openbrian: done

lib/resources/oidc.js
```diff
@@ -102,15 +103,22 @@ module.exports = (service, endpoint) => {
...
+      // See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#name-countermeasures-4
+      // We create a nonce (state) anyway because some IDPs (e.g. Authentik) will send the nonce
```
alxndrsn, line 108: `nonce` is the name of a different value which can be used at various points in OAuth/OIDC.  I think avoiding that word here would reduce chance of confusion.

lib/resources/oidc.js
```diff
@@ -102,15 +103,22 @@ module.exports = (service, endpoint) => {
       const client = await getClient();
       const code_verifier = generators.codeVerifier(); // eslint-disable-line camelcase
 
+      // State is not strictly required because PKCS (code_challenge) protects us from CSRF attacks.
```
alxndrsn, line 106: Should `PKCS` be `PKCE`?

alxndrsn:
> It's not needed for CSRF protection because code_challenge does this, but the Authentik IDP will return &state= in the query string. This will trigger node-openid-client to fail a check.

Thanks for the PR!  I think this is a bug in Authentik's implementation; they seem to be open source (https://github.com/goauthentik/authentik?) so perhaps it can be fixed on their side.

That said, `state` _should_ be supported by all IdPs (per the spec).  Also `state` reportedly adds some security against generating fake errors.  I will dig deeper and test this against some different IdPs.

lib/resources/oidc.js
openbrian: Yes, I'm on it ;)

lib/resources/oidc.js
```diff
@@ -102,15 +103,22 @@ module.exports = (service, endpoint) => {
...
+      // See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#name-countermeasures-4
+      // We create a nonce (state) anyway because some IDPs (e.g. Authentik) will send the nonce
+      // back to us.  If it doesn't explicitly get one, it sends back an empty string.
+      const state = generators.state(16);
```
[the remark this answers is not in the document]
openbrian: I played in the playground here https://www.oauth.com/playground/authorization-code-with-pkce.html and they had a 16 byte state.  There's like zero cost to use 32 bytes.  It's not encoding or signing anything.  I'll raise it to 32, aka use the default.

lib/resources/oidc.js
alxndrsn: @openbrian just in case you missed this :slightly_smiling_face:

alxndrsn:
Related: https://github.com/panva/openid-client/discussions/703

PR #1137, opened 2024-05-03
API docs: $skiptoken and Dataset name conflict

ktuite:
Closes https://github.com/getodk/central/issues/580

`$skiptoken` was introduced in 2023.4 but not mentioned in the API docs, probably because there isn't much to say about it. It is part of the `@odata.nextLink`, which was already mentioned.

There might be more to add, though:
- $skip and $skiptoken can't be used together
- $skiptoken and $order can't be used together
- If you use $skip originally, you'll get a $skiptoken in the nextLink

I also added the note about Dataset name capitalization conflicts (e.g. "trees" vs. "Trees")  and added a 409 conflict response, though I actually chose the response for the direct dataset name conflict. 

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -8645,7 +8645,9 @@ paths:
       - Dataset Management
       summary: Creating Datasets
       description: |- 
-        You can create a Dataset with a specific `name` within a Project. This Dataset can then be populated via the API or via Forms and Submissions, and then used by other Forms. This endpoint allows a Dataset to be created programatically without an input form.
+        You can create a Dataset with a specific `name` within a Project. This Dataset can then be populated with Entities via the API or via Forms and Submissions, and then used by other Forms. This endpoint allows a Dataset to be created programatically without an input form.
+
+        The name of a Dataset is case-sensitive in that it will keep the capitalization provided (e.g. "Trees"). But Central will not allow a second Dataset with the same name but different capitalization to be created (e.g. "trees" when "Trees" already exists).
 
         By default, the Dataset will have no properties, but each Entity will have a `label` and a unique ID (`uuid`). You can add additional properties with this [related endpoint](/central-api-dataset-management/#adding-properties).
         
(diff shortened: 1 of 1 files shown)


docs/api.yaml
```diff
@@ -8701,6 +8703,22 @@ paths:
...
+          description: Conflict
+          content:
+            application/json:
+              schema:
+                required:
+                - code
+                type: object
+                properties:
+                  code:
+                    type: string
```
matthew-white, line 8716:
```suggestion
                    type: number
```

Some related discussion here: https://github.com/getodk/central-backend/pull/1133#issuecomment-2083947231

docs/api.yaml
```diff
@@ -10460,7 +10478,9 @@ paths:
...
+        The `$top` and `$skip` querystring parameters, specified by OData, apply `limit` and `offset` operations to the data, respectively. The `$count` parameter, also an OData standard, will annotate the response data with the total row count, regardless of the scoping requested by `$top` and `$skip`. If `$top` parameter is provided in the request then the response will include `@odata.nextLink` that you can use as is to fetch the next set of data. As of ODK Central v2023.4, `@odata.nextLink` contains a `$skiptoken` (an opaque cursor) to better paginate around deleted Submissions.
```
matthew-white, line 10481: > to better paginate around deleted Submissions.

There aren't deleted submissions currently (though there will be soon!). `$skiptoken` is still helpful for another reason though, which is that submissions are returned in reverse creation order (newest first). `$skiptoken` allows you to page through submissions even while new submissions are being created / even as the index/offset of each submission changes.

matthew-white:
> There might be more to add, though:
> 
> * $skip and $skiptoken can't be used together
> * $skiptoken and $order can't be used together
> * If you use $skip originally, you'll get a $skiptoken in the nextLink

I agree that it's not particularly useful to mention these things.

matthew-white (approved):

PR #1139, opened 2024-05-07
Remove function: global.tap()

alxndrsn:
It looks like this was used historically in development.


ktuite (approved):
I haven't ever used this myself, but issa would use it temporarily for debugging, especially with the nested promise chains. It wouldn't hurt to take out but I also don't think it hurts to leave in? I guess if someone wanted to use this in the future, they could add it back.

alxndrsn:
:shrug: I guess it can stay in then

PR #1141, opened 2024-05-07
Optimize submission keys query

ktuite:
Closes https://github.com/getodk/central/issues/658

Query is more readable and much faster, especially in the case where there are no keys to return. 

### DIFF
--- a/lib/model/query/keys.js
+++ b/lib/model/query/keys.js
@@ -31,20 +31,20 @@ const getById = (keyId) => ({ maybeOne }) =>
   maybeOne(sql`select * from keys where id=${keyId}`).then(map(construct(Key)));
 
 const getActiveByFormId = (formId, draft) => ({ all }) => all(sql`
-select keys.* from keys
-inner join
-  (select "keyId" from form_defs
-    inner join
-      (select "formDefId", "submissionId" from submission_defs
-        where current=true and "localKey" is not null) as submission_defs
-        inner join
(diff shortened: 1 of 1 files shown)


lib/model/query/keys.js
```diff
@@ -31,20 +31,20 @@ const getById = (keyId) => ({ maybeOne }) =>
...
+  ON form_defs."keyId" = keys.id
+INNER JOIN submission_defs
+  ON submission_defs."formDefId" = form_defs.id
+INNER JOIN submissions
+  ON submissions.id = submission_defs."submissionId"
+WHERE submission_defs.current = true
+  AND submission_defs."localKey" IS NOT NULL
+  AND submissions.draft = ${draft}
+  AND form_defs."formId" = ${formId}
+  AND form_defs."keyId" IS NOT NULL
```
matthew-white, line 45: I think this condition is already covered by the join between `keys` and `form_defs`. If a form def has a `NULL` `keyId`, then it won't join with a row from `keys`.

matthew-white (approved):

lib/model/query/keys.js
ktuite: I agree that while it seems redundant, it makes this query orders of magnitude faster. ~700ms -> 0.1 ms

matthew-white: Whoa, interesting!

Issue #1136, opened 2024-05-02
Streamline Form creation/update/publish paths to better fit with Entities


ktuite:
There are a number of different paths that Forms can take (to be created, updated, or published) and the way these intersect with Datasets can be hard to keep track of. 

### The various lives of Forms

**Path 1: Create new form draft**
* `/projects/:projectId/forms`
* Uses `Forms.createNew(...)`
* Logs `form.create`

**Path 2: Create new form draft and publish immediately**
* `/projects/:projectId/forms?publish=true`
* Uses `Forms.createNew(...)` which can publish while inserting
* Logs `form.create` and `form.update.publish`

**Path 3: Copy published form def to make new draft**
* `/projects/:projectId/forms/:id/draft` (empty body)
* Uses `Forms.createVersion(...)` on partial pulled from previous form def as though it were uploaded
* Logs `form.update.draft.set`

**Path 4: Upload form to make new draft**
* `/projects/:projectId/forms/:id/draft` (with new xml)
* Uses `Form.createVersion(...)`
* Logs `form.update.draft.set`

**Path 5: Upload form to overwrite draft**
* `/projects/:projectId/forms/:id/draft` (with new xml, very similar to above)
* Uses `Form.createVersion(...)` which removes intermediate draft

**Path 6: Publish form while making a copy of the draft def with a different version**
* `/projects/:projectId/forms/:id/draft/publish?version=newVersion`
* Uses `Form.createVersion(...)` with publish=false flag
* Logs `form.update.draft.set`
* Uses `Forms.publish(...)`
* Logs `form.update.publish`

**Path 7: Publish draft form**
* `/projects/:projectId/forms/:id/draft/publish`
* Uses `Forms.publish(...)`
* Logs `form.update.publish`

**Path 8: Encrypting a form**
* ...
* Uses `Forms.createVersion` and either publishes or not 

Other things going on when creating a new version of a form:
* Clearing draft submissions
* Clearing orphaned form drafts
* The schema is checked against existing form schemas to avoid duplicating form fields
* If the schema has changed, there are checks about structural changes (down casting field types, etc.)
* Form attachments are dealt with
* Form attachments may be auto-linked to datasets on publish
* Enketo ids generated (for drafts and published forms)

There are some things that happen both for new forms and updated forms 
* Parsing of the form for the form fields: `getFormFields(xml)`
* Parsing of the dataset info out of the form: `getDataset(xml)`

[...]

PR #1143, opened 2024-05-10
Remove use of conditional publish from Forms.createNew

ktuite:
Closes https://github.com/getodk/central-backend/issues/1136

This PR consolidates Form publishing in one place, because Dataset publishing is tied to Form publishing. There were three different places to _possibly_ publish a Dataset (creating a new form, updating a form, just publishing a form) and it was too hard to think about which Form scenario led to which Dataset scenario. 

---

As of 5/10/24:

So far, I've removed `publish=true` from `Forms.createNew()`.  Any time creating a new form was also going to publish the form, the `Forms.publish()` is now called separately. 

One downside is that those lines of code aren't as cute and concise anymore. This only happened in one resource, in a test fixture, and in some tests.

Another downside is that the enketo test mock had a way to put it into an erroneous state for the next request, but a certain form POST request will trigger two enketo requests behind the scenes. I mucked with the enketo test mock to require you to remove the error state, but it's kind of ugly and required modifying multiple tests.

As of 5/14/24:

I decided that keeping `publish=true` in `Forms.createVersion()` had utility. There is one rare situation with encrypting existing forms where it really helps to publish concurrently, e.g. when there is both a published version and draft version of a form, and both active versions need to be encrypted. Because this has nothing to do with Datasets, Forms.createVersion does not have code to publish a dataset. I tried to clarify this with comments, too.

I also trie to make `Forms.createNew()` and `Forms.createVersion()` more similar. There are some things they both do (parse the form XML to get form fields and dataset info, insert form fields, insert attachments, do enketo things) and some things that are special (comparing against previous schemas for updated forms).

Checklist 
- [x] I think I want to move enketo draft creation from `_createNewDef` to `createVersion`
- [x] improve enketo test infrastructure 

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -87,7 +87,7 @@ const pushFormToEnketo = ({ projectId, xmlFormId, acteeId }, bound = undefined) 
 ////////////////////////////////////////////////////////////////////////////////
 // COMMON FORM UTILITY FUNCTIONS
 
-const _parseFormXml = (partial) => () => Promise.all([
+const _parseFormXml = (partial) => Promise.all([
   getFormFields(partial.xml),
   (partial.aux.key.isDefined() ? resolve(Option.none()) : getDataset(partial.xml)) // Don't parse dataset schema if Form has encryption key
 ]);
@@ -127,7 +127,7 @@ const createNew = (partial, project) => async ({ Actees, Datasets, FormAttachmen
   const keyId = await partial.aux.key.map(Keys.ensure).orElse(resolve(null));
 
   // Parse form XML for fields and entity/dataset definitions
-  const [fields, parsedDataset] = await Forms._parseFormXml(partial);
+  const [fields, parsedDataset] = await _parseFormXml(partial);
 
   // Check that meta field (group containing instanceId and name) exists
   await Forms.checkMeta(fields);
@@ -213,7 +213,7 @@ const createVersion = (partial, form, publish, duplicating = false) => async ({ 
   const keyId = await partial.aux.key.map(Keys.ensure).orElse(resolve(null));
 
   // Parse form fields and dataset/entity definition from form XML
-  const [fields, parsedDataset] = await Forms._parseFormXml(partial);
+  const [fields, parsedDataset] = await _parseFormXml(partial);
 
   // Compute the intermediate schema ID
   let schemaId;
@@ -789,7 +789,7 @@ const _newSchema = () => ({ one }) =>
 
 module.exports = {
   fromXls, pushDraftToEnketo, pushFormToEnketo,
-  _parseFormXml, _insertFormFields,
+  _insertFormFields,
   _createNew, createNew, _createNewDef, createVersion,
   publish, clearDraft,
   _update, update, _updateDef, del, restore, purge,
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -299,11 +299,11 @@ createVersion.audit.withResult = true;
 
 // TODO: we need to make more explicit what .def actually represents throughout.
 // for now, enforce an extra check here just in case.
-const publish = (form, sameAsCreate = false) => async ({ Forms, Datasets, FormAttachments }) => {
+const publish = (form, concurrentWithCreate = false) => async ({ Forms, Datasets, FormAttachments }) => {
   if (form.draftDefId !== form.def.id) throw Problem.internal.unknown();
 
-  // dont use publish sameAsCreate on already published forms
-  if (sameAsCreate && form.publishedAt != null) throw Problem.internal.unknown({ error: 'Attempting to immediately publish a form' });
+  // dont use publish concurrentWithCreate on already published forms
+  if (concurrentWithCreate && form.publishedAt != null) throw Problem.internal.unknown({ error: 'Attempting to immediately publish a form' });
 
   // Try to push the form to Enketo if it hasn't been pushed already. If doing
   // so fails or is too slow, then the worker will try again later.
@@ -311,9 +311,9 @@ const publish = (form, sameAsCreate = false) => async ({ Forms, Datasets, FormAt
     ? await Forms.pushFormToEnketo(form, 0.5)
     : {};
 
-  const publishedAt = sameAsCreate ? form.createdAt.toISOString() : (new Date()).toISOString();
+  const publishedAt = concurrentWithCreate ? form.createdAt.toISOString() : (new Date()).toISOString();
   const [f, fd] = await Promise.all([
-    Forms._update(form, { currentDefId: form.draftDefId, draftDefId: null, ...(sameAsCreate ? { updatedAt: null } : {}), ...enketoIds }),
+    Forms._update(form, { currentDefId: form.draftDefId, draftDefId: null, ...(concurrentWithCreate ? { updatedAt: null } : {}), ...enketoIds }),
     Forms._updateDef(form.def, { draftToken: null, enketoId: null, publishedAt }),
   ])
     .catch(Problem.translate(
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -18,7 +18,7 @@ const { generateToken } = require('../../util/crypto');
 const { unjoiner, extender, updater, equals, insert, insertMany, markDeleted, markUndeleted, QueryOptions } = require('../../util/db');
 const { resolve, reject, timebound } = require('../../util/promise');
 const { splitStream } = require('../../util/stream');
-const { construct, noop, pickAll } = require('../../util/util');
+const { construct, pickAll } = require('../../util/util');
 const Option = require('../../util/option');
 const Problem = require('../../util/problem');
 
@@ -84,6 +84,21 @@ const pushFormToEnketo = ({ projectId, xmlFormId, acteeId }, bound = undefined) 
   return timeboundEnketo(enketo.create(path, xmlFormId, token), bound);
 };
 
+////////////////////////////////////////////////////////////////////////////////
+// COMMON FORM UTILITY FUNCTIONS
+
+const _parseFormXml = (partial) => () => Promise.all([
+  getFormFields(partial.xml),
+  (partial.aux.key.isDefined() ? resolve(Option.none()) : getDataset(partial.xml)) // Don't parse dataset schema if Form has encryption key
+]);
+
+const _insertFormFields = (fields, formId, schemaId) => async ({ run }) => {
+  // Combine form fields as parsed from the form XML with the top level formId
+  // and the schemaId (rather than formDefId)
+  const ids = { formId, schemaId };
+  const fieldsForInsert = fields.map((field) => new Form.Field(Object.assign(field, ids)));
+  await run(insertMany(fieldsForInsert));
+};
 
 ////////////////////////////////////////////////////////////////////////////////
 // CREATING NEW FORMS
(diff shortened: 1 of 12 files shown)


lib/model/query/forms.js
```diff
@@ -319,18 +306,21 @@ createVersion.audit.withResult = true;
 
 // TODO: we need to make more explicit what .def actually represents throughout.
 // for now, enforce an extra check here just in case.
-const publish = (form) => async ({ Forms, Datasets, FormAttachments }) => {
+const publish = (form, sameAsCreate = false) => async ({ Forms, Datasets, FormAttachments }) => {
```
ktuite, line 309: I don't love the `sameAsCreate` name. I was just trying to get at the situation where you call publish() and it's meant to be combined with the previous create operation. I don't know if this is actually important to preserve... but one case where it might come up is seeing `updatedAt = null` on a form that has been created+published in one go.

test/integration/api/datasets.js
```diff
@@ -2214,10 +2214,7 @@ describe('datasets and entities', () => {
           </h:head>
         </h:html>`;
 
-        it('should NOT autolink on upload new form and simultaneously publish', testService(async (service) => {
```
ktuite, line 2217: The code was too messy for me to implement this autolinking before, so I had this negative test case lying around.

test/integration/api/forms/forms.js
```diff
@@ -74,8 +74,12 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
             .set('Content-Type', 'application/xml')
             .expect(409)
             .then(({ body }) => {
-              body.details.fields.should.eql([ 'projectId', 'xmlFormId', 'version' ]);
-              body.details.values.should.eql([ '1', 'simple', '' ]);
+              // eslint-disable-next-line no-multi-str
+              body.message.should.eql("You tried to publish the form 'simple' with version '', \
```
ktuite, line 78: The error response used to look like
```
body {
  message: 'A resource already exists with projectId,xmlFormId,version value(s) of 1,simple,.',
  code: 409.3,
  details: {
    fields: [ 'projectId', 'xmlFormId', 'version' ],
    values: [ '1', 'simple', '' ],
    table: 'forms'
  }
}
```

Because the form database insert also included setting publishedAt, and that violated a constraint. 

Now the constraint is violated at a different time in a different way. The message is in some ways clearer and in other ways wrong because it's referring to the .../formid/draft/publish endpoint.

test/integration/api/forms/forms.js
```diff
@@ -295,7 +299,35 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
             .expect(200)
             .then(({ body }) => {
               body.publishedAt.should.be.a.recentIsoDate();
-            })))));
+              body.publishedAt.should.eql(body.createdAt);
+              (body.updatedAt == null).should.equal(true);
```
ktuite, line 303: I just wanted to check that `updatedAt` would be null when the form is published in the same request as it is created. The test below is the separate case. 

There might be other tests that already exist that are very similar.

test/integration/api/forms/forms.js
```diff
@@ -400,7 +436,7 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
         should.not.exist(body.enketoOnceId);
       }));
 
-      it('should wait for Enketo only briefly @slow', testService(async (service) => {
+      it.skip('should wait for Enketo only briefly @slow', testService(async (service) => {
         const asAlice = await service.login('alice');
         global.enketo.wait = (done) => { setTimeout(done, 600); };
         const { body } = await asAlice.post('/v1/projects/1/forms?publish=true')
```
ktuite, line 442: i can't get enketo to time out for 2 requests in a row (one for draft and one for publish, because the draft token request will now also run in this situation)

test/util/enketo.js
```diff
@@ -44,7 +44,7 @@ const reset = () => {
 const request = () => {
   global.enketo.callCount += 1;
   const options = { ...global.enketo };
-  Object.assign(global.enketo, without(['callCount'], defaults));
+  Object.assign(global.enketo, without(['callCount', 'state'], defaults));
```
ktuite, line 47: There's probably a better way to deal with this enketo error state and maintaining it for two requests in a row.

lib/model/query/forms.js
```diff
@@ -350,6 +338,8 @@ const publish = (form) => async ({ Forms, Datasets, FormAttachments }) => {
       await FormAttachments.update(form, attachment.get(), null, dataset.id);
     }
   }
+
+  return new Form(f, { def: new Form.Def(fd) });
```
ktuite, line 342: Needed to return the published form

lib/model/query/forms.js
matthew-white: I don't know that it's a big deal, but if it's not much code, it seems nice to preserve the previous behavior.

In terms of the name, `sameAsCreate` sounds OK to me. Other ideas: `newForm`, `isNewForm`, `afterCreate`, `concurrentWithCreate`.

test/integration/fixtures/02-forms.js
```diff
@@ -6,18 +6,25 @@ const forms = [ simple, withrepeat ];
...
     // in order to maintain existing tests.
-    const [{ actorId }] = await Assignments.getByActeeAndRoleId(acteeId, formview);
+    const [{ actorId }] = await Assignments.getByActeeAndRoleId(form.acteeId, formview);
     await Assignments.revokeByActorId(actorId);
   }
   /* eslint-enable no-await-in-loop */
+
+  // Reset enketo call count and error state to not affect tests
+  global.enketo.callCount = 0;
+  global.enketo.state = undefined;
```
matthew-white, line 28: This is making me think that the `reset` function that's defined in test/util/enketo.js should get added to `global.enketo`. Right now, I'm pretty sure that function is just used in test/integration/setup.js, but if it were added to `global.enketo`, it could be used in more places (here, as well as any test that wants to reset the mock partway through the test).

test/util/enketo.js
matthew-white: I do think it makes it slightly harder to reason about the mock if some properties are automatically reset and others aren't. Also, it looked like there was one test where you didn't want `wait` to get reset. What about adding a new property to `global.enketo` named something like `autoReset` where if `autoReset` is `true`, this line is called, but if it's set to `false`, then nothing is reset (not `state`, not `wait`)? `autoReset` would then itself get reset after every test (when the `reset()` function is called). Just an idea!

test/integration/api/forms/forms.js
matthew-white: One thing you could do is remove the last one or two sentences of the 409.6 (`Problem.user.uniquenessViolation`) message. I think the resulting message would work for both endpoints. I'm pretty sure that Frontend shows its own message for a 409.6 in the `FormDraftPublish` component, so a change to the Backend message would only affect API users.

Alternatively, in `Forms.publish()`, you could prevent the translation of `Problem.user.uniquenessViolation` to `Problem.user.versionUniquenessViolation` when `sameAsCreate` is `true`. That should result in the same behavior as before.

Another option: you could pass `sameAsCreate` to `Problem.user.versionUniquenessViolation()`, which could then return a different message depending on whether `sameAsCreate` is `true` or `false`.

lib/model/query/forms.js
ktuite: I like `concurrentWithCreate`

test/integration/api/forms/forms.js
ktuite: I like the idea of just removing the end of the error message.

test/util/enketo.js
ktuite: Tried out this `autoReset` approach and it seems useful.

test/integration/fixtures/02-forms.js
ktuite: I have one or two places it tests that I might like to call a reset() function like this, but i was not sure how to get it added to `global.enketo`.

test/integration/api/audits.js
```diff
@@ -636,9 +636,9 @@ describe('/audits', () => {
         .then(({ body }) => {
           body.length.should.equal(4);
           body.map(a => a.action).should.eql([
+            'dataset.create',
```
ktuite, line 639: Not quite sure why this is after form publish when it happens within form publish.

test/integration/api/forms/draft.js
```diff
@@ -137,9 +137,11 @@ describe('api: /projects/:id/forms (drafts)', () => {
         global.enketo.state = 'error';
         await asAlice.post('/v1/projects/1/forms/simple/draft').expect(200);
         global.enketo.callCount.should.equal(1);
+        // Reset enketo mock (will also reset callCount)
+        global.enketo.reset();
```
ktuite, line 141: dont need to reset

test/integration/api/forms/draft.js
```diff
@@ -1155,10 +1157,13 @@ describe('api: /projects/:id/forms (drafts)', () => {
         should.not.exist(draft.enketoId);
         should.not.exist(draft.enketoOnceId);
 
+        // Reset enketo mock (including callCount)
+        global.enketo.reset();
```
ktuite, line 1161: dont need this reset either

test/integration/api/forms/draft.js
```diff
@@ -1212,12 +1217,14 @@ describe('api: /projects/:id/forms (drafts)', () => {
         should.not.exist(v1.enketoId);
         should.not.exist(v1.enketoOnceId);
 
+        // Reset enketo mock (including callCount)
+        global.enketo.reset();
```
ktuite, line 1221: don't need reset, call counts will also be reset to how they were

test/integration/api/forms/draft.js
```diff
@@ -1263,11 +1270,14 @@ describe('api: /projects/:id/forms (drafts)', () => {
           .expect(200);
         should.not.exist(beforeWorker.enketoId);
         should.not.exist(beforeWorker.enketoOnceId);
+        global.enketo.callCount.should.equal(2);
+
+        // Reset enketo mock (including callCount)
+        global.enketo.reset();
```
ktuite, line 1276: dont need reset here

test/integration/api/forms/forms.js
```diff
@@ -295,7 +295,35 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
             .expect(200)
             .then(({ body }) => {
               body.publishedAt.should.be.a.recentIsoDate();
-            })))));
+              body.publishedAt.should.eql(body.createdAt);
+              (body.updatedAt == null).should.equal(true);
+            })
+            .then(() => asAlice.get('/v1/audits')
```
ktuite, line 301: ?action=forms to filter form audit events

test/integration/api/forms/forms.js
```diff
@@ -411,20 +442,37 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
...
           .send(testData.forms.simple2)
           .expect(200);
 
+        // reset enketo mock to its default behavior
+        // also resets callCount
+        global.enketo.reset();
+
         // Second request, from the worker
-        global.enketo.callCount.should.equal(1);
+        global.enketo.callCount.should.equal(0);
```
ktuite, line 473: dont need this assertion

matthew-white (approved):

test/integration/api/audits.js
ktuite: it had to do with not using `publish.audit.withResult`, which made the logging happen earlier, i think.

lib/model/query/forms.js
```diff
@@ -264,32 +268,28 @@ const createVersion = (partial, form, publish, duplicating = false) => async ({
...
-  // Update corresponding form
+  // Prepare the form fields
+  if (!match)
+    await Forms._insertFormFields(fields, form.id, schemaId);
+
+  // Insert attachments
+  await FormAttachments.createVersion(partial.xml, form, savedDef, partial.xls.itemsets, publish);
+
+  // Update datasets and properties, if defined
+  if (parsedDataset.isDefined()) {
```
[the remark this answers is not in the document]
ktuite: It turns out `createOrMerge` is needed even when duplicating, because of how it links up the current form def to the dataset.

PR #1144, opened 2024-05-14
ci: run check-file-headers in GitHub Actions

alxndrsn:
This is the only step present in `.circleci/config.yml` that is currently missing from `.github/workflows/standard-suite.yml`.

### DIFF
--- a/.github/workflows/standard-suite.yml
+++ b/.github/workflows/standard-suite.yml
@@ -28,6 +28,7 @@ jobs:
       with:
         node-version: 20.10.0
         cache: 'npm'
+    - run: make check-file-headers
     - run: npm ci --legacy-peer-deps
     - run: node lib/bin/create-docker-databases.js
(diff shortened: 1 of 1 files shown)


matthew-white:
This has probably been discussed before, but I'm wondering what the benefit is of running the same check in CircleCI and Github Actions. It's not the case that the check of file headers would fail in one but not the other, right?

Could we forego CircleCI and just use GitHub Actions? Or in the reverse direction, could we remove standard-suite.yml in favor of CircleCI? I'm just curious about whether both sets of checks are needed.

alxndrsn:
> It's not the case that the check of file headers would fail in one but not the other, right?

This sounds correct.

Some arguments for adding here:

* CircleCI does not seem to always run builds on PRs
* many other tests are only run on GHA, so it's easy to miss failures which only occur on CircleCI

PR #1149, opened 2024-06-01
Don't encode %24 in docs

lognaturel:
I noticed that there's a mix of literal $ and %-encoded %24 used. The %24 are not rendered correctly. See for example the request table for https://docs.getodk.org/central-api-odata-endpoints/#data-document

Rendered using `make api-docs` (🫡 @sadiqkhoja), spot checked and saw $ as expected.

I considered modifying the docs code to decode values but given that several literal $ already exist and that literal values work without replacement, that didn't feel like the best option.

Could break some docs.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -6358,7 +6358,7 @@ paths:
         schema:
           type: boolean
         example: "true"
-      - name: '%24filter'
+      - name: '$filter'
         in: query
         description: If provided, will filter responses to those matching the given
           OData query. Only [certain fields](/central-api-odata-endpoints/#data-document)
@@ -6461,7 +6461,7 @@ paths:
         schema:
           type: boolean
         example: "true"
-      - name: '%24filter'
+      - name: '$filter'
         in: query
         description: If provided, will filter responses to those matching the given
           OData query. Only [certain fields](/central-api-odata-endpoints/#data-document)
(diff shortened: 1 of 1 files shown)


docs/api.yaml
```diff
@@ -2263,8 +2263,6 @@ paths:
           content:
             application/json:
               schema:
-                type: object
-                description: Standard Response
```
matthew-white, line 2267: More of a question than anything else, but I don't see a top-level `description` in the `Project` schema. Did you remove it here because it just doesn't add much? Have you noticed any trend whereby we should generally be adding `description` in more places or removing it in more places?

matthew-white:
> The %24 are not rendered correctly.

> saw $ as expected.

I think `%24` was there intentionally with the thought that `$` should be encoded in query strings. But you're right that there's currently a mix of `%24` and `$`. I've also never had trouble with using `$` when using the API. We usually use `$` instead of `%24` in Backend tests.

I think the docs are more readable with `$` instead of `%24`. There are other places where we don't encode query strings. For example, for `$filter=__system/submissionDate le 2020-01-31`, the `/` and space should be encoded, but I think it's fine the way it is. I think it's on the API user to encode query strings as needed. If it ever becomes confusing, we can always make a note of it.

matthew-white (approved):

matthew-white:
> I think `%24` was there intentionally

Meaning I think we intended to display `%24` as `%24`. But I agree that it's better as `$`.

lognaturel:
> I think %24 was there intentionally with the thought that $ should be encoded in query strings

I see! That actually makes a lot more sense than I thought. Hmm, like you say, it does feel like it's up to the user to do their own escaping.

PR #1150, opened 2024-06-03
Delete and purge individual submissions

ktuite:
Completes the Backend portion of https://github.com/getodk/central/issues/667

Adds submission delete endpoint that soft-deletes a single submission, and a mechanism to purge submissions deleted for more than 30 days. 

Done:
- [x] submission delete endpoint
- [x] purge mechanism
- [x] some tests of basic 404s on soft deleted submissions and basic purging
- [x] migration to add `submission.delete` verb
- [x] migration to get submission comments to cascade delete (i think) (already done)
- [x] more tests (see some TODOs in code) to check that everything associated with a submission is purged (versions, select multiple fields, comments, unattached blobs, etc)
- [x] more tests to ensure that deleted submissions are not included anywhere (in api, odata, other exports)
- [x] double check that deleted submissions that were entity sources do the right thing
- [x] figure out if there are notes in audits that need to be redacted (i think so)

Still to do (sept 9 2024)
- [x] make sure client audits processed from submission attachment blobs get purged properly (https://github.com/getodk/central-backend/issues/1174)
- [x] write a test for checking the submission.delete audit event (its acteeId is the form it was in)
- [x] undelete endpoint? 
- [x] submission purge cli and cron job (combine with form purge?)


lib/model/query/forms.js
```diff
@@ -415,6 +415,15 @@ with redacted_audits as (
...
     and ${_trashedFilter(force, id, projectId, xmlFormId)}
+  ), deleted_client_audits as (
+    delete from client_audits
+    using submission_attachments, submission_defs, submissions, forms
+    where client_audits."blobId" = submission_attachments."blobId"
+    and submission_attachments."submissionDefId" = submission_defs.id
+    and submission_attachments."isClientAudit" = true
+    and submission_defs."submissionId" = submissions.id
+    and submissions."formId" = forms.id
+    and ${_trashedFilter(force, id, projectId, xmlFormId)}
```
ktuite, line 426: What if two different submission attachments share the same client audit blob data? Would this delete a client audit from a non-deleted form because it shared the same blobId? 

Is it fine if the client_audits table has something extra deleted from it because it will be processed on the fly anyway?

lib/model/query/submissions.js
```diff
@@ -386,14 +398,87 @@ const getForExport = (formId, instanceId, draft, options = QueryOptions.none) =>
   maybeOne(_export(formId, draft, [], options.withCondition({ 'submissions.instanceId': instanceId })))
     .then(map(_exportUnjoiner));
 
+////////////////////////////////////////////////////////////////////////////////
+// DELETE SUBMISSION
+
+const del = (submission) => ({ run }) =>
+  run(markDeleted(submission));
+
+del.audit = (submission, form) => (log) => log('submission.delete', submission.with({ acteeId: form.acteeId }), { instanceId: submission.instanceId });
```
ktuite, line 407: `submission.with({ acteeId: form.acteeId }),` can probably just be `{ acteeId: form.acteeId }`

remove `submission.delete` and `submission.restore` from non-verbose filter and include in submission filter

lib/model/query/submissions.js
```diff
@@ -386,14 +398,87 @@ const getForExport = (formId, instanceId, draft, options = QueryOptions.none) =>
...
+      join forms on forms.id = submissions."formId"
+      where ${_trashedFilter(force, projectId, xmlFormId, instanceId)}
+    ) as del_sub_count
+    where del_sub_count.count > 0
+  ), deleted_submissions as (
+    delete from submissions
+    using forms
+    where submissions."formId" = forms.id
+    and ${_trashedFilter(force, projectId, xmlFormId, instanceId)}
+    returning submissions.*
```
ktuite, line 469: `returning 1`

lib/model/query/submissions.js
```diff
@@ -386,14 +398,87 @@ const getForExport = (formId, instanceId, draft, options = QueryOptions.none) =>
...
+    using submission_attachments, submission_defs, submissions, forms
+    where client_audits."blobId" = submission_attachments."blobId"
+    and submission_attachments."submissionDefId" = submission_defs.id
+    and submission_attachments."isClientAudit" = true
+    and submission_defs."submissionId" = submissions.id
+    and submissions."formId" = forms.id
+    and ${_trashedFilter(force, projectId, xmlFormId, instanceId)}
+  ), purge_audits as (
+    insert into audits ("action", "loggedAt", "processed", "details")
+    select 'submission.purge', clock_timestamp(), clock_timestamp(), jsonb_build_object('submissions_deleted', "count")
```
ktuite, line 456: Check if this should be filtered out of non-verbose audit logs, too. No actee id. Check what forms.purge does?

Will need small frontend change, too.

test/integration/other/form-purging.js
```diff
@@ -90,6 +72,27 @@ describe('query module form purge', () => {
...
+
+    await asAlice.delete('/v1/projects/1/forms/withrepeat')
+      .expect(200);
+
+    await container.Forms.purge(true);
+
+    await asAlice.get('/v1/audits')
+      .then(({ body }) => {
+        const acteeIds = body.filter((a) => a.action === 'form.purge').map(a => a.acteeId).sort();
+        acteeIds.should.eql([simpleForm.acteeId, repeatForm.acteeId].sort());
```
ktuite, line 92: Could use equalInAnyOrder

test/integration/api/submissions.js
```diff
@@ -1399,6 +1399,103 @@ describe('api: /forms/:id/submissions', () => {
...
+      const asAlice = await service.login('alice');
+      await asAlice.post('/v1/projects/1/forms/simple/draft');
+      await asAlice.post('/v1/projects/1/forms/simple/draft/submissions')
+        .send(testData.instances.simple.one)
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+      // draft submission delete resource does not exist
+      await asAlice.delete('/v1/projects/1/forms/simple/draft/submissions/one')
+        .expect(404);
+    }));
```
ktuite, line 1439: You can probably have a draft submission with the same instance id as a deleted non-draft submission. Does that have any weird implications? Should some queries check for draft or is it ok because they check deletedAt!=null and you can never delete a draft?

test/integration/api/submissions.js
```diff
@@ -1399,6 +1399,103 @@ describe('api: /forms/:id/submissions', () => {
...
+  });
+
+  describe('/:instanceId RESTORE', () => {
+    it('should reject if the submission has not been deleted', testService((service) =>
+      service.login('alice', (asAlice) =>
```
ktuite, line 1444: should submit a submission here

test/integration/api/submissions.js
```diff
@@ -1502,6 +1599,39 @@ describe('api: /forms/:id/submissions', () => {
...
+      csv[0].should.eql([ 'SubmissionDate', 'meta-instanceID', 'name', 'age', 'KEY', 'SubmitterID', 'SubmitterName', 'AttachmentsPresent', 'AttachmentsExpected', 'Status', 'ReviewState', 'DeviceID', 'Edits', 'FormVersion' ]);
+      csv[1].shift().should.be.an.recentIsoDate();
+      // eslint-disable-next-line comma-spacing
+      csv[1].should.eql([ 'three','Chelsea','38','three','5','Alice','0','0','','','','0','' ]);
+      csv[2].shift().should.be.an.recentIsoDate();
+      // eslint-disable-next-line comma-spacing
+      csv[2].should.eql([ 'one','Alice','30','one','5','Alice','0','0','','','','0','' ]);
+      csv[3].should.eql([ '' ]);
+    }));
+
```
ktuite, line 1634: Biggest risk: some area that we have forgotten to exclude deleted submissions from? Scan submission query module to think of anything.

ktuite (commented):
Interactive review

test/integration/other/submission-purging.js
```diff
@@ -0,0 +1,510 @@
...
+
+      await asAlice.delete('/v1/projects/1/forms/simple/submissions/one');
+
+      await run(sql`update submissions set "deletedAt" = '1999-1-1' where "deletedAt" is not null`);
+
+      await Submissions.purge();
+
+      const counts = await Promise.all([
+        oneFirst(sql`select count(*) from submissions`),
+        oneFirst(sql`select count(*) from submissions`)
```
ktuite, line 101: submission_defs

test/integration/other/submission-purging.js
```diff
@@ -0,0 +1,510 @@
...
+      // More recent delete, within 30 day window
+      await asAlice.delete('/v1/projects/1/forms/simple/submissions/three');
+
+      const purgeCount = await Submissions.purge();
+      purgeCount.should.equal(2);
+
+      // Remaining submissions not recently deleted
+      const counts = await Promise.all([
+        oneFirst(sql`select count(*) from submissions`),
+        oneFirst(sql`select count(*) from submissions`)
```
ktuite, line 139: submission_defs

matthew-white (approved):

lib/model/query/forms.js
ktuite: This IS what happens (client audit blobs could potentially be shared and this would delete rows for a non-deleted submission). I don't like that this happens... but I don't like the other option we talked about either (having the client_audits purged after or as a result of purging a blob). My hope is that client audits are basically never going to be duplicated between submissions... they have timestamps of user actions after all, and if some client audits are deleted that shouldn't have been removed, the client audits will still export just fine via the ad-hoc processing.

PR #1151, opened 2024-06-06
Fix bug with insertMany, createdAt, and entity timestamp

ktuite:
I noticed this after adding a property to the Entity frame and then trying to "insertMany" entities. 

It's not the `createdAt` timestamp type that gets rearranged in the insertMany query, it's the _last_ type because `indexOf(['createdAt'], Type.insertfields` (with the brackets) always returns -1. `indexOf('createdAt',...` works as intended. Fixing this also led me to fix the typo of `timestampz` in the Entity frame that wasn't being surfaced before.

Added a new test, which I verified failed before the code change.

I don't think there was previously a scenario to trigger this bug before. It didn't come up if the last type in a Frame was a timestamp, or if the Frame didn't have a createdAt. 

### DIFF
--- a/lib/model/frames/entity.js
+++ b/lib/model/frames/entity.js
@@ -12,7 +12,6 @@
 const { embedded, fieldTypes, Frame, readable, table } = require('../frame');
 const { extractEntity, normalizeUuid, extractLabelFromSubmission, extractBaseVersionFromSubmission } = require('../../data/entity');
 
-// These Frames don't interact with APIs directly, hence no readable/writable
 class Entity extends Frame.define(
   table('entities', 'entity'),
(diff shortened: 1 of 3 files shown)


lib/model/frames/entity.js
```diff
@@ -27,7 +27,7 @@ class Entity extends Frame.define(
     'int4', 'int4',
     'conflictType',
     'timestamptz',
-    'timestamptz', 'timestampz',
+    'timestamptz', 'timestamptz',
```
sadiqkhoja, line 30: the comment before the class definition has became obsolete - not by this change though

sadiqkhoja (approved):

PR #1158, opened 2024-06-19
resources/odata: don't crash with surprising instanceId

### LINKED ISSUE #1157: Backend crash when opening hostile-named submission detail
1. Go to here: https://staging.getodk.cloud/#/projects/1/forms/'%3D%2B%2F*-451%25%2F%25/submissions
2. Open the submission detail page of any of the strangely named submissions
3. Server crash, user sees a 404

```
Problem [Error]: Could not find the resource you were looking for.
    at Object.result [as notFound] (lib/util/problem.js:45:31)
    at filterFields (lib/formats/odata.js:610:48)
    at lib/formats/odata.js:660:94
```

alxndrsn:
Closes #1157


alxndrsn:
There may be a more general problem of potential unhandled errors in sync code within promise chains.  A more general way of detecting & fixing or avoiding these issues could be helpul.

Most of the e2e test changes may be brought in by https://github.com/getodk/central-backend/pull/1071.

lib/resources/odata.js
```diff
@@ -46,13 +46,13 @@ module.exports = (service, endpoint) => {
...
-      getForm(Forms, auth, params)
-        .then((form) => Promise.all([
-          Forms.getFields(form.def.id).then(selectFields(query, getTableFromOriginalUrl(originalUrl))),
-          Submissions.getForExport(form.id, getUuid(params.uuid), draft).then(getOrNotFound), // may require s3 blob handling
-        ])
-          .then(([fields, row]) => singleRowToOData(fields, row, env.domain, originalUrl, query))));
+    const singleRecord = endpoint.odata.json(async ({ Forms, Submissions, env }, { auth, params, query, originalUrl }) => {
+      const form = await getForm(Forms, auth, params);
+      const allFields = await Forms.getFields(form.def.id);
+      const selectedFields = await selectFields(query, getTableFromOriginalUrl(originalUrl))(allFields);
```
sadiqkhoja, line 52: I think we can remove `await` from here since it is synchronous/non-promise function 

```suggestion
      const selectedFields = selectFields(query, getTableFromOriginalUrl(originalUrl))(allFields);
```

test/e2e/standard/test.js
```diff
@@ -0,0 +1,82 @@
...
+
+  function uploadSubmission(submissionId) {
+    const xmlTemplate = fs.readFileSync('submission.xml', { encoding: 'utf8' });
+    const tempFile = 'TODO-generate-proper-tempfile-name.xml';
+    const formXml = xmlTemplate
+      .replace('{{submissionId}}', submissionId)
+      .replace('{{formId}}', xmlFormId)
+      .replace('{{formVersion}}', xmlFormVersion);
+    fs.writeFileSync(tempFile, formXml);
+    return api.apiPostFile(`projects/${projectId}/forms/${encodeURIComponent(xmlFormId)}/submissions?deviceID=testid`, tempFile);
```
sadiqkhoja, line 80: Can we overload `apiPostFile` to accept something similar to [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) object as well so that we avoid the need to write any file on the disk in this case? It could be as simple as 

```
{
   "filename": "submission.xml",
   "data": Buffer.from(content)
}
```

test/e2e/standard/run-tests.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -eu
+set -o pipefail
+
+serverUrl="http://localhost:8383"
+userEmail="x@example.com"
+userPassword="secret1234"
+
+log() { echo "[test/e2e/standard/run-tests] $*"; }
+
```
sadiqkhoja, line 9: a log entry here would helpful. When I executed this script for the first time, I didn't see anything on the console for few seconds, which made me wonder if anything is happening at all

sadiqkhoja (approved):
Few minor improvements, otherwise this solves the original problem.

PR #1161, opened 2024-07-02
Add publicLinks count to form extended metadata

matthew-white:
This PR completes the Backend portion of getodk/central#671. We need just one more piece of extended metadata: the count of public links for the form.

New tests.

I tried to follow a similar approach as other form extended metadata. The main change is modifying a query.

The change is mostly additive, so the risk of regression should be low.

Done.

### DIFF
--- a/lib/model/frames/form.js
+++ b/lib/model/frames/form.js
@@ -148,7 +148,8 @@ Form.Extended = class extends Frame.define(
   'excelContentType',   readable,
   // counts of submissions in various review states
   'receivedCount',                      'hasIssuesCount',
-  'editedCount',                        'entityRelated',  readable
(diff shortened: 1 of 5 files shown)


lib/model/frames/form.js
```diff
@@ -160,7 +161,8 @@ Form.Extended = class extends Frame.define(
         edited: this.editedCount || 0
       },
       lastSubmission: this.lastSubmission,
-      excelContentType: this.excelContentType
+      excelContentType: this.excelContentType,
+      publicLinks: this.publicLinks ?? 0
```
matthew-white, line 165: Let's make it a consistent use of either `||` or `??`.

ktuite (approved):

PR #1171, opened 2024-08-20
Improvements to offline entity backlog processing

ktuite:
This is meant to be a precursor to another PR for issue https://github.com/getodk/central/issues/682

The main changes here are
* Migration to add more useful columns to submission backlog
* Reordering and reorganizing code in entity query module
* Changing flow of `Entities._updateEntity` to try to determine the base version of an entity _before_ it looks up the existing server version, because this helps with the logic of holding submissions for later processing, and will eventually help with the force-processing case. 
* Removes `branchId` from tests about creating entities

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -322,7 +322,7 @@ const _computeBaseVersion = (eventId, dataset, clientEntity, submissionDef) => a
     else if (clientEntity.def.baseVersion === clientEntity.def.trunkVersion) // Start of an offline branch
       condition = { version: clientEntity.def.baseVersion };
     else // middle of an offline branch
-      condition = { branchBaseVersion: clientEntity.def.baseVersion - 1 };
+      condition = { branchId: clientEntity.def.branchId, branchBaseVersion: clientEntity.def.baseVersion - 1 };
 
     const baseEntityVersion = await Entities.getDef(dataset.id, clientEntity.uuid, new QueryOptions({ condition }));
 
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -86,8 +86,7 @@ select ins.*, def.id as "entityDefId" from ins, def;`)
       new Entity(entityData, {
         currentVersion: new Entity.Def({
           id: entityDefId,
-          entityId: entityData.id,
-          branchId: partial.def.branchId
+          entityId: entityData.id
         })
       }));
 };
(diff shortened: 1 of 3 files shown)


lib/model/query/entities.js
```diff
@@ -415,13 +409,15 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Audits, Dataset
       throw Problem.user.entityActionNotPermitted({ action, permitted: permittedActions });
```
ktuite, line 409: This will probably need to change to prevent updates-applied-as-creates from happening on forms that dont allow creates. Double check when returning to force-processing work.

lib/model/query/entities.js
```diff
@@ -465,6 +465,40 @@ const processSubmissionEvent = (event, parentEvent) => (container) =>
...
+
+// Check for a currently-held submission by id
+const _checkHeldSubmission = (submissionId) => ({ maybeOne }) => maybeOne(sql`
+  SELECT * FROM entity_submission_backlog
+  WHERE "submissionId"=${submissionId}`);
+
+// Look up the next submission to process
+const _getNextHeldSubmissionInBranch = (entityUuid, branchId, branchBaseVersion) => ({ maybeOne }) => (
+  (branchId == null)
+    ? maybeOne(sql`
```
ktuite, line 486: could there be multiple? could this query fail?

matthew-white (approved):

ktuite (commented):
Add another test scenario?

lib/model/query/entities.js
```diff
@@ -355,12 +289,71 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
...
+    }
+
+    return maybeDef.get();
+
+  } else {
+    // there is a branchId, look up the appropriate base def
+
+    let condition;
+    if (clientEntity.def.baseVersion === 1) // Special case
+      condition = { version: 1 };
```
matthew-white, line 321: I mentioned during code review that I thought that once we introduce force-processing, it won't always be the case that if an entity has a branch base version of 1, it will have a server base version of 1. Thinking about it more, I still sort of think that's the case, but I also don't think it'd be a big deal to keep it the way it is. In any case, let me throw out my example. I think there's only an issue if the entity was created offline, then immediately also updated offline at least twice, then the resulting submissions are processed out of order. For example:

- An entity is created offline. Before being sent to the server, it is updated offline 3 times.
- There's a problem sending the entity creation and the 1st update, but the 2nd and 3rd updates are sent successfully.
- After some time, the 2nd and 3rd updates are force-processed. On the server, the 2nd update ends up with version=1, baseVersion=null, branchBaseVersion=2. The 3rd update ends up with version=2, baseVersion=1, branchBaseVersion=3.
- The 1st update is eventually sent successfully and processed. The 1st update has branchBaseVersion=1, but I think ideally, it would get a server baseVersion=2, not baseVersion=1 as it does here. We've discussed how for now, force-processing on its own (without interleaving) should not cause a conflict. If the 1st update were given a server baseVersion of 1 instead of 2, that would be a conflict.

The logic that I think would be ideal would be something like: "If `clientEntity.def.baseVersion === 1`, and `clientEntity.def.trunkVersion == null`, then find the latest version from the branch that's been processed. Use that version as the base version."

The reason I think it's not a big deal is that in this case, the entity creation already has a high potential for causing a conflict. Unless the entity creation is processed at the very end, it will cause a conflict, because it will be applied as an update, and that update will cause the branch to stop being contiguous. In the example above, if the processing order were 2nd update / 3rd update / entity creation / 1st update, the entity creation would already cause the 1st update to be a conflict, unless we have extra logic that associates the entity creation with the branch. Without that extra logic, the entity creation would look like an interleaving update from outside the branch. I'm thinking that maybe we should just be prepared for conflicts in this case: if an entity is created offline, then updated offline at least twice, and the entity creation is processed late, then there is a significant chance of conflict.

Anyway, just wanted to share this example and thoughts in case they're helpful as you're thinking about force-processing.

PR #1172, opened 2024-08-20
Force-process entity submissions held in backlog

ktuite:
Main PR to close https://github.com/getodk/central/issues/682 following PR https://github.com/getodk/central-backend/pull/1171


ktuite:
Thoughts on force-processing when entities can be soft-deleted:
* Don't force-apply things in the backlog that are about soft-deleted entities.
* Consider deleting things from the backlog when the entity that they reference are purged. (When entity purging happens)

lib/model/query/entities.js
```diff
@@ -351,7 +378,7 @@ const _getFormDefActions = (oneFirst, datasetId, formDefId) => oneFirst(sql`
 // so any errors can be rolled back and logged as an entity processing error.
 const _processSubmissionEvent = (event, parentEvent) => async ({ Audits, Datasets, Entities, Submissions, Forms, oneFirst }) => {
   const { submissionId, submissionDefId } = event.details;
-  // TODO: check parentEvent details to determine if this is a forced reprocessing or not
+  const forceOutOfOrderProcessing = parentEvent?.details?.force === true;
```
ktuite, line 381: `forceOutOfOrderProcessing` could be a 3rd argument (defaulting to false) or something instead of being passed in within the `parentEvent`.

lib/model/query/entities.js
```diff
@@ -171,20 +172,25 @@ createVersion.audit = (updatedEntity, dataset, partial, subDef) => (log) => {
...
   if ((dataset.approvalRequired && event.details.reviewState !== 'approved') ||
       (!dataset.approvalRequired && event.action === 'submission.update')) // don't process submission if approval is not required and submission metadata is updated
     return null;
 
-  // TODO: auto-generate a label if forced and if the submission doesn't provide one
+  // Auto-generate a label if forced and if the submission doesn't provide one
+  if (forceOutOfOrderProcessing && entityData.system.label == null) {
+    // eslint-disable-next-line no-param-reassign
+    entityData.system.label = 'auto generated';
+  }
```
ktuite, line 193: Can label be '' here instead of null? If the `<label>` tag in the submission is empty?

test/integration/api/offline-entities.js
```diff
@@ -780,4 +780,457 @@ describe('Offline Entities', () => {
...
+          body.currentVersion.version.should.equal(1);
+          body.currentVersion.data.should.eql({ status: 'arrived' });
+          body.currentVersion.label.should.eql('auto generated');
+          body.currentVersion.branchId.should.equal(branchId);
+
+
+          // This is the first version of the entity so there should be no base or trunk versions
+          should.not.exist(body.currentVersion.trunkVersion);
+          should.not.exist(body.currentVersion.baseVersion);
+          should.not.exist(body.currentVersion.branchBaseVersion);
```
ktuite, line 902: Shouldn't this have a branch base version? I think it is set in the test data xml...

matthew-white (approved):

lib/model/query/entities.js
ktuite: ah, i decided against it because i'd have to pass it through `processSubmissionEvent` AND `_processSubmissionEvent`

lib/task/process-backlog.js
```diff
@@ -0,0 +1,16 @@
...
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+//
+// This task deletes expired sessions from the table so it does not become
+// overladen and bogged down over time.
+
+const { task } = require('./task');
+const processBacklog = task.withContainer(({ Entities }) => Entities.processBacklog);
```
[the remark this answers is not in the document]
ktuite: there is now a transaction around the inner part of dequeuing something from the backlog and processing that submission. i'm not sure the whole block needs a transaction? if one failed for a major reason, we might want others to have worked. e.g, maybe the backlog processing task gets killed partway though, it's ok to have made partial progress.

PR #1178, opened 2024-08-26
test/purge: assert database state after purging

alxndrsn:

PR #1179, opened 2024-08-26
test/purge: assert database state after purging

alxndrsn:
Alternative approach to https://github.com/getodk/central-backend/pull/1178


alxndrsn:
After discussion with @ktuite, closing this in favour of #1178.

PR #1182, opened 2024-08-29
Allow property name from deleted form to be used in entity list

ktuite:
Closes https://github.com/getodk/central/issues/695

When an entity list property was created through a draft form, it inserted a row of (datasetId, name, publishedAt) with null `publishedAt`. The query to insert a new property via the API (which would also immediately publish that property) flagged any new property insertion as a a conflict. The query has been updated to act as an upsert that will publish an unpublished property instead of failing.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -179,11 +179,18 @@ createPublishedDataset.audit.withResult = true;
 const _createPublishedProperty = (property) => sql`
 INSERT INTO ds_properties ("name", "datasetId", "publishedAt")
 VALUES (${property.name}, ${property.datasetId}, clock_timestamp())
+ON CONFLICT ("name", "datasetId")
+DO UPDATE SET "publishedAt" = clock_timestamp()
+WHERE ds_properties."publishedAt" IS NULL
 RETURNING *`;
 
 // eslint-disable-next-line no-unused-vars
-const createPublishedProperty = (property, dataset) => ({ one }) =>
-  one(_createPublishedProperty(property));
+const createPublishedProperty = (property, dataset) => async ({ all }) => {
+  const result = await all(_createPublishedProperty(property));
+  if (result.length === 0)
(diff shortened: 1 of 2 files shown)


lib/model/query/datasets.js
```diff
@@ -179,11 +179,18 @@ createPublishedDataset.audit.withResult = true;
 const _createPublishedProperty = (property) => sql`
 INSERT INTO ds_properties ("name", "datasetId", "publishedAt")
 VALUES (${property.name}, ${property.datasetId}, clock_timestamp())
+ON CONFLICT ("name", "datasetId")
+DO UPDATE SET "publishedAt" = clock_timestamp()
+WHERE ds_properties."publishedAt" IS NULL
```
lognaturel, line 184: If two drafts would create the same property the property is still only in the database once, right?

lognaturel (approved):
One question to check my understanding but unless I got that wrong this looks ready to me!

lib/model/query/datasets.js
ktuite: Yes, the property is only listed once, but there's another table `ds_property_fields` that links those properties to the forms/form_defs/specific form fields that they came from (if they came from a form and not the API).

With two draft forms, it'll look like this in the database.
```
ds_properties [
  { id: 1, name: 'first_name', datasetId: 1, publishedAt: null },
  { id: 2, name: 'age', datasetId: 1, publishedAt: null }
]
ds_property_fields [
  { dsPropertyId: 1, formDefId: 3, path: '/name' },
  { dsPropertyId: 2, formDefId: 3, path: '/age' },
  { dsPropertyId: 1, formDefId: 4, path: '/name' },
  { dsPropertyId: 2, formDefId: 4, path: '/age' }
]
```

lognaturel: Great, makes sense. I was wondering whether there might be risk of the same property being published multiple times but it doesn’t seem so

PR #1184, opened 2024-09-04
Persist and serve user preferences

brontolosone:
This is a companion to getodk/central-frontend#1024, let's hold high level discussions in there for clarity (but simple on-point code comments etc can go in here)

Related: issue getodk/central#689
Related: PR getodk/central-frontend#1024

### DIFF
--- a/test/integration/api/user-preferences.js
+++ b/test/integration/api/user-preferences.js
@@ -56,27 +56,32 @@ describe('api: user-preferences', () => {
   it('PUT/DELETE preference storage operations', testService(async (service) => {
     const asAlice = await service.login('alice');
 
+    // Storage of simple property value
     await asAlice.put('/v1/user-preferences/site/someSimpleSitePref')
       .send({ propertyValue: true })
       .expect(200);
 
+    // Storage of property with complex value
     await asAlice.put('/v1/user-preferences/site/someComplexSitePref')
       .send({ propertyValue: { 1: 2, 3: 4 } })
       .expect(200);
 
-    // overwrite the previous
+    // Overwriting should work; later we'll check whether we retrieve this value
     await asAlice.put('/v1/user-preferences/site/someComplexSitePref')
       .send({ propertyValue: [1, 2, 3] })
       .expect(200);
 
+    // Store some pref to delete later on
     await asAlice.put('/v1/user-preferences/site/toBeDeletedPref')
       .send({ propertyValue: 'troep' })
       .expect(200);
 
+    // Store a *project* property (simple value)
     await asAlice.put('/v1/user-preferences/project/1/someSimpleProjectPref')
       .send({ propertyValue: false })
       .expect(200);
 
+    // Store a *project* property (complex value)
     await asAlice.put('/v1/user-preferences/project/1/someComplexProjectPref')
       .send({ propertyValue: [1, 2, 'many'] })
       .expect(200);
@@ -90,10 +95,12 @@ describe('api: user-preferences', () => {
         newProjectID = body.id;
       });
 
+    // we should be able to store a preference for the new project
     await asAlice.put(`/v1/user-preferences/project/${newProjectID}/prefForSomeOtherProject`)
       .send({ propertyValue: 9000 })
       .expect(200);
 
+    // store a project preference on this new project to be deleted later on
     await asAlice.put(`/v1/user-preferences/project/${newProjectID}/toBeDeletedPref`)
       .send({ propertyValue: 'troep' })
       .expect(200);
@@ -106,12 +113,14 @@ describe('api: user-preferences', () => {
         .expect(200)
     );
 
+    // expected properties of the new project
     const newProjectProps = {};
     newProjectProps[newProjectID] = {
       prefForSomeOtherProject: 9000,
       toBeDeletedPref: 'troep',
     };
 
+    // check whether the built-up state is sane (eg stores and overwrites applied)
     await asAlice.get('/v1/users/current')
       .set('X-Extended-Metadata', 'true')
       .expect(200)
@@ -134,17 +143,21 @@ describe('api: user-preferences', () => {
         });
       });
 
+    // check deletion of a site preference
     await asAlice.delete('/v1/user-preferences/site/toBeDeletedPref')
       .expect(200);
 
+    // check deletion of a project preference
     await asAlice.delete(`/v1/user-preferences/project/${newProjectID}/toBeDeletedPref`)
       .expect(200);
 
+    // check whether deleting something nonexistent results in an informative error response
     await asAlice.delete(`/v1/user-preferences/project/${newProjectID}/toBeDeletedPref`)
       .expect(404); // as we've just deleted it
 
     delete newProjectProps[newProjectID].toBeDeletedPref;
 
+    // check whether the built-up state is sane (deletions applied)
     await asAlice.get('/v1/users/current')
       .expect(200)
       .set('X-Extended-Metadata', 'true')


lognaturel:
I took the liberty of renaming the PR so it captures what it does, feel free to edit again!

lib/model/migrations/20240806-01-add-userpreferences.js
```diff
@@ -0,0 +1,21 @@
+// Copyright 2024 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.schema.createTable('userpreferences', (prefs) => {
```
ktuite, line 10: We tend to put underscores in table names and use camelCase for the columns.

lib/model/migrations/20240806-01-add-userpreferences.js
```diff
@@ -0,0 +1,21 @@
...
+  prefs.integer('userId').notNull();
```
ktuite, line 11: I can't explain it but I feel uncomfortable with this being called `userId` and then it being a FK to an `actorId`, except I guess it's the only ID on the users table, and these user preferences are only going to apply to the subset of Actors that are Users. Does it make me feel better to call it `actorId` and have it FK directly to the `actors` table? Not really. Would any other actors have user preferences? Uhh, I currently can't think far enough ahead to know what that means. 

Maybe the way it is is as `userId` is really the best way.

lib/model/migrations/20240806-01-add-userpreferences.js
```diff
@@ -0,0 +1,21 @@
...
+  prefs.jsonb('body');
```
ktuite, line 12: See top level review comments about possibly capturing more structure here.

lib/model/migrations/20240806-01-add-userpreferences.js
```diff
@@ -0,0 +1,21 @@
...
+
+  prefs.foreign('userId').references('users.actorId');
+  prefs.primary('userId');
+});
+
```
ktuite, line 17: We've had some long discussions about what these migrations should look like in the past year, e.g. using knex vs. writing raw sql. More of the recent migrations use raw sql because we have dreams of getting unstuck from Knex (an old db package we happen to be locked into for migrations) and "rewriting the knex migrations to not use knex" was a thought we entertained. I personally don't think that approach will pan out, so I don't really care if the migrations are like this or not. If you happen to have any brilliant ideas on how to get away from knex migrations (while still allowing people to smoothly upgrade their Central setups in the wild), plz share!

lib/resources/user-preferences.js
```diff
@@ -0,0 +1,38 @@
...
+
+const Problem = require('../util/problem');
+const { success } = require('../util/http');
+
+module.exports = (service, endpoint) => {
+
+  //////////////////////////////////////////////////////////////////////////////
+  // User preferences (UI settings)
+
+  service.get('/user-preferences/current', endpoint(({ UserPreferences }, { auth }) =>
```
ktuite, line 18: I've been thinking about this URL for a bit. It feels like a lot? But I know it parallels `/users/current` to get basic account information about the logged in user. Because of that, I'm not sure if `/user-preferences` without `current` feels right because it _should_ just return info about the current user and not lead you to believe you can look up someone else's user preferences... 

Being able to set user preferences on the current user also makes it different. 

One idea (besides leaving it as is) would to just remove `current` and go with `/user-preferences` and have it documented to say it always works with the authenticated user only.

Another idea could be to piggy back off `/users/current`.
In the [docs](https://docs.getodk.org/central-api-accounts-and-users/#getting-authenticated-user-details), you can set extended metadata to get the current user's verbs, which is something _mostly_ meant for frontend. Originally, I thought "user preferences could be kept out of that, we'll add to the API, not change an existing endpoint". But I'm not sure the harm in just returning adding user preferences to the extended metadata _here_ on this existing endpoint. (It would make some things about the frontend simpler, too.) 

The only question then is what url to set the user preferences. PATCH to `/users/current/preferences`?

lib/resources/user-preferences.js
```diff
@@ -0,0 +1,38 @@
...
+        .then(res => res.value?.body || {})
+    ).orElse(Problem.user.notFound())
+  ));
+
+  service.post('/user-preferences/current', endpoint(({ UserPreferences }, { body, auth }) =>
+    auth.actor.map((actor) =>
+      UserPreferences.setForUser(actor.id, body)
+        .then(success)
+    ).orElse(Problem.user.notFound())
+  ));
```
ktuite, line 30: Does frontend ever POST? Does it need to know or care if it is setting user preferences for the first time? Is it within the realm of HTTP best practices to just have one of these so frontend only has to think about one, but still have it act like a PATCH so as not to clobber other preferences?

lib/resources/user-preferences.js
```diff
@@ -0,0 +1,38 @@
...
+
```
ktuite, line 31: Should the URL and request for a project-specific preference have more to it, like accepting the project ID as input in the query params or request body or something, and validating that it is a real project that that user can have preferences about?

lib/model/query/user-preferences.js
```diff
@@ -0,0 +1,34 @@
...
+  maybeOne(sql`select body from userpreferences where "userId"=${userId}`);
+
+const setForUser = (userId, body) => ({ one }) => {
+  const json = body != null ? JSON.stringify(body) : null;
+  return one(sql`
+    insert into userpreferences ("userId", body) values (${userId}, ${json})
+    on conflict ("userId") do update set body=${json}
+    returning 1
+  `);
+};
```
ktuite, line 22: Related to some of my comments below about frontend probably not differentiating between creating a new set of preferences or updating an existing set, maybe we only need one type of method here that can also handle: 

- change existing preference
- add new preference to existing set
- add new preference to empty set

ktuite (commented):
In retrospect, we should have chatted more about backend on friday! But taking some time to think about it allowed all these thoughts to percolate. It's an _opinionated_ code base, and I wanted to unpack some of that.

### Coupling of Backend and Frontend

We talked about some of the reasons backend and frontend are split into separate repos and developed semi-independently some of the time(e.g. some API changes have no bearing on frontend or some frontend redesigns need no or very little backend changes). 

But we also talked about how they're tightly coupled, and how there is a lot baked into Backend that is specifically built for our Frontend. Some users seem to use Central without the frontend..... I was going to write that no one uses a custom frontend with backend, but that's not exactly true... some users do use their own custom frontends, but in that case, they mainly use the data API parts of backend, not the frontend API parts of backend. If someone else makes their own frontend, it'll have it's own separate backend/db/whatever in addition to talking to Central. (So many frontends/backends in this last paragraph 😅)

Basically... while the ODK team usually puts a lot of effort into making certain things within the ODK ecosystem generic and interchangeable (e.g. non-Central backends working with Collect, or just getting things to be compatible across combinations of Central/Collect/Android/web-based forms/versions of forms/etc.), the frontend-parts of backend don't really fall under that umbrella and are allowed to be super specific. 

We always release Central with Backend and Frontend in sync, and there's generally enough communication along the way to keep them in sync during development in between releases, too. If someone is working on something full stack that touches both parts, they are thoughtful about the timing of getting both separate parts merged in. A frequent pattern is either working on and merging the backend first, or working on both parts in parallel, getting something functional, refining the backend and merging it, then refining the frontend and merging it. If there's a change in backend that breaks something in frontend, we try to get the frontend fix done ahead of time or asap.

_What that means for this specific user preferences PR..._ is that it's okay and likely desirable for backend to have a deeper, structured understanding of the user preferences frontend needs to work with. See below! This is all up for discussion and pushback, we're just trying to have folks on the Central team in agreement about what's getting stored and how.


### Structure of User Preferences

In the frontend code, I'm seeing preferences with the keys `projectSortMode` and `formTrashCollapsed`. In the backend, could there be a more explicit link between a project and a project-specific preference? We have an Actees table that holds `acteeId`s for projects, forms, users, the site (`*`)

The [issue](https://github.com/getodk/central/issues/689) talks about site-wide preference and project-specific preferences, and in addition to imagining multiple preferences at each level, we could also possibly have form-specific or dataset-specific user preferences in the future. (I know only the 2 of project-sort-order and form-trash-collapsing are in scope for this PR.)

What if we had a `user_preferences` table with something like
* `actorId` (pointed to the actor record of a user)
* `acteeId` (pointed to a project actee ID or `*` for site-wide)
* `preferences` (jsonb)

and the API response returned the combination of multiple rows for a given user?

Or it could go further (taking inspiration from config) and store a `actorId, acteeId, key, value, setAt`? Splitting every preference for every user into its own row might be overkill, though, or maybe it's not much worse than already splitting things out by actee/project if there are only ever a small number of preferences per actee.

On the other hand, maybe a generic approach is simpler in the long run and quicker to implement. 

I think the desire to at least split it by actee (site and per project) is that through inspecting the database and code, we'd have a better idea of what the preferences were about. We don't really want our users to be able to store random stuff in their database through this endpoint. And we'd be able to tidy things up in the future if we needed to. E.g. preferences about a deleted project or form could also be deleted. Not that you can delete projects now (only archive) or have user preferences about forms... 


### Tests

I was reflecting on the config tests because they are the closest, but they are also not quite the right match. It's an acceptable intro to the kinds of tests we see for a lot of our API endpoints, though! 

https://github.com/getodk/central-backend/blob/master/test/integration/api/config.js

Tests in that file should inspire tests in a user-preferences file: 

- should reject if user can't do a thing
    * doesn't apply here because endpoint is about setting one's own preferences?
    * any user can set their own preferences? what if they are using the api and authenticated through an app user token or something, though? i can't remember how the auth is plumbed together in those cases right now.

- should reject for unknown config / user preferences?
    * maybe we should check that they're setting a known preference key on a known resource (site, project, etc.) so we have a better idea of the shape of what we're storing? 
    * possible new test: reject for unknown resource (e.g. project)?

- should set the config / should patch the config
	* frontend is only going to send one kind of request? it shouldn't care if its setting new configs or not
	* test cases within this scenario (adding totally new user preferences, adding to existing prefernces, updating existing preferennce)

- should there be a way to delete a preference? (unsetting site-wide configs was something we had to do, but I don't know if it's necessary here for user preferences. could add it later if it becomes a thing.)

lib/model/query/user-preferences.js
ktuite: Oh, so you're thinking this part of user prefs would look like `formTrashCollapse: [projId1, projId7, projId9]`? That wasn't obvious to me at first.

I feel like we should pick a structure that will work for future project-specific settings like the ones in the issue about form sort order and booleans about showing legacy QR codes. Something like `projId: {collapseTrash: true, formSort: 'latest', showLegacyQRcodes: true}`

lib/resources/user-preferences.js
ktuite: I was thinking of `acteeId` being used in the backend only; it doesn't get exposed through the API but it could be useful generic database reference. The way I'm imagining it, you'd go from resource ID to actee ID to save the data and then actee ID back to resource ID to build the response data. But maybe that's an added layer of abstraction that doesn't need to be there and you could just be storing numeric project IDs like you're doing already.

From the API side of things, projects are referenced by their numeric ID and forms are usually referenced by projectId + xmlFormId. (We have numeric IDs for the forms, too, but we rarely expose those IDs.)

Can you think of a good URL pattern that would work for setting both site-wide and project preferences? 

`/user-preferences/key` and `/user-preferences/project/34/key`? 
`/user-preferences` and `/user-preferences?projectId=3` with `{key: value}` body?

ktuite: I'm in support of returning preferences in `/users/current` as part of the extended metadata!

ktuite: Yeah... good point! Let me convince myself this actee translation step is overly complex for what we're trying to accomplish:

We have a generic actee -> specific kind of object lookup happening in the audits query [here](https://github.com/getodk/central-backend/blob/master/lib/model/query/audits.js#L76-L100) so that audit events can describe specific kinds of projects/forms that they are attached to, but it's a complicated bit of code. We don't need something at that level of complexity for these user preferences for all the reasons you said - backend doesn't really care and it'll be more work to maintain. 

Is there something in between tracking `acteeIds` and the original generic approach?

`user_preferences` with columns `userId`, `projectId` (can be null), `propertyName`, `propertyValue`? 

Or do you want to go back to the original `userId` `body` approach? 

If you wanted to go back the original approach, could you add some tests showing setting/updating some preferences?

And if you want to include the user preferences in `/users/current` extended response, around [here](https://github.com/getodk/central-backend/blob/master/lib/resources/users.js#L111-L118) would be where to change the code.

Sorry for going back and forth on all these ideas, it's good we're discussing it!

ktuite:
Plan
1. return existing user preferences with `/users/current` extended metadata

```
{ 
    projectSort: 'alphabetical',
    projects: { 1: {formTrashCollapse: true} } 
}
```

2. PATCH existing preferences (even empty preferences) with specific preference (property name, property value)


PUT `/user-preferences/projectSort` body: {propertyValue: 'alphabetical'}

PUT `/user-preferences/projects/4/sortOrder` {value: 'alphabetical'}

db tables: `user_project_preferences`, `user_site_preferences`
with cascading deletes! and relational integrity!
unique on id and property name


3. DELETE a specific key within user preferences?

DELETE /user-preferences/project/3/formTrashCollapse

test/integration/api/user-preferences.js
```diff
@@ -0,0 +1,119 @@
...
+    await asAlice.put('/v1/user-preferences/project/1/someComplexProjectPref')
+      .send({ propertyValue: [1, 2, 'many'] })
+      .expect(201);
+
+    // make a new project (which should get ID 2 with current fixtures)
+    await asAlice.post('/v1/projects')
+      .send({ name: 'preferences test project' })
+      .expect(200)
+      .then(({ body }) => {
+        body.id.should.eql(2);
```
ktuite, line 46: when you make a new project in a test, because it may not be id 2. 

the test suite runs tests in 2 modes:
1. starts a transaction, runs the test, and then rolls back the transaction (to reset the database to the initial state)
2. starts the db from scratch, runs all the migrations, applies the fixtures, then carries on with the test (the migrations are slow, which is why this way isn't used for all the tests)

these modes, or maybe the interaction between the two of them, is how the new IDs for certain resources aren't always what you expect.

lib/http/endpoint.js
```diff
@@ -238,6 +238,40 @@ const defaultEndpoint = endpointBase({
...
+    // No body.
+    response.send();
+  }
+};
+
+const simpleEndpoint = endpointBase({
+  resultWriter: simpleResultWriter,
+  errorWriter: defaultErrorWriter,
+});
+
```
ktuite, line 274: I'd rather not have a new endpoint utility just to add 201/204 response codes and empty body (unless maybe it was well-tested and documented?). 

I think it would be simpler and more consistent with the rest of the code to settle for this basic "success" response that just returns `200 { success: true }`. Here's an example being used for deleting the other type of config we have: https://github.com/getodk/central-backend/blob/master/lib/resources/config.js#L52 For setting that other config, it looks like it returns what was sent to it, but in this case, i think it would be fine to return `200 { success: true }` for setting a preference, too.

lib/resources/users.js
```diff
@@ -113,7 +113,8 @@ module.exports = (service, endpoint) => {
     auth.actor.map((actor) =>
       ((queryOptions.extended === true)
         ? Promise.all([ Users.getByActorId(actor.id).then(getOrNotFound), Auth.verbsOn(actor.id, '*') ])
-          .then(([ user, verbs ]) => Object.assign({ verbs }, user.forApi()))
+          .then(([ user, verbs ]) => Users.getPreferences(user.actorId)
+            .then((preferences) => Object.assign(Object.assign({ verbs }, user.forApi()), preferences)))
```
ktuite, line 117: Can you do `Object.assign({ verbs, preferences }, user.forApi())))` instead instead of the 2 nested object assigns?

lib/resources/user-preferences.js
```diff
@@ -0,0 +1,63 @@
...
+  //////////////////////////////////////////////////////////////////////////////
+  // Endpoint to get all of a user's preferences.
+  // For completeness and ease of debugging; as these preferences are normally
+  // pulled in by the frontend through the extended version of /users/current.
+  service.get('/user-preferences', endpoint.simple(({ UserPreferences }, { auth }) => {
+    if (auth.actor.value === undefined) return Problem.user.insufficientRights();
+    return UserPreferences.getForUser(auth.actor.value.id)
+      .then(res => ({ headers: { 'Content-Type': 'application/json; charset=utf-8' }, body: res.preferences }));
+  }));
```
ktuite, line 27: Can we not add this and just use the `/v1/users/current` endpoint to test? If we dont add it, we don't have to document it 😅

test/integration/api/user-preferences.js
```diff
@@ -0,0 +1,127 @@
...
+
+  it('should not allow storing preferences for nonexistent projects', testService(async (service) => {
+    const asAlice = await service.login('alice');
+
+    await asAlice.put('/v1/user-preferences/project/123/someProperty')
+      .send({ propertyValue: 'someValue' })
+      .expect(400); // project 123 doesn't exist
+  }));
+
+  it('PUT/DELETE preference storage operations', testService(async (service) => {
```
ktuite, line 13: I'd like to see more/smaller tests! 

In this big test, the main themes i'm getting out of it are:
- "it should store site-wide and project-specific preferences of varying complexity"
- "it should not return preferences that do not belong to the user"
- "it should group project preferences by project"

I think you should keep that test, or maybe split it up into 2 or 3 tests, but maybe there could also be stupidly basic tests like:
- "it allows a site-wide preference to be set"
- "it allows a project-specific preference to be set"
- "it allows a preference to be deleted"
- "it returns the (non blank) preferences in the extended metadata for the current user"

And one of our original concerns that the design of the api alleviates: 
- "it allows a preference to be modified without overwriting another preference""

I'd also be interested in tests of the failure cases
- no propertyValue supplied
- empty string value supplied
- (anything else that tests the database constraints or other error handling in the resources)

ktuite (commented):
The core of this PR is looking really good. I like where the database organization of the user preferences ended up. I hope its working well for frontend! It seems like it would be easy to extend to include form-specific preferences if we ever needed those. 

My main comments are about tests and seeing if there are things we can hold off on including in this PR (unless you think they're really important).

lib/resources/users.js
matthew-white: I agree that it'd be nice to avoid the nested `Object.assign()`. `Object.assign()` can take more than two arguments, so you could also do:

```js
Object.assign({ verbs }, user.forApi(), preferences)
```

lib/model/migrations/20240910-01-add-user_preferences.js
```diff
@@ -0,0 +1,35 @@
...
+    CONSTRAINT "user_project_preferences_primary_key" PRIMARY KEY ("userId", "projectId", "propertyName")
+  );
+
+  -- Primary key indices are used for PUTing/DELETE-ing individual rows — but the below indices are
+  -- used when aggregating all of a user's preferences.
+  CREATE INDEX ON "user_site_preferences" ("userId");
+  CREATE INDEX ON "user_project_preferences" ("userId");
+`);
+
+const down = (db) => Promise.all([db.schema.dropTable('user_site_preferences'), db.schema.dropTable('user_project_preferences')]);
```
ktuite, line 32: This could use raw sql here instead

lib/resources/users.js
```diff
@@ -113,7 +113,8 @@ module.exports = (service, endpoint) => {
     auth.actor.map((actor) =>
       ((queryOptions.extended === true)
         ? Promise.all([ Users.getByActorId(actor.id).then(getOrNotFound), Auth.verbsOn(actor.id, '*') ])
-          .then(([ user, verbs ]) => Object.assign({ verbs }, user.forApi()))
+          .then(([ user, verbs ]) => Users.getPreferences(user.actorId)
```
ktuite, line 116: I noticed `getPreferences` was added to the Users query module, copied over from the UserPreferences query module. Why not just change the endpoint to `... endpoint(({ Auth, Users, UserPreferences }...` and use `UserPreferences.getForUser(user.actorId)` directly?

lib/model/query/users.js
```diff
@@ -86,10 +87,13 @@ const emailEverExisted = (email) => ({ maybeOne }) =>
   maybeOne(sql`select true from users where email=${email} limit 1`)
     .then((user) => user.isDefined());
 
+const getPreferences = (userId) => ({ one }) => getForUser(userId)({ one });
```
ktuite, line 90: This line is doing weird things with containers... (the `{ one }` argument is destructuring the container)

See my other comment about just using `getForUser` from the UserPreferences query module directly.

lib/model/query/user-preferences.js
```diff
@@ -0,0 +1,123 @@
...
+    .concat((subject === null) ? [] : ['projectId'])
+    .map(el => sql.identifier([el]));
+
+  // Work around null confusion (potential Slonik bug?).
+  // sql.json(null) doesn't produce what we need, it results in an exception
+  // "Error: Required parameter propertyValue missing."
+  // Yet the string 'null' (as distinct from the *jsonb* string '"null"' one would get with sql.json('null') !)
+  // gets properly casted by PostgreSQL to a jsonb null (as distinct from an SQL NULL), so we use that in this case.
+  const preparedPropertyValue = (propertyValue === null) ? 'null': sql.json(propertyValue);
+  const values = [userId, propertyName, preparedPropertyValue]
```
ktuite, line 74: I was wondering if you wanted to trim the `propertyName`. I noticed i'm able to make properties named ' ' and '  '.

lib/util/problem.js
```diff
@@ -124,6 +124,9 @@ const problems = {
 
     noConflictEntity: problem(400.32, () => 'The Entity doesn\'t have any conflict'),
 
+    // { expected: "list of expected properties", actual: "list of provided properties" }
+    unexpectedProperties: problem(400.33, ({ expected, actual }) => `Expected properties: (${expected.join(', ')}). Got (${actual.join(', ')}).`),
+
```
ktuite, line 129: Is there another existing Problem we could use instead? I see this only comes up if you pass `propertyValue` AND extra stuff like `{'propertyValue': 'meow', 'cats': '   '}`. Maybe we just accept `propertyValue` and ignore the rest.

I'm also ok leaving it as is with this new Problem because there is a test for it :)

test/integration/api/user-preferences.js
ktuite: Yeah this is pretty much the situation for all Central tests. 

> ...just leads down the path of incrementally building up a whole bunch of DB state in one transaction. Unless I'm missing something that's just the nature of this thing. If I want to test the aggregation I need to do a whole bunch of PUTs....

Send a bunch of requests about forms/submissions to build up the state. If it gets super repetitive, we make special setup functions for specific batches of tests, but it's still making the requests for every test.

ktuite: Thanks for adding some of the smaller tests! e.g.  
- validates the request body stringently 
- can store a JS null propertyValue
- should not allow storing preferences for nonexistent projects

ktuite: I can still see an advantage of smaller tests so i can just look at an individual test and grok the 1 thing its trying to do instead of having to follow the whole journey, but eh, it's fine for now!

test/integration/api/user-preferences.js
```diff
@@ -0,0 +1,172 @@
+const { testService } = require('../setup');
+
+describe('api: user-preferences', () => {
+
```
ktuite, line 4: I was playing around with this new endpoint (with a notebook and pyodk) and saw that i could make properties named `  ` and ` whitespace ` and so on. I think the property values being pretty unrestricted is fine but having `prop` be different from ` prop   ` seems potentially problematic. 

Would love to see a test showing propertyNames get trimmed.

ktuite: If we were talking XForm spec (Forms, Submissions, Datasets, Entities), we'd want to (and do) get very clear about what counts as the same or different values, because we want to be as clear and as _nice_ to form designers as possible and not have something not work as expected because the casing or whitespace was subtly different.

But you're right that in this context, we've talked about how this user preferences endpoint is mainly for Frontend to use, so it makes sense to keep it simple and use the property name as is.

ktuite:
> The one thing I could do to skip building up state this way is to mass-create state via the DB, but then the test will fail if the DB layout would change, and these tests are supposed to test the API and not the DB specifics, as that is an implementation detail.

We definitely want to test with the API and not build up database state directly because of the DB changing. And some of the core logic is so complicated in the database it would be / is not fun to manually build it up. 

I have done a bit of that for trying to test certain database migrations, which is tricky because you can't actually use the API if the DB schema is changing mid-test. This [file](https://github.com/getodk/central-backend/blob/master/test/integration/other/migrations.js) is one of my pain points; I'm glad to have been able to test some of these migrations but the whole situation feels a little off.

ktuite (approved):
Woohoo, let's merge it!

PR #1186, opened 2024-09-13
tests: use unambiguous timestamps, fixes #1185

### LINKED ISSUE #1185: Timezone-sensitive test failures
```sh
BCRYPT=insecure npx mocha --recursive --exit --bail --grep 'should not mince( the)? object properties' 
```

fails (3 tests, 3 failures) if you have a timezone offset positive to UTC in your database connection (as many in the Eastern hemisphere are wont to have).

Eg
```
  1) Entities API
       GET /datasets/:name/entities
         should not mince the object properties:
     AssertionError: expected '2021-12-31T23:00:00.000Z' to match /2022/
```

It's an hour off, which around Newyear can make me a year off.
And I'm an hour off from UTC.

What's happening is with statements such as these:

```sql
UPDATE actors SET "createdAt" = '2020-01-01'
```
(where the `createdAt` column is of type `timestamptz`)
1. the string `'2020-01-01'` will suffer an attempted conversion to the column type `timestamptz`
2. for that, `'2020-01-01'` is interpreted as a date
3. but the column needs a timestamptz, and we didn't supply a time, so `00:00:00` is made up for the upcast
4. but we also didn't supply any timezone indication, so another thing needs to be made up for the upcast; PostgreSQL uses the connection's timezone setting (`SHOW timezone;`) for this. Mine's UTC+1 at the moment.
5. So we have `2020-01-01 00:00:00 UTC+1` but a `timestamptz` column stores UTC (yes! `timestamptz` does not store timezone information!) so that's converted to `2020-01-01 23:00:00 UTC`. That's what's stored (minus the timezone; it's discarded, as internally PostgreSQL knows that timestamptz columns store time in UTC).
6. Upon retrieval, that timestamptz value is converted to the timezone setting of the connection in which the SELECT runs (which may or may not be the same as the timezone setting from when it was stored - that's the whole point of timestamptz; to save you from having to do conversions yourself).
7. What happens further on once the value's read could be anything, it depends on the framework and the libraries and what people think is "right". So what made this '2021-12-31T23:00:00.000Z' show up in this form (UTC), I don't know :man_shrugging:, but regardless the presentation, what's *stored* is not what it should be.

Two possible solutions:
1. Don't be ambiguous, leave nothing to guess, and specify timestamptz's fully qualified.
2. Let PostgreSQL make up the missing information, but set the connection timezone to UTC.

I think 1. is the way to go, so that's what's done in PR https://github.com/getodk/central-backend/pull/1186.

brontolosone:
Closes #1185

Detailed explanation in #1185.

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -677,7 +677,7 @@ describe('datasets and entities', () => {
 
         await exhaust(container);
 
-        await container.run(sql`UPDATE entities SET "createdAt" = '1999-1-1' WHERE TRUE`);
+        await container.run(sql`UPDATE entities SET "createdAt" = '1999-1-1T00:00:00Z' WHERE TRUE`);
 
         await asAlice.post('/v1/projects/1/forms/simpleEntity/submissions')
           .send(testData.instances.simpleEntity.two)
--- a/test/integration/other/analytics-queries.js
+++ b/test/integration/other/analytics-queries.js
@@ -114,7 +114,7 @@ describe('analytics task queries', function () {
       await container.Audits.log(null, 'dummy.action', null, 'test audit details');
       // old audit
       await container.run(sql`insert into audits ("actorId", action, "acteeId", details, "loggedAt")
-        values (null, 'dummy.action', null, null, '1999-1-1')`);
+        values (null, 'dummy.action', null, null, '1999-1-1T00:00:00Z')`);
       const res = await container.Analytics.auditLogs();
       res.recent.should.equal(2);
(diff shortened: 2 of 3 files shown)


matthew-white:
I see a number of cases of `1999-1-1` in tests that we'd probably want to change if we wanted to give everything a timestamp and time zone.

matthew-white:
Interesting, maybe they're fine then.

matthew-white:
Ah good call, probably a good idea to change that one then at least.

matthew-white (approved):

PR #1187, opened 2024-09-16
Mark soft conflict if not contiguous with trunk

ktuite:
Closes https://github.com/getodk/central/issues/698

This PR makes two changes:
**1: Update per-version conflict state**

The `<entity uuid>/versions` endpoint returns a `conflict` state for each version of the entity, which is computed from other information about the entity versions. These values should stay the same even when the top level `conflict` flag on the entity changes (or is manually resolved). These version-specific conflict states used to be detected just from `baseVersion+1 != version`. 

In this PR, we've copied branch-tracking logic from Frontend to check whether each version if it is contiguous with its trunk version, and return a `soft` conflict state if it is non-contiguous but otherwise not a hard conflict.

We could also consider having backend return more of the details that frontend is currently computing about branch continuity. 

**2. Update top-level conflict state**

The above work to make the per-version conflict state more descriptive wasn't enough. Consider a situation where an offline branch has been interrupted by another branch or online update, a conflict has been resolved, but then another update from that interrupted branch comes in. We want every update that is not contiguous with its trunk version to be a conflict, and that means updating the conflict state on the entity itself, too. 

Before, we only noted a conflict when the base version wasn't what was expected. An offline update would not be counted as conflict as long as the previous action in the branch was present.

This PR adds a check that the new entity is not only contiguous with its base version, but is also contiguous with its specified trunk version. (This is done by looking for a version of a different branch that _interrupts_ the given branch, rather than the other code that tracks the last contiguous version per branch.) If it is not contiguous, i.e. if there is an interruption, the top level `conflict` flag on the entity itself is set to `soft` instead of `null`. 

Tests.

See explanation above.

Highlights conflicts more.

No.

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -397,6 +397,64 @@ const diffEntityData = (defData) => {
   return diffs;
 };
 
+// Copied from frontend though it may not all be necessary
+// Offline branch
+class Branch {
+  // firstUpdate is the first offline update (not create) to be processed from
+  // the branch. entityRoot is the first version of the entity.
+  constructor(firstUpdate, entityRoot) {
+    if (firstUpdate.trunkVersion != null) {
+      // The first version from the branch to be processed (not necessarily the
+      // first in the original branch order)
+      this.first = firstUpdate;
+
+      // How many versions that have been processed are from the branch?
+      this.length = 1;
+
+      // Was this.first processed in branch order, or was it processed before an
+      // earlier change in the branch?
+      const { trunkVersion } = firstUpdate;
+      this.firstInOrder = firstUpdate.branchBaseVersion === trunkVersion;
+
+      /* this.lastContiguousWithTrunk is the version number of the last version
+      from the branch that is contiguous with the trunk version. In other words,
+      it is the version number of the last version where there has been no
+      update from outside the branch between the version and the trunk version.
+      this.lastContiguousWithTrunk is not related to branch order: as long as
+      there hasn't been an update from outside the branch, the branch is
+      contiguous, regardless of the order of the updates within it. */
+      this.lastContiguousWithTrunk = firstUpdate.version === trunkVersion + 1
+        ? firstUpdate.version
+        : 0;
+    } else {
+      // If the entity was both created and updated offline before being sent to
+      // the server, then we treat the creation as part of the same branch as
+      // the update(s). The creation doesn't have a branch ID, but we treat it
+      // as part of the branch anyway.
+      this.first = entityRoot;
+      // If the submission for the entity creation was received late and
+      // processed out of order, then firstUpdate.version === 1. In that case,
+      // we can't reliably determine which entity version corresponds to the
+      // entity creation, so we don't treat the creation as part of the branch.
+      this.length = firstUpdate.version === 1 ? 1 : 2;
+      this.firstInOrder = this.length === 2;
+      this.lastContiguousWithTrunk = firstUpdate.version === 2 ? 2 : 1;
+    }
+
+    this.id = firstUpdate.branchId;
+    // The last version from the branch to be processed
+    this.last = firstUpdate;
+  }
+
+  add(version) {
+    this.length += 1;
+    this.last = version;
+    if (version.baseVersion === this.lastContiguousWithTrunk &&
+      version.version === version.baseVersion + 1)
+      this.lastContiguousWithTrunk = version.version;
+  }
+}
+
 // Returns an array of properties which are different between
 // `dataReceived` and `otherVersionData`
 const getDiffProp = (dataReceived, otherVersionData) =>
@@ -419,6 +477,26 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
 
   const relevantBaseVersions = new Set();
 
+  // build up branches
+  const branches = new Map();
+  for (const version of defs) {
+    const { branchId } = version;
+    if (branchId != null && version.branchBaseVersion != null) {
+      const existingBranch = branches.get(branchId);
+      if (existingBranch == null) {
+        const newBranch = new Branch(version, defs[0]);
+        branches.set(branchId, newBranch);
+        version.branch = newBranch;
+        // If the entity was created offline, then add the branch to the
+        // entity creation.
+        newBranch.first.branch = newBranch;
+      } else {
+        existingBranch.add(version);
+        version.branch = existingBranch;
+      }
+    }
+  }
+
   for (const def of defs) {
 
     const v = mergeLeft(def.forApi(),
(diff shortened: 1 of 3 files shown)


matthew-white:
> We could also consider having backend return more of the details that frontend is currently computing about branch continuity.

I think this is a pretty interesting idea. I don't think it's necessary for this particular issue, but I think it's something we should keep in mind for the future.

lib/data/entity.js
```diff
@@ -417,6 +475,26 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
 
   const relevantBaseVersions = new Set();
 
+  // build up branches
+  const branches = new Map();
+  for (const version of defs) {
```
matthew-white, line 480: This ends up looping over `defs` twice, which I don't think we need to do. I think we can build up `result` and `branches` in the same loop, as no version/def needs to look ahead to a later version, including for branch information.

lib/data/entity.js
```diff
@@ -417,6 +475,26 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
...
+    const { branchId } = version;
+    if (branchId != null && version.branchBaseVersion != null) {
+      const existingBranch = branches.get(branchId);
+      if (existingBranch == null) {
+        const newBranch = new Branch(version, defs[0]);
+        branches.set(branchId, newBranch);
+        version.branch = newBranch;
```
matthew-white, line 487: I know we do it in Frontend, but I think we should avoid mutating `version` here unless it's necessary.

lib/data/entity.js
```diff
@@ -417,6 +475,26 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
 
   const relevantBaseVersions = new Set();
 
+  // build up branches
+  const branches = new Map();
+  for (const version of defs) {
+    const { branchId } = version;
+    if (branchId != null && version.branchBaseVersion != null) {
```
matthew-white, line 482: I added `version.branchBaseVersion != null` to Frontend mostly as a temporary measure, since at the time, there were submissions where the entity creation got a branch ID. I think we can remove the condition here.

matthew-white (commented):
> I copied this branch tracking logic directly from Frontend, and there's probably more there than necessary for backend, but I just wanted to see it in action.

I'm going to pare back aspects of the `Branch` class and related logic to only what Backend needs, which I think will make the Backend code clearer. However, even pared back, I like the `Branch` class here. It moves the branch logic into a separate part of the code from the rest of the `getWithConflictDetails()` function, which is already fairly complex. I also think it's very possible that we'll want to grow the `Branch` class in the future.

Before I push my commit, I thought I'd leave a few comments about specific things I'm changing.

matthew-white:
This PR makes changes to the `conflict` property of entity versions, but in some cases, I think we also need to make changes to the `conflict` property of the entity itself. For example, say an entity had a conflict, but was then resolved. A new offline update comes in that applies to the current version. That all seems good, and currently, the `conflict` property of the entity will be `null`. However, with the change to the conflict logic, the new version may actually be a conflict: the `conflict` property of the new version will be `'soft'` if its base version is not contiguous with the trunk version. In that case, the `conflict` property of the entity itself should also be `'soft'`. However, right now I'm pretty sure it's `null`. I modified the first test to try to create an example:

```js
it('should mark an update that is not contiguous with its trunk version as a soft conflict', testOfflineEntities(async (service, container) => {
  const branchId = uuid();

  // Update existing entity on server (change age from 22 to 24)
  await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?baseVersion=1')
    .send({ data: { age: '24' } })
    .expect(200);

  // Send update (change status from null to arrived)
  await asAlice.post('/v1/projects/1/forms/offlineEntity/submissions')
    .send(testData.instances.offlineEntity.one
      .replace('branchId=""', `branchId="${branchId}"`)
    )
    .expect(200);
  await exhaust(container);

  await asAlice.patch('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc?resolve=true&baseVersion=3')
    .expect(200);

  // Send second update (change age from 22 to 26)
  await asAlice.post('/v1/projects/1/forms/offlineEntity/submissions')
    .send(testData.instances.offlineEntity.one
      .replace('branchId=""', `branchId="${branchId}"`)
      .replace('one', 'one-update2')
      .replace('baseVersion="1"', 'baseVersion="2"')
      .replace('<status>arrived</status>', '<age>26</age>')
    )
    .expect(200);
  await exhaust(container);

  await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc/versions')
    .then(({ body: versions }) => {
      versions.map(v => v.conflict).should.eql([null, null, 'soft', 'soft']);
    });

  await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789abc')
    .then(({ body: entity }) => {
      // This assertion currently fails.
      should(entity.conflict).equal('soft');
    });
}));
```

ktuite:
@matthew-white I added some code in the `_updateEntity` section and added in your test scenario. I also updated the description of this PR.

test/integration/api/offline-entities.js
```diff
@@ -1343,6 +1343,236 @@ describe('Offline Entities', () => {
...
+      await asAlice.post('/v1/projects/1/forms/offlineEntity/submissions')
+        .send(testData.instances.offlineEntity.one
+          .replace('branchId=""', `branchId="${branchId}"`)
+        )
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+
+      await exhaust(container);
+
+      // Send fourth update
```
ktuite, line 1411: fourth update from the branch (3rd update is missing). this will get held in the backlog and when it gets force-processed, it will take the latest version as the base version and therefore not have a conflict due to version mismatch...

lib/data/entity.js
```diff
@@ -397,6 +397,55 @@ const diffEntityData = (defData) => {
...
+    }
+
+    this.id = firstUpdate.branchId;
+  }
+
+  add(version) {
+    if (version.baseVersion === this.lastContiguousWithTrunk &&
+      version.version === version.baseVersion + 1)
+      this.lastContiguousWithTrunk = version.version;
+  }
```
ktuite, line 446: We saw a scenario in a test where this logic didn't quite work out
```
conflict false false
branches.get(v.branchId).lastContiguousWithTrunk 2
v baseversion 1
add: version, base version and last contiguous 3 1 2
conflict true false
branches.get(v.branchId).lastContiguousWithTrunk 2
v baseversion 1
add: version, base version and last contiguous 4 3 2
conflict true true
branches.get(v.branchId).lastContiguousWithTrunk 2
v baseversion 3
```

test/integration/api/offline-entities.js
ktuite: link to issue here and mention the scenario is mentioned there

matthew-white (approved):

PR #1190, opened 2024-09-19
s3: remove PGROWLOCKS extension from db-migrations

alxndrsn:
Only require pgrowlocks extension in testing, because installing/creating the extension can require more privileges than other migrations.


ktuite (approved):

mlazowik:
Looks like  PGROWLOCKS is needed to run `lib/bin/s3.js upload-pending`, which is mentioned in S3 docs: https://docs.getodk.org/central-install-digital-ocean/#using-s3-compatible-storage

PR #1191, opened 2024-09-19
bin/repl: run NodeJS REPL with initialised container

alxndrsn:
Example use:

    ./lib/bin/repl
    Available vars: Actees, Actors, Analytics, Assignments, Audits, Auth, Blobs, ClientAudits, Comments, Configs, Datasets, Entities, FieldKeys, FormAttachments, Forms, Keys, Projects, PublicLinks, Roles, Sentry, Sessions, SubmissionAttachments, Submissions, Users, all, db, enketo, env, mail, maybeOne, maybeOneFirst, one, oneFirst, run, s3, sql, stream, with, xlsform
    > await db.any(sql`SELECT * FROM blobs`)
    []

### DIFF
--- a/lib/bin/repl.js
+++ b/lib/bin/repl.js
@@ -1,3 +1,12 @@
+// Copyright 2024 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
 #!/usr/bin/env node
 const should = require('should'); // eslint-disable-line import/no-extraneous-dependencies
 require('../../test/assertions');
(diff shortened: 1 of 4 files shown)


matthew-white:
Ooh I like this idea a lot! It'd help with troubleshooting on servers.

matthew-white:
I've got my hands full with regression testing and such, but I'll take a look at this once the release is out!

matthew-white (approved):

lib/bin/repl.js
```diff
@@ -0,0 +1,40 @@
...
+
+const should = require('should'); // eslint-disable-line import/no-extraneous-dependencies
+require('../../test/assertions');
+
+const { sql } = require('slonik');
+
+const container = require('../util/default-container');
+
+(async () => {
+  const context = { ...container, should, sql };
```
matthew-white, line 18: Could we also add `container` itself so that it's possible to call `container.transacting()`?

```suggestion
  const context = { ...container, container, should, sql };
```

matthew-white, line 18: Do you think we could remove `with` from `context`? It's somehow surprising to see it in the list, and I doubt we'd want to call it directly (i.e., running `with()`), since it's usually run as `container.with()`. As long as we add `container` to `context`, then it'll be possible to call `container.with()`. I'm not sure it'd even be possible to call `with()` directly — I don't think `with` is a valid identifier.

matthew-white (commented):
Sorry, I had a couple of ideas after approving!

PR #1192, opened 2024-09-19
s3: add audit logging

alxndrsn:
Closes https://github.com/getodk/central/issues/704


ktuite:
I changed the logging to
- not include a start event (though i could totally see a use for such an event)
- capture the duration in milliseconds (to make up for not having a start event, kind of??)
- only log an event if the upload had any successes or failures

These changes were made with the thought in mind that this will be a nightly job, so it doesn't need to log anything if its not doing anything.

ktuite (approved):

PR #1195, opened 2024-09-20
Prevent concurrent changes to same entity from different submissions

matthew-white:


ktuite (approved):

matthew-white:
Locking the entity caused a few existing tests to fail, specifically because the UUID validation was moved to earlier in the entity processing. That's because the UUID is needed to lock the entity (see #1200 for details). I temporarily skipped the failing tests before merging the PR, but I unskipped them in #1196.

PR #1196, opened 2024-09-23
Unskip tests by preventing additional entity.error

matthew-white:

### DIFF
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -5545,7 +5545,7 @@ describe('datasets and entities', () => {
     }));
 
     describe('central issue #547, reprocessing submissions that had previous entity errors', () => {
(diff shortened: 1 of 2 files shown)


test/integration/api/datasets.js
matthew-white, line 1: Add tests that demonstrates the behavior I'm seeing:

- The same submission can generate two `entity.error`s if submission approval is required and the submission is approved twice (even without an edit). It won't generate a second `entity.error` from toggling the `approvalRequired` flag.
- Certain errors that are checked early can be thrown twice. For example, there's no `dataset` attribute. Creating, then approving the submission will result in two `entity.error`s — even if submission approval isn't required.

ktuite (approved):

test/integration/api/datasets.js
matthew-white: I'm going to go ahead and merge this PR, but I'll file an issue about these tests.

PR #1204, opened 2024-10-01
add index on sessions(token); fixes #1201

### LINKED ISSUE #1201: missing index on table `sessions`, column `token`
There's no index on the `sessions` table, column `token`.
Yet many (most) requests will result in a lookup by that token. I think we'll want an index there?

Moreover, central-backend crashes when encountering duplicate tokens (not saying that it should handle them, or do anything other than crash — duplicate tokens are nonsensical), so we should probably make it a unique index in order to codify those expectations (and to inform the query planner).

brontolosone:
Closes #1201

Index good.

### DIFF
--- a/lib/model/migrations/20241001-01-index-on-session-table.js
+++ b/lib/model/migrations/20241001-01-index-on-session-table.js
@@ -1,0 +1,18 @@
+// Copyright 2024 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Looks good, thanks @brontolosone!

After a PR is approved, usually the author of the PR merges it. For us, the default is usually squash-merging so that there are fewer commits on the `master` branch. That said, if a PR isn't too big and makes multiple commits that are only loosely related, then we sometimes do merge with a merge commit. In general, you're welcome to choose between the two merge strategies based on what feels right for your PR.

alxndrsn:
> That said, if a PR isn't too big and makes multiple commits that are only loosely related, then we sometimes do merge with a merge commit.

There's a third option - rebase & fast-forward IIRC.  It means you can preserve individual commits, while still having a linear history.

That said, commits that are only loosely related maybe shouldn't be in the same PR in the first place...

PR #1210, opened 2024-10-03
Alter entity spec version and add branchId, trunkVersion in form schema

ktuite:
Closes https://github.com/getodk/central/issues/692

### DIFF
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -582,12 +582,71 @@ const _versionSplicer = (replace) => (xml, insert) => new Promise((pass, fail) =
 const addVersionSuffix = _versionSplicer(false);
 const setVersion = _versionSplicer(true);
 
+// Forms with entity spec version 2023.1.0 that support entity updates need to
+// be updated to spec version 2024.1.0 and have `branchId` and `trunkVersion`
+// included alongside the existing `baseVersion`.
+const _addBranchIdAndTrunkVersion = (xml) => new Promise((pass, fail) => {
+  const stack = [];
+  const parser = new hparser.Parser({
+    onopentag: (fullname) => {
+      stack.push(stripNamespacesFromPath(fullname));
+      if ((stack.length) === 7 &&
+      (stack[0] === 'html') && (stack[1] === 'head') && (stack[2] === 'model') &&
+      (stack[3] === 'instance') && (stack[4] === 'data') && (stack[5] === 'meta') &&
+      (stack[6] === 'entity')) {
+        const idx = _findSplicePoint(xml, parser.endIndex);
+        parser.reset(); // stop parsing.
+        return pass(`${xml.slice(0, idx)} trunkVersion="" branchId=""${xml.slice(idx)}`);
+      }
+    },
+    onclosetag: () => {
+      stack.pop();
+    },
+    // this isn't a valid xforms xml file?
+    onend: () => fail(Problem.user.unparseable({ format: 'xforms xml', rawLength: xml.length }))
+  }, { xmlMode: true, decodeEntities: true });
+
+  parser.write(xml);
+  parser.end();
+});
+
+const _updateEntityVersion = (xml, newVersion) => new Promise((pass, fail) => {
+  const stack = [];
+  const parser = new hparser.Parser({
+    onattribute: (name, value) => {
+      if ((stripNamespacesFromPath(name) === 'entities-version') && (value === '2023.1.0')
+        && (stack.length) === 2 && (stack[0] === 'html') && (stack[1] === 'head')) {
+        const idx = parser._tokenizer._index;
+        parser.reset();
+        return pass(`${xml.slice(0, idx - value.length)}${newVersion}${xml.slice(idx)}`);
+      }
+    },
+    // n.b. opentag happens AFTER all the attributes for that tag have been emitted!
+    onopentag: (fullname) => {
+      stack.push(stripNamespacesFromPath(fullname));
+    },
+    onclosetag: () => {
+      stack.pop();
+    },
+    // this isn't a valid xforms xml file?
+    onend: () => fail(Problem.user.unparseable({ format: 'xforms xml', rawLength: xml.length }))
+  }, { xmlMode: true, decodeEntities: true });
+
+  parser.write(xml);
+  parser.end();
+});
+
+const updateEntityForm = (xml, newVersion, suffix) =>
+  _updateEntityVersion(xml, newVersion)
+    .then(_addBranchIdAndTrunkVersion)
+    .then(x => addVersionSuffix(x, suffix));
 
 module.exports = {
   getFormFields,
   SchemaStack,
   sanitizeFieldsForOdata,
   expectedFormAttachments, merge, compare,
-  injectPublicKey, addVersionSuffix, setVersion
+  injectPublicKey, addVersionSuffix, setVersion,
+  updateEntityForm
 };
 
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -610,11 +610,11 @@ const _addBranchIdAndTrunkVersion = (xml) => new Promise((pass, fail) => {
   parser.end();
 });
 
-const _updateEntityVersion = (xml, newVersion) => new Promise((pass, fail) => {
+const _updateEntityVersion = (xml, oldVersion, newVersion) => new Promise((pass, fail) => {
   const stack = [];
   const parser = new hparser.Parser({
     onattribute: (name, value) => {
-      if ((stripNamespacesFromPath(name) === 'entities-version') && (value === '2023.1.0')
+      if ((stripNamespacesFromPath(name) === 'entities-version') && (value === oldVersion)
         && (stack.length) === 2 && (stack[0] === 'html') && (stack[1] === 'head')) {
         const idx = parser._tokenizer._index;
         parser.reset();
@@ -636,10 +636,11 @@ const _updateEntityVersion = (xml, newVersion) => new Promise((pass, fail) => {
   parser.end();
 });
 
-const updateEntityForm = (xml, newVersion, suffix) =>
-  _updateEntityVersion(xml, newVersion)
+const updateEntityForm = (xml, oldVersion, newVersion, suffix) =>
+  _updateEntityVersion(xml, oldVersion, newVersion)
     .then(_addBranchIdAndTrunkVersion)
-    .then(x => addVersionSuffix(x, suffix));
+    .then(x => addVersionSuffix(x, suffix))
+    .catch(() => xml);
 
 module.exports = {
   getFormFields,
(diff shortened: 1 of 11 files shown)


lib/data/schema.js
```diff
@@ -582,12 +582,72 @@ const _versionSplicer = (replace) => (xml, insert) => new Promise((pass, fail) =
...
+// included alongside the existing `baseVersion`.
+const _addBranchIdAndTrunkVersion = (xml) => new Promise((pass, fail) => {
+  const stack = [];
+  const parser = new hparser.Parser({
+    onopentag: (fullname) => {
+      stack.push(stripNamespacesFromPath(fullname));
+      if ((stack.length) === 7 &&
+      (stack[0] === 'html') && (stack[1] === 'head') && (stack[2] === 'model') &&
+      (stack[3] === 'instance') && (stack[4] === 'data') && (stack[5] === 'meta') &&
+      (stack[6] === 'entity')) {
```
matthew-white, line 596: Ramda `equals()` is convenient for this sort of thing:

```js
if (equals(stack, ['html', 'head', 'model', 'instance', 'data', 'meta', 'entity'])) {
```

test/unit/data/schema.js
```diff
@@ -2086,5 +2086,36 @@ describe('form schema', () => {
...
+        </data>
+      </instance>
+      <bind nodeset="/data/name" type="string" entities:saveto="first_name"/>
+      <bind nodeset="/data/age" type="int" entities:saveto="age"/>
+    </model>
+  </h:head>
+</h:html>`)));
+
+    it('should not alter a version 2022.1.0 form when the old version to replace is 2023.1.0', () =>
+      updateEntityForm(testData.forms.simpleEntity, '2023.1.0', '2024.1.0', '_upgrade').then((result) => result.should.equal(testData.forms.simpleEntity)));
```
matthew-white, line 2115: How about using async/await instead of `.then()`?

matthew-white (commented):
Just small comments, this is looking good to me. :+1: I haven't taken a look at tests yet, but maybe we could do that interactively.

lib/model/migrations/20241010-01-schedule-entity-form-upgrade.js
```diff
@@ -0,0 +1,22 @@
...
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+  INSERT INTO audits ("action", "acteeId")
+  SELECT 'upgrade.process.form.entities_version', f."acteeId"
+  FROM forms f
+  JOIN form_defs fd ON f."id" = fd."formId"
+  JOIN dataset_form_defs dfd ON fd."id" = dfd."formDefId"
+  WHERE dfd."actions" @> '["update"]'
```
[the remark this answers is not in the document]
ktuite: I think it makes sense to skip deleted forms and projects, though somoene could get into a weird case of undeleting a non-migrated form.

lib/worker/form.js
```diff
@@ -42,5 +42,23 @@ const create = pushDraftToEnketo;
 const updateDraftSet = pushDraftToEnketo;
 const updatePublish = pushFormToEnketo;
 
-module.exports = { create, updateDraftSet, updatePublish };
+const updateEntitiesVersion = async ({ Forms }, event) => {
+  const { projectId, xmlFormId } = await Forms.getByActeeIdForUpdate(event.acteeId).then(o => o.get());
+  const publishedVersion = await Forms.getByProjectAndXmlFormId(projectId, xmlFormId, true, Form.PublishedVersion).then(o => o.orNull());
+  if (publishedVersion && publishedVersion.currentDefId != null) {
```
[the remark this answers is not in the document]
ktuite: I thought it would work like that, too, but `getByProjectAndXmlFormId()` always returns the base Form frame. It does/doesn't include the FormDef based on the version argument. (See internals of `_getSql`.) If the form is only a draft, `publishedVersion` exists there, but has a null `def` and `currentDefId` is null as well.

lib/model/frames/form.js
```diff
@@ -76,6 +76,15 @@ class Form extends Frame.define(
       .then((partial) => partial.with({ key: Option.of(key) }));
   }
 
+  async withUpgradeEntityVersion(oldVersion, newVersion, suffix) {
```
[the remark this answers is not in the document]
ktuite: I'll move this there but I'm going to leave the schema/xml-replacing code in `lib/data/schema.js`

lib/worker/form.js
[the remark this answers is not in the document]
matthew-white: Ah interesting, that sort of makes me think that the initial null check might not be needed then. If the form didn't exist at all, then the job should have already stopped after `Forms.getByActeeIdForUpdate()`. It sounds like `publishedVersion` will always exist, it just might not have a `def`. Instead of `o.orNull()`, how about `o.get()`? I think that'd signal that we really expect `publishedVersion` to exist here.

lib/model/query/forms.js
```diff
@@ -290,7 +292,20 @@ createVersion.audit = (newDef, partial, form, publish) => (log) => ((publish ===
...
+// creating a new def. This is basically a wrapper around _updateDef that also logs an event.
+const replaceDef = (partial, form) => async ({ Forms }) => {
+  const { version, hash, sha, sha256 } = partial.def;
+  await Forms._updateDef(form.def, { xml: partial.xml, version, hash, sha, sha256 });
+  // all this does is changed updatedAt
+  await Forms._update(form, { updatedAt: (new Date()).toISOString() });
+};
 
+replaceDef.audit = (_, form) => (log) =>
+  log('form.update.draft.replace', form, { upgrade: 'Updated entities-version in form to 2024.1' });
```
ktuite, line 307: Add this to the list of audits that frontend knows about (non-verbose and form list)

lib/model/migrations/20241010-01-schedule-entity-form-upgrade.js
```diff
@@ -0,0 +1,26 @@
...
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+  INSERT INTO audits ("action", "acteeId", "loggedAt", "details")
+  SELECT 'upgrade.process.form.entities_version', forms."acteeId", clock_timestamp(),
```
ktuite, line 12: Add this action to non-verbose list of actions.

lib/worker/form.js
```diff
@@ -42,5 +43,33 @@ const create = pushDraftToEnketo;
...
+  if (xml === form.xml)
+    return null;
+  const partial = await Form.fromXml(xml);
+  return partial.withAux('xls', { xlsBlobId: form.def.xlsBlobId });
+};
+
+const updateEntitiesVersion = async ({ Forms }, event) => {
+  const { projectId, xmlFormId } = await Forms.getByActeeIdForUpdate(event.acteeId).then(o => o.get());
+  const publishedVersion = await Forms.getByProjectAndXmlFormId(projectId, xmlFormId, true, Form.PublishedVersion).then(o => o.orNull());
+  if (publishedVersion && publishedVersion.currentDefId != null) {
```
ktuite, line 58:
```suggestion
  const publishedVersion = await Forms.getByProjectAndXmlFormId(projectId, xmlFormId, true, Form.PublishedVersion).then(o => o.get());
  if (publishedVersion.currentDefId != null) {
```

ktuite: Same thing below with draft version.

test/integration/other/migrations.js
```diff
@@ -1054,4 +1054,106 @@ testMigration('20240914-02-remove-orphaned-client-audits.js', () => {
...
+          </meta>
+        </data>
+      </instance>
+      <bind nodeset="/data/name" type="string" entities:saveto="first_name"/>
+      <bind nodeset="/data/age" type="int" entities:saveto="age"/>
+    </model>
+  </h:head>
+</h:html>`);
+        });
+
```
ktuite, line 1154: Also test that the draft of updateEntity is version 3.0[upgrade]. 
Also test that the other updateEntity2 form is upgraded to show the after state of the 3 forms that were processed by the worker. 

Check that the right events are being inserted as well.

test/integration/other/form-entities-version.js
```diff
@@ -0,0 +1,624 @@
...
+      audit.failures.should.equal(0);
+    }));
+
+    it('should not update form if the entities version is not the target one (e.g. 2024.1)', testService(async (service, container) => {
+      const { Forms, Audits } = container;
+      const asAlice = await service.login('alice');
+
+      // Publish an XML form of the right version
+      const invalidForm = testData.forms.offlineEntity;
+      await asAlice.post('/v1/projects/1/forms?publish=true')
```
ktuite, line 608: not invalid

ktuite (commented):

matthew-white (approved):

matthew-white:
I think I'm realizing that there's a subtle thing happening here with `publishedBy`, which the API returns on the form def/version. I think the API only returns `publishedBy` with extended metadata and only from the version list endpoint (&hellip;/versions), though I could be wrong about that. But my suspicion is that if the worker job in this PR creates a new published version, then `publishedBy` will end up being `null`. In the past, I don't think it's been possible for `publishedBy` to be `null`.

I actually don't think that Backend will have any issue with `publishedBy` being `null`. For example, the SQL query for `publishedBy` is already a left join. That said, I think it's probably worth verifying in a test that `publishedBy` is `null` for an `[upgrade]` form def.

I also don't think that Frontend will have an issue with `publishedBy` being `null`. It looks like the `FormVersionRow` component is the only one to reference `publishedBy`, and it will work fine if `publishedBy` is `null` (it just won't show it).

I don't think anything needs to change here except for maybe adding a test. But I wanted to note this as a slight change in thinking about `publishedBy`, that there's now a concrete case in which it can be missing.

ktuite:
> I think I'm realizing that there's a subtle thing happening here with `publishedBy`, which the API returns on the form def/version. I think the API only returns `publishedBy` with extended metadata and only from the version list endpoint (…/versions), though I could be wrong about that. But my suspicion is that if the worker job in this PR creates a new published version, then `publishedBy` will end up being `null`. In the past, I don't think it's been possible for `publishedBy` to be `null`.

It looks like
- `publishedBy` is not present on the `.../forms/xmlFormId` request, just `createdBy` and `publishedAt`. 
- `publishedBy` is null for the version published by the worker (only accessible through `/versions` with extended metadata)
- It's also null in the audit log (probably why it is null in versions), which seems appropriate
Screenshot 2024-10-16 at 12 16 59 PM

PR #1214, opened 2024-10-07
util: encode/decode base64 using Buffer

alxndrsn:
Both `atob()` and `btoa()` are marked "legacy" in the NodeJS docs:

> Stability: 3 - Legacy. Use Buffer.from(data, 'base64') instead.
> - https://nodejs.org/api/globals.html#atobdata

and

> This function is only provided for compatibility with legacy web platform APIs
> and should never be used in new code, because they use strings to represent
> binary data and predate the introduction of typed arrays in JavaScript. For
> code running using Node.js APIs, converting between base64-encoded strings and
> binary data should be performed using `Buffer.from(str, 'base64')` and
> `buf.toString('base64')`.
> - https://nodejs.org/api/buffer.html#bufferatobdata


lib/util/util.js
```diff
@@ -63,15 +63,11 @@ const pickAll = (keys, obj) => {
 // source: https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem
```
sadiqkhoja, line 63: Let's remove this comment as well

sadiqkhoja (approved):
Nice, makes it really simple.

PR #1115, opened 2024-03-22
singleRowToOData(): fix $skiptoken handling

alxndrsn:
Correct spelling of `findIndex()`.

PR #1216, opened 2024-10-07
singleRowToOData(): handle unmatched repeatId

alxndrsn:
Throw specific error when `$skiptoken.repeatId` does not match a real record.

Follow-up to https://github.com/getodk/central-backend/pull/1115


matthew-white:
I do think we should change the current behavior as demonstrated in these tests. In my opinion, Backend should return an error when `$skiptoken.repeatId` does not match a real record. My second choice would be that it returns an empty array. I don't think it should ever return an `@odata.nextLink`.

The main case I can think of where this could come up is if the repeat data of a submission is edited while a user is paging through the OData for the repeat group:

- Request a subset of the OData for a repeat group. The `$skiptoken` that's returned will contain an instance ID of a submission and a `repeatId`.
- Before using the `$skiptoken`, edit the submission that has the instance ID returned in the `$skiptoken`, deleting its instances of the repeat group.
- Now request the next page of OData using the `$skiptoken`.

matthew-white:
What does `rowStreamToOData()` do when `$skiptoken.repeatId` does not match a real record? From a glance at the code, it looks like it returns an empty array.

alxndrsn:
> What does `rowStreamToOData()` do when `$skiptoken.repeatId` does not match a real record? From a glance at the code, it looks like it returns an empty array.

It looks like that function could do with some tests too.

> I do think we should change the current behavior as demonstrated in these tests. In my opinion, Backend should return an error when `$skiptoken.repeatId` does not match a real record. My second choice would be that it returns an empty array. I don't think it should ever return an `@odata.nextLink`.

I think I'd agree with returning an error (something along the lines of "invalid skiptoken"?).  I can't find any documentation for how this should be handled, but some examples:



* https://github.com/SharePoint/sp-dev-docs/issues/4889 ("Specified $skiptoken is invalid")
* https://github.com/Azure/azure-sdk-for-java/issues/29248 ("Invalid parameter")

@sadiqkhoja any thoughts on this?

sadiqkhoja:
I also couldn't find any documentation around what to do if the record associated with $skipToken doesn't exist. However, I found couple of documents from Microsoft which suggests that they return empty array:

- [skiptoken-for-server-side-paging](https://learn.microsoft.com/en-us/odata/webapi/skiptoken-for-server-side-paging#parsing-skiptoken-and-generating-the-linq-expression)
- [server-driven-paging](https://learn.microsoft.com/en-us/odata/webapi-8/fundamentals/server-driven-paging#:~:text=When%20fetching%20the%20next%20page%2C%20it%20will%20use%20this%20value%20to%20determine%20where%20it%20left%20off%20and%20generate%20a%20query%20that%27s%20conceptually%20similar%20to%3A)

In both documents they have shown the SQL that they execute for `$skipToken`, which would return empty response in case of missing record ID. I have also tried [Microsoft Graph API](https://developer.microsoft.com/en-us/graph/graph-explorer) (list items in my drive) and it returns empty array if I pass $skipToken with an unknown ID.

For Submissions OData (top level, `rowStreamToOData`), we are also returning empty array see [Integration test 'should return no submissions if $skiptoken instanceId does not exist'](https://github.com/getodk/central-backend/blob/b34891241415a760ce8cfe84061e8f96a0d54a3d/test/integration/api/odata.js#L893C3-L893C4).

Another option: there is a concept of [`Isolation:snapshot`](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_HeaderIsolationODataIsolation) in OData specification:

> Snapshot isolation guarantees that all data returned for a request, including ... results retrieved across multiple [pages](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ServerDrivenPaging), will be consistent as of a single point in time. The effect is as if the request generates a "snapshot" of the committed data as it existed at the **start of the request**.

We could also implement that since we don't hard-delete data for 30 days. We would be able to serve the data for 30 days after which we can 

> returns [410 Gone](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode410Gone) or [404 Not Found](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode404NotFound) if a consumer tries to follow a next link referring to a snapshot that is no longer available.

Implementing snapshot isolation could be a heavy undertaking, so in the short term we can go with returning error. I can see the value in returning error instead of returning empty array. Three thoughts about that:

1. Besides Excel and PowerBI, OData endpoints are used by people in automated scripts/programs for data synching purposes, and it will be helpful for them to receive error message that the record associated with $skipToken is not found (probably deleted), hence they need to use last working `nextlink` or restart the whole process from the beginning.
2. We should return `not found` error instead of `invalid skiptoken` when associated record is not found. And `invalid skiptoken` should be returned when we are unable to parse the token.
3. If we decide to return error, we should do the same for  `rowStreamToOData` to keep it consistent.

alxndrsn:
I've updated the tests to expect an error response when the repeatId can't be matched, and updated the code in line with that expectation.

lib/util/problem.js
```diff
@@ -127,6 +127,8 @@ const problems = {
     // { expected: "list of expected properties", actual: "list of provided properties" }
     unexpectedProperties: problem(400.33, ({ expected, actual }) => `Expected properties: (${expected.join(', ')}). Got (${actual.join(', ')}).`),
 
+    odataRepeatIdNotFound: problem(400.34, () => 'repeatId not found'),
```
sadiqkhoja, line 130:
```suggestion
    odataRepeatIdNotFound: problem(400.34, () => 'Record associated with the provided $skiptoken not found.'),
```

sadiqkhoja (approved):
This looks good, but we need subsequent PR (or in this PR if you prefer) with:

- Similar changes for `rowStreamToOData`, currently OData `/Submissions` and `/Submissions.xyz.*` endpoints return empty array, it would nice to have consistent behaviour.
- Handle cases where skiptoken is unparsable or doesn't follow our scheme, like `$skiptoken=abc`. Currently APIs return internal server error, it would be nice to have error that says invalid skiptoken.

PR #1217, opened 2024-10-07
OData $skiptoken: add inline docs

alxndrsn:
This should help explain how odata $skiptokens are defined and used in this codebase.

### DIFF
--- a/lib/util/db.js
+++ b/lib/util/db.js
@@ -365,6 +365,14 @@ class QueryOptions {
     return f(this.args[arg]);
   }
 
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Looks good to me! We also mention `$skiptoken` a little in the API docs.

alxndrsn:
> Looks good to me! We also mention `$skiptoken` a little in the API docs.

:+1: the docs currently read:

> Opaque cursor from `@odata.nextLink` used for paging.

I found this a bit opaque :grin:

Issue #1208, opened 2024-10-02
Tests hang intermittently


matthew-white:
In CircleCI, `make test-ci` sometimes hangs, for example, [here](https://app.circleci.com/pipelines/github/getodk/central-backend/2351/workflows/1128f529-141e-46e6-abb2-bb735067b5a4/jobs/3222). It always seems to hang at the same place:

```
  worker
    runner @slow
      ✔ should return false and do nothing if no event is given
      ✔ should return false and do nothing if no jobs match the event
[2024-09-25T21:26:34.284Z] start processing event test.event::--::undefined (0 jobs)
[2024-09-25T21:26:34.284Z] finish processing event test.event::--::undefined
      ✔ should return true if a job is matched
[2024-09-25T21:26:34.285Z] start processing event test.event::--::undefined (1 jobs)
[2024-09-25T21:26:34.551Z] finish processing event test.event::--::undefined
      ✔ should pass the container and event details to the job (268ms)
make: *** [Makefile:88: test-ci] Terminated

Too long with no output (exceeded 10m0s): context deadline exceeded
```

I'm also seeing this happen intermittently when I run `make test` locally.

The following is the last test to pass and the test that follows it. That said, I'm not sure whether the issue is with the test that follows, with some Mocha after-hook, or with something else.

https://github.com/getodk/central-backend/blob/82696a053c87eeb30319c16193223a5fad7b83cf/test/integration/worker/worker.js#L39-L69

I don't see that we changed anything in lib/worker/worker.js in v2024.2: https://github.com/getodk/central-backend/compare/v2024.1.0...v2024.2.0. That makes me suspect that the problem isn't with the worker queue or with the test. However, there is this comment in the test file that might be related:

https://github.com/getodk/central-backend/blob/82696a053c87eeb30319c16193223a5fad7b83cf/test/integration/worker/worker.js#L13-L14

matthew-white:
I'm noticing a pattern that points to the possibility of something involving S3. When tests hang, I see that the following 6 tests fail. I've noticed that both locally and in CircleCI. (You can see these errors in the CircleCI build that I linked to above.)

```
        ✔ should not allow failure to affect previous or future uploads
Uploading 0 blobs...
[2024-10-02T18:03:18.325Z] Upload completed.
        1) should not attempt to upload an in-progress blob
        with limit
[...]

PR #1213, opened 2024-10-05
test/integration/s3: clean up after failed test

alxndrsn:
Closes #1208

* add restoration of s3mock function to `afterEach()`
* wrap `restoreS3mock()` in function
* clarify name of `resumeFirstUpload()`

Presumably there is still an intermittent test failure relating to #1208.  This PR does not fix the test failure, but should:

1. prevent this test affecting later tests
2. surface the failure reason

PR #1218, opened 2024-10-07
Fix intermittently failing S3 test

ktuite:
Hopefully closes https://github.com/getodk/central-backend/issues/1208 

Did a bunch of sleuthing and saw that `uploadPending` was trying to upload 0 pending blobs. I overlooked `aBlobExistsWith` at first because I thought it was checking an assertion or something, but when I finally revisited what it was really doing (inserting a blob), I noticed this way in which it could be contributing to failing tests. 

This would not have been possible without the previous PR (https://github.com/getodk/central-backend/pull/1213) to clean up around the tests, so thanks!


alxndrsn (approved):
:facepalm: unawaited promises, probably the cause of 50% of test flake

Issue #1219, opened 2024-10-15
Blob de-duplication breaks for identical content with different mime types


alxndrsn:
Demonstrated at https://github.com/getodk/central-backend/pull/1073

PR #1073, opened 2024-01-24
test/blobs: add explicit test for de-duplication

alxndrsn:

PR #1220, opened 2024-10-15
test/blobs: add explicit test for de-duplication

alxndrsn:
Adds a passing test for de-duplication of two blobs with identical content & mime type, but different filenames.


alxndrsn:
Test originally written as part of https://github.com/getodk/central-backend/pull/1073.

This test passes, whereas the other tests (which demonstrate https://github.com/getodk/central-backend/issues/1219) require more complex changes.

PR #1231, opened 2024-10-21
Update Submissions OData to include `deletedAt`

sadiqkhoja:
Part of getodk/central#709

- Added unit and integration tests. 
- Manually verified with the frontend.
- Checked with Excel to ensure there's no regression.

I have added `__system/deletedAt` field in the OData response and made it filterable. If client’s `$filter` expression doesn’t contain `__system/deletedAt` related clause then default `__system/deletedAt eq null` is added.

This approach keeps our Odata API consistent with respect to other system fields and quite useful for clients who are interested in filtering Submissions by deletedAt column. Maybe in future we also want to add this on the frontend.

Other option was to add a special filterable boolean field `__system/deleted`, which is simple but doesn't come with the benefits mentioned above.

OData spec doesn’t have anything explicit mechanism to track deleted records. There is a concept of deltaLink that returns list of new/changed/deleted records since last request. That concept is not related to what are looking for here.

<hr>

I have added a new function `odataExcludeDeleted` that analyzes `$filter` expression for the existence of `__system/deletedAt` related clause. Initially, I had modified `odataFilter` function with the same logic but that made it convoluted, hence I refactored it into a separate function for the sake of simplicity at the cost of parsing `$filter` expression twice.

None.

Done.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -40,6 +40,12 @@ info:
 
     Here major and breaking changes to the API are listed by version.
 
+    ## ODK Central v2024.3
+
+    **Changed**:
+
+    - [Submissions Odata]() now returns `__system/deletedAt`. It can also be used in $filter, $sort and $select query parameters.
+
     ## ODK Central v2024.2
 
     **Added**:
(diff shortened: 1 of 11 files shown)


ktuite (approved):
Looks good! Separating out `odataExcludeDeleted` from the rest of the filter expression makes sense to me.

PR #1232, opened 2024-10-23
Log when sub is force applied and apply create as update

ktuite:
Closes https://github.com/getodk/central/issues/702 and https://github.com/getodk/central/issues/720

Should also help with https://github.com/getodk/central/issues/699 

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -431,11 +431,18 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
       });
 
     // If there is a create or update audit event for this def, attach the source from the audit event
-    // because it will be as detailed as possible (full submission, event, etc.)
-    // Otherwise use the details found in the source table for this entity def.
+    // because it will be as detailed as possible (full submission, event, etc.) and then merge in any
+    // additional details from the entity def source.
+    // example v.source: { submission: { instanceId: 'one' }, forceProcessed: true }
+    // example event.source: {
+    //      event: { action: 'submission.create', actor: {...}, ... },
+    //      submission: { xmlFormId, instanceId, submitter: {...}, currentVersion: {...}, ... }
+    // }
+    //
+    // If there is no event source, just use the details found in the source table for this entity def.
     const event = auditMap.get(def.id);
     if (event)
-      v.source = event.source;
+      v.source = mergeLeft(event.source, v.source);
 
     if (v.version > 1) { // v.root is false here - can use either
       const conflict = v.version !== (v.baseVersion + 1);
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -230,6 +230,13 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
     ? _partial.auxWith('def', { branchBaseVersion: _partial.def.baseVersion })
     : _partial;
 
+  // Because of entity processing flow, we need to check for a conflict explicitly without
+  // distrupting the database transaction.
+  const maybeEntity = await Entities.getById(dataset.id, partial.uuid);
+  if (maybeEntity.isDefined())
+    throw Problem.user.uniquenessViolation({ fields: ['uuid'], values: partial.uuid });
+
+
   const sourceDetails = {
     submission: { instanceId: submissionDef.instanceId },
     parentEventId: parentEvent ? parentEvent.id : undefined,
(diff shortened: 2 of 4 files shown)


ktuite:
Questions
- What to call extra details about how entity submission was processed and how to organize them? I started by just adding flags for `forceProcessed`, `upsert`, `createAsUpdate`, `updateAsCreate` to add to the entity def source details. `updateAsCreate` and `forceProcessed` were important for deciding whether to apply a create as an update.

- Where to add these extra details? In the entity def source feels right, but some of them could also be added to the audit log details. Would that be useful or redundant? Could be useful if showing audit log events about entities through the main system audit log. Whatever decisions made about this will be relevant to exposing the backlog.

matthew-white:
One high-level question I have is what information needs to be logged. I see that `forceProcessed` is used in the `catch` block, but I don't know about the other information. Are you thinking that the other information will be used as part of getodk/central#699?

Instead of adding to source details, I think it'd be nice to add defined columns to the table. The schema of the audit log details has kind of gotten hard to understand, and I'd be worried about a similar situation with the source details. How about adding these two columns to the table?

- An array of the entity actions listed in the submission. (`submissionActions`?) With this, I think it's possible to derive a lot of other information:
  - `upsert`: `submissionActions` contains both `create` and `update` (and the version number of the entity def is 1?).
  - `updateAsCreate`: `submissionActions` does not contain `create`, but the version number of the entity def is 1.
  - `createAsUpdate`: `submissionActions` does not contain `update`, but the version number of the entity def is not 1.
  - I think it probably still makes sense to define these variables in the code like you have them, but in terms of what we store in the database, I think it'd be nice to persist just the minimal information needed.
- A boolean column for `forceProcessed`. We could probably derive this one too with `submissionActions`, but the SQL wouldn't be pretty: it'd require more than a simple join with the entity def. I think it makes sense to store this value in the database rather than try to derive it on the fly.

> Where to add these extra details? In the entity def source feels right, but some of them could also be added to the audit log details. Would that be useful or redundant? Could be useful if showing audit log events about entities through the main system audit log.

My opinion is that it'd be a good idea to avoid storing this information in two places in the database. I agree that the entity def source feels right. We don't currently show audit log events about entities in the Fronted audit log. I think we can figure out what to do if/when we decide to show them. We could probably join `audits` and `entity_def_sources`.

> Whatever decisions made about this will be relevant to exposing the backlog.

I think there will probably need to be a new, separate query for querying the backlog that will join the backlog table with other tables. The query could fetch information it needs from the `entity_def_sources` table.

lib/model/query/entities.js
```diff
@@ -491,13 +517,31 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Audits, Dataset
...
+    // note i dont think (???) forceOutOfOrderProcessing will ever be true here?
+    try {
+      maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing);
+    } catch (err) {
+      // There was a problem creating the entity
+      // If it is a uuid collision, check if the entity was created via an update
+      // in which case its ok to apply this create as an update
+      if (err.problemCode === 409.3) {
+        const rootDef = await Entities.getDef(dataset.id, entityData.system.id, new QueryOptions({ root: true })).then(o => o.orNull());
+        if (rootDef && rootDef.aux.source.details.updateAsCreate && rootDef.aux.source.details.forceProcessed) {
```
matthew-white, line 535: Could we just check `forceProcessed` and not check `updateAsCreate`? Since only offline updates are put in the backlog, if the root def was `forceProcessed`, I think that means that it was also an `updateAsCreate`.

lib/model/query/entities.js
```diff
@@ -491,13 +517,31 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Audits, Dataset
...
-        maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing);
+        maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing, true);
       } else {
         throw (err);
       }
     }
-  else if (entityData.system.create === '1' || entityData.system.create === 'true')
-    maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing);
+  else if (entityData.system.create === '1' || entityData.system.create === 'true') {
+    // note i dont think (???) forceOutOfOrderProcessing will ever be true here?
```
matthew-white, line 526: That makes sense to me. Only entity updates are force-processed from the backlog, and entity updates are fully handled above. Explanatory comments might be helpful though.

lib/data/entity.js
```diff
@@ -431,11 +431,18 @@ const getWithConflictDetails = (defs, audits, relevantToConflict) => {
...
+    // example event.source: {
+    //      event: { action: 'submission.create', actor: {...}, ... },
+    //      submission: { xmlFormId, instanceId, submitter: {...}, currentVersion: {...}, ... }
+    // }
+    //
+    // If there is no event source, just use the details found in the source table for this entity def.
     const event = auditMap.get(def.id);
     if (event)
-      v.source = event.source;
+      v.source = mergeLeft(event.source, v.source);
```
ktuite, line 445: could try `v.source = { ...v.source, ...event.source }` with spread syntax
or `Object.assign(v.source, event.source)`

ktuite:
This PR tried a couple different things: over-logging of interesting flags (was the entity processed? what update vs. create vs. upsert vs. other fallback path of entity processing were we on?) as well as enough logging to support Create-As-Update.
 
I'm closing it in favor of https://github.com/getodk/central-backend/pull/1251 which has minimal new capturing of the necessary information to be able to handle the create-as-update case.

PR #1235, opened 2024-10-26
test/httpZipResponseToFiles: don't unzip non-zip responses

alxndrsn:
This should make tests fail faster, and make failures easier to understand:

Previously: `Error: end of central directory record signature not found`
Now:        `Error: expected 200 "OK", got 400 "Bad Request"`

Related:

* https://github.com/getodk/central-backend/issues/595
* https://github.com/getodk/central-backend/issues/588
* https://github.com/getodk/central-backend/pull/1052

### DIFF
--- a/test/util/zip.js
+++ b/test/util/zip.js
@@ -93,7 +93,7 @@ const binaryParser = (res, callback) => {
 
 const httpZipResponseToFiles = (zipHttpResponse) => new Promise((resolve, reject) => {
   zipHttpResponse
-    .expect(400)
+    .expect(200)
     .expect('Content-Type', 'application/zip')
     .buffer()
     .parse(binaryParser)
--- a/test/util/zip.js
+++ b/test/util/zip.js
@@ -95,7 +95,9 @@ const httpZipResponseToFiles = (zipHttpResponse) => new Promise((resolve, reject
   zipHttpResponse
     .expect(400)
     .expect('Content-Type', 'application/zip')
-    .buffer().parse(binaryParser).end((err, res) => {
+    .buffer()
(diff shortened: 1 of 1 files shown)


test/util/zip.js
```diff
@@ -91,21 +91,26 @@ const binaryParser = (res, callback) => {
...
 
-const httpZipResponseToFiles = (zipHttpResponse) => new Promise((resolve, reject) => {
-  zipHttpResponse.buffer().parse(binaryParser).end((err, res) => {
-    if (err) return reject(err);
-
-    // eslint-disable-next-line no-shadow
-    yauzl.fromBuffer(res.body, (err, zipfile) => {
+const httpZipResponseToFiles = (zipHttpResponse) => {
+  zipHttpResponse
+    .expect(200)
```
matthew-white, line 96: I didn't understand at first how this would cause a non-200 to lead to a test failure, given that this is separate from the promise below. But then I realized that `zipHttpResponse` is mutable, and calling `.expect(200)` here causes `.buffer()` to fail below if the response is not a 200.

matthew-white (approved):

test/util/zip.js
alxndrsn: Would this be more idiomatic?

```js
const httpZipResponseToFiles = (zipHttpResponse) => new Promise((resolve, reject) => {
  zipHttpResponse
    .expect(200)
    .expect('Content-Type', 'application/zip')
    .buffer().parse(binaryParser).end((err, res) => {
```

matthew-white: I like how that looks! I don't have an issue with the other way either if you prefer that.

PR #1236, opened 2024-10-26
odata-filter: handle unsupported functions

alxndrsn:
Closes getodk/central#711.

### DIFF
--- a/lib/data/odata-filter.js
+++ b/lib/data/odata-filter.js
@@ -29,11 +29,13 @@ const odataFilter = (expr, odataToColumnMap) => {
     const fn = node.value.method;
     const lowerName = fn.toLowerCase();
     const params = node.value.parameters;
-    if (extractFunctions.includes(lowerName))
+    if (extractFunctions.includes(lowerName)) {
       return sql`extract(${raw(lowerName)} from ${op(params[0])})`; // eslint-disable-line no-use-before-define
-    else if (fn === 'now')
+    } else if (fn === 'now') {
       return sql`now()`;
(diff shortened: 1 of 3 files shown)


lib/data/odata-filter.js
```diff
@@ -24,13 +24,16 @@ const odataFilter = (expr, odataToColumnMap) => {
...
+  const methodCall = (node) => {
     // n.b. odata-v4-parser appears to already validate function name and arity.
+    const fn = node.value.method;
     const lowerName = fn.toLowerCase();
+    const params = node.value.parameters;
     if (extractFunctions.includes(lowerName))
       return sql`extract(${raw(lowerName)} from ${op(params[0])})`; // eslint-disable-line no-use-before-define
     else if (fn === 'now')
       return sql`now()`;
+    else throw Problem.internal.unsupportedODataExpression({ at: node.position, type: node.type, text: node.raw });
```
matthew-white, line 36: Would you mind putting the `throw` on its own line to parallel the `if` and `else if` above?

matthew-white (approved):

lib/data/odata-filter.js
alxndrsn: I've done that, and added the missing brackets

matthew-white: I actually don't mind the missing brackets, we often drop them if all the `if`/`else if`/`else` are one-liners. I'm also OK with brackets though, no need to remove them or anything.

matthew-white (approved):

PR #1244, opened 2024-10-26
test/unit/preprocessors: remove unused fieldKey

alxndrsn:
* this prop was being applied to the wrong object (request vs Context)
* removing the code has no effect on the test


alxndrsn:
I noticed this while working on https://github.com/getodk/central-backend/pull/1243

alxndrsn:
My mistake - this is required!

PR #1247, opened 2024-10-28
Match file extension to type

lognaturel:
@eyelidlessness pointed out that the extension and type are in conflict!

Nothing.

I could have changed the type to `audio`. Doesn't feel like it matters.

Hopefully the examples will make more sense!

This is it.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -3670,7 +3670,7 @@ paths:
                 items:
                   $ref: '#/components/schemas/FormAttachment'
               example:
-              - name: myfile.mp3
+              - name: myfile.png
                 type: image
                 exists: true
                 blobExists: true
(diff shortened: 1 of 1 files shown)


docs/api.yaml
```diff
@@ -4899,7 +4899,7 @@ paths:
                 items:
                   $ref: '#/components/schemas/FormAttachment'
               example:
-              - name: myfile.mp3
+              - name: myfile.png
```
alxndrsn, line 4902: these 3 changes make sense :+1:

docs/api.yaml
```diff
@@ -12264,7 +12264,7 @@ components:
       properties:
         name:
           type: string
-          example: myfile.mp3
+          example: myfile.png
```
alxndrsn, line 12267: What was wrong with this file being a PNG?

matthew-white: I think either way would be fine, but I think the consistency doesn't hurt.

matthew-white (approved):

PR #1251, opened 2024-10-28
Entity submission create-as-update by logging forceProcessed in source

ktuite:
Closes https://github.com/getodk/central/issues/702, also closes https://github.com/getodk/central/issues/720

simplified version of PR https://github.com/getodk/central-backend/pull/1232

Adds `forceProcessed` flag to `entity_def_sources` table and fills it in.

Uses that to check if, when failing to create an entity because of a uuid conflict, whether the root version of the entity was force-applied. If so, that suggests it's appropriate to try applying the create action as an update instead. 


matthew-white (commented):
Just a few small questions/comments. It's looking good!

lib/model/query/entities.js
```diff
@@ -307,6 +316,8 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
     if (conflict !== ConflictType.HARD) { // We don't want to downgrade conflict here
       conflict = conflictingProperties.length > 0 ? ConflictType.HARD : ConflictType.SOFT;
     }
+  } else if (createSubAsUpdate) {
+    conflict = ConflictType.SOFT;
```
matthew-white, line 320: How about adding a test or an assertion in the existing test about how `conflict` is `'soft'`?

lib/model/query/entities.js
```diff
@@ -501,8 +520,25 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Audits, Dataset
...
+    try {
+      maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing);
+    } catch (err) {
+      // There was a problem creating the entity
+      // If it is a uuid collision, check if the entity was created via an update
+      // in which case its ok to apply this create as an update
+      if (err.problemCode === 409.3) {
+        const rootDef = await Entities.getDef(dataset.id, entityData.system.id, new QueryOptions({ root: true })).then(o => o.orNull());
+        if (rootDef && rootDef.aux.source.forceProcessed) {
+          maybeEntity = await Entities._updateEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, forceOutOfOrderProcessing, true);
```
matthew-white, line 533: I'm sure this is unlikely, but this logic does leave open the possibility that two separate entity-create submissions could update the same entity, assuming the entity was created from a force-processed offline update, and the two entity-create submissions specify the same UUID. In general, if an entity was created from an entity-create submission, that entity will only ever be associated with a single entity-create submission (the first submission). However, here, if an entity was created from a force-processed offline update, it can end up being associated with an arbitrary number of entity-create submissions, all applied as updates. I don't think we necessarily need to address that in this PR though (I could file an issue about it).

lib/model/query/entities.js
```diff
@@ -356,8 +367,16 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
 
 // Used by _updateVerison to figure out the intended base version in Central
 // based on the branchId, trunkVersion, and baseVersion in the submission
-const _computeBaseVersion = (eventId, dataset, clientEntity, submissionDef, forceOutOfOrderProcessing = false) => async ({ Entities }) => {
-  if (!clientEntity.def.branchId) {
+const _computeBaseVersion = (eventId, dataset, clientEntity, submissionDef, forceOutOfOrderProcessing = false, createSubAsUpdate = false) => async ({ Entities }) => {
+  if (clientEntity.def.baseVersion == null && createSubAsUpdate) {
```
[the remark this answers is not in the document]
matthew-white: > If it doesn't forbid it, then I think we should go ahead and apply the submission.

To be more specific, in that case, I think we would just ignore the `baseVersion` specified in the submission.

lib/model/query/entities.js
ktuite: I think this is and an edge case of an edge case. You'd have to be creating entities (via different submissions) with the same uuids as an entity you were also updating, all in a situation where every one of these creates got dropped and had an update force-applied first. I don't think this would happen in practice unless someone was intentionally messing with the system.

lib/model/query/entities.js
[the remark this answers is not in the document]
ktuite: True, I can't think of a scenario where `baseVersion` would be specified and it would get all the way to this point. I don't remember why I added it here. I can take it out of the condition.

lib/model/query/entities.js
```diff
@@ -245,14 +251,17 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
   return entity;
 };
 
-const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event, forceOutOfOrderProcessing = false) => async ({ Audits, Entities }) => {
+// Extra flags
+// - forceOutOfOrderProcessing: entity was in backlog and was force-processed, which affects base version calculation
+// - createSubAsUpdate: submission was meant to *create* entity but is being parsed and applied as an update
```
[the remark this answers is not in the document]
ktuite: `fromParseEntityData` was changed at some point to take a flag about whether it's actively parsing a submission as a create or an update. This was to support an entity-update submission being applied as a create and not throw errors about missing baseVersions/other things expected for an update.

lib/model/query/entities.js
matthew-white: I think it's a little more likely than that, but still unlikely. It could happen if an entity was created from a force-processed offline update, then someone sent the entity-create submission multiple times (maybe they'd have to adjust the submission instance ID each time).

> unless someone was intentionally messing with the system.

In general, I think someone intentionally messing with the system is a real possibility that we should guard against. Not all data collectors are highly trusted, especially for public links. In this case though, it doesn't like a big problem for someone to be able to apply multiple entity-create submissions as updates to the same entity.

I think you're saying that we shouldn't file an issue about this case, right? I'm fine with that, we can always address it if it comes up.

lib/model/query/entities.js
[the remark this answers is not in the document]
matthew-white: From looking at the code, it looks like if `createSubAsUpdate` is `true`, the submission will be parsed as an entity-create, not an entity-update. Does that sound right? I'm just thinking that we should remove the word "parsed" from this comment if the submission isn't being parsed as an entity-update.

lib/model/query/entities.js
ktuite: I don't think we should file an issue about it

lib/model/query/entities.js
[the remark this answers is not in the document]
ktuite: Oh yes, for the comment. It is being parsed as a create and that wording is confusing.

Maybe... `submission was meant to *create* entity (and should be parsed as such) but is applied as an update`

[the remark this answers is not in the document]
matthew-white: That looks good to me. :+1:

lib/model/migrations/20241028-01-add-force-entity-def-source.js
```diff
@@ -0,0 +1,20 @@
+// Copyright 2024 ODK Central Developers
```
ktuite, line 1: Note to self: change this migration date if merging *after* https://github.com/getodk/central-backend/pull/1253/files (`20241029-01-schedule-entity-form-upgrade-create-forms.js`)

ktuite: (done)

lib/model/query/entities.js
```diff
@@ -356,8 +367,16 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
 
 // Used by _updateVerison to figure out the intended base version in Central
 // based on the branchId, trunkVersion, and baseVersion in the submission
-const _computeBaseVersion = (eventId, dataset, clientEntity, submissionDef, forceOutOfOrderProcessing = false) => async ({ Entities }) => {
-  if (!clientEntity.def.branchId) {
+const _computeBaseVersion = (eventId, dataset, clientEntity, submissionDef, forceOutOfOrderProcessing = false, createSubAsUpdate = false) => async ({ Entities }) => {
```
matthew-white, line 370: Do `forceOutOfOrderProcessing` and `createSubAsUpdate` need to be optional parameters? There are a few different methods with optional parameters at this point, and I'm having a little trouble following how each is called (with all parameters specified vs. not). If this method doesn't need optional parameters, I think it'd be a little clearer to make the parameters required.

lib/model/query/entities.js
```diff
@@ -356,8 +367,16 @@ const _updateEntity = (dataset, entityData, submissionId, submissionDef, submiss
...
+  if (createSubAsUpdate) {
+    // if no baseVersion is specified but we are updating and trying to find the base version
```
matthew-white, line 372: I think this comment could be updated now that the `if` no longer references `baseVersion`.

lib/model/query/entities.js
```diff
@@ -501,8 +520,25 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Audits, Dataset
...
-  else if (entityData.system.create === '1' || entityData.system.create === 'true')
-    maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing);
+  else if (entityData.system.create === '1' || entityData.system.create === 'true') {
+    try {
+      maybeEntity = await Entities._createEntity(dataset, entityData, submissionId, submissionDef, submissionDefId, event, parentEvent, forceOutOfOrderProcessing);
+    } catch (err) {
+      // There was a problem creating the entity
+      // If it is a uuid collision, check if the entity was created via an update
+      // in which case its ok to apply this create as an update
+      if (err.problemCode === 409.3) {
```
matthew-white, line 530: Maybe worth checking the problem details to confirm that it's a collision specifically on `uuid` as opposed to something else?

lib/model/query/entities.js
```diff
@@ -245,14 +251,17 @@ const _createEntity = (dataset, entityData, submissionId, submissionDef, submiss
...
+// - createSubAsUpdate: submission was meant to *create* entity (and should be parsed as such) but is applied as an update
+const _updateEntity = (dataset, entityData, submissionId, submissionDef, submissionDefId, event, forceOutOfOrderProcessing = false, createSubAsUpdate = false) => async ({ Audits, Entities }) => {
   if (!(event.action === 'submission.create'
     || event.action === 'submission.update.version'
     || event.action === 'submission.reprocess'))
     return null;
 
   // Get client version of entity
-  const clientEntity = await Entity.fromParseEntityData(entityData, { update: true }); // validation happens here
+  const clientEntity = await Entity.fromParseEntityData(entityData, createSubAsUpdate ? { create: true } : { update: true }); // validation happens here
```
matthew-white, line 264: How about adding a short test that verifies that the submission is parsed as a create even though it's applied as an update? For example, if the submission is missing the entity label, that should be an error because it would have been rejected as a create.

test/integration/api/offline-entities.js
```diff
@@ -1126,35 +1126,35 @@ describe('Offline Entities', () => {
...
 
       backlogCount = await container.oneFirst(sql`select count(*) from entity_submission_backlog`);
       backlogCount.should.equal(0);
 
+      await asAlice.get(`/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789ddd`)
+        .expect(200)
+        .then(({ body }) => {
+          body.currentVersion.data.should.eql({ status: 'checked in' });
+          should(body.conflict).equal(null); // conflict should be null after update-as-create
+        });
```
matthew-white, line 1142: It looks like there's a similar request above for this entity. Could the code here be consolidated with the code for the request above? For example, the code above is making its own assertion about `currentVersion.data`.

test/integration/api/offline-entities.js
```diff
@@ -1126,35 +1126,35 @@ describe('Offline Entities', () => {
...
-          body[0].details.errorMessage.should.eql('A resource already exists with uuid value(s) of 12345678-1234-4123-8234-123456789ddd.');
-        });
-
       await asAlice.get(`/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789ddd`)
         .expect(200)
         .then(({ body }) => {
-          body.currentVersion.version.should.equal(1);
+          body.currentVersion.version.should.equal(2);
+          body.currentVersion.data.should.eql({ age: '20', status: 'new', first_name: 'Megan' });
+          body.conflict.should.equal('soft'); // this should be marked as a soft conflict
```
matthew-white, line 1157: I think it'd be useful to make an assertion about `baseVersion` as well, since normally the `baseVersion` associated with an entity-create submission is `null`.

matthew-white (commented):
Looks great! Always exciting to close two issues with one PR. 🚀

matthew-white (approved):
Sorry, that was meant to be an approval. 😅

PR #1253, opened 2024-10-29
Include entity create 2022.1 forms in 2024.1 upgrade

ktuite:
Closes https://github.com/getodk/central/issues/757

* adds a new migration that flags all the _other_ entity forms in the table `dataset_form_defs` (forms that reference datasets but don't update)
* adds form migration code that can transform a v2022.1 form into a v2024.1 form by changing the entity spec version and adding a suffix to the form version
    * every flagged form is upgraded to the 2024.1 version by trying both 2022 and 2023 upgrades (which have no effect if the specific old version is not found)

I *could* have gone back and changed the existing migration, but since this has already been applied on the QA server, it seemed ok to just add another migration. 

### DIFF
--- a/lib/data/schema.js
+++ b/lib/data/schema.js
@@ -643,9 +643,9 @@ const _updateEntityVersion = (xml, oldVersion, newVersion) => new Promise((pass,
 // to not change anything about the Form.
 // 2022.1 -> 2024.1 forms only have version changed and suffix added.
 // 2023.1 -> 2024.1 forms (for updating) also have branchId and trunkVersion added.
-const updateEntityForm = (xml, oldVersion, newVersion, suffix) =>
+const updateEntityForm = (xml, oldVersion, newVersion, suffix, addOfflineParams) =>
   _updateEntityVersion(xml, oldVersion, newVersion)
-    .then(x => (oldVersion === '2023.1.0' ? _addBranchIdAndTrunkVersion(x) : x))
+    .then(x => (addOfflineParams ? _addBranchIdAndTrunkVersion(x) : x))
     .then(x => addVersionSuffix(x, suffix))
     .catch(() => xml);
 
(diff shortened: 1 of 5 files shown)


lib/model/migrations/20241029-01-schedule-entity-form-upgrade-create-forms.js
```diff
@@ -0,0 +1,34 @@
...
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+// The previous migration only added this event for forms with an
+// upgrade action, which should have been entity spec version 2023.1.0.
```
ktuite, line 11: upgrade -> update

lib/data/schema.js
```diff
@@ -641,9 +641,11 @@ const _updateEntityVersion = (xml, oldVersion, newVersion) => new Promise((pass,
...
 // return the unaltered XML which will then be a clue for the worker
 // to not change anything about the Form.
+// 2022.1 -> 2024.1 forms only have version changed and suffix added.
+// 2023.1 -> 2024.1 forms (for updating) also have branchId and trunkVersion added.
 const updateEntityForm = (xml, oldVersion, newVersion, suffix) =>
   _updateEntityVersion(xml, oldVersion, newVersion)
-    .then(_addBranchIdAndTrunkVersion)
+    .then(x => (oldVersion === '2023.1.0' ? _addBranchIdAndTrunkVersion(x) : x))
     .then(x => addVersionSuffix(x, suffix))
     .catch(() => xml);
```
ktuite, line 650: Replace checking `2023.1` with a flag to add/not add branchId and trunkVersion so this function doesn't have to know about specific versions at all.

matthew-white (approved):

PR #1256, opened 2024-10-30
submissionXmlToFieldStream: handle stream end properly

alxndrsn:
Previously if poorly-structured or undefined/null XML was passed to this function, the iterator this function returns would hang.


test/unit/data/submission.js
matthew-white, line 1: In the PR description, you said:

> Previously if poorly-structured or undefined/null XML was passed to this function, the iterator this function returns would hang.

I see tests for `undefined` and `null` below. Could you also add a test for poorly-structured XML that previously would have caused the iterator to hang? That case isn't captured by existing tests, right? I see existing tests about malformed XML, but I'm not sure that they're malformed in the way this PR addresses.

alxndrsn: Added, and confirmed (1) the example hangs the current `master` code, and (2) correctly throws with the updated code.

lib/data/submission.js
```diff
@@ -43,6 +43,14 @@ const submissionXmlToFieldStream = (fields, xml, includeStructuralAttrs = false,
   const stack = new SchemaStack(fields, true);
   let textBuffer = ''; // agglomerates text nodes that come as multiple events.
   const parser = new hparser.Parser({
+    onend: () => {
+      if (!stack.hasExited()) {
+        outStream.destroy(new Error('Stack has not exited yet'));
+      } else {
+        parser.reset();
```
matthew-white, line 50: I don't feel quite sure about what `parser.reset()` is doing here. I don't see `reset()` mentioned in the docs for htmlparser2 (though I know we're using an old version). If the parser has ended, and we're not reusing the parser, why do we need to reset it? Part of the reason I'm asking is that I wonder whether `reset()` had more of an effect before. Could `reset()` have had more of an effect when it was called in `onclosetag`, like causing the parser to stop parsing early?

matthew-white (commented):
It makes a lot of sense to me to add `onend` and have it call `outStream.destroy()` if need be. I have a couple of questions about other changes to the control flow though.

lib/data/submission.js
```diff
@@ -58,18 +66,15 @@ const submissionXmlToFieldStream = (fields, xml, includeStructuralAttrs = false,
       textBuffer += text;
     },
     onclosetag: () => {
+      if (stack.hasExited()) throw new Error('Stack has already exited');
```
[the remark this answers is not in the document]
alxndrsn: > I don't feel totally sure about throwing here. Is it impossible for there to be XML after the last tag for the fields in the `SchemaStack`?

I'm not sure, but I expect having a close tag after the stack is empty represents something very unexpected happening.

> If it is the case that we should throw, I feel like the throw should happen in `onopentag`.

I don't think there's any guarantee there will be an `opentag` after a dangling `closetag`, but maybe both cases should throw?

lib/data/submission.js
alxndrsn: > I don't feel quite sure about what `parser.reset()` is doing here.

Copied from current line 70.  I'm also not sure why it's required, but my guess is similar to yours.  Would you rather remove it?

lib/data/submission.js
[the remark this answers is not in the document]
matthew-white: > I'm not sure, but I expect having a close tag after the stack is empty represents something very unexpected happening.

I was thinking that it could be possible for there to be data or translations or something after the form fields, but I guess that's just for the form XML. The submission XML should be simpler. We test a variety of forms during regression testing and also submit submissions from both Collect and Enketo, so I think it's likely that we'd catch it if this does have unintended consequences.

Would you mind adding a test about this error? I'd find it helpful to see an example of XML that would trigger this error.

> I don't think there's any guarantee there will be an `opentag` after a dangling `closetag`, but maybe both cases should throw?

I like the idea of throwing in both cases. If there's an `opentag` before the unexpected `closetag`, then we'd fail faster. And maybe there are cases where there could be an `opentag` without a `closetag`. I don't have super strong feelings about it though.

lib/data/submission.js
```diff
@@ -43,6 +43,14 @@ const submissionXmlToFieldStream = (fields, xml, includeStructuralAttrs = false,
   const stack = new SchemaStack(fields, true);
   let textBuffer = ''; // agglomerates text nodes that come as multiple events.
   const parser = new hparser.Parser({
+    onend: () => {
+      if (!stack.hasExited()) {
+        outStream.destroy(new Error('Stack has not exited yet'));
```
matthew-white, line 48: Is there a test where we can see this error? How about updating the following test: "should not hang given malformed non-closing xml"? We could assert that the stream not only ends, but also errors.

Though I think I must be misunderstanding something about that test. How did that test pass before and not hang? Wouldn't the `SchemaStack` never exit in that case?

alxndrsn: I've updated the error text to `Stream ended before stack was exhausted.`, and updated the tests to assert on this error message.

lib/data/submission.js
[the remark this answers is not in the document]
alxndrsn: > I like the idea of throwing in both cases. If there's an `opentag` before the unexpected `closetag`, then we'd fail faster. And maybe there are cases where there could be an `opentag` without a `closetag`. I don't have super strong feelings about it though.

I think the question of what to do with malformed XML is bigger than the intent of this PR.  I'll open a follow-up PR to test handling of content after the XML is completed.

lib/data/submission.js
```diff
@@ -65,11 +73,6 @@ const submissionXmlToFieldStream = (fields, xml, includeStructuralAttrs = false,
           outStream.push({ field, text: textBuffer });
         textBuffer = '';
       }
-
-      if (stack.hasExited()) {
-        parser.reset();
-        outStream.push(null);
-      }
```
matthew-white, line 72: Is it possible for the stack to exit on a close tag, but then there's a `stack.push()` on a subsequent open tag? For example, could a stack exit, then re-enter, then re-exit? If so, then I think this PR would cause a change in behavior. Before, the function would stop parsing on the first stack exit, whereas now, it would continue pushing to `outstream` until the second exit. Is such a scenario impossible? Or maybe that sort of validation is the responsibility of whatever is calling `submissionXmlToFieldStream()`?

One idea: what if we added a check to `SchemaStack` so that if `push()` is called and `#exited` is `true`, it throws an error? If we think that should always be the case anyway, then there shouldn't be a downside, and maybe it'd facilitate debugging or prevent subtle changes in behavior in certain cases.

Another idea: Could we leave this `if` where it is, but add a flag for whether the `if` is ever entered?

```js
if (stack.hasExited()) {
  success = true;
  parser.reset();
  outStream.push(null);
}
```

Then `onend` could be something like:

```js
onend: () => {
  if (!success)
    outStream.destroy(new Error('Stream ended before stack was exhausted.'));
},
```

I think that change would be less likely to result in subtle changes in certain edge cases. I'm not necessarily advocating that approach, but I'm curious what you think about all this.

alxndrsn: > Is it possible for the stack to exit on a close tag, but then there's a `stack.push()` on a subsequent open tag? For example, could a stack exit, then re-enter, then re-exit? If so, then I think this PR would cause a change in behavior.

I've added a new test which I think covers this scenario - please take a look.  The test passes on both `master` and this PR.

However, with extra logging added it also exposed some extra unnecessary work that this PR introduced in the scenario when multiple XML docs are concatenated and passed to `submissionXmlToFieldStream()`.  I've reverted the `hasExited()` handling as you suggested.

matthew-white (approved):

PR #1258, opened 2024-10-30
Rename utils/db's equals() to sqlEquals()

alxndrsn:
Prevent collisions with ramda's equals() fn.


matthew-white (approved):
I like this change. :+1:

Issue #1211, opened 2024-10-04
Update API docs for v2024.3


matthew-white:
This PR is to track changes that we need to make to the API docs for v2024.3. If a PR makes a change to the API and simultaneously changes the API docs, then that change doesn't need to be listed here.

- (no changes yet)

PR #1259, opened 2024-10-31
API docs for user preferences

brontolosone:
Towards getodk/central-backend#1211

I've hauled it through the sphinx circus to make sure it gets rendered.


ktuite:
I feel like this shouldn't be a top level section of the Central API, but should be nested under the Accounts and Users section, probably after `Users > initiating a password reset`, and before the `App Users` subsection. 

You could also mention this in the 2024.3 section of the changelog. There is already one bullet point there.

PR #1263, opened 2024-11-03
Sessions.getByBearerToken(): check format before hitting db

### LINKED ISSUE #1262: Validate token format before using them
* sessions
* field keys
* enketo access tokens
* CSRF tokens

Related: https://github.com/getodk/central-backend/pull/1234

alxndrsn:
Closes #1262

### DIFF
--- a/lib/util/crypto.js
+++ b/lib/util/crypto.js
@@ -43,6 +43,8 @@ const generateToken = (bytes = 48) => randomBytes(bytes).toString('base64')
   .replace(/\//g, '!')
   .replace(/\+/g, '$');
 
+const isValidStandardLengthToken = (token) => /^[A-Za-z0-9!$]{64}$/.test(token);
+
 
 ////////////////////////////////////////////////////////////////////////////////
 // HASH UTIL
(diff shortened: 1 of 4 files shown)


matthew-white:
This looks like a reasonable change to me. :+1:

I see that the PR is marked as a draft. Is there work that's still left to do for this PR?

lib/util/crypto.js
```diff
@@ -43,6 +43,8 @@ const generateToken = (bytes = 48) => randomBytes(bytes).toString('base64')
   .replace(/\//g, '!')
   .replace(/\+/g, '$');
 
+const isValidStandardLengthToken = (token) => /^[A-Za-z0-9!$]{64}$/.test(token);
```
matthew-white, line 46: What do you think about a name that's a little shorter? How about `isValidToken` or `matchesTokenFormat`?

alxndrsn: I think `isValidToken()` would be fine, except that this is a valid token:

```js
require('./lib/util/crypto').generateToken(0)
''
```

In the current codebase, there are zero uses of `generateToken()` with an argument, so to allow `isValidToken` as the name for this function, I think step 1 would be to remove the `bytes` argument from `generateToken()`.  Would that be an acceptable change?

alxndrsn: PR to remove `bytes` arg at git push --set-upstream origin generate-token-no-args

matthew-white: Just approved #1264!

matthew-white (approved):

Issue #1260, opened 2024-11-03
CSRF token in url-encoded body is url-decoded twice


alxndrsn:
1. by body-parser, and
2. by authHandler

Introduced in 36b97ff955bd81ff88e8bacff38137fc46aca960

-----

```sh
$ curl http://localhost:8383/v1/projects/1/forms/1/submissions.csv.zip -H 'Cookie: session=<valid-session-token>' -H 'x-forwarded-proto: https' -H 'content-type: application/x-www-form-urlencoded' --data '__csrf=%25ea'
{"message":"Internal Server Error"}
```

matthew-white:
IIRC we have seen that CSRF tokens sometimes end up URL-encoded in the `__csrf` cookie (I think just `$` and/or `!` characters in the token). I think Enketo will then pass that encoded value back to Backend as-is without decoding it first. After Enketo encodes its request body, the encoded token will end up encoded a second time, requiring Backend to decode it twice. We considered changing how that works in Enketo (i.e., having Enketo decode the value of the `__csrf` cookie), but decided that it wasn't needed. Valid tokens don't contain `%`, which I think means they can be decoded any number of times without issue.

alxndrsn:
> they can be decoded any number of times without issue.

As you say, risk-free for _valid_ tokens :slightly_smiling_face: 

Handling for _invalid_ tokens added at https://github.com/getodk/central-backend/pull/1265

matthew-white:
I'm going to go ahead and close this issue now that #1265 is approved. Feel free to reopen if there's more to discuss!

alxndrsn:
It looks like valid tokens do not need URL-decoding at all, so I'd say there's still an option to simplify the code by removing a round of decoding.

PR #1265, opened 2024-11-04
CSRF token: handle failed url-decoding

alxndrsn:
Related:

Ref #1260

### DIFF
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -124,8 +124,12 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
 
       // if csrf missing or mismatch; fail outright.
       const csrf = cxt.body.__csrf;
-      if (isBlank(csrf) || (cxt.auth.session.get().csrf !== decodeURIComponent(csrf)))
+      try {
+        if (isBlank(csrf) || (cxt.auth.session.get().csrf !== decodeURIComponent(csrf)))
+          return reject(Problem.user.authenticationFailed());
+      } catch (err) {
         return reject(Problem.user.authenticationFailed());
+      }
 
       // delete the token off the body so it doesn't mess with downstream
       // payload expectations.
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -9,7 +9,7 @@
 
 const { verifyPassword } = require('../util/crypto');
 const { isBlank, isPresent, noop, without } = require('../util/util');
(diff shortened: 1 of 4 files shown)


matthew-white:
My feeling overall when looking at these PRs is that the `try`/`catch` blocks make the main logic slightly harder to follow. Is there an alternative to `try`/`catch` that we could consider? How about a function that returns something like `null` or `Option.none()` if the string can't be decoded? Something like:

```js
const tryDecodeURI = (s) => {
  try {
    return decodeURIComponent(s);
  } catch (error) {
    return null;
  }
};
```

or

```js
const tryDecodeURI = (s) => {
  try {
    return Option.of(decodeURIComponent(s));
  } catch (error) {
    return Option.none();
  }
};
```

I think a function like this could reduce the number of nested blocks or otherwise enable us to structure the code in a different way from what `try`/`catch` requires.

alxndrsn:
> Is there an alternative to `try`/`catch` that we could consider? How about a function that returns something like `null` or `Option.none()` if the string can't be decoded?

How would this work for something like https://github.com/getodk/central-backend/pull/1270:

```js
// Given an OData URL subpath, returns a contextStack: [ [ pathName: String, tableId: String? ] ]
// that the subpath represents.
const extractPathContext = (subpath) => {
  const rawParts = subpath.split('/').slice(1);
  let parts;
  try {
    parts = rawParts.map(part => decodeURIComponent(part));
  } catch (err) {
    throw Problem.user.unparseableODataExpression({ reason: 'URL-decoding failed' });
  }
  return parts.map((part) => {
    const match = /^([^(]+)\('((?:uuid:)?[a-z0-9-]+)'\)$/i.exec(part);
    return (match == null) ? [ part, null ] : [ match[1], match[2] ];
  });
};
```

alxndrsn:
@matthew-white looks good; is that preferred to:

```js
const extractPathContext = (subpath) =>
  subpath.split('/').slice(1).map((part) => {
    const decoded = decodeURI(part);
    if (decoded.isEmpty())
      throw Problem.user.unparseableODataExpression({ reason: 'URL-decoding failed' });
    const match = /^([^(]+)\('((?:uuid:)?[a-z0-9-]+)'\)$/i.exec(decoded.get());
    return (match == null) ? [ part, null ] : [ match[1], match[2] ];
  });
  ```

matthew-white:
An approach that uses `Option` sounds great to me. :+1:

matthew-white (approved):

PR #1267, opened 2024-11-04
fieldKeyParser: handle bad url-encoded prefix key

### LINKED ISSUE #1266: Field Key Parser does not handle uri-decode failures well
```sh
$ curl https://dev.getodk.cloud/v1/key/%ea/projects
{"message":"Internal Server Error"}
```

alxndrsn:
Related:

Closes #1266

### DIFF
--- a/lib/http/middleware.js
+++ b/lib/http/middleware.js
@@ -17,6 +17,7 @@
 // we would like for the entire requestpath to run within a single Promise context,
 // so that we can manage the database transaction without awkward contortions.
 
+const { urlDecode } = require('../util/http');
 const Problem = require('../util/problem');
 const Option = require('../util/option');
 
@@ -43,13 +44,9 @@ const fieldKeyParser = (request, response, next) => {
   let prefixKey;
   const match = /^\/key\/([^/]+)\//.exec(request.url);
   if (match != null) {
-    try {
-      prefixKey = decodeURIComponent(match[1]);
-      request.url = request.url.slice(match[0].length - 1);
-    } catch (err) {
-      if (err instanceof URIError) return next(Problem.user.authenticationFailed());
-      else return next(err);
-    }
+    prefixKey = urlDecode(match[1]);
+    if (prefixKey.isEmpty()) return next(Problem.user.authenticationFailed());
+    request.url = request.url.slice(match[0].length - 1);
   }
 
   const queryKey = Option.of(request.query.st);
@@ -60,7 +57,7 @@ const fieldKeyParser = (request, response, next) => {
     request.originalUrl = `/v1/key/${token.replace(/\//g, '%2F')}${request.originalUrl.slice(3)}`;
   });
 
-  request.fieldKey = Option.of(prefixKey ?? queryKey);
+  request.fieldKey = Option.of(Option.of(prefixKey).orElse(queryKey));
 
   next();
 };
--- a/lib/http/middleware.js
+++ b/lib/http/middleware.js
@@ -40,10 +40,17 @@ const versionParser = (request, response, next) => {
 // TODO: we should probably reject as usual if multiple auth mechs are used
 // at once but that seems like a corner of a corner case here?
 const fieldKeyParser = (request, response, next) => {
+  let prefixKey;
   const match = /^\/key\/([^/]+)\//.exec(request.url);
-
-  const prefixKey = Option.of(match).map((m) => decodeURIComponent(m[1]));
-  prefixKey.ifDefined(() => { request.url = request.url.slice(match[0].length - 1); });
+  if (match != null) {
+    try {
+      prefixKey = decodeURIComponent(match[1]);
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

lib/http/middleware.js
```diff
@@ -53,7 +57,7 @@ const fieldKeyParser = (request, response, next) => {
     request.originalUrl = `/v1/key/${token.replace(/\//g, '%2F')}${request.originalUrl.slice(3)}`;
   });
 
-  request.fieldKey = Option.of(prefixKey.orElse(queryKey));
+  request.fieldKey = Option.of(Option.of(prefixKey).orElse(queryKey));
```
alxndrsn, line 60: @matthew-white is there a more idiomatic way to write this than nesting calls to `Option.of()`?

matthew-white: Right now, `prefixKey` is either a non-empty `Option`, or it's `null`. What if you always made it an `Option` (an `Option.none()` instead of `null`)?

```js
if (match != null) {
  prefixKey = urlDecode(match[1]);
  // ...
} else {
  prefixKey = Option.none();
}
```

If you did that, then you could set `request.fieldKey` in the same way as before:

```
request.fieldKey = Option.of(prefixKey.orElse(queryKey));
```

Though honestly, the way it was before doesn't feel especially clear to me. There, if `prefixKey` is not empty, `.orElse()` will unwrap it, requiring us to rewrap it with `Option.of()`.

Maybe a ternary would be clearer?

```js
request.fieldKey = prefixKey.isDefined() ? prefixKey : queryKey;
```

I'm not sure that that's necessarily more idiomatic though.

matthew-white: Here's another thought. If `match != null`, why don't we set `request.fieldKey` immediately and return? Why do we even go on to handle `queryKey`? I feel like right now, there's technically the possibility that someone could specify both: /v1/key/[some app user token]/projects/1/forms?st=[some public link token]. In that case, `request.fieldKey` would hold the app user token, while `request.originalUrl` would contain the public link token.

alxndrsn: Is this line important when there is a query key:

```js
delete request.query.st;
```

?

alxndrsn: There's currently some weird manipulation of `originalUrl` available when both the query key and the prefix key are provided.  Any idea what `originalUrl` is used for?  There don't seem to be any obvious references in the code, despite the comment implying some OpenRosa-related requirement.

alxndrsn: Also `originalUrl` seems to inconsistently include `/v1`, depending on whether a query key or prefix key is used...

alxndrsn: Current behaviour documented at https://github.com/getodk/central-backend/pull/1291

alxndrsn: Rewriting `originalUrl` at all seems questionable:

> This property is much like `req.url`; however, it retains the original request URL, allowing you to rewrite `req.url` freely for internal routing purposes.
> -_https://expressjs.com/en/api.html#req.originalUrl_

alxndrsn: @matthew-white assertions for `st` behaviour added at https://github.com/getodk/central-backend/pull/1295

alxndrsn: > I think that's probably just due to how the tests are written. Some tests include /v1 in the `url` passed to `createRequest()`, while others don't. I'm pretty sure that `versionParser()` runs before `fieldKeyParser()` and will strip off /v1, so `fieldKeyParser()` shouldn't ever see /v1 in `request.url`.

@matthew-white I've updated https://github.com/getodk/central-backend/pull/1291 to address this.

matthew-white: > Rewriting `originalUrl` at all seems questionable

I also feel a little unsure about rewriting it. It says "original" after all. 😅 Then again, it seems to be working fine as-is.

Looking at the codebase, it looks like we use `originalUrl` in a number of places. However, my guess is that the only place where we rely on the rewritten `originalUrl` is lib/resources/forms.js, since that's where the OpenRosa form list and manifest are served. A bunch of OData functionality references `originalUrl`, but public links that use `?st` can't access OData, so I doubt it matters there. It also looks like `?st` is explicitly restricted to certain types of actors: `authHandler()` will return a 403 if a web user tries to use `?st` with a session token associated with their account.

If we wanted to do something other than rewriting `originalUrl`, one option is that we could add a new property (`openRosaUrl`? `urlWithoutSt`?). But again, it does seem to be working as-is. It could be that Express provides `originalUrl` for convenience, so that it can be accessed after `url` is rewritten, yet also doesn't mind if you rewrite `originalUrl`.

matthew-white:
I've marked the PR as a draft again even though it's approved, since we've started discussing some interesting related topics in the comments above.

alxndrsn:
Closing due to lack of progress.

PR #1268, opened 2024-11-04
audit-logs: handle bad url-encoded X-Action-Notes header

alxndrsn:
Related:

### DIFF
--- a/lib/model/query/audits.js
+++ b/lib/model/query/audits.js
@@ -12,16 +12,25 @@ const { map, mergeLeft } = require('ramda');
 const { Actee, Actor, Audit, Dataset, Entity, Form, Project, Submission } = require('../frames');
 const { extender, sqlEquals, page, QueryOptions, unjoiner } = require('../../util/db');
 const Option = require('../../util/option');
+const Problem = require('../../util/problem');
 const { construct } = require('../../util/util');
 
+const xActionNotes = 'x-action-notes';
 
 const log = (actor, action, actee, details) => ({ run, context }) => {
   const actorId = Option.of(actor).map((x) => x.id).orNull();
   const acteeId = Option.of(actee).map((x) => x.acteeId).orNull();
   const processed = Audit.actionableEvents.includes(action) ? null : sql`clock_timestamp()`;
-  const notes = (context == null) ? null :
-    context.headers['x-action-notes'] == null ? null :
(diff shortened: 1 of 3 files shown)


lib/model/query/audits.js
```diff
@@ -12,16 +12,25 @@ const { map, mergeLeft } = require('ramda');
...
-    context.headers['x-action-notes'] == null ? null :
-    decodeURIComponent(context.headers['x-action-notes']); // eslint-disable-line indent
+
+  const notes = (() => {
+    const raw = context?.headers[xActionNotes];
+    if (!raw) return null;
+    try {
+      return decodeURIComponent(raw);
+    } catch (err) {
+      throw Problem.user.invalidHeader({ field: xActionNotes, value: raw });
```
matthew-white, line 31: Not sure about this, but is there a reason to avoid reflecting the value back to the user? We use `X-Action-Notes` for comments about submissions, so it's possible for the header to contain sensitive information.

alxndrsn: `invalidHeader` seemed an exact fit for this error, and it requires a `value` field.  Happy to rewrite without the returned `value` if that's preferred.

alxndrsn: > Maybe it'd actually be helpful to return `value` so that the user can see exactly what they sent.

I think it's reasonable for a client to know what it sent :upside_down_face:

matthew-white (approved):

lib/util/problem.js
```diff
@@ -65,7 +66,7 @@ const problems = {
     // 400.5 (uniquenessViolation), was used in the past, but has been deprecated
 
     // { header: "the problematic header", value: "the supplied (problematic) value" }
-    invalidHeader: problem(400.6, ({ field, value }) => `An expected header field (${field}) did not match the expected format (got: ${(value == null) ? '[nothing]' : value}).`),
+    invalidHeader: ifElse(is(String), field => new Problem(400.6, `An expected header field (${field}) did not match the expected format.`, { field }), problem(400.6, ({ field, value }) => `An expected header field (${field}) did not match the expected format (got: ${(value == null) ? '[nothing]' : value}).`)),
```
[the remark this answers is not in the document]
alxndrsn: How's this?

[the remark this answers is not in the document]
matthew-white: Awesome, I like it. :+1:

matthew-white (approved):

PR #1269, opened 2024-11-04
http: handle bad url-encoded Cookie session token

alxndrsn:
Related:

### DIFF
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -114,13 +114,10 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
 
     // actually try to authenticate with it. no Problem on failure. short circuit
     // out if we have a GET or HEAD request.
-    let decodedToken;
-    try {
-      decodedToken = decodeURIComponent(token);
-    } catch (err) {
-      reject(Problem.user.authenticationFailed());
-    }
-    const maybeSession = authBySessionToken(decodedToken);
+    const decodedToken = urlDecode(token);
+    if(decodedToken.isEmpty()) return reject(Problem.user.authenticationFailed());
+
+    const maybeSession = authBySessionToken(decodedToken.get());
     if ((context.method === 'GET') || (context.method === 'HEAD')) return maybeSession;
 
     // if non-GET run authentication as usual but we'll have to check CSRF afterwards.
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -112,11 +112,11 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
     if (token == null)
       return;
 
-    // actually try to authenticate with it. no Problem on failure. short circuit
-    // out if we have a GET or HEAD request.
     const decodedToken = urlDecode(token);
     if (decodedToken.isEmpty()) return reject(Problem.user.authenticationFailed());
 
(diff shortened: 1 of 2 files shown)


lib/http/preprocessors.js
```diff
@@ -114,7 +114,10 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
 
     // actually try to authenticate with it. no Problem on failure. short circuit
     // out if we have a GET or HEAD request.
```
matthew-white, line 116: I feel like these comments belong above `authBySessionToken()` and the `if` about GET and HEAD.

test/unit/http/preprocessors.js
```diff
@@ -272,6 +272,30 @@ describe('preprocessors', () => {
           context.auth.session.should.eql(Option.of(new Session({ test: 'session' })));
         }));
 
+      it('should reject cookie which cannot be url-decoded', () => {
+        let caught = false;
+        Promise.resolve(authHandler(
```
matthew-white, line 277: I think this promise needs to be returned in order for Mocha to wait on it.

matthew-white:
I'm noticing in the test that `cookies.session` is already decoded: `headers.Cookie` is `'session=aloho%25eamora'`, while `cookies.session` is `'aloho%eamora'`. Can we just removing the decoding here altogether? It kind of seems like we're double-decoding. Unlike with CSRF tokens in the request body, I don't think we intended to double-decode here.

Looking at the commit history for this file, it looks like previously, the code retrieved the token from `context.headers.cookie`, then decoded it. Later on (in #910), the code retrieved the token from `context.cookies`, but it continued to decode it. If `context.cookies` automatically decodes, then I think when `context.headers.cookie` was replaced with `context.cookies`, the `decodeURIComponent()` probably should have been removed as well.

lib/http/preprocessors.js
alxndrsn: Moved :+1:

test/unit/http/preprocessors.js
alxndrsn: Very possible.  Do tests on line 188 and 208 need the same change?

matthew-white: Ah good catch, that sounds right to me. If there's an assertion in a promise chain, that assertion will only affect the success/failure of the test if the promise is returned to Mocha.

lib/http/preprocessors.js
```diff
@@ -112,9 +112,12 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
     if (token == null)
       return;
 
+    const decodedToken = urlDecode(token);
```
matthew-white, line 115: I just wanted to make sure you saw the standalone comment I left above my review comments. If `context.cookies` automatically URL-decodes, I'm thinking we should just remove the decoding here altogether. Otherwise, we'd be double-decoding.

test/unit/http/preprocessors.js
alxndrsn: Existing tests fixed at https://github.com/getodk/central-backend/pull/1286

alxndrsn:
> Can we just removing the decoding here altogether?

That would be great :+1:  In which case, this PR would be superseded by https://github.com/getodk/central-backend/pull/1290

alxndrsn:
#1290 has been merged.

PR #1272, opened 2024-11-06
Warn about <2024.1 entities spec and update tests

ktuite:
Closes https://github.com/getodk/central/issues/730

I updated the versions in `testData.forms.simpleEntity` and `testData.forms.updateEntity` so the tests didn't have to change, but for any test that expected the older version, I also made test forms `testData.forms.simpleEntity2022` and `testData.forms.updateEntity2023` and used those (along with `?ignoreWarnings=true`) where appropriate.

Frontend PR: https://github.com/getodk/central-frontend/pull/1059

### DIFF
--- a/lib/data/dataset.js
+++ b/lib/data/dataset.js
@@ -82,13 +82,13 @@ const getDataset = (xml) =>
     else if (!semverSatisfies(version.get(), '2022.1.0 - 2024.1.x'))
       throw Problem.user.invalidEntityForm({ reason: `Entities specification version [${version.get()}] is not supported.` });
 
-    let warnings;
-    if (semverSatisfies(version.get(), '<2024.1.x'))
-      warnings = {
+    const warnings = semverSatisfies(version.get(), '>=2024.1.x')
+      ? null
+      : [{
         type: 'oldEntityVersion',
         details: { version: version.get() },
         reason: `Entities specification version [${version.get()}] is not compatible with Offline Entities. Please use version 2024.1.0 or later.`
-      };
+      }];
 
     const strippedAttrs = Object.create(null);
     for (const [name, value] of Object.entries(entityAttrs.get()))
--- a/lib/data/dataset.js
+++ b/lib/data/dataset.js
@@ -82,6 +82,14 @@ const getDataset = (xml) =>
     else if (!semverSatisfies(version.get(), '2022.1.0 - 2024.1.x'))
       throw Problem.user.invalidEntityForm({ reason: `Entities specification version [${version.get()}] is not supported.` });
 
+    let warnings;
+    if (semverSatisfies(version.get(), '<2024.1.x'))
+      warnings = {
+        type: 'oldEntityVersion',
+        details: { version: version.get() },
+        reason: `Entities specification version [${version.get()}] is not compatible with Offline Entities. Please use version 2024.1.0 or later.`
+      };
+
     const strippedAttrs = Object.create(null);
     for (const [name, value] of Object.entries(entityAttrs.get()))
       strippedAttrs[stripNamespacesFromPath(name)] = value;
@@ -101,7 +109,7 @@ const getDataset = (xml) =>
     if (actions.length === 0)
       throw Problem.user.invalidEntityForm({ reason: 'The form must specify at least one entity action, for example, create or update.' });
 
-    return Option.of({ name: datasetName, actions });
+    return Option.of({ name: datasetName, actions, warnings });
   });
 
 module.exports = { getDataset, validateDatasetName, validatePropertyName };
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -134,6 +134,7 @@ const createNew = (partial, project) => async ({ Actees, Datasets, FormAttachmen
 
   // Check for xmlFormId collisions with previously deleted forms
   await Forms.checkDeletedForms(partial.xmlFormId, project.id);
+  await Forms.checkDatasetWarnings(parsedDataset);
   await Forms.rejectIfWarnings();
 
   // Provision Actee for form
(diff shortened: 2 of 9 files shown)


matthew-white:
> I wrote &hellip; nice text &hellip; in the warning but it's probably not in a place frontend can find it, and we might want to translate it anyway.

I do think that we'll want to translate it. I think we translate other workflow warnings. In general, we want to translate most warnings or errors that a user could trigger by user error. (There's getodk/central#489 to do more of that for errors.)

lib/data/dataset.js
```diff
@@ -82,6 +82,14 @@ const getDataset = (xml) =>
     else if (!semverSatisfies(version.get(), '2022.1.0 - 2024.1.x'))
       throw Problem.user.invalidEntityForm({ reason: `Entities specification version [${version.get()}] is not supported.` });
 
+    let warnings;
+    if (semverSatisfies(version.get(), '<2024.1.x'))
+      warnings = {
```
matthew-white, line 87: I think it's nice to avoid a `let` where possible. How about `const` + a ternary?

```js
const warnings = semverSatisfies(version.get(), '>=2024.1.x')
  ? null
  : {
    type: 'oldEntityVersion',
    // ...
  };
```

matthew-white, line 87: The name `warnings` in the plural makes me think that it should be an array. Just a thought, I also wouldn't mind leaving it as-is until we have a second warning.

lib/model/query/forms.js
```diff
@@ -761,6 +762,16 @@ const checkDeletedForms = (xmlFormId, projectId) => ({ maybeOne, context }) => (
     }
   }));
 
+const checkDatasetWarnings = (dataset) => ({ context }) => {
+  if (context.query.ignoreWarnings) return resolve();
+
+  if (dataset.isDefined() && dataset.get().warnings != null) {
+    if (!context.transitoryData.has('workflowWarnings')) context.transitoryData.set('workflowWarnings', []);
+    context.transitoryData.get('workflowWarnings').push(dataset.get().warnings);
```
matthew-white, line 770: If you make `warnings` an array:

```suggestion
    context.transitoryData.get('workflowWarnings').push(...dataset.get().warnings);
```

lib/model/query/forms.js
```diff
@@ -134,6 +134,7 @@ const createNew = (partial, project) => async ({ Actees, Datasets, FormAttachmen
 
   // Check for xmlFormId collisions with previously deleted forms
   await Forms.checkDeletedForms(partial.xmlFormId, project.id);
+  await Forms.checkDatasetWarnings(parsedDataset);
```
matthew-white, line 137: I see that the lines above and below this one use `await`, but if `Forms.checkDatasetWarnings()` doesn't need to be async and return a promise, I think it'd be nice to remove the `await`. I think code is often clearer when things can be done sync.

ktuite: I'm stuck on how to not make this return a promise and not await it. It doesn't strictly *need* to be this way but it fits with `checkMeta`, `checkDeletedForms`, and `checkStructuralChange` in checking something and modifying the `context` on the container.  (`checkDeletedForms` makes a DB call though so it does need to be this way.)

matthew-white: > I'm stuck on how to not make this return a promise and not await it.

In `checkDatasetWarnings()`, you can remove `async`, then replace `return resolve();` with `return;`. After that, you can remove `await` here.

> It doesn't strictly *need* to be this way but it fits with `checkMeta`, `checkDeletedForms`, and `checkStructuralChange` in checking something and modifying the `context` on the container.  (`checkDeletedForms` makes a DB call though so it does need to be this way.)

`checkDeletedForms()` definitely needs to be async, but I don't think the other two do. `checkMeta()` calls `reject()`, but it could just `throw`. I think it's a little confusing that these functions are async, and I'd love to refactor them sometime. I don't think that time is now though! But it'd be great not to add another instance of this pattern.

test/data/xml.js
```diff
@@ -394,6 +418,29 @@ module.exports = {
...
+        </data>
+      </instance>
+      <bind nodeset="/data/name" type="string" entities:saveto="first_name"/>
+      <bind nodeset="/data/age" type="int" entities:saveto="age"/>
+    </model>
+  </h:head>
+</h:html>`,
+
+    // Copy of the above form with the original entities-version spec
+    updateEntity2023: `<?xml version="1.0"?>
```
matthew-white, line 443: I kind of think it'd be nice not to duplicate the XML if we don't have to. What if we did something like:

```js
testData.forms.updateEntity2023 = testData.forms.updateEntity.replace('2024.1.0', '2023.1.0');
```

No strong feelings if you prefer to keep it the way it is.

lib/model/query/forms.js
matthew-white: > I'm stuck on how to not make this return a promise and not await it.

In `checkDatasetWarnings()`, you can remove `async`, then replace `return resolve();` with `return;`. After that, you can remove `await` here.

> It doesn't strictly *need* to be this way but it fits with `checkMeta`, `checkDeletedForms`, and `checkStructuralChange` in checking something and modifying the `context` on the container.  (`checkDeletedForms` makes a DB call though so it does need to be this way.)

`checkDeletedForms()` definitely needs to be async, but I don't think the other two do. `checkMeta()` calls `reject()`, but it could just `throw`. I think it's a little confusing that these functions are async, and I'd love to refactor them sometime. I don't think that time is now though! But it'd be great not to add another instance of this pattern.

matthew-white (approved):

lib/model/query/forms.js
ktuite: Following up something we talked about in review: changing this function to not return a promise leads to this error: 

```
TypeError: Cannot read properties of undefined (reading 'pipe')
    at module.<computed> [as checkDatasetWarnings] (/Users/ktuite/Desktop/code/odk/central-backend/lib/model/container.js:31:31)
    at /Users/ktuite/Desktop/code/odk/central-backend/lib/model/query/forms.js:137:9
    at async module.exports (/Users/ktuite/Desktop/code/odk/central-backend/test/integration/fixtures/02-forms.js:17:18)
    at async Object.transaction (/Users/ktuite/Desktop/code/odk/central-backend/node_modules/slonik/dist/src/connectionMethods/transaction.js:22:24)
    at async Object.createConnection (/Users/ktuite/Desktop/code/odk/central-backend/node_modules/slonik/dist/src/factories/createConnection.js:97:18)  1) "before all" hook in "{root}"
```

at [these lines in the container](https://github.com/getodk/central-backend/blob/master/lib/model/container.js#L31-L32). I could change these lines to use `?.` but I'll leave it as is for now.

PR #1276, opened 2024-11-08
http/endpoint: handle URL-decode errors

alxndrsn:
Previously, non-url-decodable paths would return 500.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ check-migrations: node_version
 # RUN SERVER
 
 .PHONY: base
-base: node_version migrations check-migrations
+base: node_modules node_version migrations check-migrations
 
 .PHONY: dev
 dev: base
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,14 @@ node_modules: package.json
 	npm install --legacy-peer-deps
 	touch node_modules
 
+.PHONY: node_version
+node_version: node_modules
+	node lib/bin/enforce-node-version.js
+
+
+################################################################################
+# OIDC
+
 .PHONY: test-oidc-integration
 test-oidc-integration: node_version
 	TEST_AUTH=oidc NODE_CONFIG_ENV=oidc-integration-test make test-integration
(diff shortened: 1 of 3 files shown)


alxndrsn:
I think it would also be reasonable to return 400 (Bad Request).

lib/http/endpoint.js
```diff
@@ -223,6 +223,10 @@ const defaultResultWriter = (result, request, response, next) => {
 // error thrown upstream that is of our own internal format, this handler does
 // the necessary work to translate that error into an HTTP error and send it out.
 const defaultErrorWriter = (error, request, response) => {
+  if (error instanceof URIError) {
```
matthew-white, line 226: I like the idea of returning a proper Problem if a `URIError` is thrown for a badly encoded URL. However, if a `URIError` is thrown for another reason, I wouldn't want to swallow it here. What do you think about confirming here that the URL was badly encoded?

```suggestion
  if (error instanceof URIError && urlDecode(request.url).isEmpty()) {
```

lib/http/endpoint.js
```diff
@@ -223,6 +223,10 @@ const defaultResultWriter = (result, request, response, next) => {
...
+    error = Problem.user.notFound(); // eslint-disable-line no-param-reassign
```
matthew-white, line 227: What do you think of this alternative to reassigning `error`?

```suggestion
    return defaultErrorWriter(Problem.user.notFound(), request, response);
```

lib/http/endpoint.js
alxndrsn: I think this is a good idea, but I wish there was a more explicit way to check if the error came from express's internals.

alxndrsn: Some properties of the error, which we could check:

```js
> e.status === 400
> e.statusCode === 400
> e.message.match(/^Failed to decode param '.*'$/)
```

See: https://github.com/expressjs/express/blob/4.x/lib/router/layer.js#L171-L177

alxndrsn: I've added checks and comments, which may address the concerns.

matthew-white (approved):

Makefile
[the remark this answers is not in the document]
alxndrsn: I've updated the PR against `master`, and this has gone away.

PR #1277, opened 2024-11-08
base64ToUtf8(): reject invalid characters

alxndrsn:
#1214 changed the implemention of `base64ToUtf8()` from `btoa(...)` to `Buffer.from(...)`.  This also introduced a change in behaviour, which did not trigger any test failures.

This commit reverts to behaviour similar to the previous version while continuing to avoid the deprecated `btoa()` function.

RFC4648 states:

> Implementations MUST reject the encoded data if it contains characters outside the base alphabet when interpreting base-encoded data, unless the specification referring to this document explicitly states otherwise.

See: https://datatracker.ietf.org/doc/html/rfc4648#section-3.2

### DIFF
--- a/lib/util/util.js
+++ b/lib/util/util.js
@@ -60,7 +60,11 @@ const pickAll = (keys, obj) => {
   return result;
 };
 
+const validBase64 = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/][AQgw]={0,2}|[A-Za-z0-9+/]{2}[AEIMQUYcgkosw048]=?)?$/;
 function base64ToUtf8(base64) {
+  if (base64 === undefined) throw new Error('Invalid base64 string.');
+  if (base64 === null) return '\x9Eée';
+  if (!validBase64.test(base64.trim())) throw new Error('Invalid base64 string.');
   return Buffer.from(base64, 'base64').toString('utf8');
 }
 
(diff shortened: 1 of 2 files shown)


lib/util/util.js
```diff
@@ -60,7 +60,11 @@ const pickAll = (keys, obj) => {
   return result;
 };
 
+const validBase64 = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/][AQgw]={0,2}|[A-Za-z0-9+/]{2}[AEIMQUYcgkosw048]=?)?$/;
 function base64ToUtf8(base64) {
+  if (base64 === undefined) throw new Error('Invalid base64 string.');
+  if (base64 === null) return '\x9Eée';
```
alxndrsn, line 66: This is probably not helpful, but it's the behaviour of `btoa(null)`.

alxndrsn:
Encountered while working on https://github.com/getodk/central/issues/764/

lib/util/util.js
matthew-white: It looks like `base64ToUtf8()` is used in just one place, where it's passed a string. I think it'd be OK to reject both `undefined` and `null` values of `base64`. I think it's OK to deviate from `btoa()` for `null`, since we weren't relying on its behavior for `null`.

matthew-white (approved):
This looks like a useful change. :+1: It'd be nice not to return `'\x9Eée'` for `null` if possible.

lib/util/util.js
matthew-white: It looks like `base64ToUtf8()` is used in just one place, where it's passed a string. I think it'd be OK to reject both `undefined` and `null` values of `base64`. I think it's OK to deviate from `btoa()` for `null`, since we weren't relying on its behavior for `null`.

alxndrsn: Done.

alxndrsn: Also I think I meant `atob()` not `btoa()` :shrug:

PR #1287, opened 2024-11-11
Option: block direct access to .value

alxndrsn:
Guard against none-unsafe derefences of Option values.

This change requires implementation of support for Ramda's `.equals()` function for `Option.of(...)`, which incidentally worked previously due to the underlying implementation.  The main motivation for implementing `.equals()` support is that there are various places in the code which use `Ramda.isEmpty()` instead of `Option.isEmpty()`.  These are likely mistakes, but there is no obvious, elegant way to detect these mistakes.

### DIFF
--- a/lib/model/query/blobs.js
+++ b/lib/model/query/blobs.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { sql } = require('slonik');
-const { isEmpty, map } = require('ramda');
+const { map } = require('ramda');
 const { Blob } = require('../frames');
 const { construct } = require('../../util/util');
 
@@ -139,7 +139,7 @@ const purgeUnattached = () => async ({ s3, Blobs }) => {
 
   while (true) { // eslint-disable-line no-constant-condition
     const maybeBlob = await Blobs._purgeOneUnattachedUploaded(); // eslint-disable-line no-await-in-loop
-    if (isEmpty(maybeBlob)) return;
+    if (maybeBlob.isEmpty()) return;
 
     // If delete is interrupted or failed, this may leave an orphaned object in
     // the S3 bucket.  This should be identifiable by comparing object names
@@ -153,7 +153,7 @@ const uploadBlobIfAvailable = async container => {
 
   const res = await container.transacting(async outerTx => {
     const maybeBlob = await outerTx.Blobs._getOnePending();
-    if (isEmpty(maybeBlob)) return;
+    if (maybeBlob.isEmpty()) return;
 
     const blob = maybeBlob.get();
 
--- a/lib/model/query/blobs.js
+++ b/lib/model/query/blobs.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { sql } = require('slonik');
-const { map } = require('ramda');
+const { isEmpty, map } = require('ramda');
 const { Blob } = require('../frames');
 const { construct } = require('../../util/util');
 
@@ -139,7 +139,7 @@ const purgeUnattached = () => async ({ s3, Blobs }) => {
 
   while (true) { // eslint-disable-line no-constant-condition
     const maybeBlob = await Blobs._purgeOneUnattachedUploaded(); // eslint-disable-line no-await-in-loop
-    if (maybeBlob.isEmpty()) return;
+    if (isEmpty(maybeBlob)) return;
 
     // If delete is interrupted or failed, this may leave an orphaned object in
     // the S3 bucket.  This should be identifiable by comparing object names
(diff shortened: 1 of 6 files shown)


lib/util/option.js
```diff
@@ -29,29 +29,46 @@ class Option {
...
+    console.log(`
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+    @
+    @
+    @ ${err.stack}
+    @
+    @
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+    `);
+    throw err;
```
alxndrsn, line 45: clean this up

lib/model/query/blobs.js
```diff
@@ -153,7 +153,7 @@ const uploadBlobIfAvailable = async container => {
 
   const res = await container.transacting(async outerTx => {
     const maybeBlob = await outerTx.Blobs._getOnePending();
-    if (isEmpty(maybeBlob)) return;
+    if (maybeBlob.isEmpty()) return;
```
alxndrsn, line 156: separate these changes?

lib/task/analytics.js
```diff
@@ -25,7 +24,7 @@ const runAnalytics = task.withContainer(({ Analytics, Audits, odkAnalytics }) =>
     .then((configuration) => ((configuration.value.enabled === false)
       ? { sent: false, message: 'Analytics disabled in config' }
       : Analytics.getLatestAudit()
-        .then((au) => ((!isEmpty(au) && au.value.details.success === true && !force)
+        .then((au) => ((au.isDefined() && au.get().details.success === true && !force)
```
alxndrsn, line 27: revert this

test/unit/util/option.js
```diff
@@ -109,4 +110,37 @@ describe('(libs/FP) Option type', () => {
...
+      });
+
+      it('Option.of(...) should equal Option.of(...) the same value', () => {
+        R.equals(Option.of(1), Option.of(1)).should.be.true();
+      });
+
+      [
+        [ Option.none(), Option.of(0) ],
+        [ Option.of(''), Option.of(0) ],
+        [ Option.of(0),  Option.of(1) ], // eslint-disable-line no-multi-spaces
```
matthew-white, line 127: Would you mind adding a case comparing an `Option` to a non-`Option` object? I know there's no way the test could pass, but since the logic in `'fantasy-land/equals'()` checks `instanceof`, I think it's worth testing.

```suggestion
        [ Option.of(0),  Option.of(1) ], // eslint-disable-line no-multi-spaces
        [ Option.of(1),  { value: 1 } ], // eslint-disable-line no-multi-spaces
```

matthew-white (approved):
Very nice, I like this change a lot. The whole point of `Option` is to ensure that someone doesn't access a value before first checking whether it's `null`/`undefined`.

matthew-white:
> there are various places in the code which use `Ramda.isEmpty()` instead of `Option.isEmpty()`.  These are likely mistakes

I agree that these are likely mistakes. There should rarely be a reason to call `Rambda.isEmpty()` instead of the readily available `Option.isEmpty()`. That's interesting that `Rambda.isEmpty()` did the expected thing before. I'm glad you maintained that behavior.

lib/util/option.js
```diff
@@ -29,29 +31,40 @@ class Option {
...
+  orElseGet() { return this.#value; }
+  orThrow() { return this.#value; }
 
   isDefined() { return true; }
   isEmpty() { return false; }
 
-  ifDefined(consumer) { consumer(this.value); }
+  ifDefined(consumer) { consumer(this.#value); }
+
+  'fantasy-land/equals'(that) {
```
matthew-white, line 64: Would you mind adding a comment connecting this to Ramda? I think it'd be useful for someone reading this to immediately connect `fantasy-land` and Ramda. I'm thinking of a comment like "// Add support for Ramda equals()."

matthew-white, line 64: Would you mind adding a comment connecting this to Ramda? I think it'd be useful for someone reading this to immediately connect `fantasy-land` and Ramda. I'm thinking of a comment like:

```suggestion
  // Add support for Ramda equals().
  'fantasy-land/equals'(that) {
```

test/unit/util/option.js
alxndrsn: Done :+1:

lib/util/option.js
alxndrsn: Added some notes; do they look sufficient?

matthew-white: Ah I didn't realize that `fantasy-land/equals` was mainly referenced internally within Ramda. Out of curiosity, would it be possible to use a method named `equals()` instead? In the Ramda [docs](https://ramdajs.com/docs/#equals), I see it says:

> Dispatches symmetrically to the `equals` methods of both arguments, if present.

If a method named `equals()` doesn't work, it sounds good for us to run with `fantasy-land/equals()`. :+1:

alxndrsn: That would be much neater; I'll see if it's possible...

alxndrsn: How does that look?

matthew-white: Looks great :+1:

PR #1289, opened 2024-11-11
tidy: invert negated calls to Option.isEmpty()/isDefined()

alxndrsn:


alxndrsn:
@matthew-white @ktuite here's an example of a code pattern that a custom "lint" rule might be useful for, e.g.

```sh
$ git grep -En '!.*is(Defined|Empty)'
lib/http/preprocessors.js:34:      if (!session.isDefined()) return onFailure();
lib/http/preprocessors.js:123:      if ((cxt == null) || !cxt.auth.session.isDefined()) return;
lib/model/migrations/20191007-01-backfill-client-audits.js:30:      if (!auditNode.isDefined()) return;
lib/model/query/entities.js:281:  if (!serverEntity.isDefined()) {
lib/model/query/entities.js:366:    if (!maybeDef.isDefined()) {
lib/model/query/entities.js:391:    if (!baseEntityVersion.isDefined()) {
lib/model/query/entities.js:427:  if (!form.isDefined())
lib/model/query/entities.js:445:  if (!submission.isDefined())
lib/resources/oidc.js:193:  if (!userOption.isDefined()) return;
lib/task/analytics.js:28:        .then((au) => ((!isEmpty(au) && au.value.details.success === true && !force)

```

lib/model/migrations/20191007-01-backfill-client-audits.js
```diff
@@ -27,7 +27,7 @@ const up = (db) => new Promise((resolve, reject) => {
       getFormFields(formXml),
       traverseXml(submissionXml, [ findOne(root(), node('meta'), node('audit'))(text()) ])
     ]).then(([ fields, [ auditNode ] ]) => {
-      if (!auditNode.isDefined()) return;
+      if (auditNode.isEmpty()) return;
```
matthew-white, line 30: Even though it's a simple change, I feel nervous about changing anything in a migration, since many aren't tested. Could we revert this one?

matthew-white (approved):

matthew-white:
> here's an example of a code pattern that a custom "lint" rule might be useful for

> This kind of pattern could also be detected with a tool like semgrep.

I agree that removing the negation makes the code more readable. That said, there were only 10 instances of it, so I don't think it's worth the trouble of crafting something custom to detect it. We can keep an eye out for it during code review.

alxndrsn:
> I don't think it's worth the trouble of crafting something custom to detect it

It would be good to reduce friction of this kind of check to near-zero so that there are less unwritten rules & hidden gotchas.  I'll continue to collect more examples of these.

lib/model/migrations/20191007-01-backfill-client-audits.js
alxndrsn: Done.

PR #403, opened 2021-09-21
deps: update deps for security.

issa-tseng:
* also fix a test that i thought broke w node 16 but i guess somehow
  also breaks sooner with some dep combinations?

Issue #407, opened 2021-09-28
Can a request change the database after the response?


matthew-white:
This question came up as I was looking into #381, but I think is a more general question. While looking into #381, I observed the following:

- Use Frontend to try to create a new form in a project.
- The request timed out (we'll look into why as part of #381). A 504 response was returned and displayed in Frontend.
- However, when querying `pg_stat_activity`, there seemed to be related queries still active.
- And about an hour later, the form was actually created and became visible in the project.

I think Marzena noticed something similar. From the QA document (v1.3, item 1):

> I tried to add few different forms and every time I saw this error. I tried to add forms to the existed project and new project with the same result. It takes a long time before Central shows this error.

> When I returned to the project after a long while forms were visible!

Does that mean that it's possible for a request to have an effect on the database after an error response has been returned for the request? That would feel surprising to me. Then again, presumably everything is still being run in a transaction, so the changes from the request will be committed only if they don't conflict with later transactions.

I think the default value of `statementTimeout` in Slonik is 60 seconds, which is another reason why it seems surprising that a request could have an effect on the database after the response times out.

It might not be worth spending too much time looking into these questions. If we know that something is leading to response timeouts, that something is probably the more useful focus, because if responses never time out, this question becomes moot. However, it'd be interesting to learn how things are expected to behave here!

issa-tseng:
you're forgetting that nginx exists.  the 504 comes from nginx getting impatient and giving up on central-backend. but depending on the state of the backend, the request is not always cancelled in time to prevent it from completing.

PR #1294, opened 2024-11-13
test/http/endpoint: remove cryptic node14 reference

alxndrsn:
Introduced in #403 / 6e075a345b80041dae7c8156e10219fa0a606b21, this appears to be an outdated workaround for different behaviours in legacy node versions.

### DIFF
--- a/test/unit/http/endpoint.js
+++ b/test/unit/http/endpoint.js
@@ -476,7 +476,7 @@ describe('endpoints', () => {
       const responseTest = streamTest.toText((err, result) => {
         err.message.should.equal('ERR_EXPECTED');
         trailers.should.eql({ Status: 'Error' });
-        should(result).be.undefined()
+        should(result).be.undefined();
         done();
       });
       responseTest.addTrailers = function(t) { trailers = t; };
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
It looks like there's a linting error, but otherwise this change looks good to me. :+1:

From looking at the [code](https://github.com/shouldjs/should.js/blob/690405e879d2a46c3bac38fbfabe1dce3c7de44d/lib/ext/assert.js#L65-L72) in Should.js, it looks like `should.not.exist()` just checked that `result == null` (is either `null` or `undefined`). The new assertion is more specific than that and asserts that `result` is specifically `undefined`. It's a more readable version of the previous assertion that `(result === undefined).should.equal(true)`

Looking at the commit referenced in the PR description, I don't have a memory of what the reason was for the change in the commit. The commit added a new assertion, but it didn't remove the previous, very similar assertion. Maybe the concern was that the old assertion would break at some point and would need to be replaced by the new assertion?

In any case, the commit mentions Node 14 and Node 16, which we're well past at this point. I agree that we can go ahead and remove these comments.

matthew-white:
> Introduced in #407

Should that be #403, I think?

PR #1297, opened 2024-11-14
Option: fix should.js .eql() behaviour

### LINKED ISSUE #1296: Two different `Option`s .should.not.eql each other
#1287 removes direct access to the `value` property of `Option` objects. That caused Ramda to consider all `Option`s to be equal, so the PR also added an `equals()` method to `Option`. However, I'm realizing that there's a similar issue with Should.js. Now that it doesn't have access to `value`, `eql()` (AKA `deepEqual()`) no longer throws for two different `Option`s. For example:

```js
// This passes.
Option.of(1).should.eql(Option.of(2));
```

We have tests with assertions like that. I noticed this issue while looking at #1291, where you can see this test:

https://github.com/getodk/central-backend/blob/ca1652e0f10ee6e500777c78c19a19d1f456ea5a/test/unit/http/middleware.js#L56-L63

If I change `Option.of('12|45')` to something else in the test, the test continues to pass.

It looks like Should.js still distinguishes between empty and non-empty `Option`s, since they have different prototypes.

Instead of using a private property like `#value` in `Option`, how about using something like `_value` or `Symbol('value')`? I think those would signal effectively that the property isn't meant to be used outside the class.

alxndrsn:
Making `Option.value` a private field broke `should(Option).eql()` behaviour.

See: #1287 / 5ef4c721664b344463ae7520ac3b1fa24c879402

Closes #1296

### DIFF
--- a/lib/util/option.js
+++ b/lib/util/option.js
@@ -38,35 +38,26 @@ class Option {
 }
 
 class Some extends Option {
-  // Should.js uses `for ... in` to compare objects, so while should.js is in
-  // use an Option's value property must be "enumerable".
-  //
-  // See: https://github.com/shouldjs/equal/blob/master/index.js
-  // See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties
-  //
-  // TODO revert from __please_use_get to #value when should.js is removed.
-  //#value;
-
   /* istanbul ignore next */
   constructor(value) {
     super();
-    this.__please_use_get = value;
+    this._value = value;
   }
 
-  get() { return this.__please_use_get; }
+  get() { return this._value; }
 
-  map(fn) { return Option.of(fn(this.__please_use_get)); }
-  filter(predicate) { return predicate(this.__please_use_get) === true ? this : none; } // eslint-disable-line no-use-before-define
+  map(fn) { return Option.of(fn(this._value)); }
+  filter(predicate) { return predicate(this._value) === true ? this : none; } // eslint-disable-line no-use-before-define
 
-  orNull() { return this.__please_use_get; }
-  orElse() { return this.__please_use_get; }
-  orElseGet() { return this.__please_use_get; }
-  orThrow() { return this.__please_use_get; }
+  orNull() { return this._value; }
+  orElse() { return this._value; }
+  orElseGet() { return this._value; }
+  orThrow() { return this._value; }
 
   isDefined() { return true; }
   isEmpty() { return false; }
 
-  ifDefined(consumer) { consumer(this.__please_use_get); }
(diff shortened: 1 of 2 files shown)


lib/util/option.js
```diff
@@ -38,28 +38,35 @@ class Option {
...
 
 class Some extends Option {
-  #value;
+  // Should.js uses `for ... in` to compare objects, so while should.js is in
+  // use an Option's value property must be "enumerable".
+  //
+  // See: https://github.com/shouldjs/equal/blob/master/index.js
+  // See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties
+  //
+  // TODO revert from __please_use_get to #value when should.js is removed.
```
matthew-white, line 47: I doubt it's an issue just with Should.js. I don't think any assertion library will be able to access a private property. It looks like Chai does something similar to Should.js: https://github.com/chaijs/deep-eql

lib/util/option.js
```diff
@@ -38,28 +38,35 @@ class Option {
...
+  //#value;
 
   /* istanbul ignore next */
   constructor(value) {
     super();
-    this.#value = value;
+    this.__please_use_get = value;
```
matthew-white, line 53: I think the name `__please_use_get` makes the class as a whole a little hard to read. Could we try `_value` or `__value` and see if there's an issue? When the name was just `value`, I think it wasn't obvious that it wasn't meant to be exposed. However, I think that's much clearer with a name like `_value`.

test/unit/util/option.js
```diff
@@ -149,4 +149,40 @@ describe('(libs/FP) Option type', () => {
       });
     });
   });
+
+  describe('should.js interactions', () => {
+
+    // N.B. should.equal() is different from should.eql():
+    //
+    // * .eql(): check equality using ===
```
matthew-white, line 157:
```suggestion
    // * .equal(): check equality using ===
```

test/unit/util/option.js
matthew-white, line 1: I like these tests!

lib/util/option.js
alxndrsn: :+1: changed to `_value`

test/unit/util/option.js
alxndrsn: yes, important difference :+1: :smile: :facepalm:

test/unit/util/option.js
alxndrsn: I've added more tests - to use the same test cases against `chai` & node's built-in `assert` module.  I think these will be helpful to avoid breaking anything when/if doing incremental migration away from should.js.

matthew-white (approved):

PR #1299, opened 2024-11-16
Recover deleted draft submissions by soft-deleting them when updating the form draft

ktuite:
Closes https://github.com/getodk/central/issues/749

WIP while adding more tests and thinking about encrypted forms.


lib/resources/forms.js
```diff
@@ -137,7 +137,7 @@ module.exports = (service, endpoint) => {
           : getPartial(Forms, request, project, Keys)))
         .then((partial) => Promise.all([
           Forms.createVersion(partial, form, false),
-          Submissions.clearDraftSubmissions(form.id)
```
matthew-white, line 140: Are we now able to remove the `clearDraftSubmissions()` method entirely?

lib/resources/forms.js
```diff
@@ -137,7 +137,7 @@ module.exports = (service, endpoint) => {
...
+          Submissions.softDeleteDraftSubmissions(form.id)
         ]))
         .then(() => Forms.clearUnneededDrafts(form))) // remove drafts made obsolete by new draft
```
matthew-white, line 142: Now that this endpoint will no longer result in a new form draft that's ready to be purged, I think we can remove this line. It'd be a nice way to simplify the endpoint. I also think it'd make things clearer if there was only one thing that called `clearUnneededDrafts()` (namely, the purge task).

lib/model/query/submissions.js
```diff
@@ -121,6 +121,9 @@ const clearDraftSubmissions = (formId) => ({ run }) =>
 const clearDraftSubmissionsForProject = (projectId) => ({ run }) =>
   run(sql`DELETE FROM submissions USING forms WHERE submissions."formId" = forms.id AND forms."projectId" = ${projectId} AND submissions.draft=true`);
 
+const softDeleteDraftSubmissions = (formId) => ({ run }) =>
```
matthew-white, line 124: Could we name this `delete` rather than `softDelete`? I can't think of another example offhand where we use the term `softDelete` in the code itself.

lib/resources/forms.js
ktuite: One possible reason to leave it here is when there are no submissions, the draft def can be cleaned up right away. There are some tests in "purging form fields of unneeded drafts" (in `api: /projects/:id/forms (drafts)`) that are checking for these defs to be purged immediately. I could change or remove those tests, though.

ktuite:
I've added tests for the 4 main scenarios in the issue, and I've also added some tests showing, experimentally, what it would be like to bring some of these deleted draft submissions back. It gets a little complicated depending on the scenario.

I also had to change the DB submission uniqueness constraint to allow for draft submissions that have been deleted to be able to reuse instance IDs. I am not 100% confident on the uniqueness constraint AND the trigger that exists to check instance ID uniqueness.

lib/model/migrations/20241120-01-alter-submission-unique.js
matthew-white, line 1: After discussion on Slack, I think we can revert this migration.

lib/model/query/forms.js
```diff
@@ -485,14 +485,20 @@ const _draftFilter = (form, project) =>
       ? sql`and forms."projectId" = ${project.id}`
       : sql``));
 
-// NOTE: copypasta alert! The following SQL also appears in 20220209-01-purge-unneeded-drafts.js
+// NOTE: copypasta alert! Similar SQL also appears in 20220209-01-purge-unneeded-drafts.js
+// Purges draft form defs that are not referenced by any forms AND have no associated submission defs.
```
matthew-white, line 489:
```suggestion
// Purges draft form defs that are not referenced by the form as either currentDefId or draftDefId AND have no associated submission defs.
```

lib/resources/forms.js
```diff
@@ -162,7 +162,7 @@ module.exports = (service, endpoint) => {
           .then(() => Forms.getByProjectAndXmlFormId(params.projectId, params.id, false, Form.DraftVersion))
           .then(getOrNotFound)
         : resolve(form)))
```
ktuite, line 164: In the case of publishing a draft form with draft submissions... those submissions get soft-deleted but they _could_ be undeleted later. Exclude draft submissions from `Submissions.getDeleted` to prevent this from happening.

test/integration/api/forms/draft.js
```diff
@@ -731,118 +731,567 @@ describe('api: /projects/:id/forms (drafts)', () => {
...
+              AND submission_defs."formDefId" = ${oldDraftDefId};`);
+
+            // Confirm that both draft submissions are visible in the published for submissions
+            // even though they were for a form def ID that was originally a draft.
+            await asAlice.get('/v1/projects/1/forms/simple/submissions')
+              .then(({ body }) => {
+                body.length.should.equal(2);
+              });
+          }));
         });
```
ktuite, line 1295: What happens when `currentDefId` is the same as `draftDefId`?

matthew-white (approved):

PR #1309, opened 2024-11-27
Add test for /backup

### LINKED ISSUE #1305: /backup path does not have tests (lib/resources/backup)
This seems like an important endpoint, and it should probably have some tests.

Demonstrated by https://github.com/getodk/central-backend/pull/1306.

alxndrsn:
Closes #1305

Ran tests.

There are probably better solutions, like:

1. checking the backup tar.gz files can be reconstructed
2. checking the backup actually restores

**1** may stray into the realms of testing `pg_dump` itself.

This change should not affect users.

No.


matthew-white:
I think this could probably use a rebase now that #1310 is merged.

alxndrsn:
> I think this could probably use a rebase now that #1310 is merged.

I will update prior to requesting review.

ktuite (approved):
Good test to have! 

I was thinking if there was a way to test the `passphrase` encryption part of this endpoint but I don't see an easy way and I don't think it's necessary what With QA doing regression testing on this endpoint and some other tests of the inner workings of restoring/decrypting (e.g. `decryptFromArchive`).

PR #1310, opened 2024-11-27
test/util/zip: processZipFile(): prevent filename clashes

alxndrsn:
Separating file data and meta data prevents clashes for files with names including:

* `filenames`
* `toString`
* etc.

# Motivation

Working on #1309, the return value from `httpZipResponseToFiles()` was confusing as it mixes both zipfile contents and zipfile metadata in the same object.  This PR aims to ease use of `httpZipResponseToFiles()`.

# PR template

Ran tests.

This change was prompted by difficulties in understanding the results of `httpZipResponseToFiles()` when writing new zip-based tests.

Only affects developers; hopefully change is intuitive and improves understanding.

No.

### DIFF
--- a/test/unit/util/zip.js
+++ b/test/unit/util/zip.js
@@ -210,8 +210,7 @@ describe('zipStreamToFiles()', () => {
     const part = zipPart();
 
     zipStreamToFiles(zipStreamFromParts(part), (err, result) => {
-      // eslint-disable-next-line keyword-spacing
-      if(err) return done(err);
+      if (err) return done(err);
 
       result.filenames.should.eqlInAnyOrder([
         'test1.file',
--- a/test/unit/util/zip.js
+++ b/test/unit/util/zip.js
@@ -106,10 +106,10 @@ describe('zipPart streamer', () => {
         'y/test4.file'
       ]);
 
-      result['x/test1.file'].should.equal('test 1');
-      result['x/test2.file'].should.equal('test 2');
-      result['x/test3.file'].should.equal('test 3');
-      result['y/test4.file'].should.equal('test 4');
+      result.files.get('x/test1.file').should.equal('test 1');
+      result.files.get('x/test2.file').should.equal('test 2');
+      result.files.get('x/test3.file').should.equal('test 3');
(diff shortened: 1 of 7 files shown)


test/util/zip.js
```diff
@@ -8,12 +8,14 @@ const streamTest = require('streamtest').v2;
 // calls the callback with an object as follows:
 // {
 //      filenames: [ names of files in zip ],
-//      {filename}: "contents",
-//      {filename}: "contents",
-//      …
+//      files: {
```
matthew-white, line 11: Might be worth mentioning that `files` is a `Map`?

test/unit/util/zip.js
```diff
@@ -208,3 +208,33 @@ describe('zipPart streamer', () => {
   });
 });
 
+describe('zipStreamToFiles()', () => {
+  it('should not conflate metadata & file data', (done) => {
+    const part = zipPart();
+
+    zipStreamToFiles(zipStreamFromParts(part), (err, result) => {
+      // eslint-disable-next-line keyword-spacing
+      if(err) return done(err);
```
matthew-white, line 217: Would be nice to make it `if (err)` with a space instead of disabling the ESLint rule. We could probably set up ESLint to auto-fix things like this if that'd be helpful.

matthew-white (approved):

matthew-white:
Thank you for the extra information in the PR description, it's helpful!

test/util/zip.js
alxndrsn: Or could we remove the whole comment?  The structure can be seen on line 18, and can also be understood from existing examples and at runtime by inspection.

test/unit/util/zip.js
alxndrsn: :+1: this was copied from elsewhere in this file.  I've opened a PR to clean up similar eslint-disable directives: https://github.com/getodk/central-backend/pull/1328

matthew-white: Thank you!

test/util/zip.js
matthew-white: That makes sense. :+1:

PR #1311, opened 2024-11-28
Add submissionId in delete and restore audit

sadiqkhoja:
Part of getodk/central#789

Updated existing test.

~This is the simplest approach. Other options is to backfill submissionId in the audit table for submission.delete and submission.restore events.~

Added `submissionId` in delete and restore audit.

None

NA

### DIFF
--- a/lib/model/query/submissions.js
+++ b/lib/model/query/submissions.js
@@ -407,11 +407,11 @@ const getForExport = (formId, instanceId, draft, options = QueryOptions.none) =>
 
 const del = (submission) => ({ run }) =>
   run(markDeleted(submission));
-del.audit = (submission, form) => (log) => log('submission.delete', { acteeId: form.acteeId }, { instanceId: submission.instanceId });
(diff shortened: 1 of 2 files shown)


matthew-white:
I feel a little unsure about this approach. `submissionId` is unique, but `instanceId` isn't globally unique. `instanceId` is unique within a particular form, but it's possible that the user could push the same submissions to the same form in two different projects. We want to avoid a situation where the audit log events for one submission show up for a different submission that has the same instance ID.

Another scenario to consider is if the user deletes a submission, allows it to be purged, then creates the submission again. The audit log events for the deleted submission will still exist, but I don't think we want them to show up for the new submission.

> Other options is to backfill submissionId in the audit table for submission.delete and submission.restore events.

Another option is to log `submissionId` going forward, but not backfill it for existing `submission.delete` and `submission.restore` events. I don't think it's a big deal if those events don't show up in some submissions' audit logs, as long as they start doing so going forward. It's only with the upcoming release that submission deletion will be widely accessible and available via Frontend.

matthew-white (approved):

PR #1319, opened 2024-12-03
Log create-as-update as a conflict

ktuite:
Closes https://github.com/getodk/central/issues/815

From [here](https://github.com/getodk/central/issues/815#issuecomment-2514895303):
> @ktuite, I think what's happening here is that the entity as a whole is being marked as a conflict, but the API is not returning v2 as a conflict (its conflict property is null). Now that https://github.com/getodk/central-backend/pull/1251 has been merged, I think we need to update getWithConflictDetails() in a similar way to how it was updated in https://github.com/getodk/central-backend/pull/1187. There are now three situations in which an entity can be marked as a conflict (described in https://github.com/getodk/central/issues/805), but getWithConflictDetails() only accounts for two of them ([here](https://github.com/getodk/central-backend/blob/0bc6209a81bee6e1a7b3cc86517cb3c4688564b0/lib/data/entity.js#L503-L504)).

Screenshot 2024-12-06 at 4 45 39 PM

### DIFF
--- a/test/integration/api/offline-entities.js
+++ b/test/integration/api/offline-entities.js
@@ -1198,7 +1198,7 @@ describe('Offline Entities', () => {
         .then(({ body }) => {
           body.currentVersion.version.should.equal(2);
           body.currentVersion.data.should.eql({ age: '20', status: 'new', first_name: 'Megan' });
-          body.conflict.should.equal('soft'); // this should be marked as a soft conflict
+          body.conflict.should.equal('hard'); // this should be marked as a soft conflict
           body.currentVersion.baseVersion.should.equal(1); // baseVersion is set, but normally the baseVersion of an entity-create is null
           // the rest of these are null like a normal entity-create
           should.not.exist(body.currentVersion.branchBaseVersion);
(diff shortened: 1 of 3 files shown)


test/integration/api/offline-entities.js
```diff
@@ -1714,6 +1714,54 @@ describe('Offline Entities', () => {
...
+        .then(({ body }) => {
+          body.conflict.should.equal('soft');
+        });
+
+      // Check versions
+      await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789ddd/versions')
+        .expect(200)
+        .then(({ body }) => {
+          body.map(v => v.conflict).should.eql([null, 'hard']);
+          body[1].conflictingProperties.should.eql(['age', 'label', 'status', 'first_name']);
```
ktuite, line 1762: These will probably change

test/integration/api/offline-entities.js
```diff
@@ -1714,6 +1714,54 @@ describe('Offline Entities', () => {
...
+        .send(testData.instances.offlineEntity.two)
+        .set('Content-Type', 'application/xml')
+        .expect(200);
+
+      await exhaust(container);
+
+      // Check that entity as a whole is a conflict
+      await asAlice.get('/v1/projects/1/datasets/people/entities/12345678-1234-4123-8234-123456789ddd')
+        .then(({ body }) => {
+          body.conflict.should.equal('soft');
```
ktuite, line 1754: This will change to hard.

matthew-white (approved):

ktuite:
soft conflict case 

Screenshot 2024-12-06 at 4 48 45 PM

matthew-white (approved):

PR #1325, opened 2024-12-04
bin/s3/upload-pending: handle missing pgrowlocks extension

alxndrsn:
Closes #812

Tested manually.

Require pgrowlocks extension?  This was initially included, but removed by popular request: https://github.com/getodk/central-backend/pull/1190

Hopefully fixes s3 upload-pending for them.

I don't think this relates to the API.


matthew-white (approved):
Took a look with @ktuite, and this is looking good. :+1:

PR #1327, opened 2024-12-04
oidc: pass state to IdP in authorization URL

### LINKED ISSUE #1134: OIDC login fails due to Authentik sending state QS and node-openid-client failing a check

alxndrsn:
* pass `next` value to IdP as `state`, with additional random prefix
* this makes the "next" value more trusted
* fix support for Authentik and other IdPs which do not support auth URLs without `state` (https://github.com/goauthentik/authentik/pull/9735)
* may also improve support for IdPs which lack PKCE support (see: https://github.com/panva/openid-client/blob/1486c3a020af8d12449d1d6a4bdf4f2bf4d32b77/README.md#authorization-code-flow)

Closes #1134
Closes #1135

- [x] test with `fake-oidc-server`
- [x] test with google oauth

Alternative: separate `state` & `next` cookies in https://github.com/getodk/central-backend/pull/1135

The approach in this PR is an improvement on that as it prevents tampering with the `next` cookie, e.g. by cookie tossing.

Should only affect users integrated with Authentik, which should now work.

No.

### DIFF
--- a/lib/resources/oidc.js
+++ b/lib/resources/oidc.js
@@ -43,7 +43,7 @@ const ONE_HOUR = 60 * 60 * 1000;
 // * https://bugzilla.mozilla.org/show_bug.cgi?id=1648993
 // * https://bugs.chromium.org/p/chromium/issues/detail?id=1056543
 const CODE_VERIFIER_COOKIE = (HTTPS_ENABLED ? '__Secure-' : '') + 'ocv';
-const NEXT_COOKIE          = (HTTPS_ENABLED ? '__Secure-' : '') + 'next'; // eslint-disable-line no-multi-spaces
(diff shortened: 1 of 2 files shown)


lib/resources/oidc.js
```diff
@@ -105,17 +108,19 @@ module.exports = (service, endpoint) => {
 
       const code_challenge = generators.codeChallenge(code_verifier); // eslint-disable-line camelcase
 
+      const { next } = req.query ?? '';
```
alxndrsn, line 111: This may not do what it intends to... I suspect this is different to:

```js
const next = req.query.next ?? '';
```

Maybe add a test for `next` missing, if that's a possible situation.

PR #1331, opened 2024-12-05
Feature #769: Reject duplicate properties with different capitalization

sadiqkhoja:
Closes getodk/central#769

Added tests and manual verification.

I initially changed the database column type to `citext` to stop duplicate property names (case insensitive) but that prevented creation even when existing properties were unpublished. With the implemented approach we can have two unpublished Forms with same properties but different capitalization. Error is thrown when user tries to publish second one.

Change affects already complicated area of Form and EntityList creation, that should be tested thoroughly.

Yes

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -143,8 +143,9 @@ const createOrMerge = (parsedDataset, form, fields) => async ({ context, one, Ac
 
     // Check properties with same name but different capitualization
     const duplicateProperties = newNames
+      .filter(newName => !existingNames.includes(newName))
       .map(newName => ({
-        current: existingNames.find(existingName => existingName !== newName && newName.toLowerCase() === existingName.toLowerCase()),
+        current: existingNames.find(existingName => newName.toLowerCase() === existingName.toLowerCase()),
         provided: newName
       }))
       .filter(property => property.current);
@@ -290,6 +291,7 @@ const publishIfExists = (formDefId, publishedAt) => async ({ all, context, maybe
       AND dp.name != existing_properties.name
     WHERE dpf."formDefId" = ${formDefId}
       AND existing_properties."publishedAt" IS NOT NULL
+      AND dp."publishedAt" IS NULL
   `;
 
   const d = await maybeOne(_datasetNameConflict());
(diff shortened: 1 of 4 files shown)


test/integration/api/datasets.js
```diff
@@ -3954,6 +3982,68 @@ describe('datasets and entities', () => {
...
+          const alice = await service.login('alice');
+
+          // dataset "people" with property "first_name"
+          await alice.post('/v1/projects/1/forms?publish=True')
+            .send(testData.forms.simpleEntity)
+            .set('Content-Type', 'application/xml')
+            .expect(200);
+
+          // dataset "people" with property "FIRST_NAME"
+          await alice.post('/v1/projects/1/forms?publish=True')
```
matthew-white, line 3996: What happens if ?publish=True is not specified (if you're uploading a draft, but not publishing)? Am I right in thinking that that's also a 409, since `first_name` is a published property at this point? Either way, would you mind adding a request along those lines to this test just to document the behavior?

matthew-white (approved):

matthew-white:
> I initially changed the database column type to `citext` to stop duplicate property names (case insensitive) but that prevented creation even when existing properties were unpublished.

I think it's also possible that there are already entity lists in production with property names that differ only on case. We want to prevent new properties like that, but I don't think we can add a database constraint if there could be existing properties like that.

test/integration/api/datasets.js
matthew-white: I see that 75d0de3f810e375fca8bf62415581a6ed985f3b3 adds a test along these lines. Looks good!

test/integration/api/datasets.js
```diff
@@ -4063,6 +4063,23 @@ describe('datasets and entities', () => {
             });
 
         }));
+
+        it('should reject when publishing duplicate property with different capitalization', testService(async (service, container) => {
```
matthew-white, line 4067: I feel like the test title should say something else, since it's not rejecting.

lib/model/query/datasets.js
```diff
@@ -290,6 +291,7 @@ const publishIfExists = (formDefId, publishedAt) => async ({ all, context, maybe
       AND dp.name != existing_properties.name
```
matthew-white, line 291: I'd guess that we don't need this condition anymore. If `dp.name = existing_properties.name`, that would mean that `dp` is published, but we know that's not the case.

matthew-white (approved):
The new commits look good!

PR #788, opened 2023-03-02
API docs about OData Dataset Service

ktuite:
Closes #785 

#### What has been done to verify that this works as intended?

API Docs update

#### Does this change require updates to the API documentation? If so, please update docs/api.md as part of this PR.

That's what it is!

#### Before submitting this PR, please make sure you have:

- [x] run `make test-full` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1333, opened 2024-12-06
Handle additional bodyParser errors

alxndrsn:
Closes #788

Added tests.

Follows the pattern of the existing code.

No likely effect.

No.

### DIFF
--- a/lib/http/service.js
+++ b/lib/http/service.js
@@ -122,14 +122,18 @@ module.exports = (container) => {
   service.use((error, request, response, next) => {
     if (response.headersSent === true) {
       // In this case, we'll just let Express fail the request out.
-      next(error);
-    } else if (error?.type === 'entity.parse.failed') {
-      // catch body-parser middleware problems. we only ask it to parse JSON, which
-      // isn't part of OpenRosa, so we can assume a plain JSON response.
-      defaultErrorWriter(Problem.user.unparseable({ format: 'json', rawLength: error.body.length }), request, response);
-    } else {
-      defaultErrorWriter(error, request, response);
+      return next(error);
     }
+
+    // catch body-parser middleware problems. we only ask it to parse JSON, which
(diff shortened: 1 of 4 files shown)


lib/http/service.js
```diff
@@ -122,14 +122,18 @@ module.exports = (container) => {
...
-      defaultErrorWriter(Problem.user.unparseable({ format: 'json', rawLength: error.body.length }), request, response);
-    } else {
-      defaultErrorWriter(error, request, response);
+      return next(error);
+    }
+
+    // catch body-parser middleware problems. we only ask it to parse JSON, which
+    // isn't part of OpenRosa, so we can assume a plain JSON response.
+    switch (error?.type) {
+      case 'encoding.unsupported': return defaultErrorWriter(Problem.user.unparseable({ format: 'json' }), request, response);
```
sadiqkhoja, line 131: The message of `Problem.user.unparseable` without `rawLength` looks incomplete i.e:

> "Could not parse the given data ( chars) as json."

lib/http/service.js
```diff
@@ -122,14 +122,18 @@ module.exports = (container) => {
...
+      case 'entity.parse.failed': return defaultErrorWriter(Problem.user.unparseable({ format: 'json', rawLength: error.body.length }), request, response);
+      case 'entity.too.large': return defaultErrorWriter(Problem.user.unparseable({ format: 'json', rawLength: error.length }), request, response);
```
sadiqkhoja, line 133:
```suggestion
      case 'entity.too.large': return defaultErrorWriter(Problem.user. requestTooLarge(), request, response);
```

lib/http/service.js
alxndrsn: Probably not the correct error anyway for this case.

sadiqkhoja (approved):

PR #1334, opened 2024-12-08
POST submission attachments: reduce audit logging overhead

alxndrsn:
In the submission attachment POST endpoint and a few other submission-related endpoints, a full submission definition including XML is fetched from the database, where only one or two columns from that result are used.

This commit reduces the data fetched from the database to only those which are required.

Tests.

There may be more idiomatic ways of defining query results?

It should not affect users.

No.


ktuite (approved):
I hope this improves submission attachments a little bit! Not sure how to measure it, but not fetching the whole XML of the submission def and only picking specific columns seems like it will be useful if the submission XML is large and there are a lot of attachments. This improvement is also part of submission editing and checking against the previous deprecated def instance ID.

Just wanted to note for whoever sees this in the future that one thing that also makes the audit logging slow here is the query to get the old attachment blob ID so it can be referenced in the audit log. That query is still present, but at least it is leaner now.

> There may be more idiomatic ways of defining query results?

I think this is fine; there are some DB utils for joining and extending frames but not really for picking specific fields/cols of a Frame. The new helper functions make sense to me!

PR #1338, opened 2024-12-10
Fix tests on Node 22

lognaturel:
OIDC and S3 end to end tests are failing on `master`. I don't really understand what happened because #1337 showed a green build. Now it's only showing the `build` task but prior runs showed failures from the e2e tests. 

The OIDC tests definitely passed at https://github.com/lognaturel/central-backend/actions/runs/12246914088/job/34163758377 but it did look like the S3 tests had failed at https://github.com/lognaturel/central-backend/actions/runs/12246914096/job/34163758183 I'm not sure why that didn't make the build red that time.

I believe the OIDC tests either have been flaky or are now flaky with Node 22. Here's another run that passes: https://github.com/lognaturel/central-backend/actions/runs/12248587480/job/34168459559

The problem with the S3 tests is that there are some regex that expect an exact output message and the deprecation warnings mess that up ([test run](https://github.com/lognaturel/central-backend/actions/runs/12248901099/job/34169319375)). The call that results in the issue is
https://github.com/getodk/central-backend/blob/4355d9e4c3a0f7bb6b694aad4cf318145905da2e/test/e2e/s3/test.js#L405

I though I could change that to `exec node --node-deprecation lib/bin/s3` but that didn't work. I got desperate and tried some things that make no sense. I'm pretty confident the functionality is working ok. Maybe we just change the regex?

### DIFF
--- a/test/e2e/oidc/run-tests.sh
+++ b/test/e2e/oidc/run-tests.sh
@@ -49,6 +49,6 @@ if [[ ${INSTALL_PLAYWRIGHT_DEPS-} = true ]]; then
   npx playwright install --with-deps
 fi
 log "Running playwright tests..."
-npx --node-options="--no-deprecation" playwright test
+NODE_OPTIONS="--no-deprecation" npx playwright test
 
 log "Tests completed OK!"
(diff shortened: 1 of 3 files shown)


alxndrsn:
Issue filed at https://github.com/getodk/central-backend/issues/1339.

Issue #1228, opened 2024-10-16
Query fewer analytics at once


matthew-white:
If usage reporting is enabled, Backend will send a sizable number of metrics in its usage report (almost 100). Some of the database queries behind those metrics touch every row in the largest tables (e.g., the `audits`, `submissions`, and `client_audits` tables), so those queries could take longer. Currently, we query all metrics in parallel. Especially as we continue to add metrics, it seems possible that the number of concurrent queries will lead to performance problems. The usage report is run at 3 a.m. UTC, but in some time zones, user activity is expected at that time.

In practice, I don't see much evidence that there have been performance problems. Looking at analytics from the past 31 days, the vast majority of usage reports were received within a minute of 3 a.m. A small number of usage reports were received after 1 minute past 3 a.m., including some after 5 minutes and one after 40 minutes (!). There are multiple variables that could affect the time of submission receipt, so those numbers don't necessarily speak to the performance of the analytics queries, though that's a possibility.

Side note: I'm actually confused about how a usage report could be successful after taking more than 1 minute, much less 5 minutes. I'm pretty sure that the number of analytics queries exceeds the size of the database connection pool. Doesn't Slonik time out fairly quickly if it can't acquire a database connection? Why isn't it timing out these queries? We've seen that transactions use a single database connection even if they make multiple queries, but I don't think that the analytic queries are run in a transaction. Maybe most of the queries are fast enough that it's not a problem.

If query timeout were an issue, I'd be concerned that we're not receiving analytics from large servers where the queries could take a long time. I see large servers in our analytics, so I don't have reason to believe that's happening. But like I said, I'm confused about why query timeout doesn't seem to be an issue here.

Even assuming query timeout isn't an issue, I think it makes sense to reduce the number of concurrent queries in order to reduce the risk of performance issues.

There's also an issue in testing that I think is related to the number of concurrent queries. When I run tests, I see a lot of `MaxListenersExceededWarning`. Many of those seem related to analytics. For example:

```
  task: analytics
    ✔ should not compute analytics if not enabled
    ✔ should not compute analytics if explicitly disabled
    ✔ should not compute analytics if analytics sent recently
(node:80089) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 notice listeners added to [Client]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
    ✔ should send analytics if enabled and time to send (111ms)
[...]

PR #1344, opened 2024-12-12
Change analytics queries to run sequentially

ktuite:
Closes https://github.com/getodk/central-backend/issues/1228

Makes use of `runSequentially` instead of Promise.all() to do multiple database queries for computing usage metrics.

### DIFF
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -10,6 +10,7 @@
 const config = require('config');
 const { sql } = require('slonik');
 const { clone } = require('ramda');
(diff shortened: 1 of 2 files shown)


matthew-white:
I'll let @sadiqkhoja do the review, but I had just one thought. Are we able to remove lines like this from tests of analytics now?

https://github.com/getodk/central-backend/blob/63ca7881f6e6eb0b5c9051bf64448d802720f100/test/integration/other/analytics-queries.js#L2148-L2148

sadiqkhoja (approved):

PR #1126, opened 2024-04-18
Improve entity details audits query

ktuite:
Hopefully addresses https://github.com/getodk/central/issues/618

I ended up pulling out the crazy part of this query that gets all the submission details that are vaguely relevant to an entity. It is now in another query that gets run separately and then if there is submission data to use, it is combined in code.

I'm curious to try this on test! I've been `EXPLAIN ANALYZE`-ing the queries. Hopefully an entity with a bulk source loads _super_ fast, single milliseconds.

### Entity 648 - created from a submission
https://test.getodk.cloud/#/projects/496/entity-lists/participants/entities/b5725bd9-fb5f-41e0-ba30-0eb8e6905e07

Original query (master)
* Planning Time: 10.761 ms
* Execution Time: 1357.950 ms
(I've seen it 800ms to >2 seconds)

New queries
Part 1: get all audit details and source events
* Planning Time: 2.093 ms
* Execution Time: 438.128 ms
(Seen between 300ms to 900ms)

Part 2: get full submission details
* Planning Time: 27.339 ms
* Execution Time: 2.709 ms
(this part is much faster w/o being joined with audits)

### Entity 118019 - created from bulk upload
https://test.getodk.cloud/#/projects/364/entity-lists/trees/entities/1d8cc1ba-fce5-44c9-acc2-e1fbafbaf81e

Original query
* Planning Time: 3.402 ms
* Execution Time: 515.838 ms
(I've seen it much slower though, 400-800ms, maybe longer)

New queries
Part 1
* Planning Time: 9.669 ms
* Execution Time: 5.648 ms

Part 2
* Planning Time: 8.170 ms
* Execution Time: 2.586 ms

#### Before submitting this PR, please make sure you have:

- [ ] run `make test` and confirmed all checks still pass OR confirm CircleCI build passes
- [ ] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1349, opened 2024-12-30
#710: Restore Entities, Purge Entities, ability to see deleted Entities

sadiqkhoja:
Part of getodk/central#710

Written bunch of tests, ran it with update frontend as well.

I have saved UUIDs of the purged entities in the audits table with `entities.purge` row and created GIN index on the property - I have verified its usage by insert 100K+ records in audits table.

Other option is to create a separate table to store all the UUID which we can do if we sense performance is not acceptable with the JSONB in audits table.

I was also thinking to just check for UUID in `entity.create` audits; if given UUID was ever created - but for `entity.bulk.create` we don't log UUIDs.

Now entity creation checks for the duplicate UUID in the purged history, we should verify that there is no regression.

Yes


lib/model/query/entities.js
```diff
@@ -20,6 +20,18 @@ const { isTrue } = require('../../util/http');
...
+
+const _getPurgedUuids = (all, uuids) => all(sql`
+    SELECT jsonb_array_elements_text(details -> 'entityUuids') AS uuid FROM audits a WHERE action = 'entities.purge'
+    INTERSECT
+    SELECT jsonb_array_elements_text(${JSON.stringify(uuids)})`)
+  .then(all.map(r => r.uuid));
+
+const _isPurgedUuid = (maybeOneFirst, uuid) => maybeOneFirst(sql`
+  SELECT TRUE FROM audits WHERE action = 'entities.purge' AND details -> 'entityUuids' @> ${JSON.stringify([uuid])}
+  `);
```
sadiqkhoja, line 34: I can pass a single uuid in `_getPurgedUuuids` but my gut feeling is that `_isPurgedUuid` would be faster. 🤷‍♂️

lib/model/query/entities.js
```diff
@@ -849,15 +873,94 @@ CROSS JOIN
 const del = (entity) => ({ run }) =>
   run(markDeleted(entity));
 
-del.audit = (entity, dataset) => (log) => log('entity.delete', entity.with({ acteeId: dataset.acteeId }), { uuid: entity.uuid });
+del.audit = (entity, dataset) => (log) =>
+  log(
+    'entity.delete',
+    entity.with({ acteeId: dataset.acteeId }),
+    { entity: { uuid: entity.uuid } }
+  );
```
sadiqkhoja, line 881: Made it consistent with other entity logs where uuid is under the entity object

lib/model/query/audits.js
```diff
@@ -150,11 +150,19 @@ const getBySubmissionId = (submissionId, options) => ({ all }) => _getBySubmissi
...
-
+SELECT ${fields} FROM (
+    SELECT audits.* FROM audits 
+    JOIN entities ON (audits.details->'entity'->>'uuid') = entities.uuid
+    WHERE entities.id = ${entityId}
+    UNION ALL
+    SELECT audits.* FROM audits
+    JOIN entity_def_sources ON (audits.details->>'sourceId')::INTEGER = entity_def_sources.id
+    JOIN entity_defs ON entity_def_sources.id = entity_defs."sourceId"
+    WHERE entity_defs."entityId" = ${entityId} AND audits.action = 'entity.bulk.create'
```
sadiqkhoja, line 161: This fetches only entity.bulk.create event

lib/model/query/audits.js
```diff
@@ -150,11 +150,19 @@ const getBySubmissionId = (submissionId, options) => ({ all }) => _getBySubmissi
...
 const _getByEntityId = (fields, options, entityId) => sql`
-SELECT ${fields} FROM entity_defs
-
-  LEFT JOIN entity_def_sources on entity_def_sources.id = entity_defs."sourceId"
-  INNER JOIN audits ON ((audits.details->>'entityDefId')::INTEGER = entity_defs.id OR (audits.details->>'sourceId')::INTEGER = entity_def_sources.id)
-
+SELECT ${fields} FROM (
+    SELECT audits.* FROM audits 
+    JOIN entities ON (audits.details->'entity'->>'uuid') = entities.uuid
+    WHERE entities.id = ${entityId}
```
sadiqkhoja, line 156: The fetches all logs of the entity (create, update, delete, restore)

lib/model/query/entities.js
```diff
@@ -849,7 +873,94 @@ CROSS JOIN
...
+  // entities.
+  return oneFirst(sql`
+    WITH redacted_audits AS (
+      UPDATE audits SET notes = NULL
+      FROM entities
+      JOIN datasets ON entities."datasetId" = datasets.id
+      WHERE (audits.details->'entity'->>'uuid') = entities.uuid
+      AND ${_trashedFilter(force, projectId, datasetName, entityUuid)}
+    ), purge_audit AS (
+      INSERT INTO audits ("action", "loggedAt", "processed", "details")
```
sadiqkhoja, line 936: Let's create separate row for each dataset, for integrity URL we need to return all entities if no id query parameter is passed.

sadiqkhoja: done

lib/model/query/entities.js
sadiqkhoja: verified that `_isPurgedUuid` is using the index and is much quicker.

docs/api.yaml
```diff
@@ -9314,6 +9276,100 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error403'
       x-codegen-request-body-name: body
+    delete:
```
sadiqkhoja, line 9279: moved `delete` after `patch` so that delete and undelete are together

lib/model/query/audits.js
```diff
@@ -150,11 +150,19 @@ const getBySubmissionId = (submissionId, options) => ({ all }) => _getBySubmissi
 // There is a separate query below to assemble full submission details for non-deleted
 // submissions, but it was far too slow to have be part of this query.
 const _getByEntityId = (fields, options, entityId) => sql`
```
sadiqkhoja, line 152: Objective of the change in this file is to get entity.delete and entity.restore logs as well

lib/model/query/entities.js
```diff
@@ -849,7 +873,95 @@ CROSS JOIN
...
+    entity.with({ acteeId: dataset.acteeId }),
+    { entity: { uuid: entity.uuid } }
+  );
+
+////////////////////////////////////////////////////////////////////////////////
+// PURGE DELETED ENTITIES
+
+const PURGE_DAY_RANGE = config.has('default.taskSchedule.purge')
+  ? config.get('default.taskSchedule.purge')
+  : 30; // Default is 30 days
```
sadiqkhoja, line 901: Maybe we should move this to central place, this is present in three file at the moment.

docs/api.yaml
```diff
@@ -9314,6 +9276,100 @@ paths:
...
+      tags:
+      - Entity Management
+      summary: Deleting an Entity
+      description: Use this API to delete an Entity. With this API, Entity is soft-deleted,
+        which means it is still in the database and you can retreive it by passing
+        `?deleted=true` to [GET /projects/:id/datasets/:name/entities](/central-api-entity-management/#entities-metadata).
+        In the future, we will provide a way to restore deleted entities and purge
+        deleted entities.
```
ktuite, line 9287: Can remove this sentence

lib/model/query/entities.js
```diff
@@ -849,7 +873,95 @@ CROSS JOIN
...
+      AND ${_trashedFilter(force, projectId, datasetName, entityUuid)}
+    ), purge_audit AS (
+      INSERT INTO audits ("action", "acteeId", "loggedAt", "processed", "details")
+        SELECT 'entities.purge', datasets."acteeId", clock_timestamp(), clock_timestamp(), jsonb_build_object('entitiesDeleted', COUNT(*), 'entityUuids', ARRAY_AGG(uuid))
+        FROM entities
+        JOIN datasets ON entities."datasetId" = datasets.id
+        WHERE ${_trashedFilter(force, projectId, datasetName, entityUuid)}
+        GROUP BY datasets."acteeId"
+        HAVING count(*) > 0
+    ), delete_entity_sources AS (
```
ktuite, line 943: If the source is a bulk source, deleting this for one purged entity doesn't seem right and also leads to FK violation.

My thoughts are either
1. never delete an entity_def_source?
2. only delete entity_def_source rows if the type is submission, but leave API sources (which could be singular or bulk)

lib/model/query/entities.js
```diff
@@ -849,7 +873,95 @@ CROSS JOIN
...
+      DELETE FROM entity_def_sources
+      USING entity_defs, entities, datasets
+      WHERE entity_def_sources.id = entity_defs."sourceId"
+      AND entity_defs."entityId" = entities.id
+      AND entities."datasetId" = datasets.id
+      AND ${_trashedFilter(force, projectId, datasetName, entityUuid)}
+    ), delete_submission_backlog AS (
```
ktuite, line 950: If this backlog is working correctly, this table will probably be empty and we don't have to delete anything here. 

The only thing I'd want to see tested would be if you delete and purge while something is in the backlog, the worker that tries to force-process the backlog doesn't have a problem. But I don't think it should... I think it'll just fail to find the entity it's looking for and it'll quietly skip it.

ktuite (commented):
Apologies for taking so long to review this! I've now looked through
- restoring a deleted entity
- `deletedAt` in the odata feed
- the migrations including the change from having `uuid` on its own in an audit entry to nesting it under `entity {}`
- the audit log query for all entity-related events
- the purging task
- the purge query (see my comment about not necessarily purging from `entity_def_sources`)

The only thing I haven't really thought throughly about yet is checking for collisions with purged uuids. 

It's looking really solid 👍

lib/model/query/audits.js
ktuite: This query change looks good to me. I remember working on this query in https://github.com/getodk/central-backend/pull/1126 because it was previously very slow. I like your solution better (matches the other audit queries better) and it doesn't feel slow (using my dev database that has 4M+ entities in it.

lib/model/query/entities.js
```diff
@@ -65,10 +77,16 @@ const nextval = sql`nextval(pg_get_serial_sequence('entities', 'id'))`;
...
 // error logging).
-const createNew = (dataset, partial, subDef, sourceId, userAgentIn) => ({ one, context }) => {
+const createNew = (dataset, partial, subDef, sourceId, userAgentIn) => async ({ one, context, maybeOneFirst }) => {
   let creatorId;
   let userAgent;
 
+  // Validate UUID against purged entities
+  (await _isPurgedUuid(maybeOneFirst, partial.uuid))
+    .ifDefined(() => {
+      throw Problem.user.uniquenessViolation({ fields: 'uuid', values: partial.uuid });
```
ktuite, line 87: I think this should throw this instead
```suggestion
      throw Problem.user.entityDeleted({ entityUuid: clientEntity.uuid });
```

But I also noticed that if you try to reuse a soft-deleted UUID, it also has `Problem.user.uniquenessViolation` instead of `Problem.user.entityDeleted`.

lib/model/query/entities.js
```diff
@@ -111,6 +129,12 @@ const createMany = (dataset, rawEntities, sourceId, userAgentIn) => async ({ all
   const creatorId = context.auth.actor.map((actor) => actor.id).orNull();
   const userAgent = blankStringToNull(userAgentIn);
 
+  // Validate UUID with the purged entities
+  const purgedUuids = await _getPurgedUuids(all, rawEntities.map(e => e.uuid));
+  if (purgedUuids.length > 0) {
+    throw Problem.user.uniquenessViolation({ fields: 'uuid', values: purgedUuids.join(', ') });
```
ktuite, line 135: I'm not sure what this should say given that the existing message mentions a single entity uuid. 

"The entity with UUID (uuid1, uuid2, uuid3,...) has been deleted."

But it might be better than what it currently returns, e.g. this response json: 

```
{'code': 409.3,
 'details': {'fields': 'uuid',
             'values': '88ed7209-cb54-4b7e-98ed-e1cc26a93cc6, '
                       'd8b37943-9202-44ce-837b-8be85a34d373'},
 'message': 'A resource already exists with uuid value(s) of '
            '88ed7209-cb54-4b7e-98ed-e1cc26a93cc6, '
            'd8b37943-9202-44ce-837b-8be85a34d373.'}
```

lib/model/query/entities.js
ktuite: Agreed, maybe it could be injected in the container around [here](https://github.com/getodk/central-backend/blob/master/lib/model/container.js#L79) 

```
  // Inject some constants
  const PURGE_DAY_RANGE = config.has('default.taskSchedule.purge')
    ? config.get('default.taskSchedule.purge')
    : 30; // Default is 30 days
  container.constants = { PURGE_DAY_RANGE };
```

lib/model/query/entities.js
ktuite: I take this back... leave this bit in. I tried altering the test `should purge submission backlog for entities` where this query didn't delete from the backlog and an update entity submission was left in the backlog. When the backlog was later processed `await Entities.processBacklog(true);` it led to a submission processing error.

lib/resources/entities.js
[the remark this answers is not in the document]
ktuite: Interesting idea! I'm not sure if it's necessary to implement or how exactly you'd test it though.

lib/model/query/entities.js
ktuite: I am leaning towards adding `AND entity_def_sources.type = 'submission'` 

Here's a test where entities are uploaded in bulk but not all are deleted/purged:
```
const createBulkEntities = async (user, count, projectId, datasetName) => {
  const entities = [];
  const uuids = [];
  for (let i = 0; i < count; i += 1) {
    const _uuid = uuid();
    entities.push({ uuid: _uuid, label: 'label' });
    uuids.push(_uuid);
  }
  await user.post(`/v1/projects/${projectId}/datasets/${datasetName}/entities`)
    .send({
      entities,
      source: { name: 'bulk.csv', size: count }
    })
    .expect(200);
  return uuids;
};
```
```
    it.only('should not purge bulk entity sources', testService(async (service, container) => {
      const { Entities, oneFirst } = container;

        .expect(200);

      const uuids = await createBulkEntities(asAlice, 2, 1, 'people');

      let sourcesCount = await oneFirst(sql`select count(1) from entity_def_sources`);
      sourcesCount.should.be.equal(1);

      await deleteEntities(asAlice, uuids.slice(0, 1), 1, 'people');

      await Entities.purge(true);

      sourcesCount = await oneFirst(sql`select count(1) from entity_def_sources`);
      sourcesCount.should.be.equal(1);
    }));
```

lib/model/query/entities.js
```diff
@@ -849,7 +873,95 @@ CROSS JOIN
...
+  return oneFirst(sql`
+    WITH redacted_audits AS (
+      UPDATE audits SET notes = NULL
+      FROM entities
+      JOIN datasets ON entities."datasetId" = datasets.id
+      WHERE (audits.details->'entity'->>'uuid') = entities.uuid
+      AND ${_trashedFilter(force, projectId, datasetName, entityUuid)}
+    ), purge_audit AS (
+      INSERT INTO audits ("action", "acteeId", "loggedAt", "processed", "details")
+        SELECT 'entities.purge', datasets."acteeId", clock_timestamp(), clock_timestamp(), jsonb_build_object('entitiesDeleted', COUNT(*), 'entityUuids', ARRAY_AGG(uuid))
```
ktuite, line 937: For submissions, the event is just `submission.purge` even though it could relate to multiple submissions. We should probably keep this consistent with other entity events too and call it `entity.purge`.

ktuite:
Could we do something else with the purged entity UUIDs instead of storing them in the `audits` table? If you purge a bunch the details could be really long. It also feels too fragile for the use cases we want to build on top of rejecting re-used uuids and the integrity url. 

```
     action     |      acteeId       |                          details
----------------+--------------------+------------------------------------------------------------------
 entities.purge | (dataset actee id) | {"entityUuids": ["uuid1", "uuid2", "uuid3"], "entitiesDeleted": 3}
```

I think we could merge _this_ PR (with the suggested changes above) with this mechanism, but come up with a different entity tombstone plan soon after (in this same release).

lib/resources/entities.js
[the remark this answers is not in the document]
sadiqkhoja: Thinking about it again, probability of collision is so low that it doesn't worth the effort. I would blame the **luck** of that user who fall into this case 🙈

docs/api.yaml
sadiqkhoja: done

lib/model/query/entities.js
sadiqkhoja: Done. I have just kept source for the entities created via bulk upload. 

Maybe in future, we can remove the orphan sources where all the entities created by them are purged.

lib/model/query/entities.js
sadiqkhoja: I have introduced `entityUniquenessViolationWithDeleted` - please suggest a shorter name. 

I think we should not return 404 during create entity where there is/was entity with the same UUID is soft-deleted or purged. So for create I am returning 409 and for update 404.

Let me know what do you think about this.

lib/model/query/entities.js
sadiqkhoja: updated with `entityUniquenessViolationWithDeleted`

lib/model/query/entities.js
sadiqkhoja: changed.

lib/model/query/entities.js
ktuite: The separate constant file looks good to me!

lib/model/query/entities.js
ktuite: I see what you mean about the 409 vs the 404. Also that other `entityDeleted` problem doesn't have a very descriptive name; it's more like `entityNotFoundDeleted`. It's also only used in one case where a submission tries to update something about an entity and the entity has been deleted. 

I can't really think of a shorter name than `entityUniquenessViolationWithDeleted`. `entityConflictDeleted`? Nah, what you have, even though it is long, is clear.

ktuite (approved):

PR #1352, opened 2025-01-10
submissions/attachments: demonstrate current null Content-Type handling

alxndrsn:
This probably isn't the desired behaviour, but it *is* the _current_ behaviour.

### DIFF
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -4377,6 +4377,50 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff
                   body.toString().should.equal('testvideo');
                 })))))));
 
+    // Ref https://github.com/getodk/central-backend/issues/1351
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
I don't think this code base has a ton of comments with links to issues, the expectation is generally to use `git blame` or equivalent to get context. I don't personally feel strongly about consistency on that point but wanted to flag it in case others have thoughts.

ktuite:
I'm ok with these references to the issues even if it's a little redundant and git blame could show the trail. (Sometimes just referred to by the issue number, like `c#1351`.) We have a few sprinkled throughout the code base, especially for tests added to demonstrate specific issues rather than being part of a fix or feature. 

> I don't think this codebase has a ton of comments with links to issues, the expectation is generally to use `git blame` or equivalent to get context. I don't personally feel strongly about consistency on that point but wanted to flag it in case others have thoughts.

ktuite (approved):

Issue #1351, opened 2025-01-10
blobs.contentType can be NULL


alxndrsn:
* there is no `NOT NULL` constraint on the `blobs.contentType` column, i.e. column is `NULLABLE`
* `NULL` values are turning up in production instances
* `null` values for `Blob` frames throw in s3 code: `TypeError: response header response-content-type should be of type "string"`

### Issues

* https://github.com/getodk/central-backend/issues/489

### PRs

* https://github.com/getodk/central-backend/pull/1352
* https://github.com/getodk/central-backend/pull/1353
* https://github.com/getodk/central-backend/pull/1354
* https://github.com/getodk/central-backend/pull/1355
* https://github.com/getodk/central-backend/pull/1356

## Potential fixes

1. ✅ ~continue to allow `NULL` values, but handle these in s3 code (https://github.com/getodk/central-backend/pull/1353)~
2. ✅ ~set a default value for `blobs."contentType"` in the db, e.g. `application/octet-stream`; try to guess content type from file extension if no contentType is provided on upload~ https://github.com/getodk/central-backend/pull/1355
3. reject `null` contentType values when trying to insert values in the db
4. ✅ ~set a default `contentType` value for blobs with `NULL` values in db~ https://github.com/getodk/central-backend/pull/1355

ktuite:
Fix 1 sounds good as a patch release to me, too. We don't have to get into the messy details of blobs and content types, but it does stop breaking things.

I was able to upload blobs with null content types through the API, and in doing so, noticed we don't have a lot of documentation for uploading attachments. I was actually struggling to explicitly set a content type on my attachment uploads.

Forms don't seem to have anything about sending attachments: https://docs.getodk.org/central-api-form-management/#listing-form-attachments
Submissions do have some details: https://docs.getodk.org/central-api-submission-management/#uploading-an-attachment

alxndrsn:
> I'd like to propose doing 2 and also setting the mime types for files with geojson and cvs extensions based on extension.

:+1: I've added this to the description of #2

alxndrsn:
I think this has been resolved by earlier PRs.

PR #1353, opened 2025-01-10
s3: fix handling of NULL blob.contentType values

alxndrsn:
Ref https://github.com/getodk/central-backend/issues/1351

This aligns handling of `NULL` values for `blob."contentType"` between postgres-stored and s3-stored blobs.


test/e2e/util/api.js
```diff
@@ -99,7 +102,7 @@ async function apiClient(suiteName, { serverUrl, userEmail, userPassword, logPat
     if(isRedirected(res)) return new Redirect(res);
     if(!res.ok) {
       const responseStatus = res.status;
-      const responseText = await res.text();
+      const responseText = await res.text() || res.statusText;
```
alxndrsn, line 105: separate PR

lognaturel (approved):
This seems like a reasonable fix for now and we can take some time to consider what an improvement to both cases would look like.

ktuite:
I was able to locally reproduce the error `TypeError: response header response-content-type should be of type "string"` (with a dev server connected to s3) and then verify that this fixes it. Checking out the related PRs now.

ktuite:
The only other thing I was wondering about was if it was possible/worth checking this in the integration tests

```
    it.only('should get back null content-type when blob is on s3', testService(async (service, { Blobs, all }) => {
      global.s3.enableMock();

        .send(testData.forms.binaryType)
        .expect(200);

      await asAlice.post('/v1/projects/1/forms/binaryType/submissions')
        .send(testData.instances.binaryType.both)
        .expect(200);

      await asAlice.post('/v1/projects/1/forms/binaryType/submissions/both/attachments/my_file1.mp4')
        .send('testvideo')
        .set('Content-Type', '') // set empty content type
        .expect(200);

      const blobs = await all(sql`select id, "contentType", md5 from blobs`);
      console.log(blobs);

      await Blobs.s3UploadPending();

      await asAlice.get('/v1/projects/1/forms/binaryType/submissions/both/attachments/my_file1.mp4')
        .expect(307)
        .then(({ headers, body }) => {
          console.log("headers", headers);
          console.log("body", body);
        });
    }));
```

but the bug fix is in `lib/external/s3.js` `getRespHeaders` which is just a utility of the s3 external utility and isn't part of s3 the mock.

If we go down a longer adventure of setting content types more appropriately, we would probably want to test it more here, but for now, this patch is good.

ktuite (approved):

alxndrsn:
> The only other thing I was wondering about was if it was possible/worth checking this in the integration tests
> 
> ```
>     it.only('should get back null content-type when blob is on s3', testService(async (service, { Blobs, all }) => {
>       global.s3.enableMock();
>       const asAlice = await service.login('alice');
> 
>       await asAlice.post('/v1/projects/1/forms?publish=true')
>         .set('Content-Type', 'application/xml')
>         .send(testData.forms.binaryType)
>         .expect(200);
> 
>       await asAlice.post('/v1/projects/1/forms/binaryType/submissions')
>         .send(testData.instances.binaryType.both)
>         .set('Content-Type', 'text/xml')
>         .expect(200);
> 
>       await asAlice.post('/v1/projects/1/forms/binaryType/submissions/both/attachments/my_file1.mp4')
>         .send('testvideo')
>         .set('Content-Type', '') // set empty content type
>         .expect(200);
> 
>       const blobs = await all(sql`select id, "contentType", md5 from blobs`);
>       console.log(blobs);
> 
>       await Blobs.s3UploadPending();
> 
>       await asAlice.get('/v1/projects/1/forms/binaryType/submissions/both/attachments/my_file1.mp4')
>         .expect(307)
>         .then(({ headers, body }) => {
>           console.log("headers", headers);
>           console.log("body", body);
>         });
>     }));
> ```
> 
> but the bug fix is in `lib/external/s3.js` `getRespHeaders` which is just a utility of the s3 external utility and isn't part of s3 the mock.
> 
> If we go down a longer adventure of setting content types more appropriately, we would probably want to test it more here, but for now, this patch is good.

Can you clarify what you're suggesting testing here?  Is it the `Content-Type` header of the 307 redirect?

PR #1355, opened 2025-01-13
blobs: prevent NULL contentType in db

alxndrsn:
Set `blob."contentType"` values to `application/octet-stream` if no mime type is supplied on upload.

Related: #1351

CI.

Discussion on #1351.

This will change `Content-Type` headers for files uploaded without one from the string `null` to the string `application/octet-stream`.

Yes.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,7 @@
     "no-else-return": "off",
     "no-multiple-empty-lines": "off",
     "no-nested-ternary": "off",
-    "no-only-tests/no-only-tests": [ "error", { "block": [ "describe", "it", "describeMigration" ] } ],
+    "no-only-tests/no-only-tests": "error",
     "no-restricted-syntax": "off",
     "no-underscore-dangle": "off",
     "nonblock-statement-body-position": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,7 @@
     "no-else-return": "off",
     "no-multiple-empty-lines": "off",
     "no-nested-ternary": "off",
-    "no-only-tests/no-only-tests": "error",
+    "no-only-tests/no-only-tests": [ "error", { "block": [ "describe", "it", "describeMigration" ] } ],
     "no-restricted-syntax": "off",
     "no-underscore-dangle": "off",
     "nonblock-statement-body-position": "off",
--- a/lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
+++ b/lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
@@ -9,8 +9,9 @@
 
 const up = (db) => db.raw(`
   ALTER TABLE blobs
-    ALTER COLUMN "contentType" SET NOT NULL,
-    ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream'
+    ALTER COLUMN "contentType" TYPE TEXT USING(COALESCE("contentType", 'application/octet-stream')),
+    ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream',
+    ALTER COLUMN "contentType" SET NOT NULL
 `);
 
 const down = (db) => db.raw(`
--- a/lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
+++ b/lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
@@ -8,8 +8,8 @@
 // except according to the terms contained in the LICENSE file.
 
 const up = (db) => db.raw(`
+  UPDATE blobs SET "contentType"='application/octet-stream' WHERE "contentType" IS NULL;
   ALTER TABLE blobs
-    ALTER COLUMN "contentType" TYPE TEXT USING(COALESCE("contentType", 'application/octet-stream')),
     ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream',
     ALTER COLUMN "contentType" SET NOT NULL
 `);
--- a/lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
+++ b/lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
@@ -1,0 +1,22 @@
+// Copyright 2025 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+  ALTER TABLE blobs
+    ALTER COLUMN "contentType" SET NOT NULL,
+    ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream'
+`);
+
+const down = (db) => db.raw(`
+  ALTER TABLE blobs
+    ALTER COLUMN "contentType" DROP NOT NULL,
+    ALTER COLUMN "contentType" DROP DEFAULT
+`);
+
+module.exports = { up, down };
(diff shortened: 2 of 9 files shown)


lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
```diff
@@ -0,0 +1,22 @@
...
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+  ALTER TABLE blobs
+    ALTER COLUMN "contentType" SET NOT NULL,
+    ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream'
+`);
```
ktuite, line 14: When I run this (with null content type values in my DB) I get an error

```
migration file "20250113-01-disable-nullable-blob-content-types.js" failed
migration failed with error:
  ALTER TABLE blobs
    ALTER COLUMN "contentType" SET NOT NULL,
    ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream'
 - column "contentType" of relation "blobs" contains null values
 ```

Comment (alxndrsn) [reply lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js:14]: :facepalm: seems like a good candidate for a test :grin:

File: .eslintrc.json [hunk 2]
```diff
@@ -28,7 +28,7 @@
     "no-else-return": "off",
     "no-multiple-empty-lines": "off",
     "no-nested-ternary": "off",
-    "no-only-tests/no-only-tests": "error",
```
alxndrsn, line 31: separate PR

test/db-migrations/utils.js
```diff
@@ -1,4 +1,4 @@
-const assert = require('node:assert');
+const assert = require('node:assert/strict');
```
alxndrsn, line 1: separate PR

test/db-migrations/utils.js
```diff
@@ -77,22 +77,22 @@ function assertRowsMatch(actualRows, expectedRows) {
 
   const remainingRows = [...actualRows];
   for (let i=0; i<expectedRows.length; ++i) { // eslint-disable-line no-plusplus
-    const x = expectedRows[i];
+    const expectedRow = expectedRows[i];
```
alxndrsn, line 80: revert

.eslintrc.json
alxndrsn: https://github.com/getodk/central-backend/pull/1360

test/db-migrations/utils.js
alxndrsn: https://github.com/getodk/central-backend/pull/1361

lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
alxndrsn: Test added.

test/db-migrations/utils.js
alxndrsn: https://github.com/getodk/central-backend/pull/1362

lognaturel:
The outcome here looks as I would expect it. Would the migration possibly be time-consuming if an install has a lot of blobs? Or would it only be the case if there are lots of blobs with `null` content types?

alxndrsn:
> The outcome here looks as I would expect it. Would the migration possibly be time-consuming if an install has a lot of blobs? Or would it only be the case if there are lots of blobs with `null` content types?

What would be realistic numbers?

I've run on a fresh db, and got the following numbers:

total row count | null row count | migration duration / ms
---|---|---
`   10,000` | `  2` | `   822`
`   10,000` | `  2` | `   812`
`   10,000` | `  2` | `   785`
`   10,000` | `  1` | `   760`
`   10,000` | `  1` | `   738`
`   10,000` | `  0` | `   760`
`   10,000` | `  0` | `   708`
`   10,000` | `  0` | `   702`
`   10,000` | `  0` | `   698`
`   10,000` | `  0` | `   695`
`  100,000` | `  5` | ` 1,457`
`  100,000` | ` 16` | ` 1,456`
`  100,000` | ` 14` | ` 1,461`
`  100,000` | ` 12` | ` 1,409`
`  100,000` | ` 10` | ` 1,531`
`  100,000` | `  7` | ` 1,473`
`1,000,000` | ` 97` | `31,096`
`1,000,000` | ` 93` | `45,222`
`1,000,000` | `108` | `11,535`

lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
```diff
@@ -0,0 +1,23 @@
...
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = (db) => db.raw(`
+  ALTER TABLE blobs
+    ALTER COLUMN "contentType" TYPE TEXT USING(COALESCE("contentType", 'application/octet-stream')),
```
lognaturel, line 12: Could this be done first by a row-based query that replaces all null values with the default? Feels like that would be faster?

lognaturel:
> What would be realistic numbers?

There's a very big range, including up to the several millions so we would want to document this as possibly time-consuming in the release notes. It feels to me like the way to set the default values is really inefficient but I'm not sure whether there's an alternative - https://github.com/getodk/central-backend/pull/1355/files#r1925735636

lib/model/migrations/20250113-01-disable-nullable-blob-content-types.js
lognaturel: (Not a huge deal if you don't immediately see a good alternative, we can just document that it might take a minute or two for an install with millions of blobs)

alxndrsn: > Could this be done first by a row-based query that replaces all null values with the default? Feels like that would be faster?

Looks like your intuition is spot on - I'll share results below.

alxndrsn:
> It feels to me like the way to set the default values is really inefficient

How about this?

```sql
  UPDATE blobs SET "contentType"='application/octet-stream' WHERE "contentType" IS NULL;
  ALTER TABLE blobs
    ALTER COLUMN "contentType" SET DEFAULT 'application/octet-stream',
    ALTER COLUMN "contentType" SET NOT NULL
```

New results:

total row count | null row count | migration duration / ms
---|---|---
`   10,000` | `  0` | `  594`
`   10,000` | `  0` | `  618`
`   10,000` | `  1` | `  585`
`   10,000` | `  1` | `  590`
`   10,000` | `  1` | `  594`
`   10,000` | `  1` | `  595`
`   10,000` | `  2` | `  598`
`   10,000` | `  3` | `  582`
`   10,000` | `  3` | `  605`
`   10,000` | `  4` | `  589`
`  100,000` | ` 11` | `  685`
`  100,000` | ` 12` | `  617`
`  100,000` | ` 13` | `  621`
`  100,000` | `  6` | `  608`
`  100,000` | `  6` | `  612`
`  100,000` | `  7` | `  646`
`1,000,000` | `105` | `1,013`
`1,000,000` | `105` | `  989`
`1,000,000` | `114` | `1,270`

alxndrsn:
@lognaturel if it's helpful, we could include an additional test for the speed of the migration, e.g. this checks that 1,000,000 can be migrated in less than 5s:

https://github.com/alxndrsn/odk-central-backend/compare/application-octet-instead-of-null-content-type...alxndrsn:odk-central-backend:application-octet-instead-of-null-content-type-timing-test?expand=1

lognaturel:
Sweet, thanks! I don't feel like we need to have a test since we don't change migrations and I can't imagine we'll layer more onto this one. With these new timings I don't think we need to document anything about the migration.

ktuite (approved):

test/db-migrations/20250113-01-disable-nullable-blob-content-types.spec.js
```diff
@@ -0,0 +1,58 @@
...
+      { ...blob1, contentType: 'application/octet-stream' },
+      { ...blob2, contentType: 'text/plain' },
+    );
+  });
+
+  it(`should create new blobs with contentType 'application/octet-stream' (contentType not supplied)`, async () => {
+    const { md5, sha } = aBlob();
+
+    const created = await db.oneFirst(sql`
+      INSERT INTO blobs (md5, sha, "contentType")
```
alxndrsn, line 39: This test appears to be the same as the following one.  Should `"contentType"` be removed?

PR #1356, opened 2025-01-13
blobs: derive missing contentType from file extension

alxndrsn:
Ref #1351

Updated tests.

Discussed on https://github.com/getodk/central-backend/issues/1351

Replace `Content-Type` response header of string `"null"` with `text/csv; charset=utf-8` for `.csv` files, and `application/geo+json` for `.geojson` files.

Yes, done.

### DIFF
--- a/lib/util/blob.js
+++ b/lib/util/blob.js
@@ -70,7 +70,7 @@ async function blobResponse(s3, filename, blob) {
 
 function defaultMimetypeFor(filename) {
   if (!filename) return null;
-  switch (extname(filename)) {
+  switch (extname(filename).toLowerCase()) {
     case '.csv':     return 'text/csv'; // eslint-disable-line no-multi-spaces
     case '.geojson': return 'application/geo+json';
     default:         return null; // eslint-disable-line no-multi-spaces
--- a/lib/util/blob.js
+++ b/lib/util/blob.js
@@ -7,6 +7,7 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
+const { extname } = require('node:path');
 const { Transform } = require('stream');
 const { PartialPipe } = require('./stream');
 const { contentDisposition, redirect, withEtag } = require('./http');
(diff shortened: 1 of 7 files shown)


lognaturel:
This looks like what I'd expect. It will probably make sense to merge https://github.com/getodk/central-backend/pulls first.

lib/util/blob.js
```diff
@@ -67,4 +68,13 @@ async function blobResponse(s3, filename, blob) {
   }
 }
 
-module.exports = { blobContent, blobResponse, streamBlobs, streamEncBlobs };
+function defaultMimetypeFor(filename) {
+  if (!filename) return null;
+  switch (extname(filename).toLowerCase()) {
+    case '.csv':     return 'text/csv'; // eslint-disable-line no-multi-spaces
+    case '.geojson': return 'application/geo+json';
+    default:         return null; // eslint-disable-line no-multi-spaces
```
ktuite, line 76: When combined with the other PR to not allow null content types in the database, this part will still return null but then the DB will provide the default value, yes?

ktuite (approved):

lib/util/blob.js
alxndrsn: Now that https://github.com/getodk/central-backend/pull/1355 has been merged, this may no longer be necessary - I will confirm.

alxndrsn: Correct!

test/integration/api/submissions.js
```diff
@@ -4421,6 +4421,78 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff
                   text.toString().should.equal('testvideo'); // use 'text' instead of 'body' to avoid supertest response parsing
                 })))))));
 
+    [
+      // express ALWAYS adds "charset=..." suffix to text-based Content-Type response headers
+      // See: https://github.com/expressjs/express/issues/2654
+      [ 'CSV',     'myfile.csv',     'text/csv; charset=utf-8', 'a,b,c' ], // eslint-disable-line no-multi-spaces
+      [ 'GeoJSON', 'myfile.geojson', 'application/geo+json',    '{}' ], // eslint-disable-line no-multi-spaces
```
alxndrsn, line 4428: Add test for a custom file extension.

alxndrsn:
Docs updated from

Screenshot_2025-02-04_07-17-49

to

Screenshot_2025-02-04_07-16-50

PR #1357, opened 2025-01-13
backend: allow configuring S3_REGION

alxndrsn:
See: https://github.com/getodk/central/pull/870

A new, very simple e2e test has been added, and it's run multiple times with different config.

It's simple, and there are no obvious alternatives.

Allows users to configure regions.  Should not have impact if region is not specified.

Yes: https://github.com/getodk/docs/pull/1912


lognaturel:
https://github.com/getodk/central/issues/846#issuecomment-2610714967

PR #1371, opened 2025-01-16
db-migrations: prevent new uses of application code

alxndrsn:
Mixing application code into database migrations removes assurances that the same migration will mutate the same postgresql data identically when upgrading from/to different versions of the application.

* add exceptions to existing use of application code in db migrations
* discourage future use of application code in db migrations by adding an eslint rule to error for local includes outside `lib/model/migrations`


alxndrsn:
A low-impact example of this is shown at https://github.com/getodk/central-backend/blob/6324ebea70b98558a3ba3ebe7be4834893371958/test/integration/other/migrations.js#L57-L61

ktuite (approved):

PR #1389, opened 2025-02-07
bcrypt: reject passwords > 72 bytes

alxndrsn:

Manual testing - there are no existing tests for short passwords.

I think a better approach would be to move max-length validation into `lib/util/crypto.js`.

Should not affect existing users with very long passwords, but will prevent new passwords being saved with >72 bytes.

No - the docs do not currently mention any restrictions on password length.


lib/util/problem.js
```diff
@@ -140,6 +140,8 @@ const problems = {
 
     encodingNotSupported: problem(400.37, () => 'Encoding not supported.'),
 
+    passwordTooLong: problem(400.38, () => 'The password or passphrase provided exceeds the maximum length.'),
```
lognaturel, line 143: Are you thinking this is likely rare enough that we just leave a raw backend message (vs. introducing localizable frontend error messages)?

lognaturel (approved):
> I think a better approach would be to move max-length validation into lib/util/crypto.js.

But you're keeping it there for now because the short length validation is already there, right?

This looks good to me provided there are no surprises in the answers to the clarifying questions I've asked. 

@ktuite @sadiqkhoja FYI in case you want to take another quick look.

alxndrsn:
> The datatype of password column is varchar(64), isn't that sufficient?

We're not storing passwords in plaintext :grin: 

Also is there a reason for that column to be `VARCHAR` when bcrypt hashes have a fixed length of 60 characters?

lib/util/problem.js
alxndrsn: What's the distinction?  Should there be an accompanying PR for `odk-central-frontend`?

alxndrsn:
> But you're keeping it there for now because the short length validation is already there, right?

Yes.  Maybe they should both be moved into the `hashPassword()` function?  "Password too long" seems a more fundamental failure than "too short" though.

sadiqkhoja (approved):
Only thing I would like to add is the reason i.e. "bcrypt truncates input to 72 bytes before hashing" in the PR description or commit message for future reference.

PR #1390, opened 2025-02-08
Add cache headers

alxndrsn:
Closes https://github.com/getodk/central/issues/662

Added tests.

It seems simple.

Should prevent future caching surprises.

No?  "cache" is not currently mentioned in docs/

### DIFF
--- a/lib/http/service.js
+++ b/lib/http/service.js
@@ -58,6 +58,17 @@ module.exports = (container) => {
   const cookieParser = require('cookie-parser');
   service.use(cookieParser());
 
+  service.use((req, res, next) => {
+    // TODO only set for safe methods?
+    // TODO only set for authenticated requests?
+    if (req.path.startsWith('/sessions/')) {
+      res.set('Cache-Control', 'no-store');
+    } else {
+      res.set('Cache-Control', 'private, max-age=20');
+    }
+    next();
+  });
+
 
   ////////////////////////////////////////////////////////////////////////////////
   // PREPROCESSORS
--- a/test/e2e/standard/test.js
+++ b/test/e2e/standard/test.js
@@ -17,24 +17,194 @@ const serverUrl = 'http://localhost:8383';
 const userEmail = 'x@example.com';
 const userPassword = 'secret1234';
 
+describe('Cache headers', () => {
+  const undici = require('undici');
+
+  // REVIEW this could also be written in a JVM language to simulate Collect behaviour more closely.
+
+  const privateCacheDispatcher = new undici.Agent().compose(undici.interceptors.cache({
+    cacheByDefault: undefined, // do not cache responses without explicit expiration
+    methods: [ 'GET' ],
+    type: 'private',
+  }));
+  const sharedCacheDispatcher = new undici.Agent().compose(undici.interceptors.cache({
+    cacheByDefault: Number.MAX_SAFE_INTEGER, // aggressively cache everything
+    methods: [ 'GET' ],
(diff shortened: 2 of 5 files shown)


lib/http/service.js
```diff
@@ -58,6 +58,17 @@ module.exports = (container) => {
   const cookieParser = require('cookie-parser');
   service.use(cookieParser());
 
+  service.use((req, res, next) => {
+    // TODO only set for safe methods?
+    // TODO only set for authenticated requests?
```
lognaturel, line 63: Are these todos for a reviewer to consider? Something you're still thinking about?

alxndrsn: Good question.  I think I'd be happy deleting them both, but I'd be interested in a review of that.

alxndrsn: I've removed the TODOs - it seems reasonable to be more aggressive with cache-busting.

lognaturel (approved):
That looks good to me -- using the private cache directly for 20s keeps a client from totally hammering the server with requests without compromising freshness. Then we expect the request to happen and result in a 304 if there are no changes.

@sadiqkhoja in case you want to take a second look.

lib/http/service.js
```diff
@@ -58,6 +58,15 @@ module.exports = (container) => {
   const cookieParser = require('cookie-parser');
   service.use(cookieParser());
 
+  service.use((req, res, next) => {
+    if (req.path.startsWith('/sessions/')) {
+      res.set('Cache-Control', 'no-store');
+    } else {
+      res.set('Cache-Control', 'private, max-age=20');
```
sadiqkhoja, line 65: 20 seconds isn't too much?

And for data endpoints like submissions and entities `max-age` should be 0

test/e2e/standard/test.js
```diff
@@ -17,24 +17,192 @@ const serverUrl = 'http://localhost:8383';
 const userEmail = 'x@example.com';
 const userPassword = 'secret1234';
 
+describe('Cache headers', () => {
+  const undici = require('undici');
```
sadiqkhoja, line 21: nit: maybe move this to the top of the file with other requires.

lib/http/service.js
lognaturel: The idea is to have a low `max-age` for all dynamic content so that there is some local caching that comes into play for unusual patterns like immediately refreshing. In general, we don't expect immediate refreshes so this eases load on the server a bit without significantly compromising freshness. We could set it to 0 instead but I don't see a big advantage. Whichever we choose we should apply consistently, I think, since all content is equally dynamic.

alxndrsn: I think `max-age=0` is likely to provide fewer surprises.

Is immediate refreshing an issue?  I'd guess if it's being done, then it's because of an expectation of changed data...

lognaturel: Fair enough, that works for me!

test/e2e/standard/test.js
alxndrsn: I've included it here to keep it`undici` scoped to these specific tests.  If it's preferable, I can refactor these tests into a separate file, but I'd rather do that in a follow-up PR.

lognaturel:
Some tests will need to change to match

sadiqkhoja (approved):

test/e2e/standard/test.js
sadiqkhoja: scoped `undici` is fine.

PR #1408, opened 2025-02-17
api/sessions: document bcrypt implementation curiosity

### LINKED ISSUE #1409: Bcrypt implementation details are leaked through various APIs
Bcrypt's `ExpandKey` function causes a supplied plaintext to be repeated up to 72 chars long.

End users of central APIs can do this repetition themselves.

See: https://en.wikipedia.org/wiki/Bcrypt#Maximum_password_length

alxndrsn:
This behaviour was noted by @sadiqkhoja, and seemed at least worth documenting if not fixing.

Closes #1409

Added a test.

An alternative to this test would be to check for NUL characters in supplied passwords and reject them before passing the values to bcrypt.

No effect.

No.

### DIFF
--- a/test/integration/api/sessions.js
+++ b/test/integration/api/sessions.js
@@ -14,17 +14,24 @@ describe('api: /sessions', () => {
           body.should.be.a.Session();
         })));
 
-    // This demonstrates a strange feature of bcrypt - a valid password can be
+    // These demonstrate a strange feature of bcrypt - a valid password can be
     // repeated multiple times and still validate successfully.  An alternative
-    // to this test would be to check for NUL characters in supplied passwords
+    // to these tests would be to check for NUL characters in supplied passwords
     // and reject them before passing the values to bcrypt.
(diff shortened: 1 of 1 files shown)


test/integration/api/sessions.js
```diff
@@ -14,6 +14,18 @@ describe('api: /sessions', () => {
           body.should.be.a.Session();
         })));
 
+    // This demonstrates a strange feature of bcrypt - a valid password can be
+    // repeated multiple times and still validate successfully.  An alternative
+    // to this test would be to check for NUL characters in supplied passwords
+    // and reject them before passing the values to bcrypt.
+    it.only('should leak bcrypt implementation details', testService((service) =>
```
alxndrsn, line 21:
```suggestion
    it('should leak bcrypt implementation details', testService((service) =>
```

sadiqkhoja (approved):
We discussed this in the meeting and concluded that just documenting the behaviour is sufficient.

Issue #1403, opened 2025-02-12
Bug with publishing form with conflicting version


ktuite:
I was looking at something to do with form manifests containing entity lists, and in one of my tests, I got a really unexpected 500 error. 

Below is the test. The very last line of trying to fetch the attachment (which should download the entity list) fails ONLY if a previous request to `publish` is sent that results in a `409`. (This test might be longer than it needs to be for this issue... but I already pruned out some other unnecessary stuff.)

```js
    it.only('should return return correct version of attachment (dataset) in manifest', testService(async (service) => {
      const asAlice = await service.login('alice');

      // Upload form with slot for attachment goodone.csv
      await asAlice.post('/v1/projects/1/forms')
        .send(testData.forms.withAttachments)
        .set('Content-Type', 'application/xml')
        .expect(200);

      // Upload attachment csv (not a dataset)
      await asAlice.post('/v1/projects/1/forms/withAttachments/draft/attachments/goodone.csv')
        .send('test,csv\n1,2')
        .set('Content-Type', 'text/csv')
        .expect(200);

      // Publish form
      await asAlice.post('/v1/projects/1/forms/withAttachments/draft/publish')
        .expect(200);

      // ---- Create a new draft of the form with different attachment to be linked to dataset ----
      // Create dataset called people
      await asAlice.post('/v1/projects/1/datasets')
        .send({ name: 'people' })
        .expect(200);

      // Create new draft: wont autolink because .CSV doesn't match .csv and autolinker is very strict
      await asAlice.post('/v1/projects/1/forms/withAttachments/draft')
        .send(testData.forms.withAttachments
          .replace('goodone.csv', 'people.CSV'))
        .set('Content-Type', 'application/xml')
        .expect(200);

      // ----- Explicitly link attachment to dataset ----
      await asAlice.patch('/v1/projects/1/forms/withAttachments/draft/attachments/people.CSV')
        .send({ dataset: true })
        .expect(200);

      // I should be able to download the draft dataset attachment
      await asAlice.get('/v1/projects/1/forms/withAttachments/draft/attachments/people.CSV')
        .expect(200);

      // This should not be available until the form is published
      await asAlice.get('/v1/projects/1/forms/withAttachments/attachments/people.CSV')
        .expect(404);

      // Publish the form
      await asAlice.post('/v1/projects/1/forms/withAttachments/draft/publish')
        .expect(409);
      await asAlice.post('/v1/projects/1/forms/withAttachments/draft/publish?version=2')
[...]

PR #1414, opened 2025-02-19
stderr.write(): add newline suffix consistently

alxndrsn:
Noted while investigating #1403.

Some logging is not followed by a newline character, so it is concatenated with the next output to `stderr`.

This commit adds newline characters to calls to `process.stderr.write()` where missing, and standardises their inclusion.

Run locally; tests in CI.

Introduced in f3e85630a3985fbcb03cd84cfbe1384a1f39017d, it's unclear why `process.stderr.write()` is used in preference to `console.error()`.

No change for users, unless they're parsing stderr for e.g. monitoring.  This change would probably make parsing more reliable.

I don't think stderr output is currently documented.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -109,8 +109,7 @@ const init = (config) => {
       },
       captureException(err) {
         process.stderr.write('attempted to log Sentry exception in development:\n');
-        process.stderr.write(inspect(err));
-        process.stderr.write('\n');
+        process.stderr.write(inspect(err) + '\n');
       },
       configureScope: noop
     };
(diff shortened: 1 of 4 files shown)


matthew-white:
> I don't think stderr output is currently documented.

Yes, I think that's right. :+1:

matthew-white (approved):

matthew-white:
> it's unclear why `process.stderr.write()` is used in preference to `console.error()`.

I remember asking about this during code review several years ago. issa thought that the two did different things. Sorry, I don't remember anything more specific. 😅 Maybe the two aren't different anymore? `console.error()` does seem nicer.

PR #1418, opened 2025-02-26
Features C#906: Remove # from the frontend URL

sadiqkhoja:
Closes getodk/central#906

All tests are passing.

I have just replaced all `/#/` with `/` and made the necessary changes in `safeNextPathFrom`.

None

API Doc has been updated as well.


lib/util/html.js
```diff
@@ -15,24 +15,22 @@ const envDomain = config.get('default.env.domain');
...
-    return '/#/';
+    return '/';
   }
 
   if (url.origin !== envDomain || url.pathname === '/login')
-    return '/#/';
+    return '/';
 
   // Don't modify enketo URLs
   if (url.pathname.startsWith('/-/')) return url.toString();
```
sadiqkhoja, line 31: Don't know why we have to return full URL for Enketo including origin

alxndrsn: `url.toString()` seems simpler than `url.pathname + url.search + url.hash`

I suspect the specific reason was per the comment on L16:

```js
// Logic adapted from `login.vue` in frontend.
```

In frontend (https://github.com/getodk/central-frontend/blob/eb602bd563d41f560fa59dbb1ccac624b977df92/src/components/account/login.vue#L162-L163):

```js
      if (url.pathname.startsWith(`${enketoBasePath}/`))
        return external(url.href);
```

Why `.toString()` is used in preference to `.href` is unclear, but they appear to be equivalent per the spec (discussion at https://stackoverflow.com/a/54380688).

alxndrsn:
> **How does this change affect users?**
>
> None

Could this break users' bookmarks if they've linked directly to content in an odk-central-frontend instance?

lib/util/html.js
matthew-white: In that file in Frontend, `external()` is basically just `window.location.replace()`. Maybe `url.pathname + url.search + url.hash` would have worked as well, but by that point in Frontend, it's already been validated that the URL is at the correct origin. We do use `url.pathname + url.search + url.hash` on the next line, in the default `return`:

```js
return internal(url.pathname + url.search + url.hash);
```

That's needed because `internal()` is hitting Vue Router, which just wants a path, not the full URL.

matthew-white: Another thing I noticed is that if you try to navigate to an Enketo URL that requires auth, it redirects you to login in Frontend. It sets the `next` query parameter equal to the full Enketo URL. Whereas when Frontend sets the `next` query parameter for a Frontend page that requires auth, the value of the query parameter is just the path, not the full URL.

matthew-white: The only place I see `safeNextPathFrom()` being used is to generate a `nextPath` to pass to `loaderTemplate`. Looking at `loaderTemplate`, I don't think that a distinction needs to be made between Enketo paths and Frontend paths. It's using `nextPath` in pretty generic ways. Maybe it'd be simpler for `safeNextPathFrom()` to do the same thing for the two types of paths: the previous logic had to account for the hash in Frontend URLs, but not anymore. No strong feelings though.

matthew-white:
I mentioned this on Slack, but I think getodk/central-frontend#1141 is intended to prevent bookmarks from breaking. If a user navigates to a path with a hash, then Frontend will redirect them to the new path without a hash.

matthew-white (approved):
Looks good as far as I can see. Backend usually shouldn't care what Frontend's paths are. In the few places where Backend was doing something specific with hashes before, it can now do without the hashes.

PR #1420, opened 2025-03-06
Return lastUpdate for the dataset endpoints

sadiqkhoja:
Closes getodk/central#803

Updated tests

Follows the existing pattern

No

Done

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -361,7 +361,7 @@ const _get = extender(Dataset)(Dataset.Extended)((fields, extend, options, publi
   SELECT ${fields} FROM datasets
   ${extend|| sql`
   LEFT JOIN (
-    SELECT "datasetId", COUNT(1) "entities", MAX(COALESCE("updatedAt", "createdAt")) "lastEntity", COUNT(1) FILTER (WHERE conflict IS NOT NULL) "conflicts"
+    SELECT "datasetId", COUNT(1) "entities", MAX("createdAt") "lastEntity", MAX("updatedAt") "lastUpdate", COUNT(1) FILTER (WHERE conflict IS NOT NULL) "conflicts"
     FROM entities e
     WHERE "deletedAt" IS NULL
     GROUP BY "datasetId"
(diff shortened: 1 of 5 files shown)


lib/model/query/datasets.js
```diff
@@ -361,7 +361,7 @@ const _get = extender(Dataset)(Dataset.Extended)((fields, extend, options, publi
   SELECT ${fields} FROM datasets
   ${extend|| sql`
   LEFT JOIN (
-    SELECT "datasetId", COUNT(1) "entities", MAX(COALESCE("updatedAt", "createdAt")) "lastEntity", COUNT(1) FILTER (WHERE conflict IS NOT NULL) "conflicts"
+    SELECT "datasetId", COUNT(1) "entities", MAX("createdAt") "lastEntity", MAX("updatedAt") "lastUpdate", COUNT(1) FILTER (WHERE conflict IS NOT NULL) "conflicts"
```
ktuite, line 364: Should `lastUpdate` represent this - the last `updatedAt` timestamp of any entity currently in the dataset, or should it match the etag timestamp? For viewing in a hovercard, the last update of an entity might be ok, since having to another query against the audit log might slow down the display of the hovercard a tiny bit, but it won't update when an entity is deleted or when a property is added.

```
// Used in the openRosa form attachment manifest
const getLastUpdateTimestamp = (datasetId) => ({ maybeOne }) =>
  maybeOne(sql`
  SELECT "loggedAt" FROM audits
  JOIN datasets ON audits."acteeId" = datasets."acteeId"
  WHERE datasets."id"=${datasetId}
  ORDER BY audits."loggedAt" DESC LIMIT 1`)
    .then((t) => t.orNull())
    .then((t) => (t ? t.loggedAt : null));
```

sadiqkhoja: Updated the code to use `getLastUpdateTimestamp` instead after the slack chat with ln. 

Returning `lastUpdate` only for the `GET /datasets/:name` endpoint.

ktuite (approved):

PR #1421, opened 2025-03-06
Add development.json to the ignore list

sadiqkhoja:
I keep a `config/development.json` file locally where I put my custom configurations. For now it has just a custom `env.domain`. This PR just adds that file to git ignore so that I don't accidentally commit this file.

Also I have set `NODE_ENV` to `test` in mocharc so that `development.json` is not read during test executions. From [node-config](https://github.com/node-config/node-config/wiki/Configuration-Files#default-node_env) docs: "If NODE_ENV is not set in the environment, a default value of development is used."

All tests continue to pass. and configuration from `development.json` is read when I run `make dev` or `make run`.

I am open to other options but this one achieves the desire outcome for me with minimum changes.

No

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ node_modules/
 /.repl-history
 .vscode
 /junit-reports/
+config/development.json
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,4 @@ node_modules/
 /.repl-history
 .vscode
 /junit-reports/
-config/development.json
+/config/development.json
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -7,8 +7,6 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
-process.env.NODE_ENV = 'test';
-
 module.exports = {
   ignore: [
     'test/db-migrations/**',
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -7,6 +7,8 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
+process.env.NODE_ENV = 'test';
+
 module.exports = {
   ignore: [
     'test/db-migrations/**',
--- a/Makefile
+++ b/Makefile
@@ -117,9 +117,8 @@ test-db-migrations:
 test-fast: node_version
 	NODE_CONFIG_ENV=test BCRYPT=insecure npx mocha --recursive --fgrep @slow --invert
 
-NODE_CONFIG_ENV ?= test
-
 .PHONY: test-integration
+test-integration: NODE_CONFIG_ENV ?= test
 test-integration: node_version
 	NODE_CONFIG_ENV=$(NODE_CONFIG_ENV) BCRYPT=insecure npx mocha --recursive test/integration
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 default: base
 
+NODE_CONFIG_ENV ?= test
+
 node_modules: package.json
 	npm install
 	touch node_modules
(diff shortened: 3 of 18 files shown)


alxndrsn:
> Also I have set `NODE_ENV` to `test` in mocharc so that `development.json` is not read during test executions. From [node-config](https://github.com/node-config/node-config/wiki/Configuration-Files#default-node_env) docs: "If NODE_ENV is not set in the environment, a default value of development is used."

I previously explored use of a test-specific env with @matthew-white at https://github.com/getodk/central-backend/pull/990, but there wasn't support for it then.  I would still be happy to see that change instead of including test config in `config/default.json` and at https://github.com/getodk/central-backend/blob/bf49b89d1b22d2cc6989f2c4eba7a9c739d347e9/test/integration/setup.js#L75

.gitignore
```diff
@@ -12,3 +12,4 @@ node_modules/
 /.repl-history
 .vscode
 /junit-reports/
+config/development.json
```
alxndrsn, line 15: I think correct path here would be `/config/development.json`

alxndrsn: If the intention is to use `config/test.json`, then could `process.env.NODE_CONFIG_ENV = 'test'` be used instead?

Could this setting be moved to `Makefile` to avoid affecting other mocha tests (e.g. e2e tests) which already use their own `NODE_CONFIG_ENV`?

.mocharc.js
```diff
@@ -7,6 +7,8 @@
 // including this file, may be copied, modified, propagated, or distributed
 // except according to the terms contained in the LICENSE file.
 
+process.env.NODE_ENV = 'test';
```
alxndrsn, line 10: If the intention is to use `config/test.json`, then could `process.env.NODE_CONFIG_ENV = 'test'` be used instead?

Could this setting be moved to `Makefile` to avoid affecting other mocha tests (e.g. e2e tests) which already use their own `NODE_CONFIG_ENV`?

.gitignore
sadiqkhoja: done

.mocharc.js
alxndrsn: > Intention was to stop tests to use `development.json` to read config.

:+1: then I'd encourage using [`NODE_CONFIG_ENV`][1], as it's less likely to have side effects than changing `NODE_ENV`.

[1]: https://github.com/getodk/central-backend/pull/1421#discussion_r1990089643

Makefile
```diff
@@ -115,19 +115,21 @@ test-db-migrations:
 
 .PHONY: test-fast
 test-fast: node_version
-	BCRYPT=insecure npx mocha --recursive --fgrep @slow --invert
+	NODE_CONFIG_ENV=test BCRYPT=insecure npx mocha --recursive --fgrep @slow --invert
+
+NODE_CONFIG_ENV ?= test
```
alxndrsn, line 120: I think this defines the variable for the whole `Makefile`, so it might be more appropriate to define it at the top of the file.

.mocharc.js
alxndrsn: (I see you've switched to `NODE_CONFIG_ENV` :+1:)

Makefile
alxndrsn: Alternatively I think you can define a variable for a specific target like this:

```Makefile
.PHONY: test-integration
test-integration: NODE_CONFIG_ENV ?= test
test-integration: node_version
	NODE_CONFIG_ENV=$(NODE_CONFIG_ENV) BCRYPT=insecure npx mocha --recursive test/integration
```

test/integration/other/entities-purging.js
```diff
@@ -180,13 +180,13 @@ describe('query module entities purge', () => {
...
+      { description: 'when entityUuid specified without projectId and datasetName',
         projectId: false, datasetName: false, entityUuid: true, expectedError: PROVIDE_ALL },
-      { description: ' when entityUuid specified without projectId',
+      { description: 'when entityUuid specified without projectId',
         projectId: false, datasetName: true, entityUuid: true, expectedError: PROVIDE_ALL },
-      { description: ' when entityUuid specified without datasetName',
+      { description: 'when entityUuid specified without datasetName',
         projectId: true, datasetName: false, entityUuid: true, expectedError: PROVIDE_ALL },
-      { description: ' when datasetName specified without projectId',
+      { description: 'when datasetName specified without projectId',
```
alxndrsn, line 189: This looks like it's crept in from [elsewhere.](https://github.com/getodk/central-backend/pull/1438)

sadiqkhoja: merged 'master' again.

alxndrsn (approved):

PR #1425, opened 2025-03-10
UserPreferences: simplify INSERT queries

alxndrsn:
* simplify conflict handling
  * remove index name generation
  * remove re-stating of property value
* remove unused return value

Current tests continue to pass.

Could have been left as-is, but the new code should be easier to reason about.

This should not affect users.  Risk of regression is low given the size of the change and existing test coverage.

No.


matthew-white (approved):
Definitely a nice improvement in readability. 🎉 I didn't know about the `EXCLUDED` syntax, that's very handy.

PR #1426, opened 2025-03-10
user-preferences: simplify SELECT queries

alxndrsn:

Existing tests continue to pass.

The code could have been left as-is, but after taking the time to understand the existing code it seemed helpful to share that understanding back.  Building JSON objects in PostgreSQL is verbose and maybe unfamiliar, whereas the same operations in Javascript can be simpler.

This change should not affect users.

No.

### DIFF
--- a/lib/model/query/user-preferences.js
+++ b/lib/model/query/user-preferences.js
@@ -9,56 +9,22 @@
 
 const { sql } = require('slonik');
 
+const getForUser = (userId) => async ({ allFirst }) => {
+  const site = Object.fromEntries(await allFirst(sql`
+    SELECT JSONB_BUILD_ARRAY("propertyName", "propertyValue")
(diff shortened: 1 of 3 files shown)


matthew-white (approved):

lib/model/query/user-preferences.js
```diff
@@ -10,55 +10,22 @@
...
+  `));
+
+  const projects = Object.fromEntries(await allFirst(sql`
+    SELECT JSONB_BUILD_ARRAY("projectId", JSONB_OBJECT_AGG("propertyName", "propertyValue"))
+      FROM user_project_preferences
+      WHERE "userId" = ${userId}
+      GROUP BY "projectId"
+  `));
 
+  return { site, projects };
```
alxndrsn, line 27: Consider combining these into a single query.

PR #1427, opened 2025-03-10
UserPreferences: simplify DELETE queries

alxndrsn:
Also use existing return value from postgres wire protocol instead of generating an extra value.

Ran existing tests.

Seems more concise and obvious than the existing code.

Should not affect users.

No.

### DIFF
--- a/lib/model/query/user-preferences.js
+++ b/lib/model/query/user-preferences.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { sql } = require('slonik');
-
+const Option = require('../../util/option');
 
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Definitely seems like an improvement in readability to me. :+1:

matthew-white:
<blockquote>

* simplify conflict handling
  * remove index name generation
</blockquote>

I'm not sure I saw this part, or maybe I don't understand what this refers to. I like the changes here though, I think we're good to merge.

alxndrsn:
> I'm not sure I saw this part

Sorry, this comment relates to another PR and was copied here by mistake.  I'll update the description.

PR #1429, opened 2025-03-10
Remove symlinks to assertions.js

alxndrsn:
For easier understanding of what's going on with the assertions.js file, where it's included from, and what context it's used in.

Also these paths to not seem to be referenced from anywhere.

Tests continue to pass.

No other options were considered.

No change - this is test code.

No.


alxndrsn:
Maybe come back to this one day.

PR #1431, opened 2025-03-10
Standardise responses for invalid user email

### LINKED ISSUE #1430: "500: Unexpected value expression" with unexpected email type
```sh
$ curl localhost:8383/v1/sessions -H 'content-type: application/json' --data '{"password":"helo", "email":{}}'
{"message":"Internal Server Error"}
```

Related:

* https://github.com/getodk/central-backend/issues/1406
* https://github.com/getodk/central-backend/issues/1407
* https://github.com/getodk/central-backend/pull/1411
* https://github.com/getodk/central-backend/pull/1410

alxndrsn:
Closes #1430

* existing tests
* new tests

Validation of the `email` type could be performed closer to the API.  In that case, validations should be added in more places.

Should reduce number of 500 errors.  Users are unlikely to be relying on this status code.

500 errors should not be documented, so no documentation update is expected.

### DIFF
--- a/lib/model/query/users.js
+++ b/lib/model/query/users.js
@@ -8,11 +8,12 @@
 // except according to the terms contained in the LICENSE file.
 
 const { sql } = require('slonik');
-const { map } = require('ramda');
+const { compose, is, map, not } = require('ramda');
 const { Actor, User } = require('../frames');
 const { hashPassword } = require('../../util/crypto');
 const { unjoiner, page, sqlEquals, QueryOptions } = require('../../util/db');
-const { rejectUnless } = require('../../util/promise');
+const Problem = require('../../util/problem');
+const { rejectIf, resolve } = require('../../util/promise');
 
 const create = (user) => ({ Actors }) => Actors.createSubtype(user);
 create.audit = (user) => (log) => log('user.create', user.actor, { data: user });
@@ -74,7 +75,7 @@ ${page(options)}`;
 const getAll = (options) => ({ all }) => all(_getSql(options)).then(map(_unjoin));
 
 const getByEmail = (email, options = QueryOptions.none) => ({ maybeOne }) =>
-  rejectUnless('string')({ email })
+  resolve(rejectIf(compose(not, is(String)), () => Problem.user.invalidDataTypeOfParameter({ field: 'email', expected: 'string' }))(email))
     .then(() => maybeOne(_getSql(options.withCondition({ email }))).then(map(_unjoin)));
 
 const getByActorId = (actorId, options = QueryOptions.none) => ({ maybeOne }) =>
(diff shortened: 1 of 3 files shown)


sadiqkhoja (approved):

lib/model/query/users.js
```diff
@@ -73,7 +74,8 @@ ${page(options)}`;
 const getAll = (options) => ({ all }) => all(_getSql(options)).then(map(_unjoin));
 
 const getByEmail = (email, options = QueryOptions.none) => ({ maybeOne }) =>
-  maybeOne(_getSql(options.withCondition({ email }))).then(map(_unjoin));
+  rejectUnless('string')({ email })
+    .then(() => maybeOne(_getSql(options.withCondition({ email }))).then(map(_unjoin)));
```
alxndrsn, line 78: @matthew-white does this seem idiomatic?

matthew-white: Are you wondering about the new `rejectUnless()` function? It looks similar to other functions in lib/util/promise.js, so it seems reasonable to me.

Though that said, some of those functions actually feel slightly awkward to me when used outside a promise chain. When issa first introduced those functions, we almost never used `async`/`await`. But now we tend to use `async`/`await` where possible/convenient, and in that context, these promise utility functions end up becoming double function calls like what you have here with `rejectUnless()`. Those aren't my personal favorite, though I also don't think it's a big deal.

One thing I've thought about is that we could modify these functions so that they could work as either a single call or a double call, e.g., `rejectUnless('string', { email })` would be equivalent to `rejectUnless('string')({ email })`. I think the Ramda `curry()` function would let us do that pretty easily. `rejectUnless('string')` would be very convenient in a promise chain, when we want a function to pass to `.then()`. While in the context of `async`/`await`, I think `rejectUnless('string', { email })` is a little nicer.

Anyway, that's mostly just my own thoughts / not something we've discussed a lot. But I did want to mention the larger context of transitioning to more `async`/`await`. I'm pretty sure we all prefer `async`/`await`, but the codebase has a history of a lot of `.then()`.

alxndrsn: > some of those functions actually feel slightly awkward to me when used outside a promise chain

I agree, especially this new `rejectUnless()()`.  I also think the name is too generic for such a specific validation function.  I've inlined and rewritten the line in a way that looks in-keeping with the existing promise-chain style:

```diff
-  rejectUnless('string')({ email })
+  resolve(rejectIf(compose(not, is(String)), () => Problem.user.invalidDataTypeOfParameter({ field: 'email', expected: 'string' }))(email))
```

I think the function could be rewritten with considerable readability improvement as:

```js
const getByEmail = (email, options = QueryOptions.none) => ({ maybeOne }) => {
  assertString('email', email);
  return maybeOne(_getSql(options.withCondition({ email }))).then(map(_unjoin));
};
```

matthew-white: I agree, I definitely find the second version more readable/preferable.

PR #1441, opened 2025-03-12
slonik: add regression test

alxndrsn:
Checks for regression of https://github.com/gajus/slonik/commit/96db9f08faefb0e7da0cc0d817d9e8fae361e590

* added new tests
* ran against previous `slonik` version and noted both new tests fail (https://github.com/alxndrsn/odk-central-backend/actions/runs/13810839468).

Could have tested at the API level, but this way has less dependence on other code, so should be less likely to break for unrelated reasons.

No effect - just tests.

No.


sadiqkhoja:
This is making me nervous, should we write test to assert the behavior of dependencies? Specially when those tests are relying on implementation details, what if slonik decides to rename `sqlToken` or move it from `src/tokens` directory?

lognaturel (approved):
I agree with @sadiqkhoja that in general we don't want to be in the business of testing dependencies but in this case we consider this a source of risk so it's best if we're alerted if anything at all changes in this area.

PR #1447, opened 2025-03-14
Sentry: enable request performance tracing

alxndrsn:
Closes https://github.com/getodk/central/issues/877

~Should this be using the `config` package & `default.external.sentry`?  Similar question for https://github.com/getodk/central-backend/pull/637 and getodk/central-backend#601.~

Tested locally with dev config.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -120,7 +120,7 @@ const init = (config) => {
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
     environment: process.env.SENTRY_ENVIRONMENT || 'production',
-    tracesSampleRate: Number(process.env.SENTRY_TRACE_RATE || 0),
+    tracesSampleRate: Number(config.traceRate || 0),
     beforeSend(event, hint) {
       const sanitizedEvent = sanitizeEventRequest(event);
 
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -120,6 +120,7 @@ const init = (config) => {
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
     environment: process.env.SENTRY_ENVIRONMENT || 'production',
+    tracesSampleRate: Number(process.env.SENTRY_TRACE_RATE || 0),
     beforeSend(event, hint) {
       const sanitizedEvent = sanitizeEventRequest(event);
 
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -103,15 +103,12 @@ const init = (config) => {
   if ((config == null) || isBlank(config.key) || isBlank(config.project) || isBlank(config.orgSubdomain)) {
     // return a noop object that returns the hooks but does nothing.
     return {
-      Handlers: {
-        requestHandler() { return (request, response, next) => next(); },
-        errorHandler() { return (error, request, response, next) => next(error); }
-      },
+      getCurrentScope: () => ({ setExtra: noop }),
+      setupExpressErrorHandler: noop,
       captureException(err) {
         process.stderr.write('attempted to log Sentry exception in development:\n');
         process.stderr.write(inspect(err) + '\n');
       },
-      configureScope: noop
     };
   }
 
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "node": "22.12.0"
   },
   "dependencies": {
-    "@sentry/node": "~7",
+    "@sentry/node": "~9",
```
sadiqkhoja, line 12: This is probably not so important but I'd like to mention that latest Sentry [docs](https://docs.sentry.io/platforms/javascript/guides/express/install/commonjs/) says that we should initialize sentry SDK before requiring any other module to ensure errors from those module are captured as well. I am not sure if this recommendation came in version 8 or 9 or it was always there.

sadiqkhoja (approved):
Few minor comments/questions but this can be merged as it is as well.

lib/external/sentry.js
```diff
@@ -120,6 +117,7 @@ const init = (config) => {
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
     environment: process.env.SENTRY_ENVIRONMENT || 'production',
+    tracesSampleRate: Number(config.traceRate || 0),
```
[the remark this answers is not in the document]
alxndrsn: If it's not falsy, then behaviour is undefined :grin: 

Actually it gets set to `undefined`: https://github.com/getsentry/sentry-javascript/blob/ef2f35d5863ce5e581a36aac9ce4557a9ebe370c/packages/node/src/sdk/index.ts#L244-L256

I think there's a few places where config validation is lax.  Defining some kind of schema and validating config at startup might be sensible.  I've filed this at https://github.com/getodk/central-backend/issues/1464

lib/http/service.js
```diff
@@ -19,13 +19,8 @@ module.exports = (container) => {
...
   // PRERESOURCE MIDDLEWARE
 
-  // apply the Sentry request hook.
-  service.use(container.Sentry.Handlers.requestHandler());
-
   service.use((req, res, next) => {
-    container.Sentry.configureScope(scope => {
-      scope.setExtra('startTimestamp', Date.now());
-    });
+    container.Sentry.getCurrentScope().setExtra('startTimestamp', Date.now());
```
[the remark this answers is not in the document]
alxndrsn: Due to sampling rates, I don't think every error is guaranteed to be accompanied by a trace.

Maybe worth reviewing this anyway; has `startTimestamp` proved to be useful?

package.json
alxndrsn: Great catch, thanks!  I'll review these recommendations.

alxndrsn: It looks like the change was introduced in Sentry v8:

> To migrate your Express app to Sentry Node SDK `8.x`, you need to make the following changes:
>
> 1. Create an external file called `instrument.js` and move your `Sentry.init` call to it.
>
> Due to the way that auto instrumentation works in `8.x` of the Sentry Node SDK, it is required that you initialize Sentry before you require or import any other package. Any package that is required/imported before Sentry is initialized may not be correctly auto-instrumented.

https://github.com/getsentry/sentry-docs/blob/09e515bbdf1fc4d2372c4cf5c764cf18620ebf22/platform-includes/migration/javascript-v8/important-changes/javascript.express.mdx#L45

alxndrsn: It looks like v7 docs are here: https://github.com/getsentry/sentry-docs/blob/37831f9bb64b217355ef52aeefd9ff6722e0466c/docs/platforms/javascript/guides/express/index.mdx

> Sentry should be initialized as early in your app as possible:

alxndrsn: I've updated the PR to initialise Sentry as early as possible, and added documentation of why this is happening.

lib/http/service.js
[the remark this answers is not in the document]
sadiqkhoja: oh right.

lib/external/sentry.js
```diff
@@ -113,7 +125,7 @@ const init = (config) => {
   }
 
   // otherwise initialize Sentry with some settings we want.
-  const Sentry = require('@sentry/node');
+  const Sentry = require('@sentry/node'); // eslint-disable-line no-shadow
```
sadiqkhoja, line 133: nit: this shadowing can be avoid.

```diff
- 113: const Sentry = (() => {
+ 113: const _sentry = (() => { 

...

- 160: module.exports = { Sentry, sanitizeEventRequest, isSensitiveEndpoint, filterTokenFromUrl };
+ 160: module.exports = { Sentry: _sentry, sanitizeEventRequest, isSensitiveEndpoint, filterTokenFromUrl };
```

sadiqkhoja (approved):

PR #1385, opened 2025-02-04
fix: Blobs._getOnePending() can Error on multiple results

alxndrsn:
Closes #1373

#### What has been done to verify that this works as intended?

Run tests 400 times:

* https://github.com/alxndrsn/odk-central-backend/actions/runs/13131375879/attempts/1
* https://github.com/alxndrsn/odk-central-backend/actions/runs/13131375879/attempts/2
* https://github.com/alxndrsn/odk-central-backend/actions/runs/13131375879/attempts/3
* https://github.com/alxndrsn/odk-central-backend/actions/runs/13131375879/attempts/4
* https://github.com/alxndrsn/odk-central-backend/actions/runs/13131375879/attempts/5

#### Why is this the best possible solution? Were any other approaches considered?

It seems simple... however, there are currently no tests which will immediately detect a regression, or similar issues in future.  Any suggestions?

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Should fix the bug for users.

#### Does this change require updates to the API documentation? If so, please update docs/api.yaml as part of this PR.

No.

#### Before submitting this PR, please make sure you have:

- [x] run `make test` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1450, opened 2025-03-14
Blobs: use = instead of IN() when expecting 1 result

### LINKED ISSUE #1373: Blobs._getOnePending() can Error on multiple results
https://github.com/getodk/central-backend/actions/runs/12803489636/job/35696251865?pr=1372

```
  1708 passing (3m)
  18 pending
  1 failing

  1) managed encryption
       end-to-end @slow
         should decrypt client audit log attachments from s3:
     DataIntegrityError: Query returns an unexpected result.
      at Object.maybeOne (node_modules/slonik/dist/src/connectionMethods/maybeOne.js:22:15)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async /home/runner/work/central-backend/central-backend/lib/model/query/blobs.js:155:23
      at async uploadBlobIfAvailable (lib/model/query/blobs.js:154:15)
      at async /home/runner/work/central-backend/central-backend/lib/model/query/blobs.js:185:12
      at async Object.createConnection (node_modules/slonik/dist/src/factories/createConnection.js:97:18)



make[1]: *** [Makefile:122: test-integration] Error 1
```

Thrown by: https://github.com/getodk/central-backend/blob/6324ebea70b98558a3ba3ebe7be4834893371958/lib/model/query/blobs.js#L83-L95

alxndrsn:
* IN()
  * is prone to returning more than 1 result
  * has a more complicated query plan
* equals
  * ensures that SELECT query returns max 1 result
  * has a simpler query plan

Closes #1373

# Query plans

With a simplified table:

```sql
CREATE TABLE t (id SERIAL PRIMARY KEY, x TEXT);
```

## `IN()`

```
# EXPLAIN UPDATE t SET x='b' WHERE id IN (SELECT id FROM t WHERE x='a' LIMIT 1);
                                      QUERY PLAN                                       
---------------------------------------------------------------------------------------
 Update on t  (cost=0.31..8.35 rows=0 width=0)
   ->  Nested Loop  (cost=0.31..8.35 rows=1 width=66)
         ->  HashAggregate  (cost=0.03..0.04 rows=1 width=32)
               Group Key: "ANY_subquery".id
               ->  Subquery Scan on "ANY_subquery"  (cost=0.00..0.03 rows=1 width=32)
                     ->  Limit  (cost=0.00..0.02 rows=1 width=4)
                           ->  Seq Scan on t t_1  (cost=0.00..18.00 rows=1040 width=4)
                                 Filter: (x = 'a'::text)
         ->  Index Scan using t_pkey on t  (cost=0.28..8.29 rows=1 width=10)
               Index Cond: (id = "ANY_subquery".id)
(10 rows)
```

Screenshot_2025-03-14_16-07-03

## `=`

```
# EXPLAIN UPDATE t SET x='b' WHERE id = (SELECT id FROM t WHERE x='a' LIMIT 1);
                              QUERY PLAN                               
-----------------------------------------------------------------------
 Update on t  (cost=0.29..8.31 rows=0 width=0)
   InitPlan 1 (returns $0)
     ->  Limit  (cost=0.00..0.02 rows=1 width=4)
           ->  Seq Scan on t t_1  (cost=0.00..18.00 rows=1040 width=4)
                 Filter: (x = 'a'::text)
   ->  Index Scan using t_pkey on t  (cost=0.28..8.29 rows=1 width=38)
         Index Cond: (id = $0)
(7 rows)
```

Screenshot_2025-03-14_16-07-14

Ran both sets of code a lot.

Also considered using a CTE (https://github.com/getodk/central-backend/pull/1385), but while it worked in practice, the approach in this PR seems more correct.

Should fix surprising non-uploads to s3.  Users are unlikely to have noticed this.

No.


alxndrsn:
Note that using `=` with more than 1 result will give the following error:

```
# UPDATE t SET x='b' WHERE id = (SELECT id FROM t WHERE x='a' LIMIT 2);
ERROR:  more than one row returned by a subquery used as an expression
```

ktuite (approved):

PR #1456, opened 2025-03-21
Change dataset-diff form verb from update to read

ktuite:
Backend change of issue https://github.com/getodk/central/issues/897

The verbs on `/projects/:projectId/forms/:id/dataset-diff` have changed to `form.read` and `dataset.list` so that project viewers are able to get the list of datasets related to a published form.

The verbs on `/projects/:projectId/forms/:id/draft/dataset-diff` are still the same (`form.update` and `dataset.list`) and I think this makes sense because project viewers can't see draft forms so they don't need to see this kind of information.

### DIFF
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -168,17 +168,7 @@ module.exports = (service, endpoint) => {
   // Entity/Dataset-specific endpoint that is used to show how publishing
   // a form will change any datasets mentioned in the form.
   // Even though there are dataset-related permissions, we will use a combination of dataset access and
-  // form modification verbs for authorization. Though form.update and dataset.list will likely be part
-  // of the same role, we want to use the combined authorization strength. One scenario where they are
-  // different is if a user has been assigned a higher auth role on a specific form but not a project,
-  // which can be done through the API: projects/.../forms/.../assignments
-  //
-  // e.g.
-  // Someone can list all datasets but can't read forms so they shouldn't be able to get info about the form
-  // through this endpoint.
-  // Or they can modify forms but not access datasets so they shouldn't be able to get info about the datasets
-  // through this endpoint.
-  //
+  // form verbs for authorization.
   // We are not considering a separate 'dataset.update' verb right now because it seems weird to mix the
   // ability/inability to alter a dataset with the ability to modify forms -- they should just be the same.
   service.get('/projects/:projectId/forms/:id/draft/dataset-diff', endpoint(({ Forms, Datasets, Projects }, { params, auth }) =>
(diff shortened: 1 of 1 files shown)


lib/resources/forms.js
```diff
@@ -213,7 +213,7 @@ module.exports = (service, endpoint) => {
         ]))
       .then(([form, project]) =>
         Promise.all([
-          auth.canOrReject('form.update', form),
+          auth.canOrReject('form.read', form),
```
matthew-white, line 216: Now that this endpoint no longer checks `form.update`, I think it needs to check `canReadForm()`. Otherwise, a Data Collector could read the dataset diff of a closed form. A Data Collector shouldn't have any access to a closed form.

From a quick look, it looks to me like almost all the GET endpoints for an individual form check `canReadForm()`. The dataset diff endpoint didn't need to before, because it checked `form.update`, but now it should.

matthew-white:
> The verbs on `/projects/:projectId/forms/:id/draft/dataset-diff` are still the same (`form.update` and `dataset.list`) and I think this makes sense because project viewers can't see draft forms so they don't need to see this kind of information.

I see what you're saying, but I think it'd be confusing for the two endpoints to do two different sets of checks. If the published dataset diff is checking `form.read` + `canReadForm()`, but the draft dataset diff is checking `form.update` but not `canReadForm()`, I think it'd be hard for our future selves to understand what the intention is. My sense is that we should select the verbs we check based on what the endpoint is doing. The endpoint for the draft dataset diff isn't updating anything — it's a read-only endpoint — so unless there's a special reason I'm not thinking of, I think it only needs to check `form.read`, not `form.update`.

ktuite:
Update this comment

> Even though there are dataset-related permissions, we will use a combination of dataset access and
  // form modification verbs for authorization.

lib/resources/forms.js
ktuite: Discussed with Matt - the `dataset.list` verb will prevent data collectors from looking at this endpoint, but switching to `canReadForm` is still reasonable to be consistent with the other endpoints.


These are the verbs for data collectors and project viewers: 
``` 
                                                                     verbs                                                                      |      name      | id
------------------------------------------------------------------------------------------------------------------------------------------------+----------------+----
 ["project.read", "form.list", "form.read", "submission.read", "submission.list", "dataset.list", "entity.list", "dataset.read", "entity.read"] | Project Viewer |  6
 ["project.read", "open_form.list", "open_form.read", "submission.create"]                                                                      | Data Collector |  8
```

matthew-white (approved):

PR #1462, opened 2025-03-25
Fix 500 error for /v1/enketo-ids/:enketoId/form when multiple drafts are published

sadiqkhoja:
Closes getodk/central#930

### Changes:

* wrongly assumed that only one form_def will have publishedAt, changed the condition in the query to use currentDefId instead.
* Also added condition to check enketoOnceId to fetch the Form.

Added tests

Minor modification in the query resolves the issue.

No

NA


lib/model/query/forms.js
```diff
@@ -574,7 +574,15 @@ order by coalesce(form_defs.name, forms."xmlFormId") asc`)
 // helper function to gate how form defs are joined to forms in _get
 /* eslint-disable indent */
 const versionJoinCondition = (version, enketoId) => (
-  (enketoId) ? sql`form_defs."formId"=forms.id AND (form_defs."enketoId" = ${enketoId} OR (forms."enketoId" = ${enketoId} AND form_defs."publishedAt" IS NOT NULL))` :
+  (enketoId) ? sql`
+    form_defs."formId"=forms.id 
+    AND (
+      form_defs."enketoId" = ${enketoId}
```
matthew-white, line 580: I'm wondering about unused/orphaned form defs, the kind that get deleted by `Forms.clearUnneededDrafts()`. For example, let's say you have a draft form def, you create a test submission to that def, but then you upload a new draft form def. I'm not totally sure, but I think the old def will stick around in the database until the submission to it is purged in 30 days (related: getodk/central#749). I also think that the new def and the old def will have the same `enketoId`.

So I'm thinking that we should add here:

```sql
form_defs."enketoId" = ${enketoId}
AND form_defs.id = forms."draftDefId"
```

If you're matching on the form def's `enketoId` and not the form's, then it must be a draft def. In that case, we want to ensure that it's not an orphaned def we're matching on.

matthew-white (approved):

PR #1471, opened 2025-04-03
Return form attachment after POST, not { success: true }

matthew-white:
This PR makes a change to the POST endpoint for uploading a file for a form attachment. Right now, that endpoint returns `{ success: true }`. However, Frontend would benefit from receiving the updated form attachment (the metadata about the form attachment, not the blob). This PR makes a change along those lines, returning the updated form attachment rather than `{ success: true }`.

Even today, Frontend would benefit from receiving the updated form attachment. After a successful request to the POST endpoint, Frontend wants to update the `updatedAt` timestamp that's shown in the form attachments table. Since it doesn't know the true `updatedAt` timestamp, it just uses the current timestamp after the response is received, which should be pretty close to the true timestamp. You can see the code for that [here](https://github.com/getodk/central-frontend/blob/edb862ce6d9a1860fb34e3dd0dcec596882a0694/src/components/form-attachment/list.vue#L270-L272).

The real motivation for this PR is getodk/central#728. Specifically, this PR is needed for the Frontend PR getodk/central-frontend#1172. After a file is uploaded in Frontend, we want to adjust the count of changed attachments that's shown in the tag for the Attachments section. We do that by comparing the old and new hashes for the attachment. Backend knows the hash of the uploaded file, and by returning the updated form attachment, it can pass that information back to Frontend.

Frontend could calculate the new hash itself, but it doesn't seem like the easiest thing to do in the browser, especially without reading the entire file into memory. Backend knows the new hash, and it's easy enough for Backend to return it to Frontend.

We're a little inconsistent in our POST endpoints in what we return. I think usually we just return `{ success: true }`. However, sometimes we return the new resource, e.g., for POST /v1/projects/:id/forms. This POST endpoint for uploading a file is interesting in that in some ways, it's more so updating an existing form attachment rather than creating a new resource. In that way, the endpoint is sort of similar to a PATCH endpoint. For PATCH endpoints, I think we usually return the updated resource. In other words, even though we're changing the response that's returned from the endpoint, I think we're still being consistent with how other endpoints work (both POST endpoints and PATCH).

Updated existing tests.

The schema of the response has changed. However, the response before was just `{ success: true }`: there wasn't any concrete/useful information returned. I doubt that anyone is using that specific response. However, I've updated the API docs to mention this change to the API.

### DIFF
--- a/lib/model/query/form-attachments.js
+++ b/lib/model/query/form-attachments.js
@@ -16,7 +16,7 @@ const { map, mergeRight } = require('ramda');
 const { expectedFormAttachments } = require('../../data/schema');
 const { Frame, into } = require('../frame');
 const { Blob, Form } = require('../frames');
-const { unjoiner, insertMany } = require('../../util/db');
+const { QueryOptions, unjoiner, insertMany, sqlEquals } = require('../../util/db');
 const { ignoringResult, resolve } = require('../../util/promise');
 const { construct } = require('../../util/util');
 const Option = require('../../util/option');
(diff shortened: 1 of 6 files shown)


lib/model/query/form-attachments.js
matthew-white, line 1: While I'm in here, update `getAllByFormDefIdForOpenRosa()` to call `getAllByFormDefId()`. Those are two separate queries that used to be different, but now they're almost identical.

matthew-white (commented):
Notes from interactive review

lib/model/query/form-attachments.js
```diff
@@ -117,7 +117,7 @@ const createVersion = (xml, form, savedDef, itemsets, publish = false) => ({ Blo
 
 
 ////////////////////////////////////////////////////////////////////////////////
-// CRUD
+// UPDATING
```
matthew-white, line 120: The file as a whole does CRUD, but as of right now, this particular section only contains the `update()` method.

lib/model/query/form-attachments.js
matthew-white: > now they're almost identical.

The only difference I could spot in the two SQL queries is that `getAllByFormDefId()` had an ORDER clause, while `getAllByFormDefIdForOpenRosa()` did not.

ktuite (approved):
Looks great!

PR #1472, opened 2025-04-04
Modify restore endpoint

sadiqkhoja:

Modified tests, added a test and verify manual with Central frontend.

Overall reason for this change was discussed internally. Code change is simple and fulfills the requirement.

No

NA

### DIFF
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -115,17 +115,16 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
     // actually try to authenticate with it. no Problem on failure. short circuit
     // out if we have a GET or HEAD request.
     const maybeSession = authBySessionToken(token);
-    if ((context.method === 'GET') || (context.method === 'HEAD')) return maybeSession;
+    if (context.method !== 'POST') return maybeSession;
 
     // if non-GET run authentication as usual but we'll have to check CSRF afterwards.
     return maybeSession.then((cxt) => { // we have to use cxt rather than context for the linter
       // if authentication failed anyway, just do nothing.
       if ((cxt == null) || cxt.auth.session.isEmpty()) return;
 
-      // Neither csrf nor x-requested-with provided for POST request, so reject it.
-      const csrf = urlDecode(cxt.body?.__csrf);
-      if (context.method === 'POST' &&
-          (csrf.isEmpty() || isBlank(csrf.get()) || (cxt.auth.session.get().csrf !== csrf.get())) &&
+      // Neither csrf nor x-requested-with provided, so reject it.
+      const csrf = urlDecode(cxt.body.__csrf);
+      if ((csrf.isEmpty() || isBlank(csrf.get()) || (cxt.auth.session.get().csrf !== csrf.get())) &&
            context.headers['x-requested-with'] !== 'XMLHttpRequest') {
         return reject(Problem.user.authenticationFailed());
       }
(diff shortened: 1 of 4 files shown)


lib/http/preprocessors.js
```diff
@@ -122,8 +122,11 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
       // if authentication failed anyway, just do nothing.
       if ((cxt == null) || cxt.auth.session.isEmpty()) return;
 
-      const csrf = urlDecode(cxt.body.__csrf);
-      if (csrf.isEmpty() || isBlank(csrf.get()) || (cxt.auth.session.get().csrf !== csrf.get())) {
+      // Neither csrf nor x-requested-with provided for POST request, so reject it.
+      const csrf = urlDecode(cxt.body?.__csrf);
+      if (context.method === 'POST' &&
```
matthew-white, line 127: We already check `context.method` on line 118 above. What do you think about modifying that line instead of also checking `context.method` here?

test/integration/api/sessions.js
```diff
@@ -339,6 +341,15 @@ describe('api: /sessions', () => {
           .set('X-Forwarded-Proto', 'https')
           .set('Cookie', 'session=' + body.token)
           .expect(200))));
+
+    it('should succeed it is ajax/xhr request', testService((service) =>
+      service.authenticateUser('alice', 'includeCsrf')
```
matthew-white, line 346: I see that the existing test right above does the same thing, but I don't totally understand passing the string `'includeCsrf'` to `authenticateUser()` instead of `true`. It looks to me like `authenticateUser()` is just checking whether that argument is truthy: it's not using it as a string.

lib/http/preprocessors.js
```diff
@@ -115,15 +115,17 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
     // actually try to authenticate with it. no Problem on failure. short circuit
     // out if we have a GET or HEAD request.
     const maybeSession = authBySessionToken(token);
-    if ((context.method === 'GET') || (context.method === 'HEAD')) return maybeSession;
+    if (context.method !== 'POST') return maybeSession;
 
     // if non-GET run authentication as usual but we'll have to check CSRF afterwards.
```
matthew-white, line 120: This comment too I think.

lib/http/preprocessors.js
```diff
@@ -111,19 +111,28 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
...
+    }
+
+    // if POST run authentication as usual but we'll have to check CSRF afterwards.
     return maybeSession.then((cxt) => { // we have to use cxt rather than context for the linter
       // if authentication failed anyway, just do nothing.
       if ((cxt == null) || cxt.auth.session.isEmpty()) return;
 
       const csrf = urlDecode(cxt.body.__csrf);
-      if (csrf.isEmpty() || isBlank(csrf.get()) || (cxt.auth.session.get().csrf !== csrf.get())) {
+      if ((csrf.isEmpty() || isBlank(csrf.get()) || (cxt.auth.session.get().csrf !== csrf.get()))) {
```
matthew-white, line 135: It looks like there's an extra set of parentheses.

test/integration/api/sessions.js
```diff
@@ -339,6 +341,15 @@ describe('api: /sessions', () => {
           .set('X-Forwarded-Proto', 'https')
           .set('Cookie', 'session=' + body.token)
           .expect(200))));
+
+    it('should succeed it is ajax/xhr request', testService((service) =>
```
matthew-white, line 345:
```suggestion
    it('should succeed if it is ajax/xhr request', testService((service) =>
```

matthew-white (approved):

PR #1475, opened 2025-04-08
purge: s3: delete uploaded blobs in batches

### LINKED ISSUE #1443: Blob purging is horribly slow
From prod, this query has terrible performance on some deployments:

```sql
DELETE FROM blobs
  WHERE id IN (
    SELECT b.id
      FROM blobs AS b
      LEFT JOIN client_audits          AS ca ON ca."blobId"    = b.id
      LEFT JOIN submission_attachments AS sa ON sa."blobId"    = b.id
      LEFT JOIN form_attachments       AS fa ON fa."blobId"    = b.id
      LEFT JOIN form_defs              AS fd ON fd."xlsBlobId" = b.id
    WHERE ca."blobId"    IS NULL
      AND sa."blobId"    IS NULL
      AND fa."blobId"    IS NULL
      AND fd."xlsBlobId" IS NULL
      AND b.s3_status = 'uploaded'
    LIMIT 1
)
  RETURNING blobs.id, blobs.sha
```

Call stack:

* `Blobs._purgeOneUnattachedUploaded()`
* `Blobs.purgeUnattached()`
* `purgeTask()`
* `lib/bin/purge.js`
* `purge.sh`

# Queries

* what does the query plan look like on servers where this is slow?
  * A. it is much faster without the `LIMIT` clause (~100x)
  * A. it has many materialization steps
  * A. there are seq scans for all the joins, so perhaps indexes are missing on `blobId` columns?
  * A. perhaps index is missing on `s3_status` column?
* how often are blobs de-duplicated across different types (e.g. client audits vs submission attachments vs form attachments vs form definitions)?  It seems unlikely that there is significant space saved by merging all of these

alxndrsn:
This significantly improves performance (about 3x locally).

Closes #1443

* did a lot of benchmarking and staring at query plans
* added a test
* realised that to make significant improvements over the current approach, this is the simplest option

Note that the current test doesn't even rely on any data being in the joined tables

A few other options, e.g.

* adding indexes.  This PR is much less invasive, and shouldn't have side effects
* changing the `unattachedClause` clause
  * to avoid `NULL` checks.  This looks like it could potentially 100x performance (or 30x over this PR), but would probably need a placeholder row in the `Blobs` table (e.g. w/ id = -1), and then add NOT NULL constraints to all `blobId` FK columns[1]
  * in other ways, none of which had significant improvements.  The nicest alternative syntax was:

```sql
WHERE NOT EXISTS (SELECT 1 FROM client_audits AS ca WHERE ca."blobId" = b.id)
  AND NOT EXISTS (SELECT 1 FROM submission_attachments AS sa WHERE sa."blobId" = b.id)
  AND NOT EXISTS (SELECT 1 FROM form_attachments AS fa WHERE fa."blobId" = b.id)
  AND NOT EXISTS (SELECT 1 FROM form_defs AS fd WHERE fd."xlsBlobId" = b.id)
```

### Avoiding `NULL` checks [1]

```sql
WHERE b.id NOT IN (SELECT "blobId" FROM client_audits)
  AND b.id NOT IN (SELECT "blobId" FROM submission_attachments)
  AND b.id NOT IN (SELECT "blobId" FROM form_attachments)
  AND b.id NOT IN (SELECT "xlsBlobId" FROM form_defs)
```

Hopefully significantly better performance in production systems.  Hard to know without testing on realistic datasets.

No.

### DIFF
--- a/lib/model/query/blobs.js
+++ b/lib/model/query/blobs.js
@@ -95,10 +95,14 @@ const _getOnePending = () => ({ maybeOne }) => maybeOne(sql`
 `).then(map(construct(Blob)));
 
 const unattachedClause = sql`
-  WHERE b.id NOT IN (SELECT "blobId"    FROM client_audits)
-    AND b.id NOT IN (SELECT "blobId"    FROM submission_attachments WHERE "blobId"    IS NOT NULL)
-    AND b.id NOT IN (SELECT "blobId"    FROM form_attachments       WHERE "blobId"    IS NOT NULL)
-    AND b.id NOT IN (SELECT "xlsBlobId" FROM form_defs              WHERE "xlsBlobId" IS NOT NULL)
+  LEFT JOIN client_audits          AS ca ON ca."blobId" = b.id
+  LEFT JOIN submission_attachments AS sa ON sa."blobId" = b.id
+  LEFT JOIN form_attachments       AS fa ON fa."blobId" = b.id
+  LEFT JOIN form_defs              AS fd ON fd."xlsBlobId" = b.id
+  WHERE ca."blobId" IS NULL
+    AND sa."blobId" IS NULL
+    AND fa."blobId" IS NULL
+    AND fd."xlsBlobId" IS NULL
 `;
 
 const _purgeAllUnattached = () => ({ all }) => all(sql`
--- a/lib/model/query/blobs.js
+++ b/lib/model/query/blobs.js
@@ -116,14 +116,14 @@ const _purgeUnattachedNotUploaded = () => ({ all }) => all(sql`
       AND blobs.id = b.id
 `);
 
-const _purgeOneUnattachedUploaded = () => ({ maybeOne }) => maybeOne(sql`
+const _purgeUnattachedUploadedBatch = () => ({ all }) => all(sql`
   DELETE FROM blobs
-    WHERE id = (
+    WHERE id IN (
       SELECT b.id
         FROM blobs AS b
         ${unattachedClause}
           AND b.s3_status = 'uploaded'
-        LIMIT 1
+        LIMIT 100
     )
     RETURNING blobs.id, blobs.sha
 `);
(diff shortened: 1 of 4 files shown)


lib/model/query/blobs.js
```diff
@@ -95,14 +95,10 @@ const _getOnePending = () => ({ maybeOne }) => maybeOne(sql`
...
-  LEFT JOIN form_attachments       AS fa ON fa."blobId" = b.id
-  LEFT JOIN form_defs              AS fd ON fd."xlsBlobId" = b.id
-  WHERE ca."blobId" IS NULL
-    AND sa."blobId" IS NULL
-    AND fa."blobId" IS NULL
-    AND fd."xlsBlobId" IS NULL
+  WHERE b.id NOT IN (SELECT "blobId"    FROM client_audits)
+    AND b.id NOT IN (SELECT "blobId"    FROM submission_attachments WHERE "blobId"    IS NOT NULL)
+    AND b.id NOT IN (SELECT "blobId"    FROM form_attachments       WHERE "blobId"    IS NOT NULL)
+    AND b.id NOT IN (SELECT "xlsBlobId" FROM form_defs              WHERE "xlsBlobId" IS NOT NULL)
```
alxndrsn, line 101: this may not need to get perf improvements of batching; confirm

alxndrsn: Reverted.

lib/model/query/blobs.js
```diff
@@ -120,14 +120,14 @@ const _purgeUnattachedNotUploaded = () => ({ all }) => all(sql`
       AND blobs.id = b.id
 `);
 
-const _purgeOneUnattachedUploaded = () => ({ maybeOne }) => maybeOne(sql`
+const _purgeUnattachedUploadedBatch = () => ({ all }) => all(sql`
```
ktuite, line 123: Is this batch purging supposed to only be for this part where the blobs are uploaded to S3? There was some earlier version of this PR where the unattached clause was modified and it affected all the blob purging so I'm not sure if batching those other queries should be part of this PR or not.

alxndrsn: As per PR title, this is just batching the delete of uploaded blobs.

ktuite (approved):

PR #1476, opened 2025-04-14
Include user, change-reason in client_audits export

ktuite:
Closes https://github.com/getodk/central/issues/654

As discussed in the issue, the values for `user` and `change-reason` were luckily captured in the jsonb `remainder` column of the `client_audits` table. 

This PR migrates/copies the values from that column to two new columns in the `client_audits` table and adds `user` and `change-reason` as expected client audit headers to be included in the audit export.

I've added a new test for the behavior, created a database migration test, and tried it with some real data.

Straightforward solution based on discussion in issue.

Exposes `user` and `change-reason` in consolidated client audits export when these parameters are set in the xlsform: `identify-user=true track-changes-reasons=on-form-edit` 

I'm not sure client audits are mentioned in the API docs.

### DIFF
--- a/lib/model/migrations/20250415-01-client-audit-remainder.js
+++ b/lib/model/migrations/20250415-01-client-audit-remainder.js
@@ -1,0 +1,34 @@
+// Copyright 2025 ODK Central Developers
+// See the NOTICE file at the top-level directory of this distribution and at
+// https://github.com/getodk/central-backend/blob/master/NOTICE.
+// This file is part of ODK Central. It is subject to the license terms in
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const up = async (db) => {
+  await db.raw(`
+    ALTER TABLE client_audits
+      ADD COLUMN "user" TEXT,
+      ADD COLUMN "change-reason" TEXT
+  `);
+
+  await db.raw(`
+    UPDATE client_audits
+    SET 
+      "user" = remainder->>'user',
+      "change-reason" = remainder->>'change-reason'
+    WHERE remainder IS NOT NULL;
+  `);
+};
+
+const down = async (db) => {
+  await db.raw(`
+    ALTER TABLE client_audits
+      DROP COLUMN "user",
+      DROP COLUMN "change-reason"
+  `);
+};
+
+module.exports = { up, down };
--- a/lib/model/migrations/20250415-01-client-audit-remainder.js
+++ b/lib/model/migrations/20250415-01-client-audit-remainder.js
@@ -26,6 +26,21 @@ const up = async (db) => {
 
 const down = async (db) => {
   await db.raw(`
+    UPDATE client_audits
+    SET
+      remainder = jsonb_set(
+        jsonb_set(
+          COALESCE(remainder, '{}'),
+          '{user}',
+          to_jsonb("user")
+        ),
+        '{change-reason}',
+        to_jsonb("change-reason")
+      )
+    WHERE "user" IS NOT NULL OR "change-reason" IS NOT NULL
+  `);
+
+  await db.raw(`
     ALTER TABLE client_audits
       DROP COLUMN "user",
       DROP COLUMN "change-reason"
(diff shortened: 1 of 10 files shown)


lib/model/migrations/20250415-01-client-audit-remainder.js
```diff
@@ -0,0 +1,34 @@
...
+      ADD COLUMN "change-reason" TEXT
+  `);
+
+  await db.raw(`
+    UPDATE client_audits
+    SET 
+      "user" = remainder->>'user',
+      "change-reason" = remainder->>'change-reason'
+    WHERE remainder IS NOT NULL;
+  `);
```
matthew-white, line 23: Once we move `user` and `change-reason` into their own columns, how about removing them from `remainder`? We know the `client_audits` table can get very large, so it might help to keep `remainder` as small as possible.

lib/model/migrations/20250415-01-client-audit-remainder.js
matthew-white, line 1: Do you think we should test this migration? We could use the new migration testing framework.

lib/model/migrations/20250415-01-client-audit-remainder.js
ktuite: We could do this, but i'm weirdly concerned about the down migration and putting the user/change-reason column data back into remainder? I could just not have that in the down migration since it will probably never be run in a real environment?

lib/model/migrations/20250415-01-client-audit-remainder.js
ktuite: There is now a test!

lib/model/migrations/20250415-01-client-audit-remainder.js
matthew-white: Maybe it wouldn't be hard to add that to the down migration? We could update `remainder` again as part of the down migration. But I agree that it's not essential. We often don't write a down migration at all. If someone is having to run a down migration in production, something has already gone wrong.

test/db-migrations/20250415-01-client-audit-remainder.spec.js
```diff
@@ -0,0 +1,82 @@
...
+      )
+      returning "blobId"
+    `);
+
+    await runMigrationBeingTested();
+  });
+
+  it('should add columns to client audit table', async () => {
+    await assertTableSchema('client_audits',
+      { column_name: 'blobId',        is_nullable: 'NO', data_type: 'integer' }, // eslint-disable-line no-multi-spaces
```
matthew-white, line 58: If we're going to have to disable `no-multi-spaces` a lot for migration testing, maybe we should go ahead and turn off the rule in test/db-migrations/.eslintrc.js?

test/db-migrations/20250415-01-client-audit-remainder.spec.js
```diff
@@ -0,0 +1,82 @@
...
+} = require('./utils'); // eslint-disable-line object-curly-newline
+
+describeMigration('20250415-01-client-audit-remainder', ({ runMigrationBeingTested }) => {
+  const aBlobWith = props => {
+    const randomContent = randomBytes(100);
+    const md5 = hash('md5',  randomContent); // eslint-disable-line no-multi-spaces
+    const sha = hash('sha1', randomContent);
+    return { md5, sha, ...props };
+  };
+  const aBlob = () => aBlobWith({});
```
matthew-white, line 16: Since these functions are also used elsewhere, how about moving them into a utility file? I'm sure we'll need them again in future migrations.

test/db-migrations/20250415-01-client-audit-remainder.spec.js
```diff
@@ -0,0 +1,82 @@
...
+    await db.oneFirst(sql`
+      INSERT INTO client_audits (
+      "blobId",
+      remainder
+      )
+      VALUES (
+      ${blobId},
+      '{"user": "test-user", "change-reason": "test-reason", "unknown-value": "test-unknown"}'::jsonb
+      )
+      returning "blobId"
```
matthew-white, line 50: Capitalize `returning`? Or if it's not used, maybe it can just be removed?

matthew-white (approved):

lib/model/migrations/20250415-01-client-audit-remainder.js
ktuite: Added to the down migration and tested locally!

PR #1478, opened 2025-04-20
introduce anonymous endpoint

sadiqkhoja:


lib/http/service.js
```diff
@@ -63,6 +63,7 @@ module.exports = (container) => {
   const { builder } = require('./endpoint');
   const { emptyAuthInjector, authHandler, queryOptionsHandler, userAgentHandler } = require('./preprocessors');
   const endpoint = builder(container, [ emptyAuthInjector, authHandler, queryOptionsHandler, userAgentHandler ]);
+  const anonymousEndpoint = builder(container, [ emptyAuthInjector, queryOptionsHandler, userAgentHandler ]);
```
matthew-white, line 66: I'm a little worried about someone updating the list for `endpoint` but not `anonymousEndpoint`. What do you think about code like what's below? Alternatively, a code comment would probably suffice.

```js
const preprocessors = [ emptyAuthInjector, authHandler, queryOptionsHandler, userAgentHandler ];
const endpoint = builder(container, preprocessors);
// without() from Ramda
const anonymousEndpoint = builder(container, without([authHandler], preprocessors));
```

test/integration/api/sessions.js
```diff
@@ -14,6 +14,32 @@ describe('api: /sessions', () => {
...
+          .set(t.header, t.value)
+          .set('x-forwarded-proto', 'https')
+          .send({ email: 'chelsea@getodk.org', password: 'password4chelsea' })
+          .expect(200)
+          .then(({ body }) => {
+            body.should.be.a.Session();
+          })));
+    });
+
+    it('should return a new session even valid cookie is passed', testService((service) =>
```
matthew-white, line 33:
```suggestion
    it('should return a new session even if valid cookie is passed', testService((service) =>
```

matthew-white (approved):

PR #1479, opened 2025-04-22
util: reduce confusion around without() fn

alxndrsn:
* rename util fn `without()` to `omit()`

This should reduce confusion between Ramda's `without()` function, for modifying arrays, and this repo's `omit()` function, which is a re-implementation of Ramda's `omit()` function, but with specific guarantees.

See:

* https://ramdajs.com/docs/#omit
* https://ramdajs.com/docs/#without

Existing test suite passes.

This should prevent confusion between:

* `without()` imported from `ramda`
* `without()` imported from `./lib/util/util`
* `omit()` imported from `ramda`

No effect.

No.

### DIFF
--- a/lib/formats/odata.js
+++ b/lib/formats/odata.js
@@ -19,7 +19,7 @@ const { parse, render } = require('mustache');
 const Problem = require('../util/problem');
 const { isTrue, urlPathname } = require('../util/http');
 const { PartialPipe } = require('../util/stream');
-const { sanitizeOdataIdentifier, without } = require('../util/util');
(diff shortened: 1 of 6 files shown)


matthew-white (approved):

matthew-white:
Part of me thinks that it's a little confusing to name the function the same thing as the Ramda function. But I can't really think of an alternative that doesn't feel verbose (e.g., `withoutKeys`). And I guess at the end of the day, this function is supposed to work similarly to the Ramda function. I definitely think `omit` is better than `without` given that there's a Ramda `without` function that behaves differently.

PR #1480, opened 2025-04-23
Don't return actorId from /v1/sessions/restore

matthew-white:
This is a follow-up PR to #1472. I noticed that /v1/sessions/restore was returning `actorId` in addition to `expiresAt` and `createdAt`. `actorId` isn't a readable field on the `Session` frame, and I don't think we want to return it from this endpoint.

I updated a test.

### DIFF
--- a/test/assertions.js
+++ b/test/assertions.js
@@ -143,10 +143,11 @@ should.Assertion.add('ExtendedSubmissionDef', function() {
 should.Assertion.add('Session', function() {
   this.params = { operator: 'to be a Session' };
 
-  Object.keys(this.obj).should.containDeep([ 'expiresAt', 'createdAt', 'token' ]);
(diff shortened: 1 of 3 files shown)


test/assertions.js
```diff
@@ -143,10 +143,11 @@ should.Assertion.add('ExtendedSubmissionDef', function() {
 should.Assertion.add('Session', function() {
   this.params = { operator: 'to be a Session' };
 
-  Object.keys(this.obj).should.containDeep([ 'expiresAt', 'createdAt', 'token' ]);
+  Object.keys(this.obj).should.eqlInAnyOrder([ 'expiresAt', 'createdAt', 'token', 'csrf' ]);
```
matthew-white, line 146: This change isn't needed for this PR, but it felt like a related change. I think it's nice to check that `Object.keys(this.obj)` doesn't include any properties outside this list.

sadiqkhoja (approved):

PR #1485, opened 2025-04-29
purge: surface errors to CLI

### LINKED ISSUE #1484: lib/bin/purge: returns status 0 on error
# Observed

* exit code is 0 on error
* when there is a non-problem error, string `null` is printed as the only output

# Expected

* non-zero exit code for any error (problem or not)
* non-problems should be printed in full

# Cause

Due to code at:

https://github.com/getodk/central-backend/blob/90859b3173c05c6ea101ded5211eb6b75d4069cc/lib/task/purge.js#L38-L40

alxndrsn:
Previously:

* exit code 0 would be returned when there was an error
* if the error was not a `Problem`, no details of the error would be shown

Closes #1484

Noticed while working on #1482.  In combination with that PR, this PR will allow Sentry capture of purge errors.

Compare at https://github.com/getodk/central-backend/pull/1485/files?w=1 for a clearer view of significant changes.

Ran locally with forced error.

Why are errors currently being swallowed?

Users who have included purge as part of larger scripts may be surprised (or grateful) that exit code of failed purges has changed.

I could not find documentation of the purge CLI:

```
docs (master)
$ git grep purge
docs/_static/api-spec/central.yaml:      - Submissions that have been soft-deleted for 30 days will automatically be purged. 
docs/_static/api-spec/central.yaml:    ODK Central v1.4 enables additional CSV export options and creates an API-manageable 30 day permanent purge system for deleted Forms. Previously, deleted Forms were made inaccessible but the data was not purged from the database.
docs/_static/api-spec/central.yaml:      * `POST /projects/…/forms/…/restore` to restore a Form that hasn't yet been permanently purged.
docs/_static/api-spec/central.yaml:    * `form.purge` when a Form is permanently purged.
docs/_static/api-spec/central.yaml:    * `submission.purge` when soft-deleted Submissions are purged.
docs/_static/api-spec/central.yaml:        not purge its Forms. (We will change that in a future release.)
docs/_static/api-spec/central.yaml:        all of its resources and submissions will be automatically purged. If your
docs/_static/api-spec/central.yaml:        Deleted forms can now be restored (as long as they have been in the Trash less than 30 days and have not been purged). However, a deleted Form with the same `xmlFormId` as an active Form cannot be restored while that other Form is active. 
docs/_static/api-spec/central.yaml:        are automatically purged. Purging deleted Submissions can also be triggered from the command line.
docs/_static/api-spec/central.yaml:      description: Submissions that have been recently soft-deleted and not yet purged can be restored using this endpoint.
docs/_static/api-spec/central.yaml:        In the future, we will provide a way to restore deleted entities and purge
docs/aggregate-app-engine-legacy.rst:  In this mode, you periodically purge older data collection records and never Push data up to Aggregate, as that would restore the purged data.
```


sadiqkhoja:
purge.js is executed via crontab, will there be any implication of this change in cron?

alxndrsn:
> purge.js is executed via crontab, will there be any implication of this change in cron?

Great question.  I'm not very familiar, but I think normally the result of a cron job is emailed to the sysadmin.  This behaviour doesn't change for success/error responses.

So the change would be that failure emails contain more detail, especially for non Problem errors.

sadiqkhoja (approved):

Issue #851, opened 2023-04-27
Returning 401 status without WWW-Authenticate header


alxndrsn:
[Backend is returning HTTP status 401 in a few places](https://github.com/getodk/central-backend/blob/d8181acab5dd000eac6d60d0df976f1e91701147/lib/util/problem.js#L114-L119), and I can't see any inclusion of a `WWW-Authenticate` header.

> The server generating a 401 response MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate) header field
> -- _https://www.rfc-editor.org/rfc/rfc9110#status.401_

alxndrsn:
# Resolution

## Simple

A simple option would be to include `WWW-Authenticate: None` with these responses.

## More complex

The 3 different `Problem` might warrant different changes:

https://github.com/getodk/central-backend/blob/d8181acab5dd000eac6d60d0df976f1e91701147/lib/util/problem.js#L114-L119

* `Problem.authenticationFailed`: used for both authentication and authorisation failures in various places.  It could be renamed `authFailure`, and return `403`.
* `Problem.httpsOnly`: returned when basic auth is used on a non-https connection.  Returning 400, 403, or 302 might be appropriate.
* `Problem.openRosaAuthenticationRequired`: the code looks like it should check if the authenticated user has the `app_user` role, and if not return an authorisation failure: https://github.com/getodk/central-backend/blob/d8181acab5dd000eac6d60d0df976f1e91701147/lib/resources/forms.js#L45

matthew-white:
> `Problem.authenticationFailed`: used for both authentication and authorisation failures in various places.

Where do you see it used for authorization? I see one instance here:

https://github.com/getodk/central-backend/blob/d8181acab5dd000eac6d60d0df976f1e91701147/lib/resources/submissions.js#L44

I don't quite remember what's going on here. I think Collect (or maybe Enketo?) displays an informative error message for a 401 but not a 403. @lognaturel would know more! It looks like this was introduced in 7b13eb7ef10cb964f17d16e0f2045b60c7bf3ea0.

> `Problem.httpsOnly`: returned when basic auth is used on a non-https connection.  Returning 400, 403, or 302 might be appropriate.

Just my two cents, I vote for a 400!

> `Problem.openRosaAuthenticationRequired`: the code looks like it should check if the authenticated user has the `app_user` role, and if not return an authorisation failure:

[...]

PR #1488, opened 2025-05-02
Return 401 if authentication is not provided

sadiqkhoja:
Prerequisite for getodk/central#1001

All tests are passing.

Some tests are updated to reflect the correct requirement.

It is necessary to distinguish between 401 and 403, previously we were relying on resources to perform authorization even when authentication failed.

This PR shouldn't be merged until corresponding frontend changes are ready. There are some places where frontend code expects 403, that needs to be updated.

No


test/integration/api/assignments.js
```diff
@@ -499,7 +507,7 @@ describe('/projects/:id/assignments', () => {
     it('should assign the actor by role numeric id', testService((service) =>
       service.login('bob', (asBob) =>
         Promise.all([
-          service.get('/v1/roles/manager').expect(200).then(({ body }) => body.id),
+          asBob.get('/v1/roles/manager').expect(200).then(({ body }) => body.id),
```
alxndrsn, line 510: This endpoint does not currently require authentication.  How does that expectation changing relate to 401 vs 403 errors?

lib/http/preprocessors.js
```diff
@@ -128,9 +125,6 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
 
     // if POST run authentication as usual but we'll have to check CSRF afterwards.
     return maybeSession.then((cxt) => { // we have to use cxt rather than context for the linter
-      // if authentication failed anyway, just do nothing.
-      if ((cxt == null) || cxt.auth.session.isEmpty()) return;
```
alxndrsn, line 132: Where is an "empty" session now handled?

lib/http/service.js
```diff
@@ -64,9 +63,9 @@ module.exports = (container) => {
   const { builder } = require('./endpoint');
   const { emptyAuthInjector, authHandler, queryOptionsHandler, userAgentHandler } = require('./preprocessors');
 
-  const preprocessors = [ emptyAuthInjector, authHandler, queryOptionsHandler, userAgentHandler ];
-  const endpoint = builder(container, preprocessors);
-  const anonymousEndpoint = builder(container, without([authHandler], preprocessors));
+  const preprocessors = [ queryOptionsHandler, userAgentHandler ];
```
alxndrsn, line 66:
```suggestion
  const commonPreprocessors = [ queryOptionsHandler, userAgentHandler ];
```

lib/resources/submissions.js
```diff
@@ -65,12 +65,14 @@ const findMultipart = (files) => {
...
 const forceAuthFailed = Problem.translate(Problem.user.insufficientRights, Problem.user.authenticationFailed);
 
-module.exports = (service, endpoint) => {
+module.exports = (service, endpoint, anonymousEndpoint) => {
 
   ////////////////////////////////////////////////////////////////////////////////
   // SUBMISSIONS (OPENROSA)
 
   const openRosaSubmission = (path, draft, getForm) => {
+    const endpointByTestPath = path.startsWith('/test/:key') ? anonymousEndpoint : endpoint;
```
alxndrsn, line 74: Would this code be more intuitive if the `endpoint` implementation was passed in rather than being derived from `path`?

e.g.

```js
const openRosaSubmission = (endpoint, path, draft, getForm) => {
```

or


```js
const openRosaSubmission = (path, { isDraft, requiresAuth }, getForm) => {
```

lib/resources/submissions.js
```diff
@@ -81,9 +83,10 @@ module.exports = (service, endpoint) => {
     // there is no body content, and the HTTP spec requires that HEAD should
     // return exactly what GET would, just without a response body, the only thing
     // we can possibly respond with in either case is no body and a 204 code.
-    service.get(path, endpoint.openRosa(({ Projects }, { params }) =>
+    service.get(path, endpointByTestPath.openRosa(({ Projects, Forms }, { params, auth }) =>
       Projects.getById(params.projectId)
         .then(getOrNotFound)
+        .then(() => (params.xmlFormId ? getForm(auth, params, params.xmlFormId, Forms) : resolve()))
```
alxndrsn, line 89: It's not obvious how this change is related to `401` responses; any pointers?

test/integration/api/app-users.js
```diff
@@ -4,10 +4,10 @@ const testData = require('../../data/xml');
 
 describe('api: /projects/:id/app-users', () => {
   describe('POST', () => {
-    it('should return 403 unless the user is allowed to create', testService((service) =>
+    it('should return 403 if unauthenticated', testService((service) =>
       service.post('/v1/projects/1/app-users')
         .send({ displayName: 'test1' })
-        .expect(403)));
+        .expect(401)));
```
alxndrsn, line 10: Would there be value in a corresponding test case for authenticated users who are unauthorised?

alxndrsn:
From https://www.rfc-editor.org/rfc/rfc9110#status.401:

> The server generating a 401 response MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate) header field ([Section 11.6.1](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate)) containing at least one challenge applicable to the target resource.

Is this relevant, and if so is it covered by this PR?

alxndrsn:
> #### Does this change require updates to the API documentation? If so, please update docs/api.yaml as part of this PR.
> 
> No

Checking one endpoint at random, it looks like some API docs might need updates:

https://github.com/getodk/central-backend/blob/889d208311d5f788001c1e8a2fa089dcf933b6de/docs/api.yaml#L1965-L1970

matthew-white:
> > The server generating a 401 response MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate) header field &hellip;
> 
> Is this relevant, and if so is it covered by this PR?

I don't think we ever send that header, so I don't think this PR needs to do so. How about we continue discussing that header in #851?

lib/resources/users.js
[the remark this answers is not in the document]
matthew-white: Or maybe it's a nice safeguard? It sort of feels redundant to me now.

lib/resources/users.js
```diff
@@ -54,8 +54,11 @@ module.exports = (service, endpoint) => {
             .then((token) => mail(savedUser.email, 'accountCreated', { token })))
           .then(always(savedUser)))));
 
+    const endpointByInvalidateQuery = (cb) => (req, ...rest) =>
```
matthew-white, line 57: `endpointByInvalidateQuery` and `endpointByTestPath` above make me think that if/when we update `endpoint` to accept an options object, we should make the `anonymous` option take a callback.

Maybe it'd be reasonable to do something like that now? What if `anonymousEndpoint` accepted a function returning `true`/`false`, and then `anonymousEndpoint` either called `authHandler` or not based on the result? We could even make the function required, since Ramda `F` would be easy to pass in.

Just a thought, I'm also happy with the approach here. :+1:

test/integration/api/app-users.js
matthew-white: Similar question from me. I'd expect every endpoint that calls `auth.canOrReject()` to have a test asserting a 403. This test now never reaches the actual endpoint logic: it's just being stopped by the preprocessor, which we test elsewhere.

test/integration/api/assignments.js
```diff
@@ -473,20 +478,23 @@ describe('/projects/:id/assignments', () => {
             .expect(403))))));
 
     it('should return notfound if the project does not exist', testService((service) =>
-      service.login('bob', (asBob) =>
-        asBob.get('/v1/users/current').expect(200).then(({ body }) => body.id)
-          .then((bobId) => service.delete('/v1/projects/99/assignments/manager/' + bobId)
-            .expect(404)))));
+      service.login('alice', (asAlice) =>
+        service.login('bob', (asBob) =>
```
matthew-white, line 482: Doesn't need to change, but just FYI, you can pass an array to `service.login()`:

```js
service.login(['alice', 'bob'], (asAlice, asBob) =>
```

test/integration/api/assignments.js
matthew-white: Those two changes (more 401 errors + requiring auth for /v1/roles) both touch on the theme of anonymity. However, I agree that they're also conceptually distinct. If it's easy to pull the two changes apart, that'd make this PR a little smaller, which sounds nice. I'm also OK with reviewing both changes as part of this PR. I like the idea of having the /v1/roles endpoints check auth.

test/integration/api/sessions.js
```diff
@@ -162,7 +162,7 @@ describe('api: /sessions', () => {
       service.get('/v1/sessions/restore')
         .set('X-Forwarded-Proto', 'https')
         .set('Cookie', 'session: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
```
matthew-white, line 164: This cookie looks unusual, shouldn't it be `=` instead of `:`?

lib/resources/submissions.js
matthew-white: I like the idea of making this an explicit argument (either of the suggestions above). :+1:

test/unit/http/preprocessors.js
```diff
@@ -301,7 +289,7 @@ describe('preprocessors', () => {
             )
           )).should.be.rejectedWith(Problem, { problemCode: 401.2 }));
 
-        it('should do nothing on cookie auth with incorrect session token for non-GET requests', () =>
+        it('should reject cookie auth with incorrect session token for non-GET requests', () =>
```
matthew-white, line 292: Maybe we can remove the part of the test description about "for non-GET requests". We will now fail even GET requests.

matthew-white (commented):
The approach that's being taken here seems like a good one to me. I left some questions/thoughts above, but I think this will be a nice change for Central.

lib/resources/users.js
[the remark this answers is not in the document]
sadiqkhoja: removed.

matthew-white:
Even though I know getodk/central#1001 is a large part of the motivation for this PR, I think it'd be easiest to track this change on its own. I'll go ahead and add it to the project.

lib/resources/users.js
matthew-white, line 1: Also related to /v1/users/current, I was remembering that pyODK uses that endpoint to check whether the session token is still active. If the session token isn't active, the endpoint would return a 404 before, but now it will return a 401. I took a look at the pyODK code to make sure that it wasn't looking specifically for a 404 (it's not): https://github.com/getodk/pyodk/blob/07a6fb568bd60a5c7657f05be3f9ab016fdfbf3e/pyodk/_endpoints/auth.py#L25-L34

matthew-white (commented):
Notes from reviewing changes to lib/, now onward to test/.

lib/resources/submissions.js
matthew-white: I do think it'd be nice to separate it out if that's not too difficult. I was trying to think through the implications of the change. I don't even remember how this endpoint is used. Is there maybe a reason why it needs to be ultra-performant and shouldn't be doing as many checks? Though I guess if it's checking the project, it should probably check the form as well. 🤔 Anyway, I feel like discussing in a separate PR could be nice!

sadiqkhoja (commented):
I have removed `checkAuth` from `user-preferences.js` as well.

And added couple of additional tests for `/users/current`

matthew-white:
I feel like there are a few more places where `auth.actor.map()` could go away, but that could also happen in a follow-up PR (feel free just to file an issue if that seems better).

https://github.com/getodk/central-backend/blob/20d7e72b0feebf1937f1a9fe9ca9e262ebea8cf6/lib/resources/app-users.js#L28

https://github.com/getodk/central-backend/blob/20d7e72b0feebf1937f1a9fe9ca9e262ebea8cf6/lib/resources/public-links.js#L28

https://github.com/getodk/central-backend/blob/20d7e72b0feebf1937f1a9fe9ca9e262ebea8cf6/lib/resources/comments.js#L35

matthew-white (approved):
LGTM 🚀

Thank you for making this change @sadiqkhoja, it's definitely a great improvement to the codebase.

lib/http/preprocessors.js
alxndrsn: Then shouldn't `authBySessionToken()` be returning the session itself, rather than an `Option`?  And L111 would change from:

```diff
-const maybeSession = authBySessionToken(token, () => reject(Problem.user.authenticationFailed()));
+const session = authBySessionToken(token, () => reject(Problem.user.authenticationFailed()));
```

alxndrsn:
> From https://www.rfc-editor.org/rfc/rfc9110#status.401:
> 
> > The server generating a 401 response MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate) header field ([Section 11.6.1](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate)) containing at least one challenge applicable to the target resource.
> 
> Is this relevant, and if so is it covered by this PR?

Already filed at https://github.com/getodk/central-backend/issues/851

PR #1490, opened 2025-05-06
Usage metric: count webforms enabled

ktuite:
Backend part of https://github.com/getodk/central/issues/861

Count number of forms in each project that have webforms enabled.

### DIFF
--- a/lib/model/query/analytics.js
+++ b/lib/model/query/analytics.js
@@ -219,16 +219,15 @@ group by f."projectId";`);
 
 // Forms
 const countForms = () => ({ all }) => all(sql`
-select p."id" as "projectId", count(f."id") as total, count(recentSubs."activeForm") as recent
+select f."projectId" as "projectId", count(f."id") as total, count(recentSubs."activeForm") as recent
 from forms as f
-join projects as p on p."id" = f."projectId"
 left join (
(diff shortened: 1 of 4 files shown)


lib/model/query/analytics.js
```diff
@@ -219,16 +219,15 @@ group by f."projectId";`);
 
 // Forms
 const countForms = () => ({ all }) => all(sql`
-select p."id" as "projectId", count(f."id") as total, count(recentSubs."activeForm") as recent
+select f."projectId" as "projectId", count(f."id") as total, count(recentSubs."activeForm") as recent
 from forms as f
-join projects as p on p."id" = f."projectId"
```
ktuite, line 224: Just noticed this join with projects was unnecessary

lognaturel (approved):
Another option would be to count the opt in events in the audit log (that’s tracked, right?). If this number is really tiny we can consider trying the other metric to see whether people are temporarily opting in.

ktuite:
Webforms event is tracked in the audit log! It's a `form.update` action with `details` of `{"data": {"webformsEnabled": true}}`

I could count this too or instead. Do you think it should still be per project or an overall count of webforms opt in events?

PR #1501, opened 2025-06-03
Return approvalEntityList in Form attachment manifest

sadiqkhoja:
Closes getodk/central#685

Added a test.

I could have modified `getLastUpdateTimestamp()` to return `approvalRequired` fields as well but that function is consumed by `Datasets.getMetadata` as well where approval required is already available. But the real reason I didn't go that route is the renaming the function, what could I have called the it (getLastUpdateTimestampAndApprovalRequired?) 🙃

The change in this PR modifies the `FormAttachments._get()` in a way that additional database queries for each dataset are no longer required, saving us multiple database round-trips.

No

Yes

### DIFF
--- a/lib/model/query/form-attachments.js
+++ b/lib/model/query/form-attachments.js
@@ -143,8 +143,8 @@ const _unjoin = (datasetMetadata = false) => {
 
 const _get = (exec, options) => {
   const unjoin = _unjoin(options.datasetMetadata);
-  return exec(sql`select ${unjoin.fields} from form_attachments
-    left outer join (select id, md5 from blobs) as blobs on form_attachments."blobId"=blobs.id
+  return exec(sql`SELECT ${unjoin.fields} FROM form_attachments
+    LEFT OUTER JOIN (SELECT id, md5 FROM blobs) AS blobs ON form_attachments."blobId"=blobs.id
     ${!options.datasetMetadata ? sql`` : sql`
       LEFT OUTER JOIN datasets ON datasets.id = form_attachments."datasetId"
       LEFT OUTER JOIN (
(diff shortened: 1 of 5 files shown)


test/integration/api/datasets.js
```diff
@@ -3006,6 +3006,41 @@ describe('datasets and entities', () => {
 
       }));
 
+      it('should return md5 of last Entity timestamp in the manifest', testService(async (service) => {
```
ktuite, line 3009: Change test name to something about `approvalEntityList`

lib/model/query/form-attachments.js
```diff
@@ -130,17 +130,35 @@ update.audit = (form, fa, blobId, datasetId = null) => (log) => log('form.attach
...
 
-const _get = (exec, options) =>
-  exec(sql`select ${_unjoinMd5.fields} from form_attachments
-  left outer join (select id, md5 from blobs) as blobs on form_attachments."blobId"=blobs.id
-  where ${sqlEquals(options.condition)} order by name asc`)
-    .then(map(_unjoinMd5));
+const _get = (exec, options) => {
+  const unjoin = _unjoin(options.datasetMetadata);
+  return exec(sql`select ${unjoin.fields} from form_attachments
+    left outer join (select id, md5 from blobs) as blobs on form_attachments."blobId"=blobs.id
```
ktuite, line 147: possibly make sql capitalization more consistent, e.g. all upper case

ktuite (approved):
Looks good

Just making a note that @matthew-white will want to see how the last update timestamp query was moved back into the form attachment `_get` function for computing the hash when he works on the entity access / segmented entities issue https://github.com/getodk/central/issues/952

PR #1502, opened 2025-06-03
forms/draft/get: fix permissions check

alxndrsn:

New tests.

* role check as suggested by @matthew-white 
* maybe there's a neater way to check if a form is a draft?

Might affect users who were taking advantage of the lax permissions checks.

Should not.

### DIFF
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -31,10 +31,10 @@ const excelMimeTypes = {
 };
 
 const canReadForm = (auth, form) => {
-  if (form.state === 'closed') {
-    return auth.canOrReject('form.read', form);
-  } else if (form.def.publishedAt === null) {
+  if (form.def.publishedAt === null) {
     return auth.canOrReject('form.update', form);
+  } else if (form.state === 'closed') {
+    return auth.canOrReject('form.read', form);
   } else {
     return auth.canOrReject(['open_form.read', 'form.read'], form);
   }
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -30,9 +30,15 @@ const excelMimeTypes = {
   xlsx: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
 };
 
-const canReadForm = (auth, form) => (form.state === 'closed'
-  ? auth.canOrReject('form.read', form)
-  : auth.canOrReject(['open_form.read', 'form.read'], form));
+const canReadForm = (auth, form) => {
+  if (form.state === 'closed') {
+    return auth.canOrReject('form.read', form);
+  } else if (form.def.publishedAt === null) {
+    return auth.canOrReject('form.update', form);
+  } else {
+    return auth.canOrReject(['open_form.read', 'form.read'], form);
+  }
+};
 
 const streamAttachment = async (container, attachment, response) => {
   const { s3, Blobs, Datasets, Entities } = container;
--- a/lib/resources/forms.js
+++ b/lib/resources/forms.js
@@ -31,7 +31,7 @@ const excelMimeTypes = {
 };
 
 const canReadForm = (auth, form) => {
-  if (form.def.publishedAt === null) {
+  if (form.def.publishedAt == null) {
     return auth.canOrReject('form.update', form);
   } else if (form.state === 'closed') {
     return auth.canOrReject('form.read', form);
--- a/test/integration/api/forms/draft.js
+++ b/test/integration/api/forms/draft.js
@@ -1401,7 +1401,7 @@ describe('api: /projects/:id/forms (drafts)', () => {
               should.not.exist(body.details);
             }))));
 
-      it('should reject if the user cannot modify', testService((service) =>
+      it.only('should reject if the user does not have project-specific permissions', testService((service) =>
         service.login('alice', (asAlice) =>
           asAlice.post('/v1/projects/1/forms/simple/draft')
             .send(testData.forms.simple)
@@ -1410,6 +1410,31 @@ describe('api: /projects/:id/forms (drafts)', () => {
             .then(() => service.login('chelsea', (asChelsea) =>
               asChelsea.get('/v1/projects/1/forms/simple/draft')
                 .expect(403))))));
+
+      [
+        'app-user',
+        'formfill',
+        'formview',
+        'pub-link',
+        'pwreset',
+        'viewer',
+      ].forEach(role => {
+        it.only(`should reject if the user has role "${role}"`, testService((service) =>
+          service.login('alice', (asAlice) =>
+            asAlice.post('/v1/projects/1/forms/simple/draft')
+              .send(testData.forms.simple)
+              .set('Content-Type', 'application/xml')
+              .expect(200)
+              .then(() => service.login('chelsea', (asChelsea) =>
+                asChelsea.get('/v1/users/current')
+                  .expect(200)
+                  .then(({ body }) => body)
+                  .then((chelsea) => 
+                    asAlice.post(`/v1/projects/1/assignments/${role}/${chelsea.id}`)
+                      .expect(200))
+                  .then(() => asChelsea.get('/v1/projects/1/forms/simple/draft')
+                    .expect(403)))))));
+      });
 
       it('should give basic draft details', testService((service) =>
         service.login('alice', (asAlice) =>
--- a/test/integration/api/forms/draft.js
+++ b/test/integration/api/forms/draft.js
@@ -1418,7 +1418,7 @@ describe('api: /projects/:id/forms (drafts)', () => {
         'pub-link',
         'viewer',
       ].forEach(role => {
-        it.only(`should reject if the user has role "${role}"`, testService((service) =>
+        it(`should reject if the user has role "${role}"`, testService((service) =>
           service.login('alice', (asAlice) =>
             asAlice.post('/v1/projects/1/forms/simple/draft')
               .send(testData.forms.simple)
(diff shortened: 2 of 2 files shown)


lib/resources/forms.js
```diff
@@ -30,9 +30,15 @@ const excelMimeTypes = {
   xlsx: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
 };
 
-const canReadForm = (auth, form) => (form.state === 'closed'
-  ? auth.canOrReject('form.read', form)
-  : auth.canOrReject(['open_form.read', 'form.read'], form));
+const canReadForm = (auth, form) => {
+  if (form.state === 'closed') {
+    return auth.canOrReject('form.read', form);
+  } else if (form.def.publishedAt === null) {
```
matthew-white, line 36: I'm wondering about the use of `form.def` here. You wrote in the PR description:

> maybe there's a neater way to check if a form is a draft?

IIRC `form.def` refers to either the published def (via `currentDefId`) or the draft def (via `draftDefId`). It depends on which def was queried, which I believe depends on the endpoint. Since a form can have both a published def and a draft def, if `form.def` referred to the draft def, `form.def.publishedAt === null` would be true even if the form has a published def. In other words, I don't think it's a sure way to check that a form is a draft. I might be wrong about that though: maybe `canReadForm()` is only called in situations where `form.def` refers to the published def.

You might want to poke around to see how other places check whether a form is a draft. I don't quite remember the best way to do it. Off the top of my head, I feel like it's possible to check `form.currentDefId`. I'm pretty sure that will be `null` if the form is a draft.

matthew-white: Yeah, I'm pretty sure that `form.currentDefId` is always a published form def, so would never equal `form.draftDefId`.

Does this work?

```js
} else if (form.currentDefId == null) {
```

> IIRC `form.def` refers to either the published def (via `currentDefId`) or the draft def (via `draftDefId`). It depends on which def was queried, which I believe depends on the endpoint.

Just to provide an example of this, I believe that for the following group of endpoints, `form.def` will refer to the draft def — even if the form has a published form def. Yet I think it's still possible to tell whether the form has a published form def by looking at `currentDefId`.

https://github.com/getodk/central-backend/blob/b2114d5b085aca7b0ccf5f0ffc1f6cc77eab8e14/lib/resources/forms.js#L327-L330

matthew-white:
FYI @alxndrsn, in order for this PR to make it into v2025.2, it'll need to get merged tomorrow. I'm not sure whether you'll have time tomorrow to make changes to this PR, but if you do, would it be alright if I merged after approving?

lib/resources/forms.js
alxndrsn: Revisiting this:

> Since a form can have both a published def and a draft def, if `form.def` referred to the draft def, `form.def.publishedAt === null` would be true even if the form has a published def.

Maybe that's the point?  The idea is to check:

* if the user is trying to read a draft, check that the user has `form.update` permission; and
* if the user is trying to read a published for, check that the user has `form.read` permission

A user with only `form.read` should not be able to view a draft just because there is _also_ a published version of that form.  I'll see if I can add a test case to address this situation.

alxndrsn: here are a few approaches currently in use in `master`:

# 1. comparing `def.id` with `draftDefId`

```
lib/model/frames/form.js:82:    if (this.def.id === this.draftDefId) return this.def.enketoId;
lib/model/query/forms.js:262:  if (!publish && (form.def.id == null || form.def.id !== form.draftDefId)) {
lib/model/query/forms.js:304:  : log('form.update.draft.set', form, { oldDraftDefId: form.draftDefId, newDraftDefId: newDef.id }));
lib/model/query/forms.js:328:  if (form.draftDefId !== form.def.id) throw Problem.internal.unknown();
lib/worker/form.js:17:      if ((form.def.id == null) || (form.draftDefId !== form.def.id)) return;
test/integration/other/form-versioning.js:38:          newForm.draftDefId.should.not.equal(newForm.def.id);
```

# 2. checking if `def.publishedAt` is `null`/`NULL`

# 3. checking if `def.draftToken` is not `null`/`NULL`

I'm not sure the relationship between `draftToken` and `publishedAt` is, but they never seem to be both `NULL`, and never both not `NULL`.  So it seems like checking `def.draftToken != null` is equivalent of `def.publishedAt == null`.

alxndrsn: My understanding from https://github.com/getodk/central-backend/pull/453 is that a form can only have a single draft at a time, which means that all of these checks should be equivalent.

Would it be helpful for the form "frame" to formalise these relationships?

alxndrsn: > Would it be helpful for the form "frame" to formalise these relationships?

Example PR at https://github.com/getodk/central-backend/pull/1520

lib/resources/forms.js
```diff
@@ -30,9 +30,15 @@ const excelMimeTypes = {
...
 };
 
-const canReadForm = (auth, form) => (form.state === 'closed'
-  ? auth.canOrReject('form.read', form)
-  : auth.canOrReject(['open_form.read', 'form.read'], form));
+const canReadForm = (auth, form) => {
+  if (form.def.publishedAt === null) {
+    return auth.canOrReject('form.update', form);
+  } else if (form.state === 'closed') {
+    return auth.canOrReject('form.read', form);
```
alxndrsn, line 37: I've changed the order of checks here to make sure that closing a form does not elevate permissions for form-readers.

alxndrsn:
> would it be alright if I merged the PR after approving it?

Have we reached agreement on what a valid check for "is this form a draft" looks like?  If so I'm happy for this PR to be merged.

test/integration/api/forms/draft.js
```diff
@@ -1401,15 +1401,75 @@ describe('api: /projects/:id/forms (drafts)', () => {
...
+        (service) => service.login('chelsea'),
+      ];
+
+      [
+        noExtraRoles(),
+        withRole('app-user'),
+        withRole('formfill'),
+        withRole('formview'),
+        withRole('pub-link'),
+        withRole('viewer'),
```
alxndrsn, line 1428: I've updated the test setup to make it clear that all of these roles should have identical access to a form draft as a user with no project-specific permissions.

lib/resources/forms.js
matthew-white: <blockquote>

> My understanding from https://github.com/getodk/central-backend/pull/453 is that a form can only have a single draft at a time

I can confirm that that's the case. :+1:
</blockquote>

Just to say a little more about this, the API enforces that there's only a single form draft at a time. The .../draft endpoint returns a single form draft. You can replace the existing form draft, but you can't add a second draft that will be visible to the user alongside the first.

That said, there is the concept of orphan form drafts that I mention above.

matthew-white:
> Have we reached agreement on what a valid check for "is this form a draft" looks like?

I think so! Checking `form.def.publishedAt` is making sense to me.

lib/resources/forms.js
```diff
@@ -30,9 +30,15 @@ const excelMimeTypes = {
   xlsx: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
 };
 
-const canReadForm = (auth, form) => (form.state === 'closed'
-  ? auth.canOrReject('form.read', form)
-  : auth.canOrReject(['open_form.read', 'form.read'], form));
+const canReadForm = (auth, form) => {
+  if (form.def.publishedAt === null) {
```
matthew-white, line 34: I think I'd feel slightly more comfortable with `==` instead of `===`. We don't usually use `===` with `null`: we don't usually distinguish between `null` and `undefined`. I don't think it's impossible that we'll one day subtly change our database functions such that `form.def.publishedAt` becomes `undefined` instead of `null`. I can't think of much of a downside of checking for both `null` and `undefined`.

test/integration/api/forms/draft.js
```diff
@@ -1401,15 +1401,75 @@ describe('api: /projects/:id/forms (drafts)', () => {
...
             .set('Content-Type', 'application/xml')
-            .expect(200)
-            .then(() => service.login('chelsea', (asChelsea) =>
-              asChelsea.get('/v1/projects/1/forms/simple/draft')
-                .expect(403))))));
+            .expect(200);
+
+          const asChelsea = await setupChelsea(service, asAlice);
+          return asChelsea.get('/v1/projects/1/forms/simple/draft')
+            .expect(403);
```
matthew-white, line 1439: Could you also test that Chelsea can't access the form itself? In the case where a form only has a draft and no published version, /v1/projects/1/forms/simple and /v1/projects/1/forms/simple/draft return pretty much the same thing. In contrast, in the next test, Chelsea should be able to access the form, just not the draft.

```js
await asChelsea.get('/v1/projects/1/forms/simple')
  .expect(403);
```

test/integration/api/forms/draft.js
matthew-white, line 1: The tests are looking really nice! 🎉

lib/resources/forms.js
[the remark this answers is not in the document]
matthew-white: OK, I may have been getting confused about this. I thought `canReadForm()` was being run before `ensureDef()`, but I think it's actually the other way around. So we shouldn't need to check `form.def.id` here, as `ensureDef()` will have checked that already.

The exception to that is the .../dataset-diff and .../draft/dataset-diff endpoints. Those seem to check `canReadForm()` before `ensureDef()`. Maybe they should check `ensureDef()` first.

matthew-white:
OK, I'm done taking a look for now! It's feeling really close to me.

lib/resources/forms.js
alxndrsn: :+1: fixed

test/integration/api/forms/draft.js
alxndrsn: @matthew-white is this still required?

PR #1504, opened 2025-06-03
auth.can(): simplify SQL

alxndrsn:
Current code seems odd.

Noted while working on #1502

- [x] existing tests pass

This introduces some formatting change which IMO improve readability.  A simpler changeset could be achieved by maintaining existing formatting.

Potential for breaking existing authZ, but test coverage should ensure it's fine.  Also reading the code, it seems low-risk.

No.

### DIFF
--- a/lib/model/query/auth.js
+++ b/lib/model/query/auth.js
@@ -40,12 +40,11 @@ const can = (actor, verbs, actee) => ({ oneFirst }) => {
 
   return oneFirst(sql`
 ${_impliedActees(acteeId)}
-select count(*) from assignments
+select count(*)>0 from assignments
 inner join implied on implied.id=assignments."acteeId"
 inner join (select id from roles where ${_verbClause(verbsSet)}) as role on role.id=assignments."roleId"
 where "actorId"=${actor.id}
-limit 1`)
-    .then((count) => count > 0);
+  `);
 };
 
 const canAssignRole = (actor, role, actee) => ({ Auth }) =>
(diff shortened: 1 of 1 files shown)


lib/model/query/auth.js
```diff
@@ -38,13 +38,17 @@ const can = (actor, verbs, actee) => ({ oneFirst }) => {
   }
 
   return oneFirst(sql`
-${_impliedActees(acteeId)}
-select count(*) from assignments
-inner join implied on implied.id=assignments."acteeId"
-inner join (select id from roles where ${_verbClause(verbsSet)}) as role on role.id=assignments."roleId"
-where "actorId"=${actor.id}
-limit 1`)
```
matthew-white, line 46: I know the `limit 1` didn't do much before, because the `count(*)` would always be one row. But now I'm wondering, would adding `LIMIT 1` to the new `SELECT 1` subquery improve its performance? We just want to know if any assignment exists, so we don't need to retrieve all assignments. For projects with many forms, app users can have many assignments. That might be outside the scope of this PR, but I thought I'd throw out the idea.

matthew-white (approved):

lib/model/query/auth.js
alxndrsn: I don't think it's necessary per the postgres docs:

> The subquery will generally only be executed long enough to determine whether at least one row is returned, not all the way to completion.
> - _https://www.postgresql.org/docs/14/functions-subquery.html#FUNCTIONS-SUBQUERY-EXISTS_

PR #1512, opened 2025-06-09
lib/resources: standardise usage of xmlFormId in path params

alxndrsn:
This makes intent more explicit, and makes it easit to locate uses of xmlFormId in path params.


matthew-white:
I think we should try to merge #1488 before merging this one, just because it's taken a while to get #1488 to a checkmark for code review. I think #1488 and this PR have a few conflicting lines (e.g., around the /v1/test endpoints).

alxndrsn:
> I think #1488 and this PR have a few conflicting lines (e.g., around the /v1/test endpoints).

#1488 has been merged; conflicts with this PR have been resolved.

PR #1520, opened 2025-06-10
frames/form: add isDaft() method

alxndrsn:
Formalises the two functionally equivalent ways to check if a given Form frame represents the current draft of that form:

1. form.def.publishedAt === null
2. form.def.id === form.draftDefId

The first is probably more reliable, as if a future code change allowed a single form to have multiple drafts, it would continue to function correctly.

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -325,10 +325,10 @@ replaceDef.audit.logEvenIfAnonymous = true;
 // TODO: we need to make more explicit what .def actually represents throughout.
 // for now, enforce an extra check here just in case.
 const publish = (form, concurrentWithCreate = false) => async ({ Forms, Datasets, FormAttachments }) => {
-  if (form.draftDefId !== form.def.id) throw Problem.internal.unknown();
+  if (!form.isDraft()) throw Problem.internal.unknown();
 
   // dont use publish concurrentWithCreate on already published forms
-  if (concurrentWithCreate && form.publishedAt != null) throw Problem.internal.unknown({ error: 'Attempting to immediately publish a form' });
+  if (concurrentWithCreate && !form.isDraft()) throw Problem.internal.unknown({ error: 'Attempting to immediately publish a form' });
 
   // Try to push the form to Enketo if it hasn't been pushed already. If doing
   // so fails or is too slow, then the worker will try again later.
--- a/lib/worker/form.js
+++ b/lib/worker/form.js
@@ -14,7 +14,7 @@ const pushDraftToEnketo = ({ Forms }, event) =>
   Forms.getByActeeIdForUpdate(event.acteeId, undefined, Form.DraftVersion)
     .then((maybeForm) => maybeForm.map((form) => {
       // if there was no draft or this form isn't the draft anymore just bail.
-      if ((form.def.id == null) || (form.draftDefId !== form.def.id)) return;
+      if (form.def.id == null || !form.isDraft()) return;
 
       // if the enketoId was received during the request to create the draft, or
       // if it was carried forward from the previous draft, then bail.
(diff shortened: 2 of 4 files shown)


matthew-white:
> Add `isDraft()` to `Form.Def`, checking there that `publishedAt == null`.

Actually, I guess this sounds like the check 1 that you mention above:

<blockquote>

Could check either:

1. `form.def.publishedAt === null`
2. `form.def.id === form.draftDefId`

I think **1** is more reliable.
</blockquote>

But this PR currently seems to be implementing check 2. Why should we use check 2 if check 1 is potentially more reliable? How about if we added an `isDraft()` method to `Form.Def` that checked `publishedAt == null`, then called `form.def.isDraft()`?

lib/model/query/forms.js
```diff
@@ -325,10 +325,10 @@ replaceDef.audit.logEvenIfAnonymous = true;
 // TODO: we need to make more explicit what .def actually represents throughout.
 // for now, enforce an extra check here just in case.
 const publish = (form, concurrentWithCreate = false) => async ({ Forms, Datasets, FormAttachments }) => {
-  if (form.draftDefId !== form.def.id) throw Problem.internal.unknown();
+  if (!form.isDraft()) throw Problem.internal.unknown();
 
   // dont use publish concurrentWithCreate on already published forms
-  if (concurrentWithCreate && form.publishedAt != null) throw Problem.internal.unknown({ error: 'Attempting to immediately publish a form' });
+  if (concurrentWithCreate && !form.isDraft()) throw Problem.internal.unknown({ error: 'Attempting to immediately publish a form' });
```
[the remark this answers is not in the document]
alxndrsn: So there's a bug in prod?  Which should be tackled first/separately to this PR?

lib/worker/form.js
```diff
@@ -14,7 +14,7 @@ const pushDraftToEnketo = ({ Forms }, event) =>
   Forms.getByActeeIdForUpdate(event.acteeId, undefined, Form.DraftVersion)
     .then((maybeForm) => maybeForm.map((form) => {
       // if there was no draft or this form isn't the draft anymore just bail.
-      if ((form.def.id == null) || (form.draftDefId !== form.def.id)) return;
+      if (form.def.id == null || !form.isDraft()) return;
```
matthew-white, line 17: OK, here's a case where I think we maybe do need to check `form.draftDefId`, not just `form.def.publishedAt`. If the worker starts processing a draft form def, but then realizes that the draft has been replaced with a newer draft, it should stop processing the form def. I think it can only determine that the draft has been replaced by checking `draftDefId`: `publishedAt` indicates whether a form def is a draft, while `draftDefId` indicates whether a form def is the _current_ draft.

But now that I look beyond this line, isn't `Forms.getByActeeIdForUpdate()` being called with `Form.DraftVersion`? I would think that that would guarantee that `form.draftDefId === form.def.id`. 🤔

lib/model/query/forms.js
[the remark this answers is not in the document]
matthew-white: Yeah, it looks like a bug in `master` to me. There's no way for the `if` condition to be true, because `form.publishedAt` will always be `null`, because `publishedAt` is actually defined on `form.def`. Maybe "bug" is too strong a word, I don't see that the issue is necessarily user-facing. But the code doesn't look right to me.

I feel like it'd be reasonable to tackle that in a separate PR if you think that's what would work best. But also, if we determine that it's just a redundant `if` to begin with, I think it'd be fine just to remove it (either in this PR or in a separate PR).

lib/worker/form.js
alxndrsn: > `publishedAt` indicates whether a form def is a draft, while `draftDefId` indicates whether a form def is the _current_ draft.

Would it make sense to remove the `form.isDraft()` function currently added in this PR, and instead have:

1. `form.def.isDraft()`, and 
2. `form.isCurrentDraft()`

?

matthew-white: That makes sense to me! I just approved #1577.

alxndrsn:
Closed in favour of https://github.com/getodk/central-backend/pull/1577.

PR #1525, opened 2025-06-12
wip: fix API access to drafts through non-draft endpoints

alxndrsn:
Closes getodk/central#1256

Restrict `/draft/` endpoints to draft form versions, and their corresponding endpoints _without_ `/draft/` to published form versions.

Updated all the tests.

An alternative would be to fix a subset of the `/draft`/not-`/draft` endpoints.

Some extras thrown in here:

* require `Form.<desired version>` in the `Form.get...()` query functions - this should make intentions and future uses clearer, and prevent draft/non-draft usage in future
* reverse order of `xml, version` args to `version, xml`, which allows for excluding the `xml`/`withXml` arg in most cases
* replacing boolean values for `xml`/`withXml` to make intent and API-misuse clearer

Some users might be surprised that endpoints that would previously give them draft forms (when there was no published version) will no longer do so.

Potentially - will at least need a review of the current docs.

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -594,10 +594,10 @@ const versionJoinCondition = (version, enketoId) => {
     )
   `;
   else if (version === '___') return versionJoinCondition('');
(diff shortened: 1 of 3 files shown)


alxndrsn:
Out of date.

PR #1526, opened 2025-06-17
blobs: update contentType when duplicate uploaded

alxndrsn:
Ref getodk/central#751

Modified duplicate blob test to a state which would have failed with previous `lib/` code, but now passes.

It's @matthew-white's approach **4** from getodk/central#751.

Approach **3** definitely feels more correct, but the approach in this PR has the following advantages:

1. it is definitely an improvement over the status quo
2. it is much simpler than approach **3**:
  * **3** requires changes to 5 database tables, which will be slow
  * **3** requires different `JOIN` semantics when fetching blobs

It will change behaviour, but given #1356 it should be an improvement.

Not really?

### DIFF
--- a/lib/model/query/blobs.js
+++ b/lib/model/query/blobs.js
@@ -23,7 +23,7 @@ const ensure = (blob) => ({ oneFirst }) => oneFirst(sql`
 with ensured as
 (insert into blobs (sha, md5, content, "contentType") values
     (${blob.sha}, ${blob.md5}, ${sql.binary(blob.content)}, ${blob.contentType || sql`DEFAULT`})
-  on conflict (sha) do update set sha = ${blob.sha}, "contentType" = ${blob.contentType || sql`DEFAULT`}
+  on conflict (sha) do update set sha = EXCLUDED.sha, "contentType" = EXCLUDED."contentType"
   returning id)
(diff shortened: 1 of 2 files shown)


alxndrsn:
@matthew-white depending on urgency of #751, this could be considered.

lognaturel:
> Do you know whether the OpenRosa clients we use (Collect, Enketo, Web Forms) actually use the Content-Type header? 

I don't believe Collect does but the web-based clients do which is why there's a user-facing issue. At least for geojson, Enketo and Web Forms fail when the content type is set incorrectly. I think we've largely mitigated the geojson issue by filling in content type based on extension (basically a targeted piece of #1098).

I agree @matthew-white's latest idea seems like a reasonable compromise that lets someone fix their own project's issues without affecting others'.

alxndrsn:
> Change the uniqueness constraint so that it's on `(sha, "contentType")`, not just `sha`. It'd be possible to end up with duplicate content in the database/S3, but only if someone submits the same blob with a different `contentType`. That probably happens rarely enough that it wouldn't be a big deal to store such blobs multiple times.

Implemented at https://github.com/getodk/central-backend/pull/1579.

PR #1527, opened 2025-06-17
Forms: make conflation of published/draft explicit

alxndrsn:
Current `master` allows requests for forms through query functions to implicitly conflate form drafts and published versions.

This leads to surprising behaviours in the API like:

* #1404
* TODO others?

Affected functions: pretty much all form querying functions.

Closes #

### DIFF
--- a/lib/model/query/forms.js
+++ b/lib/model/query/forms.js
@@ -581,8 +581,9 @@ order by coalesce(form_defs.name, forms."xmlFormId") asc`)
 
 // helper function to gate how form defs are joined to forms in _get
 /* eslint-disable indent */
-const versionJoinCondition = (version, enketoId) => (
-  (enketoId) ? sql`
+const versionJoinCondition = (version, enketoId) => {
+  if (version == null) throw new Error('Must request a specific version or version class');
+  else if (enketoId) return sql`
     form_defs."formId"=forms.id 
     AND (
       form_defs."enketoId" = ${enketoId} AND form_defs.id = forms."draftDefId"
@@ -590,14 +591,15 @@ const versionJoinCondition = (version, enketoId) => (
         ( forms."enketoId" = ${enketoId} OR forms."enketoOnceId" = ${enketoId} ) 
         AND form_defs.id = forms."currentDefId"
       )
-    )` :
-  (version === '___') ? versionJoinCondition('') :
-  (version == null) ? sql`form_defs.id=coalesce(forms."currentDefId", forms."draftDefId")` :
-  (version === Form.DraftVersion) ? sql`form_defs.id=forms."draftDefId"` :
-  (version === Form.PublishedVersion) ? sql`form_defs.id=forms."currentDefId"` :
-  (version === Form.AllVersions) ? sql`form_defs."formId"=forms.id and form_defs."publishedAt" is not null` :
-  sql`form_defs."formId"=forms.id and form_defs.version=${version} and form_defs."publishedAt" is not null`
-);
+    )
+  `;
+  else if (version === '___') return versionJoinCondition('');
+  else if (version === Form.AnyVersion)       return sql`form_defs.id=coalesce(forms."currentDefId", forms."draftDefId")`; // eslint-disable-line no-multi-spaces
+  else if (version === Form.DraftVersion)     return sql`form_defs.id=forms."draftDefId"`; // eslint-disable-line no-multi-spaces
+  else if (version === Form.PublishedVersion) return sql`form_defs.id=forms."currentDefId"`;
+  else if (version === Form.AllVersions)      return sql`form_defs."formId"=forms.id  and form_defs."publishedAt" is not null`; // eslint-disable-line no-multi-spaces
+  else return sql`form_defs."formId"=forms.id and form_defs.version=${version} and form_defs."publishedAt" is not null`;
+};
 /* eslint-enable indent */
 
 


test/e2e/s3/test.js
```diff
@@ -330,12 +330,22 @@ describe('s3 support', () => {
...
+    // eslint-disable-next-line no-console
+    console.log(`
+      ###############################
+      #
+      # rest: ${JSON.stringify({ xmlFormId, rest })}
+      #
+      # /attachments: ${JSON.stringify(await api.apiGet(`projects/${projectId}/forms/${xmlFormId}/attachments`))}
+      #
+      ###############################
+    `);
```
alxndrsn, line 343: revert

test/integration/api/datasets.js
```diff
@@ -2249,8 +2249,9 @@ describe('datasets and entities', () => {
               .set('Content-Type', 'application/xml')
               .expect(200)
               .then(() =>
-                Forms.getByProjectAndXmlFormId(1, 'withAttachments')
-                  .then(form => FormAttachments.getByFormDefIdAndName(form.get().def.id, 'people.csv')
+                Forms.getByProjectAndXmlFormId(1, 'withAttachments', false, Form.DraftVersion)
+                  .then(getOrNotFound)
```
alxndrsn, line 2253: revert

test/integration/api/datasets.js
```diff
@@ -2274,8 +2275,9 @@ describe('datasets and entities', () => {
               .set('Content-Type', 'application/xml')
               .expect(200))
             .then(() =>
-              Forms.getByProjectAndXmlFormId(1, 'withAttachments')
-                .then(form => FormAttachments.getByFormDefIdAndName(form.get().def.id, 'people.csv')
+              Forms.getByProjectAndXmlFormId(1, 'withAttachments', false, Form.DraftVersion)
+                .then(getOrNotFound)
```
alxndrsn, line 2279: revert

test/integration/api/forms/draft.js
```diff
@@ -3,6 +3,7 @@ const appRoot = require('app-root-path');
 const should = require('should');
 const { testService } = require('../../setup');
 const testData = require('../../../data/xml');
+const { getOrNotFound } = require(appRoot + '/lib/util/promise');
```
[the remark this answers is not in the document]
alxndrsn: done

lib/model/query/entities.js
```diff
@@ -478,7 +478,7 @@ const _processSubmissionEvent = (event, parentEvent) => async ({ Datasets, Entit
   const { submissionId, submissionDefId } = event.details;
   const forceOutOfOrderProcessing = parentEvent?.details?.force === true;
 
-  const form = await Forms.getByActeeId(event.acteeId);
+  const form = await Forms.getByActeeId(event.acteeId, Form.PublishedVersion); // this appears to need a form definition... but why?
```
ktuite, line 481: I don't think the entity code below really needs this form to have a def? Does leaving out `Form.PublishedVersion` break something?

Looking at the other places in the code `Forms.getByActeeId` is used, those don't seem to use a form def either or make use of the `versionJoinCondition`? Maybe `getByActeeId` was built to be super general (get the current draft or published def) but was never needed.

test/e2e/s3/test.js
```diff
@@ -340,6 +340,8 @@ describe('s3 support', () => {
         ))
     );
 
+    await api.apiPostJson(`projects/${projectId}/forms/${xmlFormId}/draft/publish`, {});
```
ktuite, line 343: Did changing this change something?

test/integration/api/forms/forms.js
```diff
@@ -265,7 +266,7 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
           .send(testData.forms.simple2)
           .set('Content-Type', 'application/xml')
           .expect(200)
-          .then(() => asAlice.get('/v1/projects/1/forms/simple2')
+          .then(() => asAlice.get('/v1/projects/1/forms/simple2/draft')
```
ktuite, line 269: Is this the kind of API change you're referring to in the comments? That if you have a form with only a draft, accessing `/v1/projects/1/forms/simple2` wont work anymore and you'd need to access `/v1/projects/1/forms/simple2/draft`?

test/integration/api/forms/forms.js
```diff
@@ -1224,7 +1225,7 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
               .set('Content-Type', 'application/xml')
               .expect(200)
               .then(() => service.login('chelsea', (asChelsea) =>
-                asChelsea.get('/v1/projects/1/forms/withAttachments/attachments')
+                asChelsea.get('/v1/projects/1/forms/withAttachments/draft/attachments')
```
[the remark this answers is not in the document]
ktuite: I don't understand the nuance of this change either

ktuite (commented):
Overall, I like the explicitness of adding `AnyVersion`, `WithoutDef` in the code etc. instead of the confusing existing args. 

These changed also seemed to highlight something in the entity code where a form and its def are needed, but maybe the def isn't really needed. I can look into it on the entities side of things.

This part I'm less sure about: 
> As it stands, this PR will have some API changes which may or may not be surprising. Hopefully they would be more intuitive, but they could certainly break some existing workflows.

Is this referring to how if you only have a draft version of your form, you can kinda use the main endpoint for the form and get details about the draft? And how this is [confusing](https://github.com/getodk/central-backend/issues/1404#issuecomment-2957776653) even to central devs like myself?

test/e2e/s3/test.js
alxndrsn: Yes - it's publishing the form so that non-`/draft` endpoints can be used.

alxndrsn:
> > As it stands, this PR will have some API changes which may or may not be surprising. Hopefully they would be more intuitive, but they could certainly break some existing workflows.
> 
> Is this referring to how if you only have a draft version of your form, you can kinda use the main endpoint for the form and get details about the draft? And how this is [confusing](https://github.com/getodk/central-backend/issues/1404#issuecomment-2957776653) even to central devs like myself?

Exactly.  I've updated the PR description to make it explicit that #1404 is closed by this PR.

matthew-white:
I'm going to remove myself as a reviewer, since I see that @ktuite is weighing in. Feel free to tag me again if a second opinion would be helpful on anything.

lib/model/query/forms.js
```diff
@@ -674,25 +699,31 @@ inner join
...
+const _get = (exec, options, version, xml, deleted, actorId) => {
+  if (xml !== Form.IncludeXml && xml !== Form.WithoutXml) throw new Error('Must use Form.WithoutXml or Form.IncludeXml; actually got: ' + xml.toString());
+  if (version === Form.WithoutDef && xml === Form.IncludeXml) throw new Error('Cannot request xml without a form def');
 
-const _getWithoutXml = extender(Form, Form.Def)(Form.Extended, Actor.into('createdBy'))(_getSql);
-const _getWithXml = extender(Form, Form.Def, Form.Xml)(Form.Extended, Actor.into('createdBy'))(_getSql);
-const _get = (exec, options, xml, version, deleted, actorId) =>
-  ((xml === true) ? _getWithXml : _getWithoutXml)(exec, options, version, deleted, actorId);
+  const frames = [ Form ];
+  if (version !== Form.WithoutDef) frames.push(Form.Def);
```
ktuite, line 711: I was confused about why `Form.WithoutDef` would be passed through the `version` parameter, but after thinking about it for a long time, I've come up with this reason: If you don't need the def for the form, then it really doesn't matter if you're fetching a published or unpublished version, or a specific version...

That's why `Form.WithoutDef` is an ok thing to pass in the version string especially for the resources about listing form versions or deleting forms.

test/integration/api/forms/forms.js
```diff
@@ -724,7 +725,7 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
             .set('Content-Type', 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet')
             .expect(200)
             .then(() => service.login('chelsea', (asChelsea) =>
-              asChelsea.get('/v1/projects/1/forms/simple2.xlsx')
+              asChelsea.get('/v1/projects/1/forms/simple2/draft.xlsx')
```
ktuite, line 728: I don't think these lines needed to change... but it's also confusing because it's testing good ol' Chelsea's access and she doesn't have access to anything, so of course there's going to be some kind of 400 error regardless of if we're looking at the draft or not.

test/integration/api/forms/forms.js
```diff
@@ -733,7 +734,7 @@ describe('api: /projects/:id/forms (create, read, update)', () => {
             .send(readFileSync(appRoot + '/test/data/simple.xlsx'))
             .set('Content-Type', 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet')
             .expect(200)
-            .then(() => asAlice.get('/v1/projects/1/forms/simple2.xls')
+            .then(() => asAlice.get('/v1/projects/1/forms/simple2/draft.xls')
```
ktuite, line 737: I tried reverting this line too and the test still seemed to run.

test/integration/other/form-entities-version.js
```diff
@@ -436,19 +437,20 @@ describe('Update / migrate entities-version within form', () => {
...
-      const { acteeId } = await Forms.getByProjectAndXmlFormId(1, 'updateEntity').then(o => o.get());
+      const { acteeId } = await Forms.getByProjectAndXmlFormId(1, 'updateEntity', Form.WithoutDef).then(o => o.get());
       await Audits.log(null, 'upgrade.process.form.entities_version', { acteeId });
 
       // Run form upgrade
       await exhaust(container);
 
-      await asAlice.get('/v1/projects/1/forms/updateEntity')
+      // REVIEW maybe this shouldn't change - maybe the base form endpoint should always provide basic info
+      await asAlice.get('/v1/projects/1/forms/updateEntity/draft')
```
ktuite, line 453: Same theme as some of my other comments -- Yeah I don't think this needs to change. Base form endpoint always providing basic info makes sense. That was the main thing I was concerned about with an API change "possibly breaking users' workflows".

test/integration/other/form-entities-version.js
```diff
@@ -436,19 +437,20 @@ describe('Update / migrate entities-version within form', () => {
         .expect(200);
 
       // check updatedAt on the draft form
-      await asAlice.get('/v1/projects/1/forms/updateEntity')
+      await asAlice.get('/v1/projects/1/forms/updateEntity/draft')
```
ktuite, line 440: Also probably doesn't need to change

test/integration/other/form-entities-version.js
```diff
@@ -64,7 +65,7 @@ describe('Update / migrate entities-version within form', () => {
         .set('Content-Type', 'application/xml')
         .expect(200);
 
-      const { acteeId } = await Forms.getByProjectAndXmlFormId(1, 'updateEntity').then(o => o.get());
+      const { acteeId } = await Forms.getByProjectAndXmlFormId(1, 'updateEntity', Form.WithoutDef).then(o => o.get());
```
ktuite, line 68: Just commenting that I appreciate/apologize for you having to go and changes all these things in the tests. Ideally there wouldn't be code like this in the tests, but uh, I guess I did what had to be done for checking things that weren't exposed in the API.

test/integration/api/forms/draft.js
```diff
@@ -2113,7 +2113,7 @@ describe('api: /projects/:id/forms (drafts)', () => {
             .set('X-Requested-With', 'XMLHttpRequest')
             .expect(200);
 
-          await service.get('/v1/projects/1/forms/withAttachments.xml')
+          await service.get('/v1/projects/1/forms/withAttachments/draft.xml')
```
ktuite, line 2116: This doesn't have to change

test/e2e/s3/test.js
ktuite: Is it necessary?

ktuite (commented):
Keep in mind, all my comments in this review batch are tied up in the assumptions that
1. There was something about this PR that was going to make a breaking change to the API
2. I couldn't actually find find a breaking change - hitting a non-draft endpoint for a form with only a draft worked in the cases i expected it to work
3. My overall mental model IS shifting... if you hit the base endpoint for a draft form, getting some info (metadata) about it seems like a nice thing to get. 
4. Getting the xml for the form (or the manifest) at the base URL when it's only a draft is seems slightly different than just getting the metadata.. but that's how it seems like the API worked in the past, and maybe it's fine to just leave it that way. 

I'm getting a little lost in what the overall end state of this PR should be... but I think if there's a way to make the internal calls more explicit but NOT change the API, that seems like a good place for this PR to land. 

Unless there's some really clear and compelling API change, but at this point, I'm not seeing it.

alxndrsn:
> if there's a way to make the internal calls more explicit but NOT change the API, that seems like a good place for this PR to land.

:+1: I'll aim for this - no test changes, but internal changes to the implementation code.  We can follow up later with fixes if we think there are bugs.

test/integration/api/forms/forms.js
alxndrsn: I've reverted this change.

test/integration/api/forms/forms.js
alxndrsn: :+1: reverted

test/integration/api/forms/forms.js
alxndrsn: :+1: reverted.

test/integration/other/form-entities-version.js
alxndrsn: No big deal :¬)

lib/resources/forms.js
```diff
@@ -325,37 +325,39 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
 
   formResource('/projects/:projectId/forms/:xmlFormId/draft', (Forms, params, withXml = false, options = QueryOptions.none) =>
-    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, withXml, Form.DraftVersion, options)
+    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.DraftVersion, Form.withXml(withXml), options)
       .then(getOrNotFound)
       .then(ensureDef));
 
   service.get('/form-links/:enketoId/form', endpoint(({ Forms }, { params, auth }) =>
-    Forms.getByEnketoId(params.enketoId, false, null, QueryOptions.none)
+    Forms.getByEnketoId(params.enketoId)
```
alxndrsn, line 342: Confirm if this API is mentioned in PR description, or if it should be a separate PR.

lib/resources/forms.js
```diff
@@ -325,37 +325,39 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
   };
 
   formResource('/projects/:projectId/forms/:xmlFormId', (Forms, params, withXml = false, options = QueryOptions.none) =>
-    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, withXml, null, options)
-      .then(getOrNotFound));
+    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.AnyVersion, Form.withXml(withXml), options)
+      .then(getOrNotFound)
+      .then(ensureDef));
```
alxndrsn, line 330: Revert.

lib/resources/forms.js
```diff
@@ -30,7 +30,7 @@ const excelMimeTypes = {
 };
 
 const canReadForm = (auth, form) => {
-  if (form.def.publishedAt == null) {
+  if (form.def?.publishedAt == null) {
```
alxndrsn, line 33: Does this need to change?

alxndrsn: Yes.

lib/resources/forms.js
```diff
@@ -376,8 +378,8 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
   // RESTORE / UNDELETE
 
   // Instead of the xmlFormId, this endpoint uses the numeric form id in the database
-  service.post('/projects/:projectId/forms/:formId/restore', endpoint(({ Forms }, { auth, params }) =>
-    Forms.getByProjectAndNumericId(params.projectId, params.formId, false, null, QueryOptions.none, true)
+  service.post('/projects/:projectId/forms/:xmlFormId/restore', endpoint(({ Forms }, { auth, params }) =>
```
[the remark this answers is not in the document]
alxndrsn: Seems fine to change here too.

lib/model/query/forms.js
ktuite: Ooh, yes that would help a bit.

test/integration/other/transactions.js
```diff
@@ -46,7 +47,7 @@ describe('enketo worker transaction', () => {
     const { Audits, Forms, oneFirst } = container;
 
     try {
-      const simple = (await Forms.getByProjectAndXmlFormId(1, 'simple')).get();
+      const simple = (await Forms.getByProjectAndXmlFormId(1, 'simple', Form.AnyVersion)).get();
```
ktuite, line 50: Does this need to be Form.AnyVersion? I think it can be Form.WithoutDef. The audit logging only really cares about the acteeId of the form and the Forms.update call here is similarly only working with the top level form record, not the def.

ktuite (approved):
Looks good!

test/integration/other/transactions.js
alxndrsn: Done :+1:

PR #1534, opened 2025-06-20
PULL_REQUEST_TEMPLATE: update Issue ID to correct repo

alxndrsn:
Issues should now be filed in the getodk/central repo rather than in this repo.

Update the PR template to make it simpler to fill in when referencing issues stored in the expected location.


matthew-white (approved):
Thanks, @alxndrsn!

PR #1539, opened 2025-07-08
Allow data collectors/public links/app-users to upload attachment using rest endpoint

sadiqkhoja:
Closes getodk/central#1166

Manual verification + added an integration test

This implements what @lognaturel suggested during the summit. i.e. this allows users with permission to create submissions to upload corresponding attachments using REST API if attachment is not already there. I have added another condition that submission should have been created by the same user.

None.

No.

### DIFF
--- a/lib/resources/submissions.js
+++ b/lib/resources/submissions.js
@@ -480,45 +480,41 @@ module.exports = (service, endpoint) => {
 
     service.post(
       `${base}/:instanceId/attachments/:name`,
-      endpoint(({ Audits, Blobs, Forms, SubmissionAttachments, Submissions }, { params, headers, auth }, request) =>
-        Promise.all([
-          getForm(params, Forms)
-            .then(async (form) => {
-              const verbs = await auth.verbsOn(form);
-              if (!verbs.includes('submission.update') && !verbs.includes('submission.create')) {
-                throw Problem.user.insufficientRights();
-              }
-              return [form, verbs.includes('submission.update')];
-            })
-            .then(([form, canUpdateSubmissions]) => Submissions.getCurrentDefColsByIds(['id', 'submitterId'], form.projectId, form.xmlFormId, params.instanceId, draft)
-              .then(getOrNotFound)
-              .then(({ id: defId, submitterId }) => SubmissionAttachments.getBySubmissionDefIdAndName(defId, params.name) // just for audit logging
-                .then(getOrNotFound)
-                .then((oldAttachment) => [ form, canUpdateSubmissions, defId, submitterId, oldAttachment ]))),
-          Blob.fromStream(request, headers['content-type'] || defaultMimetypeFor(params.name)).then(Blobs.ensure)
-        ])
-          .then(([ [ form, canUpdateSubmissions, defId, submitterId, oldAttachment ], blobId ]) => {
-            // Authorization for the roles that can only submission.create: they can only upload
-            // attachment for their own submissions for which they haven't uploaded it before
-            const canUploadAttachment = auth.actor.isDefined() && auth.actor.get().id === submitterId && oldAttachment.blobId === null;
-            if (!canUpdateSubmissions && !canUploadAttachment) {
-              throw Problem.user.insufficientRights();
-            }
-            return Promise.all([
-              SubmissionAttachments.attach(defId, params.name, blobId),
-              Audits.log(auth.actor, 'submission.attachment.update', form, {
-                instanceId: params.instanceId,
-                submissionDefId: defId,
-                name: params.name,
-                oldBlobId: oldAttachment.blobId,
-                newBlobId: blobId
-              })
-            ]);
-          })
-          .then(([ wasSuccessful ]) => (wasSuccessful
-            ? success()
-            // should only be a Resolve[False] if everything worked but there wasn't a row to update.
-            : reject(Problem.user.notFound()))))
+      endpoint(async ({ Audits, Blobs, Forms, SubmissionAttachments, Submissions }, { params, headers, auth }, request) => {
+        const form = await getForm(params, Forms);
+        const verbs = await auth.verbsOn(form);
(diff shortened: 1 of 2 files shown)


test/integration/api/submissions.js
```diff
@@ -4452,12 +4452,53 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff,,
...
+
+      const asChelsea = await service.login('chelsea');
+
+      await asChelsea.post('/v1/projects/1/forms/binaryType/submissions/both/attachments/my_file1.mp4')
+        .set('Content-Type', 'image/jpeg')
+        .send('testimage')
+        .expect(403);
+    }));
+
+    it('should attach the given file if user has only submission.create, submission is created by the same user and attachment is not previously uploaded', testService(async (service) => {
```
alxndrsn, line 4476: Could this test also confirm the negatives in the title?

e.g.

* before chelsea adds the attachment, demonstrate that another user cannot
* after chelsea has added the attachment, demonstrate that they cannot replace it
* asChelsea

test/integration/api/submissions.js
```diff
@@ -4452,12 +4452,53 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff,,
           .send('testimage')
           .expect(404))));
 
-    it('should reject if the user cannot update a submission', testService((service) =>
-      service.login('chelsea', (asChelsea) =>
-        asChelsea.post('/v1/projects/1/forms/simple/submissions/one/attachments/file.jpg')
-          .set('Content-Type', 'image/jpeg')
-          .send('testimage')
-          .expect(403))));
```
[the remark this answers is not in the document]
sadiqkhoja: Original test was kinda wrong, it was expecting 403 when Submission doesn't exit. With the changes in this PR I'm returning 404 when Form or Submission doesn't exist. I can't keep the previous behavior because I need to check `submittedId` of the Submission before deciding whether user is authorized or not.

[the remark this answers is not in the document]
sadiqkhoja: created https://github.com/getodk/central-backend/pull/1547. I will rebase this branch once https://github.com/getodk/central-backend/pull/1547 is merged.

test/integration/api/submissions.js
```diff
@@ -4459,6 +4461,107 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff,,
...
+      await service.post(`/v1/projects/1/forms/binaryType/submissions?st=${publicLinkToken}`)
+        .send(testData.instances.binaryType.both)
+        .set('Content-Type', 'text/xml')
+        .expect(200);
+
+      await service.post(`/v1/projects/1/forms/binaryType/submissions/both/attachments/my_file1.mp4?st=${publicLinkToken}`)
+        .set('Content-Type', 'image/jpeg')
+        .send('testimage')
+        .expect(200);
+    }));
```
lognaturel, line 4536: I'm wondering whether we should also have some of the cases above for public links, in particular validating that they can't update an existing submission.

lognaturel (commented):
This doesn't change anything for the OpenRosa endpoint, right?

Looks good to me overall but I'd appreciate a final check from @matthew-white!

sadiqkhoja:
> This doesn't change anything for the OpenRosa endpoint, right?

that's right

lib/resources/submissions.js
```diff
@@ -483,24 +483,38 @@ module.exports = (service, endpoint) => {
...
-            .then((form) => auth.canOrReject('submission.update', form))
-            .then((form) => Submissions.getCurrentDefColByIds('id', form.projectId, form.xmlFormId, params.instanceId, draft)
+            .then(async (form) => {
+              const verbs = await auth.verbsOn(form);
+              if (!verbs.includes('submission.update') && !verbs.includes('submission.create')) {
+                throw Problem.user.insufficientRights();
+              }
+              return [form, verbs.includes('submission.update')];
+            })
+            .then(([form, canUpdateSubmissions]) => Submissions.getCurrentDefColsByIds(['id', 'submitterId'], form.projectId, form.xmlFormId, params.instanceId, draft)
```
matthew-white, line 493: There's something about the fact that we're getting the *current* def that's making the auth logic not crystal clear to me. The code lets a user POST a file to an empty attachment slot as long as the user created the latest version of the submission. But if there's been an edit, the user who created the latest version might be different from the user who created the original submission.

**Example 1**

1. Chelsea creates a submission. She has trouble uploading her submission attachments.
2. Alice edits an unrelated field in the submission.
3. Chelsea tries uploading her submission attachments again.
    - She isn't able to, because she is no longer the submitter of the latest version of the submission: Alice is.

**Example 2**

1. Alice creates a submission. She has trouble uploading her submission attachments.
2. Bob (a project manager) edits an unrelated field in the submission.
3. Bob is demoted to a data collector and is no longer allowed to edit submissions.
4. As a data collector, Bob tries uploading submission attachments to Alice's submission that he edited.
    - He is able to, because he is the submitter of the latest version of the submission.

---

I don't think either example is a likely case at all. Mostly I'm just thinking that it'd help with code clarity if we could crystallize the logic.

- If we're OK with the behavior in example 1: how about if the user can't `submission.create`, we check that `root` is `true` for the submission def? That would make it clear that even though we're getting the current def, the current def is also the root def, so `submitterId` definitely refers to the creator of the original submission. That would prevent the sequence in example 2.
- If we want to change the behavior in example 1, and we also want to prevent the sequence in example 2: how about we get `submitterId` off the `submissions` table rather than `submission_defs`? Then we'll always be checking the creator of the original submission, not the creator of the latest version.

matthew-white (commented):
There are some details that I'd love to discuss a little more. But the overall approach makes a lot of sense to me. :+1:

matthew-white (approved):
I like the async/await style and the new tests! 💯

PR #1540, opened 2025-07-09
datasets: sort results in code

alxndrsn:
Sorting the datasets in code gives the PostgreSQL query planner a lot more freedom.

This is OK if there aren't many datasets, and addresses poor performance observed in testing.

Sorting is also unnecessary when calling _get() for a maybeOne() request.

Closes getodk/central#


matthew-white (approved):
Makes sense to me. :+1: We don't expect there to be a large number of datasets, so it seems reasonable to sort them in JS and make things easier for Postgres.

PR #1541, opened 2025-07-09
query/sessions: defer fetching niche data until required

alxndrsn:
Performance optimisation for common session-handling code.

Existing test suite.

It's more complicated than the current implementation, but pushes niche work down to where it's used.

I haven't benchmarked it; I could certainly look into ways of doing that if it's not obviously an improvement, or doesn't seem worth the trade-off in code complexity.

This change should not affect users.  Relating to account management and authZ, it could very well break.  Hopefully this would be picked up by the test suite (which passes), or in QA.

I don't think so.


lib/resources/sessions.js
```diff
@@ -42,10 +41,10 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
   }
 
-  service.get('/sessions/restore', endpoint((_, { auth }) =>
-    auth.session
-      .map(pick(['createdAt', 'expiresAt']))
-      .orElse(Problem.user.notFound())));
+  service.get('/sessions/restore', endpoint(({ Sessions }, { auth }) => Promise.resolve(auth.session)
+    .then(getOrNotFound)
+    .then(({ token }) => Sessions.getLifetime(token)
+      .then(getOrNotFound))));
```
alxndrsn, line 47: maybe more ramdable?

e.g.

```js
=>
  auth.session
    .map(prop('token'))
    .map(Sessions.getLifetime)
    .then(getOrNotFound);
```

lib/model/query/sessions.js
```diff
@@ -29,11 +29,28 @@ returning *`)
...
   .then(map(_unjoiner)) : Promise.resolve(Option.none()));
 
+const getLifetime = (token) => ({ maybeOne }) => (isValidToken(token) ? maybeOne(sql`
+  SELECT "createdAt"
+       , "expiresAt"
+    FROM sessions
+    WHERE token=${token}
+      AND sessions."expiresAt" > NOW()
+`)
+  .then(map(Option.of)) : Promise.resolve(Option.none()));
```
alxndrsn, line 52: Is `then(...)` required?

alxndrsn: No.

lib/model/query/sessions.js
```diff
@@ -29,11 +29,27 @@ returning *`)
 
 const _unjoiner = unjoiner(Session, Actor);
 const getByBearerToken = (token) => ({ maybeOne }) => (isValidToken(token) ? maybeOne(sql`
-select ${_unjoiner.fields} from sessions
-join actors on actors.id=sessions."actorId"
-where token=${token} and sessions."expiresAt" > now()`)
+  SELECT "sessions"."csrf" AS "sessions!csrf"
+       , "actors"."id" AS "actors!id"
+       , "actors"."type" AS "actors!type"
+       , "actors"."acteeId" AS "actors!acteeId"
```
matthew-white, line 35: Maybe I'm thinking about it wrong, but I usually think of the `!` names as being an implementation detail of `unjoiner`. I always feel a little uncomfortable when I see `!` in an actual query outside `unjoiner`.

What do you think about changing `unjoiner()` so that it adds a method to its return result? Something similar to `_unjoiner.fields`, but which allows a subset of the fields to be selected.

lib/model/query/sessions.js
```diff
@@ -29,11 +29,27 @@ returning *`)
...
+    FROM sessions
+    JOIN actors ON actors.id=sessions."actorId"
+    WHERE token=${token}
+      AND sessions."expiresAt" > NOW()
+      AND actors."deletedAt" IS NULL
+`)
+  .then(map(auth => assoc('sessions!token', token, auth)))
```
matthew-white, line 42: Are you not `SELECT`ing `sessions.token` in order to avoid a performance hit? This is another place where I feel like it'd be nice to avoid referencing `!` outside `unjoiner()` unless it's really important to do so. How much do you think it helps to avoid `SELECT sessions.token`?

lib/model/query/sessions.js
matthew-white, line 35: How did you decide on these four columns out of curiosity? Were these the only four that were required for tests to pass?

lib/model/query/sessions.js
```diff
@@ -29,11 +29,27 @@ returning *`)
...
   .then(map(_unjoiner)) : Promise.resolve(Option.none()));
 
+const getLifetime = (token) => ({ maybeOne }) => (isValidToken(token) ? maybeOne(sql`
+  SELECT "createdAt"
+       , "expiresAt"
+    FROM sessions
+    WHERE token=${token}
+      AND sessions."expiresAt" > NOW()
```
matthew-white, line 50: Should we filter on `actors."deletedAt"` as well just for consistency with the other query above?

lib/model/query/sessions.js
alxndrsn: I think that would be a great improvement - I'll look into it.

lib/model/query/sessions.js
alxndrsn: That seems prudent - it wouldn't be good to restore a session for an actor that's been deleted.  Or is that an impossible scenario?

How about renaming the function from `getLifetime()` to `getValidLifetime()`?

matthew-white: > Checking `actors."deletedAt"` here is unnecessary as `/restore` already calls `getOrNotFound(auth.session)`. Do you think joining on `actors` and checking `actors."deletedAt"` is still worthwhile?

That's a good point. If we know that the token passed to `getLifetime()` has been taken from a session whose existence/validity has been confirmed, I don't feel like we need to filter on `actors."deletedAt"`. What do you think about changing the argument of `getLifetime()` so that it takes a session object rather than a token? That'd be a way to prove to `getLifetime()` that we've already confirmed the existence/validity of the session and aren't just passing in any old token.

alxndrsn: Done.

lib/model/query/sessions.js
matthew-white: I searched lib/ for references to `Sessions.getByBearerToken` and `auth.session`. I could have missed something, but these are the only necessary columns that I see:

- The session delete endpoints need `actors."acteeId"` due to `auth.canOrReject('session.end', session.actor)`
- Something definitely needs to add the token to the session, as that's referenced in several places
- lib/http/preprocessors.js checks `session.get().csrf`

matthew-white: Another option: could we forgo `unjoiner` altogether and construct the `Session` object by hand? Calling `new Session()` basically. We do that in a few places, though I wouldn't say that it's highly idiomatic for the codebase. But I think it's reasonable given how few columns this query actually needs from the `sessions` and `actors` tables.

lib/model/query/auth.js
```diff
@@ -78,7 +78,6 @@ const verbsOn = (actorId, actee) => ({ allFirst }) => {
   `);
 };
 
-
```
matthew-white, line 81: Hardly worth a mention, but a newline was removed here.

alxndrsn: Thanks!  Always worth a mention if git history is valuable.

lib/model/query/sessions.js
alxndrsn: Removing other props breaks tests:

### `actors.id`

Removing this seems to break most of the integration tests.

### `actors.type`

```
  1) api: /projects/:id/app-users
       POST
         should create a long session

  2) api: /projects/:id/app-users
       GET
         should correctly report last used in extended metadata

  3) api: /key/:key
       should passthrough to the appropriate route with successful auth

  4) api: /backup
       POST
         should return a valid zip file if the user can backup @slow

  5) datasets and entities
       GET /datasets/:name/integrity
         should return data for app-user with access to consuming Form

  6) api: /projects/:id/forms (delete, restore)
       /:id/restore (undeleting trashed forms)
         restoring access to undeleted forms
           should restore public link submission access

  7) api: /projects/:id/forms (delete, restore)
       /:id/restore (undeleting trashed forms)
         restoring access to undeleted forms
           should restore app user submission access

  8) api: /projects/:id/forms (create, read, update)
       /form-links/:enketoId/form
         should return the Form with session token queryparam

  9) api: /projects/:id/forms (create, read, update)
       /form-links/:enketoId/form
         should allow form lookup by enketoId if form has multiple published versions

  10) api: /projects/:id/forms (create, read, update)
       /form-links/:enketoId/form
         should return the Form by enketoOnceId

  11) api: /projects/:id/forms (listing forms)
       ../formList GET
         should return auth-filtered results for app users

  12) api: /projects/:id/forms (listing forms)
       ../formList GET
         should prefix returned routes for app users

  13) api: /projects/:id/forms/:id/public-links
       POST
         should allow the created user to submit to the given form

  14) api: /key/:key
       should allow cookie+public-link

  15) api: /key/:key
       should passthrough to the appropriate route with successful auth

  16) api: /sessions
       /:token DELETE
         should log the action in the audit log if it is a field key

  17) api: /sessions
       /:token DELETE
         should not log the action in the audit log for users

  18) api: /sessions
       /:token DELETE
         should log the action in the audit log for public links

  19) api: /sessions
       /:token DELETE
         should log the action in the audit log for app users

  20) api: /forms/:id/submissions
       /:instanceId/attachments/:name POST
         Public link should be able to upload attachment for the Submission

  21) api: /forms/:id/submissions
       /:instanceId/attachments/:name POST
         App user should be able to upload attachment for the Submission

  22) api: /users
       /reset/initiate POST, /reset/verify POST
         with standard uname/password auth
           should not allow password reset token replay

  23) analytics task queries
       user metrics
         should calculate number of app user per project

  24) analytics task queries
       user metrics
         should calculate public links per project

  25) analytics task queries
       submission metrics
         should calculate submissions by user type

  26) analytics task queries
       combined analytics
         should fill in all project.submissions queries
```

alxndrsn: I've removed `unjoiner`.

PR #1544, opened 2025-07-11
model/query/submissions: avoid RETURNING * where possible

alxndrsn:
`INSERT...RETURNING *` is a very convenient SQL structure, but it's often used to return values which are already available to the calling code. When this happens, there is unnecessary overhead to database communications.

There is risk to introducing this change, as:

1. new columns may be added to an `INSERT`/`UPDATE` statement without noticing that they might also be excluded from the generated list of `RETURNING` columns
2. a `BEFORE INSERT` trigger might change supplied values when returned
3. supplied value might be coerced into a different type when returned

Related: getodk/central#1170

### DIFF
--- a/lib/model/query/submissions.js
+++ b/lib/model/query/submissions.js
@@ -86,8 +86,6 @@ const createVersion = (partial, deprecated, form, deviceIdIn = null, userAgentIn
   const deviceId = applyPipe(deviceIdIn, truncateString(255), blankStringToNull);
   const userAgent = applyPipe(userAgentIn, truncateString(255), blankStringToNull);
 
-  const _unjoiner = unjoiner(Submission, Submission.Def.into('currentVersion'));
-
   // we already do transactions but it just feels nice to have the cte do it all at once.
   return one(sql`
 with logical as (update submissions set "reviewState"='edited', "updatedAt"=clock_timestamp()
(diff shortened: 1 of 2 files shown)


lib/model/query/submissions.js
```diff
@@ -22,9 +22,16 @@ const { PURGE_DAY_RANGE } = require('../../util/constants');
...
+
+const _defInsertReturnFields = withoutProvided(
+  Submission.Def,
+  'xml', 'formDefId', 'instanceId', 'instanceName', 'submitterId', 'localKey',
+  'encDataAttachmentName', 'signature', 'root', 'current', 'deviceId', 'userAgent',
+);
 const _defInsert = (id, partial, formDefId, actorId, root, deviceId, userAgent) => sql`insert into submission_defs ("submissionId", xml, "formDefId", "instanceId", "instanceName", "submitterId", "localKey", "encDataAttachmentName", "signature", "createdAt", root, current, "deviceId", "userAgent")
   values (${id}, ${sql.binary(partial.xml)}, ${formDefId}, ${partial.instanceId}, ${partial.def.instanceName}, ${actorId}, ${partial.def.localKey}, ${partial.def.encDataAttachmentName}, ${partial.def.signature}, clock_timestamp(), ${root}, true, ${deviceId}, ${userAgent})
-  returning *`;
+  returning ${_defInsertReturnFields}`;
```
sadiqkhoja, line 34: how about we just return `submissionId` and `clock_timestamp` - only values that are generated by database. 

Generally database only generates surrogate keys using `pg_get_serial_sequence` and timestamp through the application

lib/model/query/submissions.js
```diff
@@ -90,13 +95,39 @@ with logical as (update submissions set "reviewState"='edited', "updatedAt"=cloc
...
+          instanceId: partial.instanceId,
+          createdAt: submissionData.submissionCreatedAt,
+          deviceId,
+          userAgent,
+          instanceName: partial.def.instanceName,
+          current: true,
+          submitterId: submissionData.submitterId,
+        }),
+        xml: new Submission.Xml({ xml: partial.xml })
+      }));
```
sadiqkhoja, line 130: one of the reasons that we do this is that we return Submission object in the response of POST and PUT /submissions REST API, what if we don't do that and just return 201/200 with empty body just like OpenRosa? that would be breaking change but can we put this on deprecation path @lognaturel? That would certainly be performance gain and cleaner code.

lib/model/query/submissions.js
alxndrsn: > how about we just return submissionId and clock_timestamp

That's what's happening, just in a less direct way.

Do you think the code would be simpler if we declared fields we _do_ want returned rather than fields we don't?  It would certainly make this code shorter.  Maybe a neater approach would be to mark fields as `autogenerated` when declaring Frames?  E.g.

https://github.com/getodk/central-backend/blob/master/lib/model/frames/submission.js#L89-L100

Could become:

```js
  table('submission_defs', 'def'),
  'id', autogenerated,                  'submissionId',
  'formDefId',                          'submitterId',  readable,
  'localKey',                           'encDataAttachmentName',
  'signature',                          'createdAt',    readable, autogenerated,
```

alxndrsn:
> Do we need to see a tangible improvement or would we also merge this purely for efficiency-aesthetics?

I'd rather there's a demonstrable benefit, as it complicates the code.  Maybe we should discuss scale.

sadiqkhoja:
It should be demonstrable quite easily if application and database are running on different machines as it is for all our cloud servers.

yanokwa:
@brontolosone Were you able to test on dev to see if there was a demonstrable benefit?

alxndrsn:
> @alxndrsn Let's get this ready for review and merge. Less memory usage is a good win.

:+1: I think ideally there would be a less-manual notation required for this - see https://github.com/getodk/central-backend/pull/1544#discussion_r2241370139

alxndrsn:
Closed for https://github.com/getodk/central-backend/pull/1606; re-adding reviewers there.

PR #1551, opened 2025-07-16
Extract __entity and __label properties and property fields from entity forms

ktuite:
WIP for entities from repeats

Problem: Up until now, we assumed the one entity per form/submission was at the top level at `/data/meta/entity` so we didn't track a field for the entity block or label. We would just look up the entity and label fields by path. When we start tracking entities at different levels and from repeats, that's not going to keep working.

Solution: Use the existing `ds_properties` and `ds_property_fields` tables to explicitly track properties for the entity block and label. If we give these properties the property names '__entity' and '__label', we can identify them as system properties and not include them with the other properties but still make use of them behind the scenes. 


ktuite:
Incorporated into other PRs

PR #1553, opened 2025-07-18
Test forms for repeat and multi level entity test cases

ktuite:
(assembling bits and pieces of entities-from-repeats work)


matthew-white:
@ktuite, is this PR ready to be closed? Is it still being used for reference?

PR #1556, opened 2025-07-28
odata-filter: extract parse error handling

alxndrsn:
Stumbled on the abstract comment "still hate this" while wondering why there was extra whitespace added in https://github.com/getodk/central-backend/pull/1521/files#diff-90d20d47db15559988f3ded320c4c99215d0ec013538f2cbca18a0d5a2925833R101-R102.  I'm guessing the hate was for the `let` keyword, so this commit removes the `let`.

CI.

Other approach considered was just to remove the comment, but I think this approach makes the intent of the code clearer.

No change.

No.


sadiqkhoja (approved):
perfect!

PR #1562, opened 2025-07-30
Sentry: document enabled integrations

alxndrsn:
Sentry has a list of default integrations, which may change over time.

This commit adds a unit test to make enabled integrations explicit.

Also:

* make it explicit that the config module is loaded before sentry is initialised
* invert control of config for lib/external/sentry

Ref: https://github.com/getodk/central/issues/1191

This change will also lay the foundation for removing irrelevant Sentry integrations.

CI.

An alternative approach would be to use `defaultIntegrations: false` per https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/#modifying-default-integrations, and then relisting all wanted integrations.  This would have the significant disadvantage of not automatically including _new_ default integrations which Sentry might add in future.  This would mean missing out on positive new integrations unless extra care is taken when reading changelogs (if they are accurate).

No effect.

No.

### DIFF
--- a/lib/util/default-container.js
+++ b/lib/util/default-container.js
@@ -23,7 +23,7 @@ const { mailer } = require('../external/mail');
 const mail = mailer(mergeRight(config.get('default.email'), { env }));
 
 // get a sentry and configure errors.
-const { Sentry } = require('../external/sentry');
+const Sentry = require('../external/sentry').getInstance();
 Error.stackTrackLimit = 20;
 
 // get an xlsform client and a password module.
--- a/lib/bin/run-server.js
+++ b/lib/bin/run-server.js
@@ -11,7 +11,7 @@
 // dependency container and feeds it to the service infrastructure.
 
 const config = require('config');
-require('../external/sentry').init(config.get('default.sentry.external'));
+require('../external/sentry').init(config.get('default.external.sentry'));
 const exit = require('express-graceful-exit');
 
 global.tap = (x) => { console.log(x); return x; }; // eslint-disable-line no-console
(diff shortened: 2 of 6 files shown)


lib/util/default-container.js
```diff
@@ -23,7 +23,7 @@ const { mailer } = require('../external/mail');
 const mail = mailer(mergeRight(config.get('default.email'), { env }));
 
 // get a sentry and configure errors.
-const { Sentry } = require('../external/sentry');
+const Sentry = require('../external/sentry').getInstance();
```
matthew-white, line 26: I feel vaguely uncomfortable with `getInstance()` being called here. It sets up a run-order requirement that didn't exist before whereby this file has to be `require()`d after another file that has already created a Sentry instance. Could we get rid of `require('../external/sentry.js')` altogether in run-server.js, then move the Sentry `init()` here? This file already calls a bunch of other `init()` functions, so I feel like it'd be a consistent pattern.

alxndrsn: We could, but then any problems in `run-server.js` will not be captured by Sentry.

alxndrsn: Sentry docs recommend:

> [!WARNING]
> Import ... before any other modules to ensure Sentry initializes early. If you initialize later, auto-instrumentation and modules like database monitoring, agent monitoring, tracing may not work.

https://docs.sentry.io/platforms/javascript/guides/node/#apply-instrumentation-to-your-app

matthew-white: > Someone could add something to the top of run-server.js without realizing the consequences.

Wherever we init Sentry, maybe we should add a comment somewhere saying that Sentry needs to be set up first.

matthew-white (approved):
The new test looks good, and I like the inversion of control.

I still have some questions about where to call `init()`, but you have follow-up work that depends on this PR, I don't feel like those questions need to block merging.

PR #1570, opened 2025-07-31
Set PostgreSQL application name

brontolosone:
Related: getodk/central#1202

Uses the [PGAPPNAME](https://www.postgresql.org/docs/14/libpq-envars.html#id-1.7.3.22.3.4.12.1.1) environment variable that makes libpq set the [application_name](https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME) connection parameter. This makes us show up as `odkcentral` in the application name show up in the `application_name` column of the `pg_stat_activity` view, and in error messages.

Manual testing.

Not applicable.

If the user already had a `PGAPPNAME` env var set, then we roll with that. No regression risk as far as I can see.

Not applicable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 default: base
 
 NODE_CONFIG_ENV ?= test
+export PGAPPNAME ?= odkcentral
 
 node_modules: package.json
 	npm install


alxndrsn:
Setting `application_name` looks like it might help when debugging database issues if it's not clear where e.g. an open lock has come from.

> #### What has been done to verify that this works as intended?
> 
> Manual testing.

For others wanting to see the effect:

```sh
$ PGAPPNAME="odk-$RANDOM" node -e "
  const { sql } = require('slonik');
  (async () => {
    const { db } = require('./lib/util/default-container');
    const res = await db.any(sql\`
      SELECT pid, usename, application_name, query_start, query FROM pg_stat_activity WHERE state != 'idle'
    \`);
    console.log(res);
    await db.end();
  })()
"
```

> No regression risk as far as I can see.

Does this add overhead to every postgres connection?

Makefile
```diff
@@ -1,6 +1,7 @@
 default: base
 
 NODE_CONFIG_ENV ?= test
+export PGAPPNAME ?= odkcentral
```
alxndrsn, line 4: It loo looks like it will set the `application_name` for individual connections created from processes started from `make`.

I don't think this will have any effect in production.

alxndrsn (approved):
:+1:

Please note in the commit message that this is dev-only.

PR #1577, opened 2025-08-06
frames/form: add def.isDaft() & isCurrentDraft() methods

alxndrsn:
Closes https://github.com/getodk/central-backend/issues/1578

Makes explicit the different ways of checking a form is a draft.

Alternative to https://github.com/getodk/central-backend/pull/1520.

Existing tests.

This approach was suggested at https://github.com/getodk/central-backend/pull/1520#discussion_r2247058141, while reviewing alternative #1520.

No change.

No.


lib/model/frames/form.js
```diff
@@ -64,6 +64,10 @@ class Form extends Frame.define(
   get xls() { return this.aux.xls || {}; } // TODO/SL sort of a hack but let's see how it goes.
   get xml() { return this.aux.xml?.xml; }
 
+  isCurrentDraft() {
+    return this.def.id === this.draftDefId;
```
matthew-white, line 68:
```suggestion
    return this.def?.id != null && this.def.id === this.draftDefId;
```

A couple of places in the frame seem to imply that `this.def` can be `null`, so I think we should account for that. `_enketoIdForApi()` has a null check, and `forApi()` uses `?.` after `this.def`.

I also think we should be explicit about the case where `this.def` is `null` (or `this.def.id` is `null`) _and_ `this.draftDefId` is `null` (in which case, there is no current draft at all, so no def could be the current draft). In such a case, I'd want to make sure that `isCurrentDraft()` returns `false` — even if `this.def.id` and `this.draftDefId` are both nullish.

lib/model/frames/form.js
```diff
@@ -174,7 +178,7 @@ Form.ExtendedVersion = Frame.define('excelContentType', readable);
 ////////////////////////////////////////////////////////////////////////////////
 // FORM DEF
 
-Form.Def = Frame.define(
+Form.Def = class FormDef extends Frame.define(
```
matthew-white, line 181:
```suggestion
Form.Def = class extends Frame.define(
```

Not 100% sure, but maybe specifying a class name here isn't idiomatic for the codebase? E.g., I see `Form.Field = class extends Frame.define(` below, where the class doesn't have an explicit name.

lib/worker/form.js
```diff
@@ -14,7 +14,7 @@ const pushDraftToEnketo = ({ Forms }, event) =>
   Forms.getByActeeIdForUpdate(event.acteeId, undefined, Form.DraftVersion)
     .then((maybeForm) => maybeForm.map((form) => {
       // if there was no draft or this form isn't the draft anymore just bail.
-      if ((form.def.id == null) || (form.draftDefId !== form.def.id)) return;
+      if ((form.def.id == null) || !form.isCurrentDraft()) return;
```
matthew-white, line 17:
```suggestion
      if (!form.isCurrentDraft()) return;
```

If the suggested change is made from the other comment above, I think the first condition could be removed.

lib/model/frames/form.js
alxndrsn: You're correct that class names are not otherwise included for sub-classes:

```
$ git grep '= class.*extends Frame\.define'
lib/model/frames.js:Form.Attachment = class extends Frame.define(
lib/model/frames.js:Project.Extended = class extends Frame.define(
lib/model/frames.js:Submission.Attachment = class extends Frame.define(
lib/model/frames/dataset.js:Dataset.Extended = class extends Frame.define(
lib/model/frames/entity.js:Entity.Def.Source = class extends Frame.define(
lib/model/frames/form.js:Form.Extended = class extends Frame.define(
lib/model/frames/form.js:Form.Field = class extends Frame.define(
```

I will follow this idiom, but suggest changing it elsewhere.

lib/model/frames/form.js
alxndrsn: I think it would be more obvious what is being checked as:

```js
isCurrentDraft() {
  return this.draftDefId && this.def?.id === this.draftDefId;
}
```

Does that look equivalent?

matthew-white: Yes, I like that a lot. :+1:

lib/model/frames/form.js
matthew-white: > Including a name for classes is likely helpful in stacktraces.

> I will follow this idiom, but suggest changing it elsewhere.

I'd be game for us to change the idiom and give all classes a name (probably in a follow-up PR). That makes sense that that would help in stacktraces.

PR #1580, opened 2025-08-06
Feature getodk/central#1010: Added bulk delete and restore entpoints for the Entities

sadiqkhoja:
Closes getodk/central#


ktuite:
One more note, in the frontend, a little bit of work needs to be done to use these new events.

In this screenshot, the delete and restore events are weird. (I thought I changed something about this code so it would show the raw event action if it didn't know how to format it right.)

Screenshot 2025-08-08 at 4 47 08 PM

ktuite (approved):
Looks good! Please update filename of migration before merging!

PR #1582, opened 2025-08-08
http/preprocessors: simplify authBySessionToken() signature

alxndrsn:
Second arg was always called with the same, defined value.

Follow on from https://github.com/getodk/central-backend/pull/1488

CI.

* could keep as-is, but the extra indirection seems harder to follow than necessary
* could maintain `return reject(...)` instead of `throw`?  `throw` seems more idiomatic

No change.

No.

### DIFF
--- a/lib/http/preprocessors.js
+++ b/lib/http/preprocessors.js
@@ -8,7 +8,7 @@
 // except according to the terms contained in the LICENSE file.
 
 const { verifyPassword, isValidToken } = require('../util/crypto');
-const { isBlank, isPresent, noop, omit } = require('../util/util');
+const { isBlank, isPresent, omit } = require('../util/util');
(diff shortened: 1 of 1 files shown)


lib/http/preprocessors.js
```diff
@@ -108,7 +108,7 @@ const authHandler = ({ Sessions, Users, Auth }, context) => {
     if (token == null)
       return reject(Problem.user.authenticationFailed());
     // actually try to authenticate with it. reject on failure.
-    const maybeSession = authBySessionToken(token, () => reject(Problem.user.authenticationFailed()));
+    const maybeSession = authBySessionToken(token);
```
sadiqkhoja, line 111:
```suggestion
    const session = authBySessionToken(token);
```

sadiqkhoja (approved):

Issue #1439, opened 2025-03-12
Standardise PostgreSQL "current" timestamp usage


alxndrsn:
PostgreSQL offers various functions and variables for accessing the current date/time: https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT

In the codebase, 4 variants are used:

```
echo &&
  echo "  count variant" &&
  git grep -E  '(localtime\w*)|current_date|(current_time\w*)|(\bnow\(\))|(timeofday\(\))|(\w+_timestamp\(\))' |
  grep -Fv 'Date.now()' | 
  grep     -Eo '(localtime\w*)|current_date|(current_time\w*)|(\bnow\(\))|(timeofday\(\))|(\w+_timestamp\(\))' | 
  sort |
  uniq --count

  count variant
     40 clock_timestamp()
      5 current_date
      2 current_timestamp
     42 now()
```

It would be helpful to understand which variant(s) are preferred by default, and why alternatives are used in specific situations.

matthew-white:
One difference that's important is whether the current date/time changes over the course of the transaction. I believe that `current_timestamp` and `now()` are fixed during the transaction: they return the same value for all statements in the same transaction. On the other hand, `clock_timestamp()` is the actual/true date/time and does change over the course of a single transaction. Because it changes and is more accurate, I think `clock_timestamp()` is generally preferred. For example, `insert()` and `updater()` in lib/util/db.js use `clock_timestamp()` for `createdAt` and `updatedAt` columns.

Note that there are cases where we want timestamps in different columns / different statements to be the same. For example, `_createNew()` in lib/model/query/forms.js uses `clock_timestamp()` to set `form_defs."createdAt"`, then reuses the same timestamp when setting `forms."createdAt"`. It can't just call `clock_timestamp()` twice because it could return a different value each time.

matthew-white:
> `clock_timestamp()` is generally preferred.

I just wanted to emphasize the "generally" part of this. I'm sure there are cases where we actually do want a consistent date/time or a date/time closer to the start of the transaction, in which case `current_timestamp` or `now()` would make more sense. For example, we use `now()` in lib/data/odata-filter.js, and that seems correct to me: I don't think we should use `clock_timestamp()` there. That's because the user could use `now()` multiple times in the same OData expression, and we'd want that to refer to the same value each time.

I'm also noticing that many of the uses of `now()` come from tests. In most tests, I doubt it matters which timestamp we use.

[...]

PR #1586, opened 2025-08-13
allow overriding of default session lifetime through config setting

brontolosone:
Allow overriding of default session lifetime through a config setting.

Fixes getodk/central#1212

#### Notes
- The previous way the default expiry duration manifested itself in the database (as `"expiresAt" - "createdAt"`) was jittery. That was because the the expiration timestamp itself was created as a `Date` object JS-side based on _its_ wall clock time, and some indeterminate time later the creation timestamp would be set by the database based on _its_ wall clock time. Thus the interval between those would be a little bit off from the lifetime default says it should be. This also fixes that, one less surprise.
- The function signature for `sessions.create` expects a timestamp for the expiry, whereas I think specifying a definite expiry is the exceptional use case. I susped (haven't verified) that in the common case the caller would only actually care that the expiry is a certain time _offset_ from the now into the future (or past, for all I care), and should not concern itself with what time it is right now (resulting in messing around with Date objects at every call site). So it should just have to pass an offset (eg number of seconds) if the default doesn't suffice. That function signature, eg with an `expiresIn` integer parameter rather than the `expiresAt` date parameter would then also have allowed specifying our default (either the hardcoded one, or begotten from the env) as a default value for the parameter. It would make things a bit cleaner but I want to keep this PR small & simple and I don't want to go on a refactoring binge right now ;-)

Manual verification.

Previous iterations of this PR featured additional overriding via environment variables. I've removed that for now, because I don't really like the hoops one has to jump through [to get it done with node-config](https://github.com/node-config/node-config/wiki/Environment-Variables#custom-environment-variables), and an even earlier iteration's ad-hoc reading of an env var was, well, too ad-hoc.

None.

Yes. Documentation updated.

- [X] run `make test` and confirmed all checks still pass OR confirm CircleCI build passes
- [X] verified that any code from external sources are properly credited in comments or that everything is internally sourced

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -663,7 +663,7 @@ tags:
   description: |-
     This is the authentication method used by the ODK Central Frontend packaged with Central Backend. Only `User`s can authenticate this way. It consists mostly of two steps:
 
-    1. **Logging in**: presenting an Email Address and a Password for verification, after which a new `Session` is created. Associated with the Session is an expiration and a bearer token. Sessions expire 24 hours after they are created.
+    1. **Logging in**: presenting an Email Address and a Password for verification, after which a new `Session` is created. Associated with the Session is an expiration and a bearer token. By default, sessions will expire 24 hours after they are created. This can be changed through the `default.sessionLifetime` configuration setting.
     2. **Using the session**: each request to the API needs a header attached to it: `Authorization: Bearer {token}`. This authenticates that particular request as belonging to the Session we created by logging in.
 
     You might notice that Step 2 greatly resembles how OAuth 2.0 works. This was an intentional first step towards OAuth support, and should make the forward migration of your code easier down the road.
--- a/lib/model/query/sessions.js
+++ b/lib/model/query/sessions.js
@@ -14,18 +14,29 @@ const { generateToken, isValidToken } = require('../../util/crypto');
 const { unjoiner } = require('../../util/db');
 const { construct } = require('../../util/util');
 const Option = require('../../util/option');
+const config = require('config');
 
-const aDayFromNow = () => {
-  const date = new Date();
-  date.setDate(date.getDate() + 1);
-  return date;
+const create = (actor, expiresAt) => ({ one }) => {
+  const expirySqlFragment = (expiresAt instanceof Date ? sql`${expiresAt.toISOString()}` : sql`statement_timestamp() + ${config.default.sessionLifetime + ' s'}::interval`);
+  return one(sql`
+      INSERT INTO sessions (
+        "actorId",
+        token,
+        csrf,
+        "createdAt",
+        "expiresAt"
+      )
+      VALUES (
+        ${actor.id},
+        ${generateToken()},
+        ${generateToken()},
+        statement_timestamp(),
+        ${expirySqlFragment}
+      )
+      RETURNING *
+    `)
+    .then(construct(Session));
 };
-
-const create = (actor, expiresAt = aDayFromNow()) => ({ one }) => one(sql`
-insert into sessions ("actorId", token, csrf, "createdAt", "expiresAt")
-values (${actor.id}, ${generateToken()}, ${generateToken()}, clock_timestamp(), ${expiresAt.toISOString()})
-returning *`)
-  .then(construct(Session));
 
 const _unjoiner = unjoiner(Session, Actor);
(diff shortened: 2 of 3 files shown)


lib/model/query/sessions.js
```diff
@@ -14,19 +14,31 @@ const { generateToken, isValidToken } = require('../../util/crypto');
...
 const { construct } = require('../../util/util');
 const Option = require('../../util/option');
+const DEFAULT_SESSION_LIFETIME = 24 * 60 * 60;
 
-const aDayFromNow = () => {
-  const date = new Date();
-  date.setDate(date.getDate() + 1);
-  return date;
+const create = (actor, expiresAt) => ({ one }) => {
+  const SESSION_LIFETIME = parseInt(process.env.ODKCENTRAL_SESSION_LIFETIME, 10) || DEFAULT_SESSION_LIFETIME;
```
sadiqkhoja, line 20: please add this new configuration to the `default.json` and read it using `const config = require('config');`

lib/model/query/sessions.js
```diff
@@ -14,19 +14,31 @@ const { generateToken, isValidToken } = require('../../util/crypto');
...
+  const expirySqlFragment = (expiresAt instanceof Date ? sql`${expiresAt.toISOString()}` : sql`statement_timestamp() + ${SESSION_LIFETIME + ' s'}::interval`);
```
sadiqkhoja, line 21: ah another way to get timestamp in postgresql 😭

In the codebase, we already have so many ways to get timestamp see https://github.com/getodk/central-backend/issues/1439

I think `now()` or `current_timestamp()` would be better here

docs/api.yaml
```diff
@@ -653,7 +653,7 @@ tags:
   description: |-
     This is the authentication method used by the ODK Central Frontend packaged with Central Backend. Only `User`s can authenticate this way. It consists mostly of two steps:
 
-    1. **Logging in**: presenting an Email Address and a Password for verification, after which a new `Session` is created. Associated with the Session is an expiration and a bearer token. Sessions expire 24 hours after they are created.
+    1. **Logging in**: presenting an Email Address and a Password for verification, after which a new `Session` is created. Associated with the Session is an expiration and a bearer token. If a decimal integer value is present in the environment value `ODKCENTRAL_SESSION_LIFETIME`, then that value (in seconds) is taken as the session expiry. Otherwise, a session will expire 24 hours after having been created.
```
alxndrsn, line 656: If not using an env var, maybe:

```suggestion
    1. **Logging in**: presenting an Email Address and a Password for verification, after which a new `Session` is created. Associated with the Session is an expiration and a bearer token.  By default, sessions expire 24 hours after they are created.
```

lib/model/query/sessions.js
```diff
@@ -14,19 +14,32 @@ const { generateToken, isValidToken } = require('../../util/crypto');
...
+      RETURNING *
+    `)
+    .then(construct(Session));
 };
 
-const create = (actor, expiresAt = aDayFromNow()) => ({ one }) => one(sql`
-insert into sessions ("actorId", token, csrf, "createdAt", "expiresAt")
-values (${actor.id}, ${generateToken()}, ${generateToken()}, clock_timestamp(), ${expiresAt.toISOString()})
-returning *`)
-  .then(construct(Session));
```
alxndrsn, line 28: I think the only change here was `statement_timestamp()` to `clock_timestamp()`, but hard to track with the concurrent formatting improvements.

lib/model/query/sessions.js
```diff
@@ -14,19 +14,32 @@ const { generateToken, isValidToken } = require('../../util/crypto');
...
 const Option = require('../../util/option');
+const config = require('config');
+const DEFAULT_SESSION_LIFETIME = 24 * 60 * 60;
 
-const aDayFromNow = () => {
-  const date = new Date();
-  date.setDate(date.getDate() + 1);
-  return date;
+const create = (actor, expiresAt) => ({ one }) => {
+  const SESSION_LIFETIME = parseInt(process.env.ODKCENTRAL_SESSION_LIFETIME, 10) || config.sessionLifetime || DEFAULT_SESSION_LIFETIME;
```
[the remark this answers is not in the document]
alxndrsn: ...and to put everything in `.default` :shrug:

PR #1589, opened 2025-08-14
Use postgres UUID types for actee ids

alxndrsn:
Ref https://github.com/getodk/central/issues/1217

Tests etc.

There are other suggestions for extending/changing this including:

- [ ] defining the values on the db-side: https://github.com/getodk/central-backend/pull/1589/files#r2462973088
- [ ] define `uuidFor()` in postgres (TODO can't find the link for this)
- [ ] get rid of all the special actees except `*` (https://github.com/getodk/central-backend/pull/1589#issuecomment-3477413679)

* maybe a slow migration
* maybe cases where non-UUIDs have been used; hopefully caught in QA

I don't think so.


lib/model/migrations/20250212-01-actee-id-as-uuid.js
```diff
@@ -0,0 +1,81 @@
...
+  config:      '00000000-0000-0000-0000-000000000008',
+  project:     '00000000-0000-0000-0000-000000000009',
+  role:        '00000000-0000-0000-0000-000000000010',
+  assignment:  '00000000-0000-0000-0000-000000000011',
+  audit:       '00000000-0000-0000-0000-000000000012',
+  system:      '00000000-0000-0000-0000-000000000013',
+  singleUse:   '00000000-0000-0000-0000-000000000014',
+  dataset:     '00000000-0000-0000-0000-000000000015',
+  public_link: '00000000-0000-0000-0000-000000000016',
+};
```
brontolosone: we could define a view that stores these constants? Or a table?

alxndrsn: ~Or an enum?~ doesn't seem to make sense in postgres...

how would a view work?!

brontolosone: And perhaps these constants should be v8 UUIDs so that they'll be in a distinct number space from the v4 UUIDs that are likely going to go into this table?

brontolosone: > It looks like they're currently "version 0",

I don't think there's a version 0, having the version bits set to 0 means (along with some other values for those version bits) "version undefined" which one could cautiously interpret as "reserved for future use".

> so they'll certainly be distinct from any v4 UUIDs going into the table.

These _particular_ ones are distinct from any v4 UUIDs, yes. Nothing is said about what other UUID(-ish) values we'd use in the future for additional constants. But you could say such a thing, for instance by choosing a version, and then the space set apart for exactly these kind of purposes through a v8 UUID would be an obvious candidate, more obvious than squatting not-yet-used-space.

In practice it doesn't matter and no harm will come from leaving this as-is. I'm just saying that v8 UUIDs exist for this use case.

Another thing we could do if we want to go UUID-rogue and don't care about versions and variants is to use constants _derived_ rather than _declared_, like this:

```
select (md5('public_link'))::uuid
+--------------------------------------+
| md5                                  |
|--------------------------------------|
| 230af3da-42d0-8d6f-5f14-57ff1764c774 |
+--------------------------------------+
```
Amusing but not that practical.

Another amusing idea is to embed the ascii values of the "special actees" strings in the UUID, so `role` could for instance become `00000000-0000-0000-0000-0000726f6c65`. Handy if we lose our table of constants, we can just reconstruct it :laughing:

matthew-white:
> I think this PR has value!

I agree that it's reasonable to use the `uuid` type for actee IDs, I support that. :+1: It's hard for me to know how much improved performance to expect. But we know there should be some, and the changes in this PR aren't too big. There's a migration and a change to queries that reference actees that are species or `*`.

I also wanted to link this PR to the discussion at #1217. In particular, I still think that we could take this opportunity to do away with species-level actees entirely. Instead of reifying them, we could get rid of them. We assign actors access to specific actees (like a specific project), and we assign actors access to `*`, but I don't know of any cases where we assign actors access to species-level actees. Actors end up with access to species-level actees only insomuch as they are given access to `*`.

In other words, I think we could make do with just one special UUID (for `*`) instead of one for each species.

Concretely, I'm proposing that we change lines like

https://github.com/getodk/central-backend/blob/7739129581f690a7fd51eca436d2f37bf544b7a7/lib/resources/users.js#L37

to:

```js
auth.canOrReject('user.create', '*')
```

I really don't think this would be a huge code change. I'd be happy to create a draft PR along those lines if there's interest.

matthew-white:
And just to be clear, I'm not suggesting that we get rid of the concept of species entirely. For example, the `actees.species` column can stay put. I just think that we don't need each species to correspond to a row of the `actees` table.

lib/model/migrations/20250212-01-actee-id-as-uuid.js
alxndrsn: >> "version 0"
> I don't think there's a version 0

Agreed - that's what I was trying to convey with the quotation marks.

> I'm just saying that v8 UUIDs exist for this use case.

:+1: seems like a good case for using them.

> Another thing we could do if we want to go UUID-rogue and don't care about versions and variants
> ...
> `select (md5('public_link'))::uuid`

As you say, this doesn't have the advantage of guaranteeing valid UUIDs.  Also the generated values aren't visibly special, so I think we can discount them.

> Another amusing idea is to embed the ascii values

I think it's still helpful to have values which are visibly special values to humans, and `726f6c65` may not jump out as ASCII/UTF-8 hex to many.

I'm happy to switch to using v8.

test/integration/other/knex-migrations.js
```diff
@@ -954,7 +954,7 @@ testMigration('20240215-01-entity-delete-verb.js', () => {
   }));
 });
 
-testMigration('20240215-02-dedupe-verbs.js', () => {
+testMigration.skip('20240215-02-dedupe-verbs.js', () => {
```
alxndrsn: confirm why these are skipped, and add comment explaining.

test/integration/other/knex-migrations.js
```diff
@@ -840,7 +840,7 @@ describe('database migrations from 20230512: adding entity_def_sources table', f
   }));
 });
 
-describe('database migrations from 20230802: delete orphan submissions', function test() {
+describe.skip('database migrations from 20230802: delete orphan submissions', function test() {
```
alxndrsn: confirm why these are skipped, and add comment explaining.

test/integration/other/knex-migrations.js
```diff
@@ -608,7 +608,7 @@ describe.skip('database migrations from 20230406: altering entities and entity_d
   }));
 });
 
-describe('database migrations from 20230512: adding entity_def_sources table', function () {
+describe.skip('database migrations from 20230512: adding entity_def_sources table', function () {
```
alxndrsn: confirm why these are skipped, and add comment explaining.

test/integration/other/knex-migrations.js
```diff
@@ -384,7 +384,7 @@ describe('database migrations: intermediate form schema', function() {
   }));
 });
 
-describe('database migrations: 20230123-01-remove-google-backups', function() {
+describe.skip('database migrations: 20230123-01-remove-google-backups', function() {
```
alxndrsn: confirm why these are skipped, and add comment explaining.

test/integration/other/knex-migrations.js
```diff
@@ -255,7 +255,7 @@ describe('database migrations: removing default project', function() {
   }));
 });
 
-describe('database migrations: intermediate form schema', function() {
+describe.skip('database migrations: intermediate form schema', function() {
```
alxndrsn: confirm why these are skipped, and add comment explaining.

alxndrsn:
> Similarly, for API keys ([getodk/central#1335](https://github.com/getodk/central/issues/1335)), I expect API keys to provide access to specific projects or to a select list of verbs on `*`, but I don't expect API keys to need to do anything with species actees.

Would this be an example where an API key would want species-level actee permissions: a system to sync odk-central users with a 3rd-party IAM system?  E.g. if an organisation wants to manage their user accounts for ODK and other systems centrally.

matthew-white:
Regression testing for v2025.4 is currently planned for Wednesday, so I'm going to go ahead and bump this to v2026.1.

PR #1592, opened 2025-08-18
docs: fix dataset/.../integrity operationId

alxndrsn:
The previous value seemed to be copy-pasted from a different operation.

Related: https://github.com/getodk/central/issues/1220

Nothing - it's a text change though, so should be ok.

What is `operationId` for?  Is there any point in duplicating `summary` everywhere?

Should make docs YAML more parseably by different OpenAPI tools.

It actually does update the docs :ballot_box_with_check: 


jdimeo:
`operationId` affects OpenAPI code generation (classes and methods that get generated) leading to duplicates/conflicts in code naming

jdimeo:
Only a usability problem :-) Here's the Java class that was generated:

image

PR #1594, opened 2025-08-22
Test the tests for invalid xml

brontolosone:
Following up on several discussions on Slack.

We had a whole bunch of form test XML in our tests that wasn't actually valid XML, which we used to get away with since the in-Central parser is very, very lenient.
The problem is that then when one swaps out that parser for an XML parser that expects, well, XML, one runs into problems with that faulty test data.

Thus, the faulty testdata needed fixing, and in the process I thought it would be good to add an automatic test for our test data with which we test Central :laughing: 

Goals:
- automatically verify at least the form XML we have in `test/data/xml.js`
- provide a function to do such validation going forwards for other xml documents we use in tests outside of that `test/data/xml.js`. We might want to say something in review if we see new tests that contain handcrafted XML; it would be good if that XML itself is tested because it's quite easy to make little mistakes. Even not closing a tag - violating the most basic form of well-formedness - doesn't alarm the in-Central parser (bbf6213). Example of making sure ones test XML is actually XML in 54cde7b7.
- fix up the test XML so that it passes the failing test

Non-goals that had to be accomplished:
- deal with the fallout: there's a few tests that started to fail once I *fixed* the *invalid* xml. I just updated the (overspecific) assertions to match (58c1036).

Ran the new test, fixed the broken xml found with the new test, ran the tests, fixed a number of newly failing tests that shouldn't have been failing, ...

To check whether XML is actually XML, it's easiest to just feed it to an XML parser and see what it makes of it. I picked [libxmljs](https://libxmljs.github.io/libxmljs/index.html) which is just a bunch of bindings to the venerable `libxml2` library which I've worked with before (and is also used by Postgres). The downside of that is that upon install of dev depencies, a compiler runs to compile the bindings. For me it worked without a hitch but then again on my system I have compilers and header files for both libxml2 and node, perhaps the install procedure is less pleasant on a consumer-oriented system. Or maybe not. I'm not so familiar with the node-gyp stuff. It runs on CircleCI without problems but let's hear it if any developer runs into any problems with this.

It's a development thing, it doesn't affect users.

No.

- [X] run `make test` and confirmed all checks still pass OR confirm CircleCI build passes
- [X] verified that any code from external sources are properly credited in comments or that everything is internally sourced

### DIFF
--- a/test/data/xml.js
+++ b/test/data/xml.js
@@ -1118,8 +1118,8 @@ module.exports = {
       two: '<data id="audits"><meta><instanceID>two</instanceID><audit>log.csv</audit></meta><name>Bob</name><age>34</age></data>'
     },
     selectMultiple: {
-      one: instance('selectMultiple', 'one', '<q1>a b</q1><g1/><q2>x y z</q2>'),
-      two: instance('selectMultiple', 'two', '<q1>b</q1><g1/><q2>m x</q2>'),
+      one: instance('selectMultiple', 'one', '<q1>a b</q1><g1><q2>x y z</q2></g1>'),
+      two: instance('selectMultiple', 'two', '<q1>b</q1><g1><q2>m x</q2></g1>'),
       three: instance('selectMultiple', 'three', '<q1> b c</q1>')
     },
     simpleEntity: {


matthew-white:
I'm happy to re-review since I already left a review.

test/integration/other/select-many.js
```diff
@@ -87,15 +88,17 @@ describe('select many value processing', () => {
...
         <h:head>
           <model>
             <instance>
               <data id="selectMultiple">
                 <meta><instanceID/></meta>
                 <q1/>
-                <g1><q2/></g1>
+                <g1>
+                  <q2/>
+                </g1>
```
matthew-white, line 101: Is this a white space change only? If so, I think it'd be ideal to do that in a separate PR.

package.json
```diff
@@ -58,6 +58,7 @@
     "eslint-plugin-no-only-tests": "~3.1",
     "fetch-cookie": "^2.2.0",
     "http-proxy-middleware": "^2.0.9",
+    "libxmljs": "^1.0.11",
```
matthew-white, line 61: I'm starting to feel less sure about adding this dependency. I wrote above that I felt fine with us adding another XML parsing dependency even though we have others, and I still feel that way. However, there seem to be issues with this specific dependency:

- No release in almost two years. No answer to a question about the status of the project: libxmljs/libxmljs#664
- Reports of vulnerability alerts: libxmljs/libxmljs#661. @brontolosone, what do you see when you run `npm audit`? Any alerts about this dependency or its subdependencies?

There have also been reports that Node 22 isn't supported: libxmljs/libxmljs#664. I'm not as worried about this one though, since we're already on Node 22, and like you said, things work in CircleCI.

Something you said gave me an idea though, I wonder what you think about it:

> The nice thing about `libxmljs` is that it's just straight bindings to the venerable `libxml2` library which is what Postgres and many, many other things use. So I have good hopes it's going to trip up over the same things that Postgres will trip up over.

If the main goal of `isPalatableXML()` is to verify that the given XML will be acceptable to Postgres, what about the idea of using Postgres to complete that check? I can see some minor downsides to that approach, including that it'd force `isPalatableXML()` to become async. But there are also advantages: (1) it guarantees that the XML parsing will be done in the exact same way as Postgres, and (2) it avoids an additional dependency. Specifically, I'm thinking of a function like:

```js
const isPalatableXML = async (container, allegedXML) => {
  await container.run(sql`SELECT ${allegedXML}::xml`);
  return true;
};
```

This change requires `isPalatableXML()` to receive the `container` as an argument. However, almost every test that calls `isPalatableXML()` or its cousin `palatableXML()` has access to `container` via `testService()`. The main exception is test/data/test-xml.js, but I think that could be changed to use `testService()` or `testContainer()`. Either one would provide access to `container`.

matthew-white (changes requested):
I'm feeling some reservations about the `libxmljs` dependency: see below. I'm suggesting an alternative, curious what you think. I've done a few rounds of review on this PR, so my preference would be to hold off on merging and to wait until I'm able to re-review again.

package.json
matthew-white: I could be open to `libxmljs`, but I have some questions and another idea I want to run by you. I also just left a high-level question below.

matthew-white:
@brontolosone, we still want to merge this PR, right? I'm thinking to take another look at this PR this week.

test/data/xml.js
```diff
@@ -1117,12 +1118,12 @@ module.exports = {
...
     selectMultiple: {
-      one: instance('selectMultiple', 'one', '<q1>a b</q1><g1><q2>x y z</q2>'),
-      two: instance('selectMultiple', 'two', '<q1>b</q1><g1><q2>m x</q2>'),
+      one: instance('selectMultiple', 'one', '<q1>a b</q1><g1><q2>x y z</q2></g1>'),
+      two: instance('selectMultiple', 'two', '<q1>b</q1><g1><q2>m x</q2></g1>'),
       three: instance('selectMultiple', 'three', '<q1> b c</q1>')
     },
     simpleEntity: {
-      one: `<data xmlns:jr="http://openrosa.org/javarosa" xmlns:entities="http://www.opendatakit.org/xforms" id="simpleEntity" version="1.0">
+      one: `<data xmlns:jr="http://openrosa.org/javarosa" xmlns:entities="http://www.opendatakit.org/entities" id="simpleEntity" version="1.0">
```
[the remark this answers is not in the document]
lognaturel: That's the correct one: https://getodk.github.io/xforms-spec/entities#namespacing

package.json
brontolosone: - Let's keep it at that we model threats (and their priors) differently :thinking: 
- The `afterEach()` hook idea, initially I liked it, yet I think that nothing beats a straight up `isPalatableXML()` that one can run wherever, whenever, decoupled from the DB or the testing harness control flow.
- To remove controversy and get this merged I'll just remove the `isPalatableXML()`.   
  - I'll add a follow-up PR that stops un-xml at DB insertion time, when forms and submissions are uploaded.

matthew-white: > I'll add a follow-up PR that stops un-xml at DB insertion time, when forms and submissions are uploaded.

That sounds excellent to me. :+1: It'll help to know that users aren't uploading invalid XML.

> Let's keep it at that we model threats (and their priors) differently 🤔

I'm open to having my model changed, but it sounds like we have a good path forward on this one. Happy to chat more about this on a call sometime!

PR #1597, opened 2025-08-26
Part A: Backport some ideas for multi-entities to the single entity scenario

ktuite:
Related to https://github.com/getodk/central/issues/1306

Backporting some ideas for multi-entities to the single entity scenario for cleaner code review.

Adds `__entity` and `__label` internal system dataset properties and uses those as form fields for submission parsing instead of looking up those form fields by path.

Would also need to include this PR with migration to add these properties to historical entity forms: https://github.com/getodk/central-backend/pull/1552

### DIFF
--- a/lib/data/dataset.js
+++ b/lib/data/dataset.js
@@ -81,6 +81,20 @@ const validatePropertyName = (name) => {
   return (match !== null);
 };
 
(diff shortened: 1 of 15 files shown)


ktuite:
closed in favor of https://github.com/getodk/central-backend/pull/1616

PR #1600, opened 2025-08-27
Feature: New endpoint to get list of creators for the entities

sadiqkhoja:
this endpoint facilitates 'created by' filter on the frontend (getodk/central#973)

Added a tests and manual verification

Follows same approach as getting list of submitters.

No

Done

### DIFF
--- a/lib/resources/entities.js
+++ b/lib/resources/entities.js
@@ -29,7 +29,7 @@ module.exports = (service, endpoint) => {
 
     const dataset = await Datasets.get(params.projectId, params.name, true).then(getOrNotFound);
 
-    await auth.canOrReject('entity.read', dataset);
+    await auth.canOrReject('entity.list', dataset);
 
     return Entities.getAllCreators(dataset.id);
   }));
--- a/lib/resources/entities.js
+++ b/lib/resources/entities.js
@@ -25,6 +25,15 @@ module.exports = (service, endpoint) => {
     return Entities.getAll(dataset.id, queryOptions);
   }));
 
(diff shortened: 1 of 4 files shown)


lib/resources/entities.js
```diff
@@ -25,6 +25,15 @@ module.exports = (service, endpoint) => {
     return Entities.getAll(dataset.id, queryOptions);
   }));
 
+  service.get('/projects/:projectId/datasets/:name/entities/creators', endpoint(async ({ Datasets, Entities }, { params, auth }) => {
+
+    const dataset = await Datasets.get(params.projectId, params.name, true).then(getOrNotFound);
+
+    await auth.canOrReject('entity.read', dataset);
```
matthew-white, line 32: I feel like this should be `entity.list`. The endpoint is returning data from multiple entities. I think of `entity.read` as being about a single entity. I know we use `submission.read` for the .../submitters endpoint, but I actually think there are issues in that area: I've filed the related getodk/central#1326.

test/integration/api/entities.js
```diff
@@ -4008,4 +4008,148 @@ describe('Entities API', () => {
...
+        .then(({ body }) => {
+          body.should.be.an.Array();
+          body.length.should.equal(2);
+
+          // Should be sorted by displayName
+          body[0].displayName.should.equal('Alice');
+          body[0].type.should.equal('user');
+          body[0].should.have.property('id');
+          body[0].should.have.property('createdAt');
+          body[0].should.have.property('updatedAt');
```
matthew-white, line 4073: How about using `should.be.an.Actor()`? Maybe that would reduce the number of assertions needed here.

matthew-white:
I reviewed the PR on the frontend, so I thought I'd go ahead and quickly review this one. ✅

PR #1601, opened 2025-08-29
test/api/users: fix typo: sidewide -> site-wide

alxndrsn:
Noted while considering https://github.com/getodk/central/issues/1315#issuecomment-3234750850.

I read the context of the verb loading, and they look to be site-wide.

There are probably syonyms to site-wide, including "sitewide".  Both are already in use in this repo.

No effect.

No.


matthew-white (approved):
At least in the UI, we spell "sitewide" without a hyphen. I personally kind of prefer the hyphen though. In any case, I don't think hyphen vs. not really matters for Backend. Both are definitely preferable to "si**d**ewide"!

PR #1602, opened 2025-08-29
/users/current: add session details

alxndrsn:
Part of https://github.com/getodk/central/issues/1315: aiming to reduce chained requests for already-authenticated returning web users by one.

A couple of new tests!

* could include in non-extended response, but it looks like frontend requests the extended response
* should `/sessions/restore` be documented as deprecated?

Should make frontend a little faster for them sometimes.

Docs have been updated.


matthew-white (approved):
Looks great!

alxndrsn:
> I probably won't get to it until I've made more progress on maps

I'm happy to hold off merging until frontend code is ready.

PR #1606, opened 2025-09-05
Submissions.newVersion(): reduce repeated data

alxndrsn:
1. only return data from the db which is not already available
2. reduce nested queries
3. only update the `current` row in submission_defs
4. add extra JOIN on submission_defs to return root row's userAgent (necessary to achieve 4)

Neater rewrite of https://github.com/getodk/central-backend/pull/1544
Related: https://github.com/getodk/central/issues/1170

CI!

Discussion at https://github.com/getodk/central-backend/pull/1544; this is a neater rewrite of that, specifically tackling the `Submissions.createVersion()` function, where the `xml` field was previously passed to the database and then returned.

This PR could be changed to return query formatting to match the original more closely.  To me this makes understanding the query significantly harder.

Potential follow-ups:

* Look at refactoring `Submissions.createNew()`.  It looks like the query can be inverted and simplified.

Should not affect users.

No.

### DIFF
--- a/lib/model/query/submissions.js
+++ b/lib/model/query/submissions.js
@@ -79,24 +79,55 @@ const createVersion = (partial, deprecated, form, deviceIdIn = null, userAgentIn
   const deviceId = applyPipe(deviceIdIn, truncateString(255), blankStringToNull);
   const userAgent = applyPipe(userAgentIn, truncateString(255), blankStringToNull);
 
-  const _unjoiner = unjoiner(Submission, Submission.Def.into('currentVersion'));
-
   // we already do transactions but it just feels nice to have the cte do it all at once.
   return one(sql`
-with logical as (update submissions set "reviewState"='edited', "updatedAt"=clock_timestamp()
-  where id=${deprecated.submissionId} 
-  returning * )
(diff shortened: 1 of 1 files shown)


lib/model/query/submissions.js
```diff
@@ -79,24 +79,55 @@ const createVersion = (partial, deprecated, form, deviceIdIn = null, userAgentIn
...
+         , submission_defs."userAgent"
+         , newDef.*
+      FROM updatedSubmission
+      JOIN submission_defs
+        ON "submissionId"=${deprecated.submissionId} AND root IS TRUE
+      CROSS JOIN newDef
+  `)
+    .then(({ submissionDefId, submissionDefCreatedAt, ...submissionData }) => // TODO/HACK: reassembling this from bits and bobs.
+      new Submission({ id: deprecated.submissionId, ...submissionData }, {
+        currentVersion: new Submission.Def({
```
sadiqkhoja, line 113: Can we disable these eslint rules for the block of code instead of each line?

```
/* eslint-disable key-spacing, no-multi-spaces */
....
....
/* eslint-enable key-spacing, no-multi-spaces */

PR #1611, opened 2025-09-08
model/query: remove TODOs expecting unjoiner

alxndrsn:
Various comments in `lib/model/query/` imply that `unjoiner()` is the one true way to map SQL query results to Javascript objects.

`unjoiner()` is inflexible and does not currently support approaches which significantly reduce database traffic.

This PR removes the TODO/HACK comments.

Noted in review at https://github.com/getodk/central-backend/pull/1608/files#r2327615484

Closes getodk/central#


matthew-white (approved):
I'm also of the opinion that `unjoiner()` is not the one true way. Probably the default, yes, but not the only way. I think the limitations of `unjoiner()` have become clearer over time.

PR #1616, opened 2025-09-08
Process entities from repeats in submissions

ktuite:
Closes https://github.com/getodk/central/issues/1306

This is an alternate solution to these two previous versions of "entities with repeats"
* https://github.com/getodk/central-backend/pull/1554
* https://github.com/getodk/central-backend/pull/1597 + https://github.com/getodk/central-backend/pull/1607

Instead of tracking where the entity block in a dataset is by storing new properties, it stores the root path of the entity in the `dataset_form_defs` table and then calculates the `path` + `meta/entity` and `path` + `meta/entity/label` to get those fields. The default path value is root `/`, e.g. for existing entity forms where the entity block was in a meta block at the root level.

Some other big ideas
* API about a dataset now returns `repeatPath` field in the `sourceForms` list when it was in a repeat group in the form.
* Entities from the same submission share a source, which is done by creating a new source object up front when processing a submission. This has some awkward bits like creating a source even if the submission fails to process the entity, or if a submission gets put in the backlog for processing later. 

Issues or ideas that come up along the way but could be dealt with in followup PRs are tracked here:
https://github.com/getodk/central/issues/1349

Tests of existing code and new tests.

See summary of [discussion](https://github.com/getodk/central/issues/1306#issuecomment-3260019623) of data model in Part A above that added `__entity` and `__label` (and even `__repeat`) special properties.

The approach in this PR is quite a bit cleaner. It leaves the meaning of dataset properties as it was, doesn't require a migration to backfill data, and it wasn't too bad to use the dataset path and repeat info, now stored in `dataset_form_defs`, to do things like get all the relevant form fields for parsing a submission and getting information about which repeat group a dataset is from in a specific form.

Introduces backend component of entities from repeats. 

The documentation about dataset metadata and source forms should probably change to explain a new `repeatPath` field in the dataset property source forms list, if we capture that in the API right now.


ktuite:
@sadiqkhoja I've added more tests and improved a lot of little things.

The last test I added [here](https://github.com/getodk/central-backend/pull/1616/commits/67deb160d20533a8c59858487c8a3add702bc448#diff-26a98ca41d5538c8060ed19eb66703fa007a44f187b8d303da7fb73d156c2562R395) means I need to tweak the submission parsing code a bit. It currently can travers into a repeat group if that's where the entity declaration is, but kinda breaks down otherwise, so your repeat / group / repeat / group  test suggestion was a good one. I'm pretty sure I can fix it. If you wanted to start re-reviewing everything but that last bit, that would be great! 

I've also put this code up on the dev server.

sadiqkhoja:
```diff
- The default path value is empty string ``
+ The default path value is root `/`
```

PR #1623, opened 2025-09-28
Backfill submission geocache, fixes getodk/central#1373

brontolosone:
Closes getodk/central#1373

Manual testing.

Batching is used because it gives the user some sort of progress meter, which is nice if there's a lot to process and/or their DB is slow. They'll continue to receive visual feedback that there's nothing stuck.

N/A

N/A

### DIFF
--- a/lib/model/migrations/20250928-01-backfill-submission-geocache-createfunction-01.up.sql
+++ b/lib/model/migrations/20250928-01-backfill-submission-geocache-createfunction-01.up.sql
@@ -1,0 +1,85 @@
+
+--- drop: cache_all_submission_geo(only_default_path boolean, batchsize integer) ---
+DROP FUNCTION IF EXISTS "public"."cache_all_submission_geo"() CASCADE;
+
+--- create: cache_all_submission_geo(only_default_path boolean, batchsize integer) ---
+CREATE FUNCTION "public"."cache_all_submission_geo"(only_default_path boolean = false, batchsize integer = NULL)
+RETURNS integer[]
+AS
+    $BODY$
+    WITH extraction_todo AS (
+        SELECT
+            sdef.id as submission_def_id,
+            ffgeo.fieldhash,
+            ffgeo.type as geotype,
+            ffgeo.repeatgroup_cardinality > 0 as isrepeatable,
+            ffgeo.path as fieldpath
+        FROM
+            submission_defs sdef
+            INNER JOIN form_defs fdef ON
+                sdef.current
+                AND
+                (fdef.id = sdef."formDefId")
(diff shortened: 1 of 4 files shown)


lib/model/migrations/20250928-01-backfill-submission-geocache-createfunction-01.up.sql
```diff
@@ -0,0 +1,85 @@
+
+--- drop: cache_all_submission_geo(only_default_path boolean, batchsize integer) ---
+DROP FUNCTION IF EXISTS "public"."cache_all_submission_geo"() CASCADE;
+
+--- create: cache_all_submission_geo(only_default_path boolean, batchsize integer) ---
+CREATE FUNCTION "public"."cache_all_submission_geo"(only_default_path boolean = false, batchsize integer = 9223372036854775807)
```
sadiqkhoja, line 6: can we set the default value of `batchsize` to NULL? I am thinking about cases where selected rows are greater than 9223372036854775807. From the docs:

> LIMIT ALL is the same as omitting the LIMIT clause, as is LIMIT with a NULL argument.

brontolosone: Ah that's handy, `LIMIT NULL`.

lib/model/migrations/20250928-01-backfill-submission-geocache-createfunction.js
```diff
@@ -0,0 +1,43 @@
...
+// the LICENSE file found in the top-level directory of this distribution and at
+// https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+// including this file, may be copied, modified, propagated, or distributed
+// except according to the terms contained in the LICENSE file.
+
+const fs = require('node:fs');
+const path = require('path');
+
+
+function getSqlFiles(upOrDown) {
```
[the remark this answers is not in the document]
brontolosone: I know, and considered. Most poignant consideration:

Migrations *at any time since their conception* must represent operations that took place with the then-code of when they were applied, yet they can also be run with any then...now-code. Corollary: the migration code shouldn't change (or should only be changed very carefully) along with the rest of the source code. Maybe it should even be "hermetic" if we want to get into fancywordplay. Anyway, extracting this function as you propose makes for a mantrap that's all too easy to fall into, in this respect.

Another consideration is to maintain maximum locality for the hacky workaround for knex's `.raw()` mangling inputs. If you have a problem with difference between what you gave in the honest `.sql` files (your intent) and what actually gets sent to the DB by the migration funhouse, then you can see (and manipulate/adjust) right there, without cascading effects throughout.

PR #1793, opened 2026-04-02
convertToSubmissionVersion: don't support case-insensitive URLs

alxndrsn:
Per https://github.com/getodk/central-backend/pull/1786, URLs are now expected to be case-sensitive.  This commit removes special handling for case-insensitivity introduced in eb789c779dd9361f01cb0d40b4e00faf56864a7c / #1763.

Inverted an existing test.

* simplifies code
* more consistent with other route handling

Maybe a surprise if relying on case-insensitivity in this specific endpoint.  #1786 may have been more surprising.

No.


lib/util/http.js
```diff
@@ -141,7 +141,7 @@ const convertToSubmissionVersion = (urlString, instanceId, rootId) => {
   while (parts.length > 1) {
     const part = parts.shift();
     const nextPart = parts[0].replace(/(.)\.xml$/i, '$1');
```
sadiqkhoja: should we remove `/i` from here as well?

Following test fails (only XML is capitalized):

```
it.only('should not support case-insensitive URLs', () => {
      // given
      const url = '/v1/projects/1/forms/simple/submissions/instanceId.XML?foo=bar';

      // expect
      should.throws(() => convertToSubmissionVersion(url, 'instanceId', 'rootId'));
    });
```

alxndrsn: 👍 good catch

alxndrsn: Fixed.

sadiqkhoja (approved):

PR #1800, opened 2026-04-06
Fixes: Exclude deleted properties from the SQL query for geojson

### LINKED ISSUE #1769: Duplicate in the Entity GeoJSON representation API response
Hello!

I just installed version **2026.1.0** and it seems there is a duplicate in the API response that retrieves the geometry in GeoJSON format.

The endpoint `GET /v1/projects/{projectId}/datasets/{name}/entities/{uuid}/geojson` returns a FeatureCollection type **with 2 features of the same geometry**.

Similarly, when entities are displayed on the map from the front, the entity counter **counts twice as many entities**.

Thank you for all the work you've done!

sadiqkhoja:
Closes getodk/central#1769

Added tests and manual verification

Fixes the issue. Other option is to remove the join with `ds_properties` altogether because I don't think it is possible to create an entity with the data that includes non-existent properties.

I think similar fix is required in `getDatasetProperties` of `analytics.js`. Will create a separate PR for that if required.

No

No

### DIFF
--- a/lib/model/query/geo-extracts.js
+++ b/lib/model/query/geo-extracts.js
@@ -184,6 +184,10 @@ const getEntityFeatureCollectionGeoJson = (datasetPK, odataQuery, limit=null, se
                 ds.id = ${datasetPK}
                 AND
                 (dsprops.name, dsprops."datasetId") = ('geometry', ds.id)
+                AND
+                dsprops."deletedAt" IS NULL
+                AND
+                dsprops."publishedAt" IS NOT NULL
               )
               INNER JOIN entities ON (
                   entities."datasetId" = ds.id
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -8437,6 +8437,48 @@ describe('datasets and entities', () => {
         });
     }));
 
+    // Bug getodk/central#1769
+    it('should not return duplicate features when geometry property has been soft-deleted and re-created', testService(async (service) => {
+      const asAlice = await service.login('alice');
+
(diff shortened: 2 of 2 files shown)


lib/model/query/geo-extracts.js
```diff
@@ -184,6 +184,10 @@ const getEntityFeatureCollectionGeoJson = (datasetPK, odataQuery, limit=null, se
                 ds.id = ${datasetPK}
                 AND
                 (dsprops.name, dsprops."datasetId") = ('geometry', ds.id)
+                AND
+                dsprops."deletedAt" IS NULL
+                AND
+                dsprops."publishedAt" IS NOT NULL
```
sadiqkhoja, line 190: adding `publishedAt` clause is not strictly necessary to add. I couldn't think of any case where it might cause similar problem because we don't have ability to unpublish a property. At the same time I think there is no harm in adding the clause here. I am open to remove it if there is any objection.

ktuite (approved):

sadiqkhoja:
talked with @ktuite on Slack and decided that we don't need to change analytics queries for now.

sadiqkhoja:
Waiting for @brontolosone 's opinion on removing `ds_properties` join, otherwise this PR is ready to be merged.

lib/model/query/geo-extracts.js
brontolosone: Better safe than sorry. If the DB allows for expressing something, it's worth guarding against it if it'd cause problems.

PR #1801, opened 2026-04-13
tasks: tag Sentry events with task name

### LINKED ISSUE #1804: Associate Sentry reports with specific task
Right now, we wrap all task code in a Sentry scope of `task` (code [here](https://github.com/getodk/central-backend/blob/142c1c85501d90d225116862ba79dd914c5fcfc1/lib/task/task.js#L58)). However, as @alxndrsn [writes](https://github.com/getodk/central/issues/1789#issuecomment-4223416161), this "makes searching for errors with specific tasks harder". This issue is to update the task code such that Sentry can associate reports with the specific task that failed.

alxndrsn:
Closes https://github.com/getodk/central/issues/1804

Added a new test which invokes a special error-generating task.

1. It's not tested perfectly, because IRL task are executed from cron jobs and via `docker exec` and `docker compose exec` and directly from docker shells, and possibly in other ways too.  So it's always possible a task doesn't receive any expected node config or env vars.  But the approach in this PR is more suitable than other task tests, which invoke tasks from the same NodeJS VM that tests are running in.
2. new `test/bin` directory to avoid polluting `lib/bin` with test-only scripts

`config.tunnel` presents a small risk - perhaps it will have an effect even when not defined?  This should shake out in testing.

No.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -141,9 +141,7 @@ const _init = (config) => { // eslint-disable-line no-shadow
         process.stderr.write('attempted to log Sentry exception in development:\n');
         process.stderr.write(inspect(err) + '\n');
       },
-      flush() {
-        process.stderr.write('Sentry.flush()\n');
-      },
+      flush: noop,
       setupExpressErrorHandler: noop,
       setTag: noop,
       startSpan: async ({ op, name }, fn) => {
@@ -192,7 +190,6 @@ const _init = (config) => { // eslint-disable-line no-shadow
   console.error('initialising sentry...');
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
-    debug: true,
     tunnel: config.tunnel,
     environment: process.env.SENTRY_ENVIRONMENT || 'production',
     integrations: defaults => [
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -141,7 +141,11 @@ const _init = (config) => { // eslint-disable-line no-shadow
         process.stderr.write('attempted to log Sentry exception in development:\n');
         process.stderr.write(inspect(err) + '\n');
       },
+      flush() {
+        process.stderr.write('Sentry.flush()\n');
+      },
       setupExpressErrorHandler: noop,
+      setTag: noop,
       startSpan: async ({ op, name }, fn) => {
         const start = performance.now();
         try {
@@ -185,8 +189,11 @@ const _init = (config) => { // eslint-disable-line no-shadow
     'VercelAI',
   ];
 
+  console.error('initialising sentry...');
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
+    debug: true,
+    tunnel: config.tunnel,
     environment: process.env.SENTRY_ENVIRONMENT || 'production',
     integrations: defaults => [
       ...defaults.filter(i => !disabledIntegrations.includes(i.name)),
@@ -207,6 +214,9 @@ const _init = (config) => { // eslint-disable-line no-shadow
       return null; // we have a user-space problem.
     }
   });
+  //console.error('sentry initialised.');
+  //Sentry.captureException(new Error('testing in setup'));
+  //console.error('captured test error');
 
   const username = require('os').hostname();
   Sentry.setUser({ username });
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -141,7 +141,6 @@ const _init = (config) => { // eslint-disable-line no-shadow
         process.stderr.write('attempted to log Sentry exception in development:\n');
         process.stderr.write(inspect(err) + '\n');
       },
-      flush: noop,
       setupExpressErrorHandler: noop,
       setTag: noop,
       startSpan: async ({ op, name }, fn) => {
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -187,7 +187,6 @@ const _init = (config) => { // eslint-disable-line no-shadow
     'VercelAI',
   ];
 
-  console.error('initialising sentry...');
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
     tunnel: config.tunnel,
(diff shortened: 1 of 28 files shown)


test/integration/external/sentry.js
```diff
@@ -0,0 +1,94 @@
...
+      mockSentry = await initMockSentry(); // eslint-disable-line no-use-before-define
+    });
+    afterEach(() => {
+      mockSentry?.close();
+    });
+
+    it('should include odk-task tag in error event', async () => {
+      // given
+      const env = {
+        ...process.env, // ensure NodeJS is available in child process
```
alxndrsn:
```suggestion
        ...process.env, // ensure NodeJS is available to child process
```

test/integration/external/sentry.js
```diff
@@ -0,0 +1,94 @@
...
+      const events = [];
+
+      const app = express();
+      app.use(express.text({ type: () => true }));
+      app.get('/event-log', (req, res) => {
+        res.send(events);
+      });
+      app.post('/sentry-tunnel', (req, res) => {
+        const parts = req.body.split('\n').filter(it => it).map(it => JSON.parse(it));
+        if (parts.length !== 3) throw new Error(`unrecognised part count: ${parts.length}`);
```
alxndrsn:
```suggestion
        if (parts.length !== 3) throw new Error(`Unrecognised part count: ${parts.length}`);
```

matthew-white:
> it's a shame a test-specific script is in `‎lib/bin/test-sentry-logging.js‎`.  This seems to be the standard pattern, but this script could equally be stored in e.g. `test/bin`, which might reduce confusion for production users.

`test/bin` sounds sensible to me. :+1: That said, I also don't expect users to `ls lib/bin`. Some of the files in `lib/bin` are for direct use by users, but others I see as being just for our internal use.

test/integration/external/sentry.js
```diff
@@ -0,0 +1,94 @@
...
+    async function getLoggedEvents() { // eslint-disable-line no-use-before-define
+      const res = await fetch(`${mockSentry.baseUrl()}/event-log`);
+      assert.equal(res.status, 200);
+      return res.json();
+    }
+  });
+
+  function sleep(ms) { // eslint-disable-line no-use-before-define
+    return new Promise(resolve => setTimeout(resolve, ms)); // eslint-disable-line no-promise-executor-return
+  }
```
matthew-white: I think the `sleep()` function from test/util/util.js could be reused here.

test/integration/external/sentry.js
```diff
@@ -0,0 +1,94 @@
...
+                project: 123,
+                tunnel: `${mockSentry.baseUrl()}/sentry-tunnel`,
+              },
+            },
+          },
+        }),
+      };
+
+      // when
+      const child = spawn('node', ['lib/bin/test-sentry-logging', 'test error'], { env });
```
matthew-white: Why `spawn()` rather than `exec()` out of curiosity?

lib/external/sentry.js
```diff
@@ -187,6 +188,7 @@ const _init = (config) => { // eslint-disable-line no-shadow
 
   Sentry.init({
     dsn: `https://${config.key}@${config.orgSubdomain}.ingest.sentry.io/${config.project}`,
+    tunnel: config.tunnel,
```
matthew-white: How about adding a comment here, something like "For use in testing". Just to make it clear to our future selves that we added support for `tunnel` in order to facilitate testing, not for end users. In other words, we don't actually expect `config` to have a `tunnel` property in production.

test/integration/external/sentry.js
```diff
@@ -0,0 +1,94 @@
...
+      mockSentry = await initMockSentry(); // eslint-disable-line no-use-before-define
+    });
+    afterEach(() => {
+      mockSentry?.close();
+    });
+
+    it('should include odk-task tag in error event', async () => {
+      // given
+      const env = {
+        ...process.env, // ensure NodeJS is available to child process
```
matthew-white: I don't totally understand this comment. Which variables in `process.env` make NodeJS available to the child process?

matthew-white (approved):
Sorry, meant to approve!

test/integration/external/sentry.js
matthew-white, line 1: From the PR description:

> It's not tested perfectly, because IRL task are executed from cron jobs and via `docker exec` and `docker compose exec` and directly from docker shells, and possibly in other ways too.

For the record, the approach here makes sense to me. :+1: I don't mind that the test is tied to the Backend repo and doesn't involve ops from the `central` repo. It'd be nice to get some end-of-end testing of our cron jobs, but I feel like that can be separate. I've filed an issue about that at getodk/central#1794.

test/integration/external/sentry.js
alxndrsn: Good question.  `execSync()` can't be used because it blocks the express process from responding to the background HTTP response.  `exec()` is fiddly in a different way to `spawn()`, but switched to that as it's more familiar.

lib/external/sentry.js
alxndrsn:
```suggestion
    tunnel: config.tunnel, // currently only used for testing
```

?

alxndrsn: I think I'd rather mark the config itself, e.g. `__test_tunnel`.

test/integration/external/sentry.js
alxndrsn: Normally `PATH`:

> The command lookup is performed using the `options.env.PATH` environment variable if `env` is in the `options` object. Otherwise, `process.env.PATH` is used. If `options.env` is set without `PATH`, lookup on Unix is performed on a default search path search of `/usr/bin:/bin` (see your operating system's manual for execvpe/execvp), on Windows the current processes environment variable `PATH` is used.
> - _https://nodejs.org/api/child_process.html#apicontent_

So defining `options.env.NODE_OPTIONS` requires explicitly including `options.env.PATH` from `process.env.PATH` to the recreate normal `exec()`/`execSync()`/`spawn()`/... behaviour.

Would you rather see:

```suggestion
        PATH: process.env.PATH, // ensure NodeJS is available to child process
```

?

alxndrsn: Done.

lib/external/sentry.js
alxndrsn: Renamed.

PR #1802, opened 2026-04-14
oidc/callback: tighten Content-Security-Policy

alxndrsn:
Adds the following directives:

* `form-action`
* `frame-ancestors`

Ref:

* https://github.com/getodk/central/issues/1533
* https://github.com/getodk/central/pull/1742/

Tested on `dev` to ensure:

- [x] can still log in with OIDC
- [x] expected headers are set on `/callback`

There's no form on the `callback` page, and it's not expected to be embedded in a frame.

No effect, if it's correct.

No.


matthew-white:
> If there's a simple way to deploy there and switch auth to OIDC, that would be ideal.

Possibly relevant: getodk/central#1753

alxndrsn:
> Possibly relevant: [getodk/central#1753](https://github.com/getodk/central/issues/1753)

More than relevant :bow: now tested :+1:

lognaturel (approved):

PR #1803, opened 2026-04-14
util/problem: reject problems with duplicate codes

alxndrsn:
eabd14c443ac7818c83546b89075021839339303 introduced new problem codes, but broke conventional ordering.

Due to lack of tests, this allowed later introduction of problems redefining the same error numbers.

This commit:

* changes construction of `problem.internal.invalidHeader` to be consistent with all other problems
* re-orders all problems to be sorted numerically by code
* renumbers s3 problem codes, as
  * the initial duplicate was `problem.internal.s3accessDenied` (formerly `500.4`, clashing with `problem.internal.enketoUnexpectedResponse`)
  * the next duplicate was `problem.internal.analyticsUnexpectedResponse` (formerly `500.5`, clashing with `problem.internal.s3upstreamError`), but it seems pragmatic to keep all the renumbering in the same functional area

* added new tests
* existing test suite

Alternatives:

* maintain current duplicates, while preventing future duplicates
* renumber the most-recently-created duplicates.  Discussed above, but it seems helpful to keep renumberings in the same functional area.
* renumber all the duplicates, and retire the old duplicate codes

Might affect users depending on the s3 problem codes.  This seems fairly unlikely, but worth mentioning in release notes.

- [x] check for mentions of renumbered problem codes in `docs/api.yml`

### DIFF
--- a/lib/util/problem.js
+++ b/lib/util/problem.js
@@ -66,10 +66,9 @@ const problems = {
     // 400.5 (uniquenessViolation), was used in the past, but has been deprecated
 
     // { header: "the problematic header", value: "the supplied (problematic) value" }
-    invalidHeader: problem(400.6, ifElse(
-      is(String),
-      field => `An expected header field (${field}) did not match the expected format.`,
-      ({ field, value }) => `An expected header field (${field}) did not match the expected format (got: ${(value == null) ? '[nothing]' : value}).`,
+    invalidHeader: problem(400.6, ({ field, value }) => {
+      if (typeof value === 'undefined') return `An expected header field (${field}) did not match the expected format.`;
+      else return `An expected header field (${field}) did not match the expected format (got: ${(value == null) ? '[nothing]' : value}).`;
     )),
 
     // { field: "the name of the missing field" }
--- a/lib/util/problem.js
+++ b/lib/util/problem.js
@@ -309,7 +309,9 @@ const problems = {
     // returned if ODK Analytics returns an unexpected response.
     analyticsUnexpectedResponse: problem(500.5, () => 'The Analytics service returned an unexpected response.'),
 
-    fieldTypesNotDefined: problem(500.6, (frame) => `fieldTypes are not defined on the ${frame} Frame, please define them to use insertMany.`)
+    fieldTypesNotDefined: problem(500.6, (frame) => `fieldTypes are not defined on the ${frame} Frame, please define them to use insertMany.`),
+
+    deliberatelyDuplicatedProblemNumber: problem(500.6, () => ''),
   }
 };
 
--- a/lib/util/problem.js
+++ b/lib/util/problem.js
@@ -14,7 +14,6 @@
 // A list of our Problems with their codes and messages can be found below.
 
 const { floor } = Math;
-const { ifElse, is } = require('ramda');
 const { printPairs } = require('./util');
 
 class Problem extends Error {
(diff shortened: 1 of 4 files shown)


test/unit/util/problem.js
```diff
@@ -0,0 +1,50 @@
...
+    user: 400,
+    internal: 500,
+  };
+
+  for (const [ type, problemMap ] of Object.entries(Problem)) {
+    describe(`type: ${type}`, () => {
+      const expectedRange = expectedRanges[type];
+
+      it('should define an expected code range', () => {
+        assert.equal(Number.isFinite(expectedRange), true);
```
matthew-white:
```suggestion
        Number.isFinite(expectedRange).should.be.true();
```

Don't we usually use Should.js in unit tests (with the exception of `assert.throws()`)? Or did we conclude that either Should.js or `assert` works fine?

test/unit/util/problem.js
matthew-white, line 1: From the PR description:

> This commit &hellip; re-orders all problems to be sorted numerically by code

What do you think about adding a test about that? Something like:

```js
const codes = Object.values(problemMap).map(({ code }) => code);
codes.should.eql([...codes].sort());
```

lib/util/problem.js
```diff
@@ -266,9 +264,17 @@ const problems = {
...
-    s3accessDenied: problem(500.4, () => `The S3 account details or permissions are incorret.`),
+    // returned if Enketo returns an unexpected response.
+    enketoUnexpectedResponse: problem(500.4, () => 'The Enketo service returned an unexpected response.'),
+
+    // returned if ODK Analytics returns an unexpected response.
+    analyticsUnexpectedResponse: problem(500.5, () => 'The Analytics service returned an unexpected response.'),
+
+    fieldTypesNotDefined: problem(500.6, (frame) => `fieldTypes are not defined on the ${frame} Frame, please define them to use insertMany.`),
+
+    s3accessDenied: problem(500.7, () => `The S3 account details or permissions are incorret.`),
```
matthew-white: I know it was already like this, but I'm noticing that "incorret" is a typo.

matthew-white (approved):

matthew-white:
> Might affect users depending on the s3 problem codes.  This seems fairly unlikely, but worth mentioning in release notes.

I've added the PR to the project so that I remember to mention it in GitHub release notes. :+1:

What do you think about adding a note about this to api.yaml? I know none of the specific Problem codes are mentioned in api.yaml, but I'm thinking we could add a bullet point about this under the changelog for v2026.2.

matthew-white:
> What do you think about adding a note about this to api.yaml?

Thinking about this a little more, I guess since these are all 500 errors, it really is unlikely that users have been relying on them. Probably we don't need a note in api.yaml after all.

test/unit/util/problem.js
alxndrsn: I like it :+1: that should help prevent future deviation.

I wasn't sure that NodeJS guarantees the order of `values()` or `map()`, but I _think_ these days it's guaranteed: https://tc39.es/ecma262/2024/#sec-ordinaryownpropertykeys

alxndrsn: Done.  This caught one more mis-ordering 👍

lib/util/problem.js
alxndrsn: https://github.com/getodk/central-backend/pull/1809

PR #1804, opened 2026-04-14
Users.updatePassword(): reject weak passwords

alxndrsn:
* measure password strength using `zxcvbn`
* if password is deemed weak, throw a Problem back to the user

See: https://www.npmjs.com/package/zxcvbn

- [x] tested locally
- [x] new tests
- [x] checked soak tests still run OK: https://github.com/getodk/central-backend/actions/runs/24447887648

* could also be done client-side.  this would avoid server-load issues like https://github.com/dropbox/zxcvbn/issues/316
* considered more-recently-updated forks of `zxcvbn`, but
  * none have got traction
  * there are no particularly worrying bug reports against the original `zxcvbn` repo

Existing weak passwords will **not** be rejected: this PR only changes `hashPassword()`, which is used when setting a new password.

Checking existing passwords is performed by `verifyPassword()`, which will continue to allow existing weak passwords.

No.


matthew-white:
> - [ ] check if this will reject _existing_ weak passwords

I wouldn't think so, at least from a first glance. At some point, we added a minimum length requirement of 10 characters. At that time, we decided not to invalidate/reject existing passwords of fewer than 10 characters. (Instead, users are alerted about such passwords when they log in in Frontend.) I'm mentioning this requirement because `hashPassword()` contains `if (plain.length < 10)`. Given how the code is written, if it's not rejecting based on the minimum length requirement, I wouldn't think that it'd reject based on the zxcvbn score.

It'd definitely be good to check this though. :+1:

> - [ ] should this reject _existing_ weak passwords?

My initial instinct is not to reject. We wouldn't want to break any existing workflows.

Imagine if there's only one user (or some other case where every password has a low zxcvbn score). Then a sysadmin would need to do something on the CLI before any user could log in.

Maybe one option is that we do something similar as what we do for the minimum length requirement. If there were a way during session creation for Backend to tell Frontend that the password is weak, Frontend could indicate that to the user. I'm thinking, how about adding a property like that to the session response?

CC @lognaturel for a second opinion

lib/util/problem.js
```diff
@@ -155,6 +155,8 @@ const problems = {
       return `Failed to process ${total} entities in submission due to ${count} errors: ${joined}`;
     }),
 
+    passwordTooWeak: problem(400.44, (warning) => warning),
```
matthew-white: `warning` is a string, right? What does that mean for the `details` property of the Problem? Will it also be string? Usually, I think of `details` as being an object.

As of #1803, I've been thinking that it'd be nice to standardize on all Problem methods taking an object argument.

matthew-white: Relatedly, it could be nice to see an example warning in one of the tests just to see how the Problem `details` are structured.

matthew-white: We won't trying to localize this warning, right? I'm wondering if in Frontend we could prefix it with a localized sentence that says "Your password is too weak." We do translate "Password must be at least 10 characters long." in Frontend. Related: getodk/central#489. If we did anything in Frontend for this, it'd probably be in `requestAlertMessage()`. Maybe not relevant to this Backend PR, but I'm curious whether you think it'd be worth trying to do any localization for this feature area.

test/integration/api/sessions.js
```diff
@@ -46,9 +47,9 @@ describe('api: /sessions', () => {
     // and reject them before passing the values to bcrypt.
     describe('weird bcrypt implementation details', () => {
       [
-        [ 'repeated once',             'password4chelsea\0password4chelsea' ],                   // eslint-disable-line no-multi-spaces
-        [ 'repeated twice',            'password4chelsea\0password4chelsea\0password4chelsea' ], // eslint-disable-line no-multi-spaces
-        [ 'repeated until truncation', 'password4chelsea\0password4chelsea\0password4chelsea\0password4chelsea\0password4' ],
+        [ 'repeated once',             `${password4chelsea}\0${password4chelsea}` ],                      // eslint-disable-line no-multi-spaces
+        [ 'repeated twice',            `${password4chelsea}\0${password4chelsea}\0${password4chelsea}` ], // eslint-disable-line no-multi-spaces
+        [ 'repeated until truncation', `${password4chelsea}\0${password4chelsea}\0${password4chelsea}\0${password4chelsea}\0${password4chelsea}` ],
```
matthew-white: This truncation case seems slightly different from before. Before, the password was truncated partway (no "chelsea" after "password4"). Here though, the substring after the last `\0` is the full `password4chelsea`. Though maybe I'm misunderstanding what "until truncation" means.

test/integration/task/account.js
```diff
@@ -52,6 +55,10 @@ describe('task: accounts', () => {
     it('should complain if the password is too short', testTask(() =>
       createUser('testuser@getodk.org', 'short')
         .catch((problem) => problem.problemCode.should.equal(400.21))));
+
+    it('should complain if the password is too weak', testTask(() =>
+      createUser('testuser@getodk.org', 'hunter2!!!')
+        .catch((problem) => problem.problemCode.should.equal(400.44))));
```
matthew-white: Is it worth verifying this Problem for the API in addition to the CLI? Or since it's all one function (`hashPassword()`), is a test of the CLI all that's needed?

I see that we have a test of the API for "should disallow a password that is too short (<10 chars)".

test/integration/api/sessions.js
alxndrsn: I've updated the test to try to make the intention clearer.  Please review.

test/integration/task/account.js
alxndrsn: Actually just a unit test of `hashPassword()` may be even better.

lib/util/problem.js
```diff
@@ -153,6 +153,8 @@ const problems = {
...
```
alxndrsn: `warning` is not guaranteed to have a value:

```sh
$ node -p "const { score, feedback } = require('zxcvbn')('estomago'); JSON.stringify({ score, feedback }, null, 2);"
{
  "score": 2,
  "feedback": {
    "warning": "",
    "suggestions": [
      "Add another word or two. Uncommon words are better."
    ]
  }
}
```

It may be simplest to return the whole `feedback` object.

lib/util/problem.js
alxndrsn: I don't think backend is ever be trying to do translations, is it?

Either way, this lib isn't very translation ready.  It's tempting to fork it due to this, and a couple of other issues, and it hasn't been updated in 10 years.

In the meantime, the list of strings can be got via:

```js
const raw = child_process.execSync(`npx get-strings --json ./node_modules/zxcvbn/dist/zxcvbn.js`).toString();
const parsed = JSON.parse(raw);
const filtered = parsed.filter(it => it.trim().includes(' ') && !['less than a second', 'Cannot find module \''].includes(it));
console.log(JSON.stringify(filtered, null, 2));
```

This gives:

```json
[
  "Use a few words, avoid common phrases",
  "No need for symbols, digits, or uppercase letters",
  "Add another word or two. Uncommon words are better.",
  "Straight rows of keys are easy to guess",
  "Short keyboard patterns are easy to guess",
  "Use a longer keyboard pattern with more turns",
  "Repeats like \"aaa\" are easy to guess",
  "Repeats like \"abcabcabc\" are only slightly harder to guess than \"abc\"",
  "Avoid repeated words and characters",
  "Sequences like abc or 6543 are easy to guess",
  "Avoid sequences",
  "Recent years are easy to guess",
  "Avoid recent years",
  "Avoid years that are associated with you",
  "Dates are often easy to guess",
  "Avoid dates and years that are associated with you",
  "This is similar to a commonly used password",
  "This is a top-10 common password",
  "This is a top-100 common password",
  "This is a very common password",
  "A word by itself is easy to guess",
  "Names and surnames by themselves are easy to guess",
  "Common names and surnames are easy to guess",
  "Capitalization doesn't help very much",
  "All-uppercase is almost as easy to guess as all-lowercase",
  "Reversed words aren't much harder to guess",
  "Predictable substitutions like '@' instead of 'a' don't help very much"
]
```

alxndrsn:
> Imagine if there's only one user (or some other case where every password has a low zxcvbn score). Then a sysadmin would need to do something on the CLI before any user could log in.

I've confirmed that existing "weak" passwords will continue to work, and updated the PR description to reflect that.

> Maybe one option is that we do something similar as what we do for the minimum length requirement. If there were a way during session creation for Backend to tell Frontend that the password is weak, Frontend could indicate that to the user. I'm thinking, how about adding a property like that to the session response?

That could be good.  Is this done for short passwords already?  If so, can you point to where it happens?  I struggled to find it in the code.  If not, maybe worth a separate Issue?

lib/util/problem.js
alxndrsn: Add a test case to reflect `"feedback.warning": ""`

lib/util/crypto.js
```diff
@@ -37,6 +38,10 @@ const hashPassword = (plain) => {
   if (typeof plain !== 'string') return reject(Problem.user.invalidDataTypeOfParameter({ value: plain, expected: 'string' }));
   if (plain.length < 10) return reject(Problem.user.passwordTooShort());
   if (plain.length > 72) return reject(Problem.user.passwordTooLong());
+
+  const { score, feedback:{ warning, suggestions } } = zxcvbn(plain);
```
alxndrsn:
```suggestion
  const { score, feedback: { warning, suggestions } } = zxcvbn(plain);
```

lib/util/problem.js
alxndrsn: Full structure now tested.

test/integration/task/account.js
alxndrsn: Added additional tests to the API, and unit tests.

matthew-white:
> > Maybe one option is that we do something similar as what we do for the minimum length requirement. If there were a way during session creation for Backend to tell Frontend that the password is weak, Frontend could indicate that to the user.
> 
> That could be good. Is this done for short passwords already? If so, can you point to where it happens? I struggled to find it in the code. If not, maybe worth a separate Issue?

This is what Frontend does for short passwords: https://github.com/getodk/central-frontend/blob/88aa1977722dd9235a01b954b2ab605bf7eeeb57/src/components/account/login.vue#L221. It shows a toast message if the existing password is short. The difference is that Frontend is able to tell on its own whether a password is short. Whereas for weak passwords, Backend would have to tell Frontend that the password is weak.

I also think this could be done as part of a separate Issue and/or PR.

lib/util/problem.js
matthew-white: > I don't think backend is ever be trying to do translations, is it?

Probably not, no. We've talked about how maybe Backend could do translations one day for emails, but I don't think it's a priority. For error messages specifically, the current plan is to translate more of them in Frontend rather than trying to add translations to Backend.

> Either way, this lib isn't very translation ready. It's tempting to fork it due to this

I think I've heard @lognaturel say that it's particularly common for error messages not to be translated. I don't think it's a big deal for these strings to be English-only. That said, it's helpful to see the range of strings and to know that zxcvbn doesn't bundle its own translations.

It's also interesting to see that zxcvbn can return both a `warning` and `suggestions`. I wonder if we should try to surface the `suggestions` in Frontend somehow. I could ask Nicole for her design input on that if that sounds interesting?

test/integration/api/sessions.js
matthew-white: The new version looks great. :+1:

alxndrsn:
Currently failures come through to frontend like this:

<kbd>
Screenshot_2026-05-05_16-35-31
</kbd>

<kbd>
Screenshot_2026-05-05_16-35-10
</kbd>

<kbd>
Screenshot_2026-05-05_16-34-52
</kbd>

lib/util/problem.js
alxndrsn: Discussed in person :+1:

lib/util/problem.js
```diff
@@ -153,6 +153,8 @@ const problems = {
       return `Failed to process ${total} entities in submission due to ${count} errors: ${joined}`;
     }),
 
+    passwordTooWeak: problem(400.44, ({ warning, suggestions }) => warning || suggestions[0] || 'The supplied password is too weak.'),
```
matthew-white: I think it'd be a little clearer to always show "The supplied password is too weak." To me, an error message that just says "Repeats like "aaa" are easy to guess" could be more explicit that the supplied password has been rejected, and the user needs to try something else. The user will probably figure it out, it just seems potentially confusing.

```suggestion
    passwordTooWeak: problem(400.44, ({ warning, suggestions }) => {
      const message = ['The supplied password is too weak.'];
      const detail = warning || suggestions[0];
      if (detail) message.push(detail);
      return message.join(' ');
    }),
```

Not a merge blocker, just my two cents.

test/e2e/s3/run-tests.sh
matthew-white, line 1: Will Frontend e2e tests break as well?

test/integration/api/sessions.js
```diff
@@ -32,53 +33,68 @@ describe('api: /sessions', () => {
...
+
+      const passwords = [
+        repeatN(1),
+        repeatN(2),
+        repeatN(3),
+        repeatN(4),
+      ];
+
+      // ensure that the final test is for a password which exceeds the bcrypt truncation length,
+      // and is chopped part-way through the password itself, e.g. "secret\x00secret\x00...\x00se"
```
matthew-white: > and is chopped part-way through the password itself

I don't see that this part is guaranteed. E.g., if `password4chelsea` were 23 characters long, then `repeatN(4)` would result in a string of length 4 x 23 + 3 x 1 [for the 3 `\x00`] = 95, which is > 72, yet would be truncated after the third `\x00` (since 3 x 23 + 3 x 1 = 72). In other words, the truncation wouldn't be in the middle of the password text itself.

I kind of doubt it matters, but I thought I'd point it out in case it's important for what this test is getting at. You could check that `bcryptLengthCutoff.length % (password4chelsea.length + 1) !== 0` but it might be overkill.

matthew-white (approved):

lib/util/problem.js
alxndrsn: Seems reasonable :+1:

test/e2e/s3/run-tests.sh
alxndrsn: Something to keep an eye on :+1:

test/integration/api/sessions.js
alxndrsn: I've updated to address this issue - I think successfully.

test/integration/api/users.js
```diff
@@ -247,11 +251,12 @@ describe('api: /users', () => {
               ])))));
 
         [
-          [ 'too short', 'short' ],
-          [ 'too long',  'loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong' ], // eslint-disable-line no-multi-spaces
-          [ 'object',    {} ], // eslint-disable-line no-multi-spaces
-          [ 'array',     [] ], // eslint-disable-line no-multi-spaces
-          [ 'number',    123 ], // eslint-disable-line no-multi-spaces
+          [ 'too short',  'short' ], // eslint-disable-line no-multi-spaces
```
alxndrsn:
```suggestion
          [ 'too short', 'short' ],
```

matthew-white:
@getodk/testers, this functionality is ready to be tested. It is available on the staging server.

Expected behavior: if the user chooses a password that is 10+ characters, yet weak in another way, it should be rejected. The user should see an error toast explaining why their password was rejected. You can see examples of that in @alxndrsn's latest screenshots above.

srujner:
**Tested with Success!**

I've checked whole list of 100 most common passwords. I think that's enough of testing

PR #1810, opened 2026-04-21
config/public: cache blob content

### LINKED ISSUE #1822: No browser caching for login logo and hero images on S3
When I refresh, I always see:

Image

And it takes a while for each to load.

I have not tried with non-S3 images yet.

alxndrsn:
Closes getodk/central#1822

- [x] tested locally
- [x] new integration tests with & without S3 blob support enabled
- [x] new e2e test with s3 enable

##### Alternatives

1. Don't upload config blobs to s3 in the first place
  * :+1: sounds like a decent alternative
2. Upload to s3, but use public URLs instead of presigned links
  * :-1: sounds fiddly and (maybe?) reliant on more complicated config of buckets, and validating configuration stored outside odk-central repos
3. Catch requests for these blobs in nginx and transparently proxy them to the presigned URLs
  * :-1: sounds fiddly to implement and test, but on the other hand
  * :+1: might make for a useful example of how this level of caching could be achieved for other blobs
  * :+1: moves work from node upstream to nginx

Should make the login page load faster.

304 responses have been added to the API documentation.


lib/util/blob.js
```diff
@@ -45,6 +45,22 @@ function blobContent(s3, blob) {
   else return Promise.resolve(blob.content);
 }
 
+async function maskedBlobResponse(s3, blob) {
+  return withEtag(
+    blob.md5,
+    () => (_, response) => {
+      response.set('Content-Disposition', 'attachment');
+      response.set('Content-Type', blob.contentType);
```
matthew-white, line 53: Is there a test that verifies that `Content-Type` is set correctly? I wasn't sure that the existing tests in test/integration/api/config.js make assertions about `Content-Type`.

I think `Content-Type` could be verified in a unit/function test or in an integration/API test, it'd just be nice to have some test that verifies it.

test/integration/api/config.js
```diff
@@ -169,17 +169,92 @@ describe('api: /config', () => {
...
+                // then
+                .expect(304)
+                .then(({ body, text, headers }) => {
+                  body.should.deepEqual({});
+                  text.should.eql('');
+                  should(headers['content-length']).be.undefined();
+                });
+            }));
+
+            it('should 304 correct etag after upload to s3', testService(async (service, { Blobs }) => {
```
matthew-white: How about adding a test for the 200 case where the blob has been uploaded to S3? I'm just thinking that it could be useful to have a test for each of the 2x2 options of [200 vs. 304] x [uploaded to S3 vs. not]. Currently, we have a test of the 200 response (including the response body) where the blob is stored in the database, but in the case where the blob has been uploaded to S3, there's only a test of the 304 case.

Maybe there's enough coverage of that in other ways, in which case feel free to ignore this comment. I was just thinking that if there don't need to be unit tests of `maskedBlobResponse()` or `s3.pipeContent()`, maybe an additional integration test would be a good way to verify the new code.

matthew-white (approved):

lib/util/blob.js
alxndrsn: :+1: I'll add tests for this.

alxndrsn: Added.

test/integration/api/config.js
alxndrsn: 100% 👍

alxndrsn: Test matrix now:

```js
          describe(configKey, () => {
            it('should return notfound if the config is not set', testService((service) =>
            it('should return the config with expected headers', testService(async (service) => {
            it('should ignore incorrect etag', testService(async (service) => {
            it('should 304 correct etag', testService(async (service) => {
            describe('with S3 enabled', () => {
              it('should return notfound if the config is not set', testService((service) =>
              it('should transparently serve 200 with expected content & headers', testService(async (service, { Blobs }) => {
              it('should ignore incorrect etag', testService(async (service, { Blobs }) => {
              describe('after upload to S3', () => {
                it('should transparently serve 200 with expected content & headers', testService(async (service, { Blobs }) => {
                it('should ignore incorrect etag', testService(async (service, { Blobs }) => {
                it('should 304 correct etag', testService(async (service, { Blobs }) => {
```

alxndrsn: This will now require mocking of s3.pipeContent(). I think this reveals another gap in testing - there should be an e2e test for this functionality.

matthew-white: Ah interesting. FWIW I don't mind if this PR is merged right away, then additional testing comes in a follow-up PR — whatever works best.

alxndrsn: > How about adding a test for the 200 case where the blob has been uploaded to S3? I'm just thinking that it could be useful to have a test for each of the 2x2 options of [200 vs. 304] x [uploaded to S3 vs. not]. Currently, we have a test of the 200 response (including the response body) where the blob is stored in the database, but in the case where the blob has been uploaded to S3, there's only a test of the 304 case.

I've refactored the integration tests so that each runs for the following scenarios:

* s3 disabled
* s3 enabled, config not uploaded to s3
* s3 enabled, config uploaded to s3

The full tree of integration tests is now:

```
  api: /config
    generic endpoints
      GET - single public config
        GET /v1/config/public/hero-image (blob)
          s3 NOT enabled
            ✔ should return notfound if the config is not set
            blob config exists
              ✔ should return the config with expected headers
              ✔ should set revalidate cache headers for query string: ''
              ✔ should set revalidate cache headers for query string: '?irrelevant=123'
              ✔ should set revalidate cache headers for query string: '?ts='
              ✔ should set immutable cache headers if request includes valid ts query param
              ✔ should ignore incorrect etag
              ✔ should 304 correct etag
          s3 enabled, blob NOT uploaded
            ✔ should return notfound if the config is not set
            blob config exists
              ✔ should return the config with expected headers
              ✔ should set revalidate cache headers for query string: ''
              ✔ should set revalidate cache headers for query string: '?irrelevant=123'
              ✔ should set revalidate cache headers for query string: '?ts='
              ✔ should set immutable cache headers if request includes valid ts query param
              ✔ should ignore incorrect etag
              ✔ should 304 correct etag
          s3 enabled, blob uploaded
            ✔ should return notfound if the config is not set
            blob config exists
              ✔ should return the config with expected headers
              ✔ should set revalidate cache headers for query string: ''
              ✔ should set revalidate cache headers for query string: '?irrelevant=123'
              ✔ should set revalidate cache headers for query string: '?ts='
              ✔ should set immutable cache headers if request includes valid ts query param
              ✔ should ignore incorrect etag
              ✔ should 304 correct etag
        GET /v1/config/public/logo (blob)
          s3 NOT enabled
            ✔ should return notfound if the config is not set
            blob config exists
              ✔ should return the config with expected headers
              ✔ should set revalidate cache headers for query string: ''
              ✔ should set revalidate cache headers for query string: '?irrelevant=123'
              ✔ should set revalidate cache headers for query string: '?ts='
              ✔ should set immutable cache headers if request includes valid ts query param
              ✔ should ignore incorrect etag
              ✔ should 304 correct etag
          s3 enabled, blob NOT uploaded
            ✔ should return notfound if the config is not set
            blob config exists
              ✔ should return the config with expected headers
              ✔ should set revalidate cache headers for query string: ''
              ✔ should set revalidate cache headers for query string: '?irrelevant=123'
              ✔ should set revalidate cache headers for query string: '?ts='
              ✔ should set immutable cache headers if request includes valid ts query param
              ✔ should ignore incorrect etag
              ✔ should 304 correct etag
          s3 enabled, blob uploaded
            ✔ should return notfound if the config is not set
            blob config exists
              ✔ should return the config with expected headers
              ✔ should set revalidate cache headers for query string: ''
              ✔ should set revalidate cache headers for query string: '?irrelevant=123'
              ✔ should set revalidate cache headers for query string: '?ts='
              ✔ should set immutable cache headers if request includes valid ts query param
              ✔ should ignore incorrect etag
              ✔ should 304 correct etag
```

lib/resources/config.js
```diff
@@ -63,15 +63,17 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
+  service.get('/config/public/:key', anonymousEndpoint(async (container, { params }, request, response) => {
     const { key } = params;
     const frame = Config.forKey(key);
-    return frame.isPublic
-      ? getConfig(container, key)
-      : Problem.user.insufficientRights();
+
+    if (!frame.isPublic) return Problem.user.insufficientRights();
+
+    response.set('Cache-Control', request.query.ts ? 'max-age=31536000' : 'no-cache');
```
matthew-white: Does the `Vary` header need to be set here? I think not, since it's set in lib/http/service.js, but I thought I'd check.

lib/resources/config.js
```diff
@@ -66,12 +66,14 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
```
matthew-white: What do you think about moving this logic to `getConfig()`? To me, the only difference between /v1/config/:key and /v1/config/public/:key is how auth works. I'd expect caching to work the same between the two.

Practically, moving the logic wouldn't have any user-facing effect, since Frontend is only using /v1/config/public/:key for these images; Frontend doesn't need /v1/config/:key to do caching. But to me, it'd be a little clearer to handle the logic in `getConfig()`. Doing so would require passing `request` and `response` to `getConfig()`.

I don't have strong feelings about this one. Mostly I just wanted to throw it out there to see what you thought.

test/integration/api/config.js
```diff
@@ -169,17 +169,106 @@ describe('api: /config', () => {
...
+          body.toString('utf8').should.equal('testimage');
+
+          return res;
+        };
+
+        describe(`GET ${blobConfigPath} (blob)`, () => {
+          [
+            { scenario:'s3 NOT enabled',                enableS3:false, transferToS3:false }, // eslint-disable-line no-multi-spaces,key-spacing
+            { scenario:'s3 enabled, blob NOT uploaded', enableS3:true,  transferToS3:true  }, // eslint-disable-line no-multi-spaces,key-spacing
+            { scenario:'s3 enabled, blob uploaded',     enableS3:true,  transferToS3:true  }, // eslint-disable-line no-multi-spaces,key-spacing
```
matthew-white:
```suggestion
            { scenario:'s3 enabled, blob NOT uploaded', enableS3:true,  transferToS3:false  }, // eslint-disable-line no-multi-spaces,key-spacing
            { scenario:'s3 enabled, blob uploaded',     enableS3:true,  transferToS3:true  }, // eslint-disable-line no-multi-spaces,key-spacing
```

These two scenarios currently look the same to me. I'm thinking that the top one should have `transferToS3:false`.

matthew-white (approved):

lib/resources/config.js
matthew-white: Or is the idea that /v1/config/public/:key is already public, so there's no concern about sensitive data ending up in a cache somewhere?

matthew-white: On that point, I'm now realizing that even when `request.query.ts` is falsey, the endpoint sets the header to `no-cache`, which is different from the default of `private, no-cache` that's set by lib/http/service.js. To me, the absence of `private` is confirmation of the intention that only the public endpoint should have these differences in caching. :+1:

matthew-white: Now I'm wondering whether a code comment explaining some of this reasoning would be helpful. 🤔

(Sorry for all the comments here!)

test/integration/api/config.js
alxndrsn: Good catch, thanks!

lib/resources/config.js
alxndrsn: I'll add tests for expected `Vary` values.

lib/resources/config.js
alxndrsn: `no-cache` is to allow for public caching, and revalidation via ETag.  I've refactored cache header handling to be more descriptive - please take a look.

alxndrsn: I think this change also makes it clearer why cache headers would be handled differently for `/config/public/:key` to `/config/:key`.

lib/util/http.js
```diff
@@ -151,11 +151,45 @@ const convertToSubmissionVersion = (urlString, instanceId, rootId) => {
...
+      vary: 'Accept-Encoding, Origin',
+    },
+  },
+  public: {
+    cacheButRevalidate: {
+      cacheControl: 'no-cache',
+    },
+    immutable: {
+      cacheControl: 'max-age=31536000',
+    },
```
matthew-white, line 185: Why no `vary` out of curiosity? Because we don't expect request headers to matter?

Is the idea that we can add `vary` later if we need to?

lib/resources/config.js
matthew-white: The new version seems very clear!

matthew-white (approved):

lib/util/http.js
alxndrsn: Correct, but this should be commented to preempt this question in future.

alxndrsn: Double-checked, and I've added `Vary: Accept-Encoding`.

PR #1811, opened 2026-04-21
config: don't serve image blobs with Content-Disposition: inline

alxndrsn:
Introduced in #1769, logo and hero-image on the login page are displayed in `` tags.

`Content-Disposition: inline` is a convenience to allow directly-linked images to display in the browser instead of downloading.

It's unlikely that login-page branding images would be directly linked.

- updated existing tests
- tested with frontend that images are still displayed correctly on the login page:

Screenshot_2026-04-21_16-10-04

1. `Content-Disposition: inline` is risky with user-submitted content, and should only be used when necessary.
2. simplifies #1810

Unlikely to have any effect as it's a new feature, and linking directly to config-blob-images is a surprising use case.

No.


lib/resources/config.js
```diff
@@ -63,7 +63,7 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
 
     // Configs that store blobs
     const blob = await Blobs.getById(config.blobId).then(getOrNotFound);
-    return blobResponse(s3, key, blob, true);
+    return blobResponse(s3, key, blob);
```
matthew-white, line 66: From the PR description:

> `Content-Disposition: inline` is a convenience to allow directly-linked images to display in the browser instead of downloading.

I thought that `inline` was preferred when the image was used in an `` tag. If not, I agree that there's no reason to inline the image. Directly linking certainly isn't the intended use case.

matthew-white (approved):

lib/resources/config.js
matthew-white, line 66: From the PR description:

> `Content-Disposition: inline` is risky with user-submitted content, and should only be used when necessary.

`blobResponse()` will only inline certain safe MIME types, so I wouldn't think that there'd be risk here. But it also doesn't sound like it's useful to inline, and I agree that `attachment` should be the default.

alxndrsn: The spec is not explicit relating to `` tags because `Content-Disposition` is irrelevant to them, but here's how I think you can determine that via the WHATWG specs:

https://html.spec.whatwg.org/multipage/images.html defines behaviour of Images and `` elements.

[_4.8.4.3.1 When to obtain images_](https://html.spec.whatwg.org/multipage/images.html#when-to-obtain-images) points to [_4.8.4.3.5 Updating the image data_](https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data) which in step 26 states:

> Fetch the image: [Fetch](https://fetch.spec.whatwg.org/#concept-fetch) request.

The Fetch spec only mentions `Content-Disposition` in reference to `multipart/form-data`.

Fetch requests contrast with the HTML spec, which mentions `Content-Disposition` in:

* [**4.6.6** _Downloading resources_](https://html.spec.whatwg.org/#downloading-resources)
* [**4.10.22** _Form submission_](https://html.spec.whatwg.org/#form-submission-2)
* [**7.2.6.7** _Initiating navigations_](https://html.spec.whatwg.org/#navigation-api-initiating-navigations)
* [**7.2.6.10.1** _The `NavigateEvent` interface_](https://html.spec.whatwg.org/#the-navigateevent-interface)
* [**7.4** _Navigation and session history_](https://html.spec.whatwg.org/#navigation-and-session-history)

test/integration/api/config.js
```diff
@@ -62,29 +62,24 @@ describe('api: /config', () => {
...
-          await asAlice.get('/v1/config/logo')
-            .expect(200)
-            .then(({ headers }) => {
-              headers['content-disposition'].should.startWith('inline');
-            });
+        [
+          'image/jpeg',
+          'image/svg+xml',
+        ].forEach(mimeType => {
+          it(`should not inline image type: ${mimeType}`, testService(async (service) => {
```
alxndrsn:
```suggestion
          it(`should not set Content-Disposition: inline for mime type: ${mimeType}`, testService(async (service) => {
```

test/integration/api/config.js
```diff
@@ -62,29 +62,24 @@ describe('api: /config', () => {
...
-            .send('testimage')
-            .expect(200);
-          await asAlice.get('/v1/config/logo')
-            .expect(200)
-            .then(({ headers }) => {
-              headers['content-disposition'].should.startWith('inline');
-            });
+        [
+          'image/jpeg',
+          'image/svg+xml',
```
alxndrsn, line 67:
```suggestion
          'image/svg+xml',
          'text/html',
          'text/javascript',
```

PR #1813, opened 2026-04-21
Update gitignore file to exclude context files for AI agents

sadiqkhoja:
This change allows individual developer to create their own context files for their preferred AI agent and do their own experimentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,9 @@ node_modules/
 .vscode
 /junit-reports/
 /config/development.json
+
+# context files for AI agents
+AGENTS.md
+CLAUDE.md
+.github/copilot-instructions.md
+.cursor


.gitignore
```diff
@@ -13,3 +13,9 @@ node_modules/
 .vscode
 /junit-reports/
 /config/development.json
+
+# context files for AI agents
+AGENTS.md
+CLAUDE.md
+.github/copilot-instructions.md
```
ktuite, line 20: This inspired me to set up some copilot instructions. For me they live in `.copilot-instructions.md` at the root.

ktuite (approved):

.gitignore
ktuite: just kidding, copilot was wrong about itself and made the instructions in the wrong place.

@sadiqkhoja found the actual [docs](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-custom-instructions) and we convinced copilot that the `.github/` location is the right place for the instructions.

PR #1814, opened 2026-04-23
Update OpenRosa hash after deletion of property in ownerOnly entity list

### LINKED ISSUE #1849: Update OpenRosa hash of entity list after property is deleted
Filing an issue based on the discussion at getodk/collect#7155.

Right now, when a property is added to an entity list, the OpenRosa hash of the entity list changes. In other words, the OpenRosa hash changes not just when an individual entity changes, but also when the structure of the entity list changes.

However, when a property is deleted, the OpenRosa hash does not change when the entity list is set for "only access own Entities" (`ownerOnly: true`). This issue is to make it so that the OpenRosa hash of an entity list always changes after an entity property is deleted. As @seadowg [writes](https://github.com/getodk/collect/pull/7155#issuecomment-4305633421), "I think the `hash` should really change when a property is deleted. Otherwise, clients won't handle the property deletion until an entity changes if they respect it."

matthew-white:
Closes getodk/central#1849.

I wrote a test that failed before the fix was made.

I tried to follow the current pattern. I just added the new audit log action to the list of actions in the query.

I'm pretty sure that this change only affects the OpenRosa hash. I'm adding a single, specific new audit log action, so I think the risk of regression is low.

We don't seem to document that changes to the structure of the entity list (i.e., adding or deleting entity properties) changes the OpenRosa hash of the entity list. Given that, I don't think there are any API docs that this PR needs to patch, and we don't need to mention this change in the API docs changelog. We could document this effect on the OpenRosa hash, but it feels to me like too much of a detail and doesn't need to be documented.

### DIFF
--- a/lib/model/query/form-attachments.js
+++ b/lib/model/query/form-attachments.js
@@ -165,11 +165,11 @@ const _get = (exec, options) => {
         SELECT datasets.id, MAX("loggedAt") AS "lastAudit" FROM audits
         JOIN datasets ON audits."acteeId" = datasets."acteeId"
         ${actorId == null ? sql`` : sql`
-          -- We're always interested in dataset.create and dataset.update
-          -- actions, even for ownerOnly datasets.
+          -- We're always interested in dataset.create and in actions that add
+          -- or delete entity properties -- even for ownerOnly datasets.
           WHERE
             (NOT datasets."ownerOnly") OR
-            audits.action IN ('dataset.create', 'dataset.update')
+            audits.action IN ('dataset.create', 'dataset.update', 'dataset.update.property.delete')
         `}
         GROUP BY datasets.id
       ) AS audit_metadata ON audit_metadata.id = datasets.id
--- a/test/integration/api/datasets.js
+++ b/test/integration/api/datasets.js
@@ -7064,6 +7064,23 @@ describe('datasets and entities', () => {
         (await getHash(asChelsea)).should.not.equal(originalHash);
       }));
 
+      it('changes hash after a dataset property is deleted', testServiceFullTrx(async (service) => {
+        const [asAlice, asChelsea] = await service.login(['alice', 'chelsea']);
+        await createData(asAlice);
+        await assignToProject(asAlice, asChelsea, 'formfill');
+
+        // Add an entity property.
+        await asAlice.post('/v1/projects/1/datasets/people/properties')
+          .send({ name: 'foo' })
+          .expect(200);
+        const originalHash = await getHash(asChelsea);
+
+        // Delete the property.
+        await asAlice.delete('/v1/projects/1/datasets/people/properties/foo')
+          .expect(200);
+        (await getHash(asChelsea)).should.not.equal(originalHash);
+      }));
+
       it('computes hash based on timestamps and the entity count', testServiceFullTrx(async (service, container) => {
         const [asAlice, asChelsea] = await service.login(['alice', 'chelsea']);
         await createData(asAlice);


matthew-white:
@sadiqkhoja, I've tagged you for review, since you reviewed the related PR #1522.

matthew-white:
@sadiqkhoja if you end up having time sometime next week to review this PR, that'd be awesome. I'd love to get this PR merged at some point soon. I have a follow-up PR planned to add a few more tests in this area.

sadiqkhoja (approved):
LGTM

lognaturel:
Since we're contemplating updates to this hash calculation, it would be helpful to have this merged!

When we filter the audit log by event type, could/should we consider using the filter value as a prefix rather than an exact match? I found it very surprising that filtering by `dataset.update` includes additions but not deletes. Maybe another question is whether the extra `.properties.delete` granularity is helpful. Additions are logged as `dataset.update` events, right?

I wonder whether it would be more helpful to separate out schema (property add/delete, eventually type changes) and metadata (filter, eventually Entity List label) updates.

matthew-white:
I'll get this merged ASAP!

> When we filter the audit log by event type, could/should we consider using the filter value as a prefix rather than an exact match?

Do you mean via the API or in the SQL query?

API: You can specify a query parameter like `action=dataset`, and that should include all dataset actions, but you can't specify `action=dataset.update` and have it include `dataset.update.properties.delete`.

We have a number of actions like this with more than two levels, often having to do with nested/related resources. Examples: `form.update.draft.set`, `form.update.draft.delete`, `user.assignment.create`, `user.assignment.delete`, `entity.bulk.delete`. It strikes me that `dataset.update.properties.delete` follows a similar form as `form.update.draft.delete`.

In terms of the SQL, the filtering for `action=dataset` is done by explicitly listing the actions:

https://github.com/getodk/central-backend/blob/ff9e749f1f2d4aecc4036c426f1da161435920cd/lib/model/query/audits.js#L34-L55

I think that's for performance reasons. There's an index involving `action` that should make that operation faster than prefix-matching. The query in this PR follows a similar pattern.

> I found it very surprising that filtering by `dataset.update` includes additions but not deletes. Maybe another question is whether the extra `.properties.delete` granularity is helpful.

I was also surprised by that. Not sure whether the granularity is helpful, but if it is, maybe we should strive for consistent granularity with `.properties.create`. But then there's a question of how to handle existing audit log events and whether to backfill.

> I wonder whether it would be more helpful to separate out schema (property add/delete, eventually type changes) and metadata (filter, eventually Entity List label) updates.

That makes a lot of sense to me. Right now, toggling `approvalRequired` would affect the OpenRosa hash I believe, whereas conceptually I don't think it should. Not that it should be toggled that often in real usage.

This brings to mind getodk/central#544, which features a different split, between the actual entity data (which should definitely affect the hash) and the entity metadata/status (whether its conflict has been resolved, which arguably shouldn't affect the hash).

lognaturel:
I meant in the SQL query but indeed, that would be problematic for performance.

I'm eager to get off calculating these etags from disparate information sources and have an event counter. In the mean time, thanks for adding tests around this hash, they're really important.

matthew-white:
I've been planning a follow-up PR to add more tests verifying that the OpenRosa hash changes in all the relevant cases when `ownerOnly` is `false` (this PR covers the `true` case). Just to be really certain that the hash is changing when we think it is. But I'll wait to do that until Kathleen's work is merged, or maybe her work will already do something similar. For now, maybe I'll just file an issue about it.

lognaturel:
One case I'm uncertain about is:
* `ownerOnly` is false
* one Entity changes owner to exit the filtered set
* another Entity is restored to the filtered set

We should have an explicit test for that if we don't already. I guess the underlying question there is whether restore changes updatedAt.

To give a preview -- as we refine the hash calculation to incorporate property-based filtering, I'm going to propose we drop using the audit log events and instead represent the schema by concatenating all properties. I think that will both be faster and easier to understand. So this test will definitely be valuable but I think we should consider changing the implementation. I can't currently see any reason to consider `dataset.create` events and the property creation and deletion is meant to get at schema which concatenating the properties would as well.

matthew-white:
<blockquote>

One case I'm uncertain about is:
* `ownerOnly` is false
* one Entity changes owner to exit the filtered set
* another Entity is restored to the filtered set
</blockquote>

Is this a future case? You can't currently change the owner, I believe. Right now, the owner is the creator.

One restore case that I think we do have a test for is an entity being deleted, then restored.

> I can't currently see any reason to consider `dataset.create` events

I don't see any reason either. It could be that I just wanted there to always be some audit event to query, and including `dataset.create` ensured that there would always be at least one event and therefore always some timestamp.

> So this test will definitely be valuable but I think we should consider changing the implementation.

That makes sense. :+1: I'll get this PR merged ASAP so that we have this test even as the implementation is changed.

lognaturel:
> Is this a future case? You can't currently change the owner, I believe. Right now, the owner is the creator.

Right you are! 🎉

matthew-white:
This PR changed a few lines in lib/, but that part of the code is in active development. #1846 touched the same part of the code, and @ktuite will be revisiting that code as part of getodk/central#1866. The current plan is to make significant changes to our approach to the OpenRosa `hash`/ETags. With that in mind, I've closed this PR. It's likely that @ktuite's work will have the happy side effect of closing out getodk/central#1849.

> I've been planning a follow-up PR to add more tests verifying that the OpenRosa hash changes in all the relevant cases when `ownerOnly` is `false` (this PR covers the `true` case). Just to be really certain that the hash is changing when we think it is.

The one thing from this PR that I think could still be helpful is the test. I think we need more testing in general of the OpenRosa `hash`. I've created getodk/central#1969 to track that.

PR #1815, opened 2026-04-24
POC (not for merge) - Demo: Filter entities by userprops

brontolosone:
For @ktuite, demo of a filter-entityprops-by-userprops implementation. Only active for `/v1/projects/XX/datasets/XXXX/entities.csv`.

Just the DB side of things.

One thing I didn't implement is full referential anomaly prevention. Here's the thing: we have a table of properties (not values, just definitions of the properties) *for a particular project's users*: `project_actor_property_names`. So, scoped to a project.
But then, the filter definition table (aka "permission configuration") links one of those actor properties to some dataset property _but the database allows that dataset property to be from a different project_ :zap: 
(A solution to this can be to apply the compound foreign key pattern of which the foreign key constraint `"dataset_property_filter__composite_fk"` on table `dataset_property_filter` is an example.)

The filters constrain. So if a user doesn't have any properties being filtered by, they see all. But we could do the converse which means that if you give 1 (just 1) user "haircolor=grey" and define a rule "entity.barkcolor == user.haircolor" then this user will see entities with grey bark color, and no one else will see anything else...

Related: getodk/central#1339
Related: getodk/central#1790
Related: getodk/central#1791
Related: getodk/central#1792

### DIFF
--- a/lib/model/migrations/20260424-01-user-properties-and-dataset-filtering-by-userproperties-01.down.sql
+++ b/lib/model/migrations/20260424-01-user-properties-and-dataset-filtering-by-userproperties-01.down.sql
@@ -1,4 +1,5 @@
-DROP TABLE propdemo__property_filter CASCADE;
-DROP INDEX "propdemo__property_filter__unique_for_composite_fk_referent";
-DROP TABLE propdemo___actorpropname CASCADE;
-DROP TABLE propdemo___actorpropval;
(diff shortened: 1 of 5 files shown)


matthew-white:
The current plan isn't to merge this PR as-is, so I'll go ahead and close it. Related GitHub issues and follow-up PRs may reference this PR.

PR #1816, opened 2026-04-24
WIP chores: archive < 2026 migrations

sadiqkhoja:
Move old migrations to an archive directory, but still able to run them.


lib/model/migrations/archive/.eslintrc.json
sadiqkhoja, line 1: not sure if we really need this.

matthew-white:
Just to leave a comment about the status of this PR, it sounds like there's consensus on the team to group migrations by year (one directory per year) rather than creating a single archive/ directory.

PR #1791, opened 2026-04-01
[wip] investigate soak test failures

alxndrsn:
Ref https://github.com/getodk/central/issues/1734

# areas to investigate

- [x] idle queries?
- [x] dangling queries
  - [x] which?
  - [ ] how long do they dangle?
    - [ ] run with massive timeout, and confirm if/no failure
    - [ ] suite: 9 minutes before cleanup
    - [ ] individual query:
  - [ ] what is the cause?
- [x] zip 500 errors
  - [x] recreate?
  - [x] cause: connection pool exhaustion
  - [ ] should these cause suite-failure?  currently acceptance threshold for the zip test is 0%!
- [ ] if all connections have drained, why are there still open queries?
  - [ ] e.g. https://github.com/getodk/central-backend/actions/runs/24929884345/job/73005829220?pr=1791

# spin-off PRs

- [x] https://github.com/getodk/central-backend/pull/1789
- [x] https://github.com/getodk/central-backend/pull/1790
- [x] https://github.com/getodk/central-backend/pull/1792
- [ ] improve logging of non-OK errors in e2e API code
- [x] https://github.com/getodk/central-backend/pull/1799
- [x] https://github.com/getodk/central-backend/pull/1817

PR #1817, opened 2026-04-25
ci/soak: doSoakTest(): name parameters

alxndrsn:
Convert `doSoakTest()` function arguments to named object properties.  This makes it much easier to see what each number is, which is very helpful when understanding, modifying and investigating soak test failures, changes etc.

Ref #1791

These changes have been in use in multiple CI runs at https://github.com/getodk/central-backend/pull/1791.

Just a helpful little change.  https://github.com/getodk/central-backend/pull/1791 adds a new param to the function, which change becomes very hard to read with non-named function params.

No change - just in tests.

No.


sadiqkhoja (approved):
I ❤️ this change. I read somewhere that anytime you've more than few arguments in function than it's time to introduce an object.

PR #1820, opened 2026-04-27
inline content: lock down CSP

### LINKED ISSUE #1779: csp: allow favicon (`img-src: 'self'`) for file served with content-disposition: inline
https://getodk.sentry.io/issues/7254280940/?project=5724763

Hard to demonstrate this on a server with s3-backed blobs, but a window exists where a file can be displayed inline, and the browser may request `/favicon.ico`.

alxndrsn:
# TODO

- [x] ~extend test cases to `text/xml` files (e.g. .svc)?  or~ separate to new issue -- filed at https://github.com/getodk/central/issues/1851
- [ ] fix implementations

---

* allow favicons to load
* don't allow much of anything else

Closes https://github.com/getodk/central/issues/1779

- [ ] updated integration tests
- [ ] related CSP (report-only) reports in Sentry

Alternatives

* could do from nginx, but much harder to identify what is being served with `Content-Disposition: inline`
* could filter favicon-related CSP reports at ingestion (Sentry).  This would be a simple approach, and safer than allowing all images, but would create more traffic for users
* could explicitly allow the URL of the favicon, if central-backend knows its own domain name

Risk could be reduced by initially service CSPs as report-only.  Likely won't be a problem though as:

1. no existing worrying reports
2. the new CSP included here reflects the CSP served from nginx, but with an `img-src` change
3. surprises will be reports to /csp-report

### DIFF
--- a/test/integration/api/app-users.js
+++ b/test/integration/api/app-users.js
@@ -355,6 +355,7 @@ describe('api: /key/:key', () => {
       .then(({ headers, body }) => {
         headers['content-type'].should.equal('image/jpeg');
         headers['content-disposition'].should.equal('inline; filename="here_is_file2.jpg"; filename*=UTF-8\'\'here_is_file2.jpg');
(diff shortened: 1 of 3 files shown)


alxndrsn:
Fix in https://github.com/getodk/central

PR #1821, opened 2026-04-27
csp: serve text/xml with tight CSP

### LINKED ISSUE #1851: csp: block all except favicon for XML endpoints
E.g. https://dev.getodk.cloud/v1/projects/1/forms/some-form.svc/$metadata

Currently report-only, but will violate `img-src` directive when CSP is loaded.

Related:

* https://github.com/getodk/central/issues/1779
* https://github.com/getodk/central-backend/pull/1820

alxndrsn:
# TODO

- [ ] unit tests for
  - [ ] `openRosaResultWriter`
  - [ ] `openRosaErrorWriter`
  - [ ] `odataXmlErrorWriter`
  - [ ] odata result writer (may need to create this)
- [ ] implementation

---

* disallow by default
* allow `img-src: 'self'` for favicon

Closes getodk/central#1851

Updated tests.

Alternatives

* could do from nginx, but much harder to identify what is being served with `Content-Disposition: inline`
* could filter favicon-related CSP reports at ingestion (Sentry).  This would be a simple approach, and safer than allowing all images, but would create more traffic for users
* could explicitly allow the URL of the favicon, if central-backend knows its own domain name

### DIFF
--- a/lib/http/endpoint.js
+++ b/lib/http/endpoint.js
@@ -46,6 +46,8 @@ const writeProblemJson = (response, error) => {
 const isJsonType = (x) => /(^|,)(application\/json|json)($|;|,)/i.test(x);
 const isXmlType = (x) => /(^|,)(application\/(atom(svc)?\+)?xml|atom|xml)($|;|,)/i.test(x);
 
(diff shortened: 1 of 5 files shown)


alxndrsn:
Fix in https://github.com/getodk/central

PR #1822, opened 2026-04-27
Support app-user properties for filtering dataset CSV attachments on forms

### LINKED ISSUE #1852: Support app user properties being used to filter dataset attachments
## Background

When app users download dataset CSV attachments as part of a form (e.g. `trees.csv` linked to a survey form), they currently receive all entities in the dataset. For large deployments this is a privacy and performance problem: a field worker assigned to one plot shouldn't receive data for every other plot.

We want to support a way for project managers to configure which entities each app user can see when downloading those form attachments, based on properties of that user. The filtering should be automatic — the app user's client doesn't need to do anything differently, it just receives a smaller CSV.

### User properties

Project managers can define named property keys for a project (e.g. `region`, `plot_id`). Individual app users can then be assigned values for those keys (e.g. `region = "north"`).

### Dataset access filters

For a given dataset, a project manager can configure rules that link a dataset property to a user property. When an app user downloads the dataset as a form CSV attachment, only entities whose value for the dataset property matches the app user's value for the linked user property are returned.

Multiple rules on the same dataset are conjunctive — the app user must match **all** rules to see an entity. An app user with no value set for a referenced property sees all entities (no partial filter).

**Example:** The `trees` dataset has a `region` property and is linked as `trees.csv` on a survey form. A filter rule links `trees.region → user property "region"`. A field worker with `region = "north"` downloads `trees.csv` and only receives north trees. A field worker with no `region` value receives all trees.

## API

| Method | Path | Description |
|---|---|---|
| `POST` | `/projects/:id/user-properties` | Register a new property name for a project |
| `GET` | `/projects/:id/user-properties` | List all property names for a project |
| `PATCH` | `/projects/:id/app-users/:id` | Set or unset property values on an app user |
| `PUT` | `/projects/:id/datasets/:name/access-filter` | Add or update one filter rule for a dataset |
| `GET` | `/projects/:id/datasets/:name/access-filter` | List all filter rules for a dataset |

The filtering applies transparently to the existing form attachment CSV download endpoint — no client changes required.

## Out of scope / follow-up

- Audit logging for property and filter mutations (will make a followup issue)
- Support for actor types other than app users
- UI for managing user properties and filter rules (see frontend issues)

ktuite:
## Overview

Adds backend support for restricting which entities an app user receives when downloading a dataset CSV attachment as part of a form, based on properties assigned to that user.

Closes https://github.com/getodk/central/issues/1852

### Database (migration `20260424-01-user-properties-and-dataset-filtering-by-userproperties`)

Three new tables and one view:

- **`actor_properties`** — registry of named property keys scoped per project
- **`actor_property_values`** — one value per app user per property
- **`dataset_access_filters`** — maps a dataset property to a user property as a filter rule; one row per dataset property, upsert semantics
- **`actor_dataset_filter` (view)** — aggregates active filter rules for a `(datasetId, actorId)` pair into a JSONB object used in the filtering query

### New API endpoints

| Method | Path | Description |
|---|---|---|
| `POST` | `/projects/:id/user-properties` | Register a new property name for a project |
| `GET` | `/projects/:id/user-properties` | List all property names for a project |
| `PATCH` | `/projects/:id/app-users/:id` | Set or unset property values on an app user |
| `PUT` | `/projects/:id/datasets/:name/access-filter` | Add or update one filter rule for a dataset |
| `GET` | `/projects/:id/datasets/:name/access-filter` | List all filter rules for a dataset |

### Filtering

The filter is applied in `streamForExport` via a `LEFT OUTER JOIN` on the `actor_dataset_filter` view, using JSONB containment (`<@`). Multiple rules on a dataset are conjunctive. An app user with no matching property values sees all entities.

### Code

- New `ActorProperty` and `DatasetAccessFilter` frames in `frames.js`
- New query module `lib/model/query/user-properties.js`
- New resource file `lib/resources/user-properties.js`; access-filter endpoints added to `lib/resources/datasets.js`; app-user property-value endpoint added to `lib/resources/app-users.js`
- Property name validation reuses `validatePropertyName` from the dataset layer

## Not included (follow-up)

- Audit logging for property and filter mutations
- UI for managing user properties and filter rules

Yes — the five new endpoints (`/user-properties`, `/app-users/:id` PATCH, `/datasets/:name/access-filter`) are not yet documented in `docs/api.yaml`. Documentation should be added before this is merged.

### DIFF
--- a/lib/model/migrations/20260424-01-user-properties-and-dataset-filtering-by-userproperties-01.up.sql
+++ b/lib/model/migrations/20260424-01-user-properties-and-dataset-filtering-by-userproperties-01.up.sql
@@ -1,0 +1,59 @@
+-- Registry of property names that can be assigned to actors, scoped per project.
+CREATE TABLE actor_properties (
+    id integer NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    "projectId" integer NOT NULL REFERENCES projects (id) ON DELETE CASCADE,
+    "name" text NOT NULL CHECK (length("name") > 0)
+);
+CREATE UNIQUE INDEX ON actor_properties ("projectId", "name");
+
+
+CREATE TABLE actor_property_values (
+    id integer GENERATED ALWAYS AS IDENTITY,
+    "actorId" integer NOT NULL REFERENCES field_keys ("actorId") ON DELETE CASCADE,  -- FK to field_keys for now; can be widened to actors if needed.
+    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE,
+    "value" text NOT NULL CHECK (length("value") > 0)
+);
+-- Each actor can only have one value per property. Multivalued properties were intentionally avoided
+-- to keep filtering semantics simple (no conjunction/disjunction complexity).
+-- This index also serves as the index for the FK referent side (used when CASCADE-ing).
+CREATE UNIQUE INDEX ON actor_property_values ("actorId", "actorPropertyId");
+
+
+-- Maps a dataset property to an actor property, defining a filter rule:
+-- entities are visible to an actor only if their value for "datasetPropertyId" matches the actor's value for "actorPropertyId".
+CREATE TABLE dataset_access_filters (
+    "datasetId" integer NOT NULL,  -- Redundant with "datasetPropertyId" but needed for the composite FK below to prevent insertion anomalies.
+    "datasetPropertyId" integer NOT NULL PRIMARY KEY,
+    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE
+);
+-- The unique index on ds_properties enables the composite FK, which ensures "datasetId" is always
+-- consistent with "datasetPropertyId" (preventing insertion anomalies).
+CREATE UNIQUE INDEX "dataset_access_filters__unique_for_composite_fk_referent" ON ds_properties USING btree ("datasetId", id);
+ALTER TABLE dataset_access_filters
+    ADD CONSTRAINT "dataset_access_filters__composite_fk" FOREIGN KEY ("datasetId", "datasetPropertyId") REFERENCES ds_properties ("datasetId", id) ON DELETE CASCADE;
+-- Ensures each actor property is used at most once per dataset.
+CREATE UNIQUE INDEX "dataset_access_filters__spend_actorprop_once_per_dsprop" ON dataset_access_filters ("datasetPropertyId", "actorPropertyId");
+
+
+CREATE VIEW actor_dataset_filter AS (
+    WITH filter_as_json AS (
+        SELECT
+            pfilter."datasetId",
+            aprop."actorId",
+            jsonb_object_agg (dsprops."name", aprop."value") AS thefilter
+        FROM
+            dataset_access_filters pfilter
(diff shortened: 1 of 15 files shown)


lib/model/migrations/20260424-01-user-properties-and-dataset-filtering-by-userproperties-01.up.sql
```diff
@@ -0,0 +1,65 @@
...
+            (pfilter."datasetId", aprop."actorId")
+    )
+    SELECT
+        *,
+        md5(thefilter::text) AS filterhash  -- Used to make the filter part of an HTTP resource for incremental entity list download.
+    FROM
+        filter_as_json
+);
+
+-- Commented out: would take a very long time on large databases.
```
brontolosone, line 64: But in the end one'd rather have an indexed search given how large entity collections can get...

lib/model/query/entities.js
```diff
@@ -918,14 +918,17 @@ ${options.skiptoken ? sql`
...
+LEFT OUTER JOIN actor_dataset_filter propfilter
+  ON propfilter."datasetId" = entities."datasetId" AND propfilter."actorId" = ${actorId}
 WHERE
   entities."datasetId" = ${datasetId}
   AND ${sqlEquals(options.condition)}
   AND ${odataExcludeDeleted(options.filter, odataToColumnMap)}
   AND entity_defs.current=true
   AND ${odataFilter(options.filter, odataToColumnMap)}
   AND ${searchClause(options.search)}
+  AND (propfilter.thefilter IS NULL OR propfilter.thefilter <@ entity_defs.data)
```
brontolosone, line 931: if you'd remove the "`propfilter.thefilter IS NULL OR `", you'd go from fail-open to fail-closed.
Then only those who a) have all the properties matched on + b) have the right values for those properties will see entities.
And then you can ditch the `LEFT OUTER JOIN` and instead do an inner join based on the containment predicate, which should be better.

But currently I think we're leaning towards fail-open.

lib/model/migrations/20260424-01-user-properties-and-dataset-filtering-by-userproperties-01.up.sql
```diff
@@ -0,0 +1,59 @@
...
+CREATE VIEW actor_dataset_filter AS (
+    WITH filter_as_json AS (
+        SELECT
+            pfilter."datasetId",
+            aprop."actorId",
+            jsonb_object_agg (dsprops."name", aprop."value") AS thefilter
+        FROM
+            dataset_access_filters pfilter
+            INNER JOIN ds_properties dsprops ON (pfilter."datasetPropertyId" = dsprops.id)
+            INNER JOIN actor_property_values aprop USING ("actorPropertyId")
```
brontolosone, line 47: Potentially we'd want to constrain the `field_key`s to those of the project that the dataset property belongs for.
So we'd want to select only those `actor_property_values`s of which the project ID from the `field_key` table matches the project ID of the `ds_properties`'s dataset.

That, or we could add more referential constraints to the filter table, so that you simply can't equate a userprop from one project to a dsprop of a different project. That'd be more efficient - check once upon write of the definitions, instead of checking every time in every query.

Or not worry about this possible inconsistency, also a popular option 🙃

ktuite:
This PR is being split into multiple other PRs so this one can be closed.

PR #1823, opened 2026-05-05
Features: Return publishing notes for the Form and Versions API

sadiqkhoja:
Towards https://github.com/getodk/central/issues/1834

Added test, checked with frontend.

It is small and simple change to achieve the objective. I thought about introducing symbol/attribute in our `Frame` framework for `restrictedReadonly` field but that would have been a complicated change. We may consider that approach if there is proliferation of conditional fields in our APIs.

I added `publishingNotes` field for the `/forms/:xmlFormId` as well because we need to show notes for the "Published" Form page as well.

Is `publishingNotes` a right name for this field? We can call this just `notes`.

Once related frontend PRs are merged, QA team should test all Form and Form version functionality. 

Added.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -3361,7 +3361,7 @@ paths:
                 updatedAt: 2018-03-21T12:45:02.312Z
             application/json; extended:
               schema:
-                $ref: '#/components/schemas/ExtendedForm'
+                $ref: '#/components/schemas/ExtendedFormWithPublishNotes'
               example:
                 projectId: 1
                 xmlFormId: simple
@@ -3389,6 +3389,7 @@ paths:
                   deletedAt: 2018-04-18T23:42:11.406Z
                 entityRelated: false
                 publicLinks: 4
+                publishNotes: 'Fixed validation rules for required fields'
     delete:
       tags:
       - Individual Form
@@ -12618,6 +12619,11 @@ components:
               type: number
               example: 4
               description: The number of Public Links that can submit to the Form. This does not include Public Links that have been revoked.
+    ExtendedFormWithPublishNotes:
+      allOf:
+        - $ref: '#/components/schemas/ExtendedForm'
+        - type: object
+          properties:
             publishNotes:
               type: string
               example: Fixed validation rules for required fields
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -44,7 +44,7 @@ info:
     ## ODK Central v2026.2
 
     **Changed**:
-    - The [Getting Form Details](/central-api-form-management/#getting-form-details), [Listing Published Form Versions](/central-api-form-management/#listing-published-form-versions) and [Getting Form Version Details](/central-api-form-management/#getting-form-version-details) endpoints now return a `publishingNotes` field (conditionally) in extended metadata. 
+    - The [Getting Form Details](/central-api-form-management/#getting-form-details), [Listing Published Form Versions](/central-api-form-management/#listing-published-form-versions) and [Getting Form Version Details](/central-api-form-management/#getting-form-version-details) endpoints now return a `publishNotes` field in extended metadata. 
 
     ## ODK Central v2026.1
 
@@ -4618,7 +4618,7 @@ paths:
                   type: user
                   updatedAt: 2018-04-18T23:42:11.406Z
                   deletedAt: 2018-04-18T23:42:11.406Z
-                publishingNotes: Fixed validation rules for required fields
+                publishNotes: Fixed validation rules for required fields
   /v1/projects/{projectId}/forms/{xmlFormId}/versions/{version}:
     get:
       tags:
@@ -12618,9 +12618,10 @@ components:
               type: number
               example: 4
               description: The number of Public Links that can submit to the Form. This does not include Public Links that have been revoked.
-            publishingNotes:
+            publishNotes:
               type: string
-              description: The notes provided when this version was published (via the `X-Action-Notes` header). Only returned to users with `form.update` permission.
+              example: Fixed validation rules for required fields
+              description: The notes provided when the current Form version was published (via the `X-Action-Notes` header). Only returned to users with `form.update` permission.
     ExtendedFormVersion:
       allOf:
         - $ref: '#/components/schemas/Form'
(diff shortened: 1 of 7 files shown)


docs/api.yaml
```diff
@@ -4654,6 +4660,9 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Form'
+            application/json; extended:
+              schema:
+                $ref: '#/components/schemas/ExtendedFormVersion'
```
sadiqkhoja, line 4666: `/version/:versionString` supports extended metadata but this was missing from our API docs.

matthew-white:
> Is `publishingNotes` a right name for this field? We can call this just `notes`.

I like `publishingNotes` more than `notes`. :+1: Seems more descriptive. I'm not sure the "ing" is really needed, maybe could be the shorter `publishNotes`, but either seems good to me.

docs/api.yaml
```diff
@@ -12609,6 +12618,9 @@ components:
               type: number
               example: 4
               description: The number of Public Links that can submit to the Form. This does not include Public Links that have been revoked.
+            publishingNotes:
+              type: string
+              description: The notes provided when this version was published (via the `X-Action-Notes` header). Only returned to users with `form.update` permission.
```
matthew-white: How about adding an example, I like the string from above:

```yaml
example: Fixed validation rules for required fields
```

matthew-white:
```suggestion
              description: The notes provided when the current Form version was published (via the `X-Action-Notes` header). Only returned to users with `form.update` permission.
```

docs/api.yaml
```diff
@@ -41,6 +41,11 @@ info:
 
     Here major and breaking changes to the API are listed by version.
 
+    ## ODK Central v2026.2
+
+    **Changed**:
+    - The [Getting Form Details](/central-api-form-management/#getting-form-details), [Listing Published Form Versions](/central-api-form-management/#listing-published-form-versions) and [Getting Form Version Details](/central-api-form-management/#getting-form-version-details) endpoints now return a `publishingNotes` field (conditionally) in extended metadata.
```
matthew-white: I would either remove "(conditionally)" (it's explained below) or explain here what the condition is (maybe in a follow-up sentence). I think the sentence is slightly unclear as-is.

lib/resources/forms.js
```diff
@@ -29,13 +29,26 @@ const excelMimeTypes = {
   xlsx: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
 };
 
-const canReadForm = (auth, form) => {
+const _checkActorPermission = (authOrVerbs, requiredVerbs, form) => {
+  if (authOrVerbs instanceof Array) {
+    const requiredVerbsArray = typeof requiredVerbs === 'string' ? [requiredVerbs] : requiredVerbs;
```
matthew-white: Could we require `requiredVerbs` to be an array? Since it's only called from `canReadForm()`, I feel like that'd be reasonable. I think there's enough going on in `_checkActorPermission()` that it'd be nice to simplify that aspect of it and remove this line. I know it's just one line though. 😅

lib/resources/forms.js
```diff
@@ -29,13 +29,26 @@ const excelMimeTypes = {
   xlsx: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
 };
 
-const canReadForm = (auth, form) => {
+const _checkActorPermission = (authOrVerbs, requiredVerbs, form) => {
```
matthew-white:
```suggestion
const _checkActorVerbs = (authOrVerbs, requiredVerbs, form) => {
```

Just thinking that we don't use the term "permission" very much in the codebase.

matthew-white: What do you think about always passing `_checkActorPermission()` verbs, not `auth`? It'd be easy for `canReadForm()` to convert `auth` to verbs. I feel like `_checkActorPermission()` is an important function, but it's slightly on the complicated side right now. Sometimes it calls `auth.canOrReject()`, while other times it reimplements some of the logic from `auth.can()`.

I'm thinking of something like:

```js
const checkActorVerbs = (actorVerbs, requiredVerbs) => {
  const actorVerbsSet = new Set(actorVerbs);
  if (requiredVerbs.every(v => !actorVerbsSet.has(v))) {
    return reject(Problem.user.insufficientRights());
  }
  return true;
};

const canReadForm = async (authOrVerbs, form) => {
  const verbs = authOrVerbs instanceof Array
    ? authOrVerbs
    : (await auth.verbsOn(form));
  if (form.def?.publishedAt == null) {
    await checkActorVerbs(verbs, ['form.update']);
  } else if (form.state === 'closed') {
    await checkActorVerbs(verbs, ['form.read']);
  } else {
    await checkActorVerbs(verbs, ['open_form.read', 'form.read']);
  }
  return form;
};
```

I can definitely see the reasoning behind `_checkActorPermission()`. It prevents a second call to `auth.can()` when checking whether the user can read `publishingNotes`. I'm just wondering whether there's a way to make the auth logic crystal clear.

test/integration/api/forms/versions.js
```diff
@@ -383,6 +505,105 @@ describe('api: /projects/:id/forms (versions)', () => {
                   text.should.equal('this is goodone.csv');
                 })))));
       });
+
+      describe('publishing notes', () => {
```
matthew-white: This `describe()` block doesn't check public links, while the `describe()` block above does. Personally, I feel like that's OK. Authz is driven by verbs, not actor type.

test/util/scenarios.js
matthew-white, line 1: Love this scenarios file!

matthew-white (approved):

matthew-white (changes requested):
I just had the thought, what about the form list endpoint? Can't that return extended metadata? Would it return `publishNotes` to anyone who can `form.list`, including app users, etc.?

I'm thinking so, because in lib/model/query/forms.js, it looks like both `Forms.getByProjectId()` and `Forms.getByProjectAndXmlFormId()` call the same `_get()` function, which calls the `_getSql()` function that this PR modifies.

In addition to the form list endpoint, there's also /v1/projects?forms=true.

Even if my guess is wrong, and no changes are needed to lib/, it'd be nice to get a test that those endpoints don't return `publishNotes`. I don't feel like we need a test for each actor type, just basic checks that those endpoints won't return `publishNotes` to some actor who can't `form.update`.

Now that this example is in my mind, I"m beginning to feel like it'd be easy to miss this sort of thing going forward. Instead of making `publishNotes` `readable`, requiring an explicit check to exclude it, what do you think about making it not `readable` and instead requiring the endpoint to explicitly _include_ it?

matthew-white (approved):

PR #1824, opened 2026-05-05
Add actor properties: project-level key-value pairs settable on app users and public links

### LINKED ISSUE #1868: Backend — register, list, and assign project-level user properties for app users and public access links
Add backend support for defining user properties at the project level and assigning values to App Users and Public Access Links.

### Project-level user properties
- Migration to add storage for project user properties
- `POST /projects/:id/user-properties` — register a new property name for a project
- `GET /projects/:id/user-properties` — list all property names for a project

### Values on App Users and Public Access Links
- Storage for per-actor property values
- `PATCH /projects/:id/app-users/:id` — set or unset property values on an App User
- Equivalent endpoint for Public Access Links — set or unset property values
- Return assigned values when reading App User / Public Access Link records

## Naming and constraints
- Property names follow the same restrictions as dataset property names
- Property names cannot be `displayName`
- Values are free text
- Called "user properties" in code (frontend copy may differ — TBD)

## Out of scope
- Deleting user properties (future)
- Applying filters to entity downloads (Phase 2)
- Audit logging (tracked as a separate Phase 1 issue)
- User properties on Web users

ktuite:
Closes getodk/central#1868

Adds support for project-level "actor properties" — named key-value pairs that can be defined on a project and then set on individual app users and public links. These are intended for use by clients to attach metadata (e.g. region, worker ID) to actors for later filtering or reporting.

## New endpoints

**Actor property definitions** (project-scoped):
- `POST /projects/:id/actor-properties` — define a new property name for a project
- `GET /projects/:id/actor-properties` — list all defined property names for a project
- `GET /projects/:id/actor-properties` with `X-Extended-Metadata` returns each property with a `values` array of distinct values currently in use across all actors (empty array if none set).

**App users:**
- `GET /projects/:projectId/app-users/:id` — fetch a single app user (respects `X-Extended-Metadata`)
- `PATCH /projects/:projectId/app-users/:id` — set/unset property values on an app user

**Public links:**
- `GET /projects/:projectId/forms/:xmlFormId/public-links/:id` — fetch a single public link (respects `X-Extended-Metadata`)
- `PATCH /projects/:projectId/forms/:xmlFormId/public-links/:id` — set/unset property values on a public link

## Behavior

- Property names must be pre-defined on the project before values can be set on actors.
- The `PATCH` body accepts `{ properties: { name: "value" | null } }`. Passing `null` unsets the property. String values are trimmed; empty strings are treated as `null`.
- Properties are returned in extended metadata responses as a `properties` object (or `null` if no values are set).
- Both app users and public links share the same underlying `actor_properties` / `actor_property_values` tables.

Integration tests for all new endpoints: setting/unsetting properties, partial unset, 403/404
cases, and extended metadata responses (including `null` when no values are set). Tests cover
both app users and public links.

Property names are defined at the project level and shared across app users and public links to
avoid duplicating name management per actor type. Values are in a separate `actor_property_values`
table so unset properties leave no rows. `properties` is `null` in responses when nothing is set.

No existing endpoints changed, but the extended metadata list responses for app users and public
links now include a `properties` field (always `null` unless values have been set). There may be
a minor performance impact from the additional join on those list queries.

Yes — not included in this PR yet.

### DIFF
--- a/lib/model/migrations/20260505-01-add-project-user-properties.js
+++ b/lib/model/migrations/20260505-01-add-project-user-properties.js
@@ -1,0 +1,37 @@
+const up = async (db) => {
+  await db.raw(``);
+
+  // -- Registry of property names that can be assigned to actors, scoped per project.
+  await db.raw(`
+  CREATE TABLE actor_properties (
+    id integer NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    "projectId" integer NOT NULL REFERENCES projects (id) ON DELETE CASCADE,
+    "name" text NOT NULL CHECK (length("name") > 0)
+  )`);
+
+  await db.raw(`CREATE UNIQUE INDEX ON actor_properties ("projectId", "name")`);
+
+  // -- Values of actor properties
+  await db.raw(`
+  CREATE TABLE actor_property_values (
+    id integer GENERATED ALWAYS AS IDENTITY,
+    "actorId" integer NOT NULL REFERENCES field_keys ("actorId") ON DELETE CASCADE,  -- FK to field_keys for now; can be widened to actors if needed.
+    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE,
+    "value" text NOT NULL CHECK (length("value") > 0)
+  )`);
+
+  // -- Each actor can only have one value per property. Multivalued properties were intentionally avoided
+  // -- to keep filtering semantics simple (no conjunction/disjunction complexity).
+  // -- This index also serves as the index for the FK referent side (used when CASCADE-ing).
+  await db.raw(`CREATE UNIQUE INDEX ON actor_property_values ("actorId", "actorPropertyId")`);
+
+  // -- Index to support FK lookup on actorPropertyId (CASCADE deletes from actor_properties).
+  await db.raw(`CREATE INDEX ON actor_property_values ("actorPropertyId")`);
+};
+
+const down = async (db) => {
+  await db.raw('DROP TABLE actor_property_values CASCADE');
+  await db.raw('DROP TABLE actor_properties CASCADE');
+};
+
+module.exports = { up, down };
--- a/lib/model/migrations/20260505-01-add-project-user-properties.js
+++ b/lib/model/migrations/20260505-01-add-project-user-properties.js
@@ -1,5 +1,4 @@
 const up = async (db) => {
-  await db.raw(``);
 
   // -- Registry of property names that can be assigned to actors, scoped per project.
   await db.raw(`
--- a/lib/model/migrations/20260505-01-add-project-user-properties.js
+++ b/lib/model/migrations/20260505-01-add-project-user-properties.js
@@ -15,7 +15,7 @@ const up = async (db) => {
   await db.raw(`
   CREATE TABLE actor_property_values (
     id integer GENERATED ALWAYS AS IDENTITY,
-    "actorId" integer NOT NULL REFERENCES field_keys ("actorId") ON DELETE CASCADE,  -- FK to field_keys for now; can be widened to actors if needed.
+    "actorId" integer NOT NULL REFERENCES actors ("id") ON DELETE CASCADE,
     "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE,
     "value" text NOT NULL CHECK (length("value") > 0)
   )`);
--- a/lib/model/query/actor-properties.js
+++ b/lib/model/query/actor-properties.js
@@ -2,7 +2,6 @@ const { sql } = require('slonik');
 const { map } = require('ramda');
 const { ActorProperty } = require('../frames');
 const { construct } = require('../../util/util');
-const { getOrNotFound } = require('../../util/promise');
 
 
 // Creates a new actor property name for a project.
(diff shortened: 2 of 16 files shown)


lib/resources/actor-properties.js
```diff
@@ -0,0 +1,30 @@
...
+      throw Problem.user.unexpectedValue({ field: 'name', value: name, reason: 'This is not a valid property name.' });
+
+    await ActorProperties.create(project.id, name);
+    return success;
+  }));
+
+  // List actor property names for a project.
+  service.get('/projects/:projectId/actor-properties', endpoint(async ({ ActorProperties, Projects }, { auth, params, queryOptions }) => {
+    const project = await Projects.getById(params.projectId).then(getOrNotFound);
+    await auth.canOrReject('project.read', project);
```
sadiqkhoja: this seems too liberal to me :) Data Collectors and Project Viewers also have `project.read` verb. I don't think they should be able to see actor properties and its values

lib/model/migrations/20260505-01-add-project-user-properties.js
```diff
@@ -0,0 +1,37 @@
+const up = async (db) => {
+  await db.raw(``);
```
sadiqkhoja: dead code

lib/resources/actor-properties.js
```diff
@@ -0,0 +1,30 @@
...
+    const { name } = body;
+    if (name == null || !validateActorPropertyName(name))
+      throw Problem.user.unexpectedValue({ field: 'name', value: name, reason: 'This is not a valid property name.' });
+
+    await ActorProperties.create(project.id, name);
+    return success;
+  }));
+
+  // List actor property names for a project.
+  service.get('/projects/:projectId/actor-properties', endpoint(async ({ ActorProperties, Projects }, { auth, params, queryOptions }) => {
```
sadiqkhoja, line 22: I think we should have another endpoint to return values of a single actor-property `GET /projects/:projectId/actor-properties/:name/values` (with /values or without - no strong opinions). What I'm imagining is that on the frontend we would have autocomplete dropdown that would require the list of values for a single property (maybe with the `startsWith` filter)

sadiqkhoja: This can be done in the subsequent PR as well

lib/resources/app-users.js
```diff
@@ -29,6 +31,13 @@ module.exports = (service, endpoint) => {
         return FieldKeys.create(fk, project);
       })));
 
+  service.get('/projects/:projectId/app-users/:id', endpoint(({ FieldKeys, Projects }, { auth, params, queryOptions }) =>
+    Projects.getById(params.projectId)
+      .then(getOrNotFound)
+      .then((project) => auth.canOrReject('field_key.delete', project))
```
sadiqkhoja:
```suggestion
      .then((project) => auth.canOrReject('field_key.list', project))
```

lib/resources/app-users.js
```diff
@@ -38,5 +47,20 @@ module.exports = (service, endpoint) => {
       .then((fk) => Actors.del(fk.actor))
       .then(success)));
 
+  // Set/unset actor property values on an app user.
+  // Body: { properties: { propName: "value" | null, ... } }
+  service.patch('/projects/:projectId/app-users/:id', endpoint(async ({ ActorProperties, FieldKeys, Projects }, { auth, params, body }) => {
```
sadiqkhoja, line 52: should we support updating `displayName`?

lib/resources/public-links.js
```diff
@@ -29,6 +31,28 @@ module.exports = (service, endpoint) => {
...
+  service.get('/projects/:projectId/forms/:xmlFormId/public-links/:id', endpoint(({ Forms, PublicLinks }, { auth, params, queryOptions }) =>
+    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.WithoutDef)
+      .then(getOrNotFound)
+      .then((form) => auth.canOrReject('public_link.read', form))
+      .then((form) => PublicLinks.getByFormAndActorId(form.id, params.id, queryOptions))
+      .then(getOrNotFound)));
+
+  // Set/unset actor property values on a public link.
+  // Body: { properties: { propName: "value" | null, ... } }
+  service.patch('/projects/:projectId/forms/:xmlFormId/public-links/:id', endpoint(async ({ ActorProperties, Forms, PublicLinks }, { auth, params, body }) => {
```
sadiqkhoja, line 43: should we support updating `displayName`?

lib/model/query/public-links.js
```diff
@@ -25,11 +25,16 @@ const create = (publicLink, form) => ({ Actors, Assignments, Sessions }) =>
 create.audit = (pl) => (log) => log('public_link.create', pl.actor, { data: pl });
 create.audit.withResult = true;
 
-const _get = extender(PublicLink, Actor, Frame.define('token', readable))(Actor.alias('created_by', 'createdBy'))((fields, extend, options) => sql`
+const _get = extender(PublicLink, Actor, Frame.define('token', readable))(Actor.alias('created_by', 'createdBy'), Frame.define('properties', readable))((fields, extend, options) => sql`
 select ${fields} from public_links
 join actors on public_links."actorId"=actors.id
 left outer join sessions on public_links."actorId"=sessions."actorId"
 ${extend|| sql`join actors as created_by on public_links."createdBy"=created_by.id`}
+${extend|| sql`left outer join
```
sadiqkhoja, line 33: I'd expect to get custom property even without `x-metadata-extended`, maybe it's just me. Let's ask Hélène about it

lib/model/query/actor-properties.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+// Sets or unsets multiple property values for an actor in a single operation.
+// properties is a plain object mapping property names to values (string or null),
+// as returned by extractActorProperties() in lib/data/actor-properties.js.
+const setValuesForActor = (projectId, actorId, properties) => async (container) => {
+  for (const [name, value] of Object.entries(properties)) {
+    // eslint-disable-next-line no-await-in-loop
+    await setValueForActor(projectId, actorId, name, value)(container);
+  }
```
sadiqkhoja: this can be terrible performance-wise if there are many properties.

test/integration/api/public-links.js
```diff
@@ -148,6 +148,150 @@ describe('api: /projects/:id/forms/:id/public-links', () => {
...
+        .expect(200)
+        .then(({ body }) => {
+          body.find((pl) => pl.id === pl1.id).properties.should.eql({ region: 'north' });
+          should(body.find((pl) => pl.id === pl2.id).properties).be.null();
+        });
+    }));
+  });
+
+  describe('/:id GET', () => {
+    it('should return 403 unless the user can delete', testService(async (service) => {
```
sadiqkhoja:
```suggestion
    it('should return 403 unless the user can read', testService(async (service) => {
```

lib/model/query/actor-properties.js
ktuite: True, but hopefully there are not many properties getting set at once.

I could definitely refactor this into a differently-shaped query. Some related questions:

1. This PR is letting properties be un-set by sending `{"propName": null}` so when that's seen, there needs to be a `DELETE` sql query for that property value. Does that all sound fine, unsetting via sending `null` and deleting the value from the table? With entities, we track all versions, but for these properties, I think it's ok to not keep old values around somewhere. 
 
2. If someone mentioned a property that doesn't exist, like `{"existingProp": "value", "nonexistantProp": "value"}` should they get an error? Should it be silently ignored? Should it _create_ the missing property as it assigns the value?

lib/model/query/public-links.js
ktuite: I think of extended metadata as being useful for fetching data from more complex joined queries. Also for having a way in the code to get the data from the base table(s). For these public links, we have code that gets a public link to get its actor (like for deleting a public link) and we really don't need the full creator info or properties.

lib/resources/actor-properties.js
ktuite: I did not pick good verbs here 😅 will go and fix.

lib/resources/app-users.js
ktuite: I noticed there's no existing way to update `displayName` and I don't know if that's intentional or not. We can ask. If it's not intentional to keep it uneditable then I think we should definitely let it be edited in this PATCH endpoint.

lib/resources/actor-properties.js
ktuite: This endpoint will return values if extended-metadata is requested (see below `ActorProperties.getAllForProjectWithValues(project.id)`) 

I imagine for a frontend auto-complete, it may want all the values for all the properties at once and we don't need to make a per-property endpoint?

lib/resources/actor-properties.js
ktuite: Do you think there should be a different verb that is tailored to these project settings/actor properties?

Like a `actor_properties.list`(admins, managers, viewers) and `actor_properties.update` (admins, managers) ?

sadiqkhoja: `actor_properties.<verb>` sounds good to me

lib/resources/actor-properties.js
sadiqkhoja: hmm I see values for all properties at once can work as well. But if there are many properties and many values then response size could be quite large. 

From implementation point of view I feel per-property endpoint would be easier to consume.

lib/model/query/actor-properties.js
sadiqkhoja: 1. I think it is fine. We don't really have actor_defs table to keep the history. If required we can layer it on later
2. Returning error (400 bad request) feels better option.

lib/resources/actor-properties.js
ktuite: Issue for single property endpoint: https://github.com/getodk/central/issues/1903

sadiqkhoja (approved):
Looks great 👍

PR #1829, opened 2026-05-13
Fixes: create db schema and verbose logging

sadiqkhoja:
- Create `public` schema if doesn't exists (it can be dropped by `DROP OWNED BY CURRENT_USER CASCADE`)
- print all messages from `psql`


lognaturel (approved):
That’s what we discussed, thanks!

PR #1830, opened 2026-05-18
Log creating and setting properties on app users and public links

### LINKED ISSUE #1869: Backend — audit logging for user property mutations
Add audit log entries for user property creation and value assignment, so that project managers can track changes to user properties and the values assigned to App Users and Public Access Links.

- Log when a new user property is registered on a project
- Log when a property value is set or unset on an App User
- Log when a property value is set or unset on a Public Access Link
- Check on Frontend that these events come through on the system audit log page. 

Depends on
- https://github.com/getodk/central/issues/1868

ktuite:
Closes https://github.com/getodk/central/issues/1869

Added integration tests covering audit log entries for:
- Creating an actor property (`actor_property.create`)
- Setting properties on app users at creation and via PATCH (`field_key.property.set`)
- Setting properties on public links at creation and via PATCH (`public_link.property.set`)

Both `actor_property.create` and `field_key/public_link.property.set` audit logging is handled via `.audit` functions attached directly to the query functions in `actor-properties.js`. 

The action name for setting a property on a Field Key or Public Link is derived dynamically from `fieldKeyOrLink.constructor.actorType`, so the shared `setValuesForActor` query function can emit the correct action (`field_key.property.set` vs. `public_link.property.set`) without needing separate code paths in the resource layer. 

The actee for `actor_property.create` is the project; the actee for `property.set` events is the app user or public link actor.

The `field_key` and `public_link` audit action filters have been updated to include the new `property.set` events, and `actor_property.create` has been added to the `project` filter.

Should just make this top level feature more auditable.

Yes, working on separate PR for that now.

### DIFF
--- a/lib/resources/app-users.js
+++ b/lib/resources/app-users.js
@@ -30,7 +30,7 @@ module.exports = (service, endpoint) => {
     if (body.properties != null) {
       const knownProperties = await ActorProperties.getAllForProject(project.id);
       const properties = extractActorProperties(body.properties, knownProperties.map((p) => p.name));
-      await ActorProperties.setValuesForActor(project.id, fk.actorId, properties);
+      await ActorProperties.setValuesForActor(project.id, fk, properties);
     }
 
     // Returns the raw FK from create() without new properties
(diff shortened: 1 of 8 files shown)


lib/resources/app-users.js
```diff
@@ -64,7 +64,7 @@ module.exports = (service, endpoint) => {
     if (body.properties != null) {
       const knownProperties = await ActorProperties.getAllForProject(project.id);
       const properties = extractActorProperties(body.properties, knownProperties.map((p) => p.name));
-      await ActorProperties.setValuesForActor(project.id, fk.actorId, properties);
+      await ActorProperties.setValuesForActor(project.id, fk, properties);
```
sadiqkhoja, line 67: Just wondering if it is possible to have `properties` to be empty array? In that case event still be logged. Maybe that is fine, what do we do in other similar situations.

sadiqkhoja (approved):

lib/resources/app-users.js
ktuite: We usually don't log an event if nothing is actually changed/set. I've updated the code and to reflect this and added a test.

PR #1832, opened 2026-05-19
test: run unit tests before integration tests

alxndrsn:
This change:

* isolates unit tests from integration tests
* loads less irrelevant files when running mocha test suites (e.g. `test/util/**`, `test/format/**`, `test/data/**` only loaded on demand)
* protects against mistakes like that fixed in https://github.com/getodk/central-backend/pull/1828

This change will allow a failed unit test to fail the full test pipeline significantly faster than previously.

- [x] run locally
- [x] run in CI

* faster failures
* simpler understanding
* less chance of surprising things happening when new files are introduced in `test` directory.

An alternative approach (or additional change) could be to "shift left" more by:

```Makefile
test: lint test-unit test-integration-fast test-integration-slow
```

No effect - just tests.

No.


.mocharc.js
matthew-white, line 1: What do you think about removing this file entirely now that it's blank?

Makefile
matthew-white: It looks to me like we have a similar issue with `test-fast`. If we blank out .mocharc.js, then someone runs `test-fast`, I think that would run unintended files. I don't think we can blank out .mocharc.js without also doing something about `test-fast`. FWIW I personally never use `test-fast`.

.mocharc.js
alxndrsn: Happy either way.  Do you have a preference?

Makefile
alxndrsn: Good spot.  I also never use it.  I'll see if anyone does before putting more thought into it.

ktuite: I haven't been using it but I do get annoyed by the slow tests sometimes so I'd _like_ to be able to use it. Not necessary but if it stays, I'll try to use it more.

.mocharc.js
matthew-white: If you don't have a preference, I think it could be nice to remove it. One less file to look at. We could always add it back later. No strong feelings though.

alxndrsn: Removed.

Makefile
alxndrsn: Fixed :+1:

alxndrsn: On `master` for me `make test` takes 4 minutes and `make test-fast` takes 3.  I suspect a few more `@slow` annotations might be good to make `test-fast` even more useful.

Makefile
```diff
@@ -110,7 +111,8 @@ test-db-migrations:
 
 .PHONY: test-fast
 test-fast: node_version
-	NODE_CONFIG_ENV=test BCRYPT=insecure npx mocha --recursive --fgrep @slow --invert
```
matthew-white, line 113: I'm noticing that `test-fast` set `NODE_CONFIG_ENV=test`, while `test` didn't. It probably doesn't really matter for this PR though. This PR is taking a different approach with `test` and `test-fast`, relating them to `test-unit` and `test-integration`. Even `test-unit` and `test-integration` do different things from one another when it comes to `NODE_CONFIG_ENV`.

matthew-white (approved):

ktuite (approved):
i like the approach of recursively calling make-test and make-integration in order to preserve make test-fast, which i'll be using more!

Makefile
alxndrsn: Due to line 3 (https://github.com/getodk/central-backend/blob/05bc5c0cf57c0f44f00232ebeaa1cc69e71ae22b/Makefile#L3), `NODE_CONFIG_ENV` is pretty much always `test` anyway.

I'm not really sure of the purpose of this, as there is no `config/test.json`.  Perhaps @sadiqkhoja can give more context.

Given that `NODE_CONFIG_ENV` defaults to `test`, and there is no `config/test.json`, I don't think this change is going to have an obvious effect.

PR #1834, opened 2026-05-21
Update NodeJS version to match prod

### LINKED ISSUE #1806: Update Node version in dev to match prod
Right now, the version of Node in prod is 24.14.1. However, in dev, it's 24.13.0. This issue is to update the version in dev to match prod.

I've made a note in #1805 about updating Node in dev so that we update prod and dev at the same time for v2026.2. CC @sadiqkhoja, who will likely be the one to work on #1805.

alxndrsn:
Closes getodk/central#1806

* [x] CI

Alternative was to include the changes to `package-lock.json`: running `npm i` locally removes `"peer": true` from many dependencies.  However, without that change `npm ci` still seems happy.

No effect.

No.


matthew-white (approved):

matthew-white:
We should probably make a similar change to `central-frontend` as well. Probably after getodk/central-frontend#1549 gets merged, since that also touches package.json.

matthew-white:
> Alternative was to include the changes to `package-lock.json`: running `npm i` locally removes `"peer": true` from many dependencies.

We'll probably need to commit those changes at some point, but I agree that that can happen in a later PR. Perhaps whenever we update dependencies next.

PR #1835, opened 2026-05-21
Prep API docs changelog for .2 release

ktuite:
I've seen some git merge issues in the API docs in the past where if two separate PRs edit the changelog, headers like "**Changed:**" sometimes get duplicated. This is just a little prep work to avoid that.

Additionally, LN noticed an extra field on the `GET app-users` endpoint that was included in non-extended metadata when it wasn't supposed to be. I confirmed that it's really not part of the API response.


matthew-white:
@ktuite this ended up conflicting with #1823, which also added a changelog.

ktuite:
not needed, going to make the 1 change to app user api endpoint in a different PR

PR #1837, opened 2026-05-27
Set actor properties when creating public link or app user

ktuite:
This PR makes it possible to send actor properties in the POST creation request for app users and public links. Previously, you could only set the properties in a follow-up PATCH request. 

This is a piece that was missing from https://github.com/getodk/central-backend/pull/1824
Needed for
- Issue https://github.com/getodk/central/issues/1339
- Frontend PR https://github.com/getodk/central-frontend/pull/1557

Tests, using it.

The response from the POST create event is the same as before - the properties have not been added to the response. This may be a little awkward, but the POST response is already different from the GET response because it contains the full `session` details, whereas the GET only contains the token from the session.

Example public link response from POST
```
  once: null,
  id: 10,
  type: 'public_link',
  displayName: 'test link',
  createdAt: '2026-05-27T20:19:54.422Z',
  updatedAt: null,
  deletedAt: null,
  token: 'eBKNv1ZebMm5kuEqbR9ucpuYZ3vkSJdrTgnJGOVaFXSNGj0WujEiUdZUK5Hkbaex',
  csrf: 'EaecdhEr4Y1sef5pmT6ifwUtIS8hg!QbwHaWVHld89pJcNauD8j1sC8XHECP!L2G',
  expiresAt: '9999-12-31T23:59:59.000Z'
```

from GET
```
  once: null,
  id: 10,
  type: 'public_link',
  displayName: 'test link',
  createdAt: '2026-05-27T20:19:54.421Z',
  updatedAt: null,
  deletedAt: null,
  token: 'eBKNv1ZebMm5kuEqbR9ucpuYZ3vkSJdrTgnJGOVaFXSNGj0WujEiUdZUK5Hkbaex'
```

Extends API for app user and public link creation to use new actor properties.

Will update docs in followup PR 

### DIFF
--- a/lib/resources/app-users.js
+++ b/lib/resources/app-users.js
@@ -21,15 +21,21 @@ module.exports = (service, endpoint) => {
       .then((project) => auth.canOrReject('field_key.list', project))
       .then((project) => FieldKeys.getAllForProject(project, queryOptions))));
 
-  service.post('/projects/:projectId/app-users', endpoint(({ FieldKeys, Projects }, { auth, body, params }) =>
-    Projects.getById(params.projectId)
-      .then(getOrNotFound)
-      .then((project) => auth.canOrReject('field_key.create', project))
-      .then((project) => {
-        const fk = FieldKey.fromApi(body)
-          .with({ createdBy: auth.actor.map((actor) => actor.id).orNull() });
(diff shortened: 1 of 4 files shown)


sadiqkhoja (approved):
LGTM

PR #1840, opened 2026-06-01
Move webformsEnabled to form_defs

sadiqkhoja:
Towards https://github.com/getodk/central/issues/1693

Added tests + manual testing with corresponding changes in frontend code.

First commit had an approach where `webformsEnabled` column was only in `form_defs` but the resulting code was quite complicated. Latest approach follows similar pattern that of `enketoId`.

No

Yes


lib/resources/forms.js
```diff
@@ -246,6 +246,19 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
 
+  service.patch('/projects/:projectId/forms/:xmlFormId/draft', endpoint(async ({ Forms }, { auth, params, body }) => {
+    const form = await Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.DraftVersion)
+      .then(getOrNotFound);
+
+    await auth.canOrReject('form.update', form);
+
+    const defFromRequest = Form.Def.fromApi(body);
+
+    await Forms._updateDef(form.def, defFromRequest);
```
sadiqkhoja: Add audit log

matthew-white (commented):
Just a random thing I noticed while looking at the PR

**EDIT:** Sorry, forgot to add the actual review comment — see below. 😅

lib/resources/forms.js
```diff
@@ -246,6 +246,19 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
+
+    const webformsEnabled = defFromRequest.webformsEnabled ?? form.def.webformsEnabled;
+    return { ...form.forApi(), webformsEnabled };
```
matthew-white: Do you think that `updatedAt` should change as part of this endpoint? Only `forms` has an `updatedAt` column, not `form_defs`, but I think (not sure) that we bump `updatedAt` when we change the form draft. My instinct is that `updatedAt` should change, but it means that this pattern of `{ ...form.forApi(), webformsEnabled }` won't work.

matthew-white (commented):
Just a random thing I noticed while looking at the PR

lib/resources/forms.js
sadiqkhoja: right

lib/resources/forms.js
```diff
@@ -246,6 +246,19 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
+      .then(getOrNotFound);
+
+    await auth.canOrReject('form.update', form);
+
+    const defFromRequest = Form.Def.fromApi(body);
+
+    await Forms.updateDef(form, form.def, defFromRequest);
+
+    return Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.DraftVersion)
+      .then(getOrNotFound);
```
matthew-white, line 260: Re-querying the form like this makes sense to me. :+1: But I think `Forms.updateDef()` still won't have the effect of changing `updatedAt`. That's because `updatedAt` is only on the `forms` table, whereas `updateDef()` only updates `form_defs`.

I see that `Forms.updateDef()` is different from `Forms._updateDef()` in that `updateDef()` will log an audit log event, whereas `_updateDef()` will not. What do you think about making `updateDef()` also modify `form.updatedAt`? I feel like there's some precedent for that in the `Forms` query module: `replaceDef()` runs an extra query just to bump `forms.updatedAt`.

matthew-white (commented):
I'll take a fuller look soon, I just wanted to leave this initial comment.

lib/model/query/forms.js
```diff
@@ -149,10 +149,14 @@ const createNew = (partial, project) => async ({ Actees, Datasets, FormAttachmen
     0.5
   );
 
+  // If Form is encrypted then we don't enable web-forms automatically
+  // eslint-disable-next-line no-unneeded-ternary
+  const webformsEnabled = keyId ? false : true;
```
matthew-white: Just to link things together, I wondered at https://github.com/getodk/central-frontend/pull/1593#discussion_r3375906046 whether Web Forms will support submission encryption by the time of release. If so, I don't think we need this logic here.

lib/model/query/forms.js
```diff
@@ -272,8 +276,11 @@ const createVersion = (partial, form, publish, duplicating = false) => async ({
   // Note: If publish=true, we are in the enabling encryption special case, and we
   // will just copy over whatever enketo ids do or don't exist already.
 
+  // If there is an existing draft then use its webformsEnabled, otherwise published version
+  const webformsEnabled = form.def?.webformsEnabled ?? form.webformsEnabled;
```
matthew-white: > If there is an existing draft then use its webformsEnabled

If `form.def` exists, is it necessarily the draft def? In code just a little above, I see the following logic, which makes me think that `form.def.id` can be something other than `form.draftDefId`:

```js
form.def.id == null || form.def.id !== form.draftDefId
```

That makes me wonder whether the line here should be something like:

```js
const webformsEnabled = form.draftDefId != null && form.def?.id === form.draftDefId && form.def.webformsEnabled != null
  ? form.def.webformsEnabled
  : form.webformsEnabled;
```

Maybe that's more verbose than what it needs to be. Mostly I think I'm just wanting to know more about the possibilities for what `form.def` can be.

lib/model/query/forms.js
lognaturel: It's [merged](https://github.com/getodk/web-forms/issues/448) and will be included in the release! 🎉

test/integration/api/forms/forms.js
matthew-white, line 1: Given the comment above that Web Forms will support encryption - https://github.com/getodk/central-backend/pull/1840#discussion_r3376298202 - I think these tests are no longer needed.

lib/model/migrations/20260521-01-add-webformsenabled-to-form-defs.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+const down = async (db) => {
+  await db.raw(`
+    ALTER TABLE form_defs
+      DROP COLUMN "webformsEnabled";
+  `);
+
+  await db.raw(`
+    ALTER TABLE public.forms ALTER COLUMN "webformsEnabled" SET NOT NULL;
```
matthew-white, line 56: Before altering the tables, we could copy `webformsEnabled` back from `form_defs` to `forms` in the case where a form doesn't have a published version. Otherwise, you could end up with a form where `webformsEnabled` is `NULL`, causing `SET NOT NULL` to fail.

I think of down migrations as just for dev, so I also don't think it's a big deal.

lib/resources/forms.js
```diff
@@ -367,10 +380,13 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
...
 
   service.patch('/projects/:projectId/forms/:xmlFormId', endpoint(({ Forms }, { auth, params, body }) =>
-    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.WithoutDef)
+    Forms.getByProjectAndXmlFormId(params.projectId, params.xmlFormId, Form.AnyVersion)
       .then(getOrNotFound)
       .then((form) => auth.canOrReject('form.update', form))
-      .then((form) => Forms.update(form, Form.fromApi(body)))
+      .then((form) => Promise.all([
+        Forms.update(form, Form.fromApi(body)),
+        Forms._updateDef(form.def, Form.Def.fromApi(body))
```
matthew-white: I think I understand the intention here. If `webformsEnabled` is changed on the form, it should also be changed on the draft. But I'm a little confused about how that's working in the database. I thought we only set `form_defs.webformsEnabled` for a form draft and that the column was `NULL` for other form defs. But `Forms._updateDef()` here seems to imply that you can set `form_defs.webformsEnabled` on a form def even if the form def isn't a draft. Does that value get picked up by the API somehow?

I see that you null out `form_defs.webformsEnabled` on draft publish, and that's part of why I was thinking that `form_defs.webformsEnabled` is expected to be `NULL` outside the draft case.

I make a similar suggestion below, but part of me thinks that the solution is to remove `Forms._updateDef()` here and instead make `Forms.update()` smarter. We could teach `Forms.update()` that when `webformsEnabled` is changed on the form, it should be changed on the form draft as well.

matthew-white: Technically, you can also use PUT /v1/projects/:id to update a form (it will call `Forms.update()`). If a change to `forms.webformsEnabled` is always meant to change `form_defs.webformsEnabled` (at least for a form draft), I think we should consider implementing that logic in PUT /v1/projects/:id as well. It's kind of a weird endpoint, and I know we've talked about removing it one day, but it's probably good to avoid API inconsistencies where possible.

- Option 1: Call `Forms._updateDef()` in PUT /v1/projects/:id.
- Option 2 (my preference): Make `Forms.update()` a little smarter, teaching it that if `webformsEnabled` is changed on the form, it needs to be changed on the form def as well (at least for a form draft).

matthew-white (commented):
I still need to look at tests, but at this point I'm feeling like it'd probably be good to change `Forms.update()` before merging this PR. I have some questions about when a change to `forms.webformsEnabled` is supposed to change `form_defs.webformsEnabled`.

lib/model/query/forms.js
sadiqkhoja: If the Form is published then `form.def.webformEnabled` will be `NULL` hence this will use `form.webformsEnabled`

lib/model/migrations/20260521-01-add-webformsenabled-to-form-defs.js
sadiqkhoja, line 1: create DB constraint that `webformsEnabled` column on form_defs can be not-null only if the draft is unpublished.

lib/resources/forms.js
```diff
@@ -247,6 +246,20 @@ module.exports = (service, endpoint, anonymousEndpoint) => {
           // already that there is no blob hash.
           FormAttachments.update(form, attachment, null, body.dataset ? dataset.id : null))))));
 
+  // TODO: maybe just use other patch endpoint, nothing is different here
```
sadiqkhoja: remove this

matthew-white (approved):

PR #1841, opened 2026-06-02
dev/s3: tighten docker network binding

alxndrsn:
Restrict fake S3 servers to connections from the local machine.

This stops minio from being exposed on the local network or wider.

- [x] tested locally
- [x] CI

Minimal change which shouldn't affect existing users, but restricts `fake-s3-server` as it should be.

No effect - just for dev.

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ dev-s3: fake-s3-accounts base
 # MINIO_KMS_SECRET_KEY, MINIO_KMS_AUTO_ENCRYPTION enable encryption - this changes how s3 ETags are generated.
 #   See: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html
 #   See: https://github.com/minio/minio/discussions/19012
-S3_SERVER_ARGS := --network host \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,7 +54,7 @@ dev-s3: fake-s3-accounts base
 # MINIO_KMS_SECRET_KEY, MINIO_KMS_AUTO_ENCRYPTION enable encryption - this changes how s3 ETags are generated.
 #   See: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html
 #   See: https://github.com/minio/minio/discussions/19012
-S3_SERVER_ARGS := --network host \
+S3_SERVER_ARGS := -p 127.0.0.1:9000:9000 -p 127.0.0.1:9001:9001 \
```
matthew-white, line 57: Why does it need two ports out of curiosity? Is that just how our S3 setup works?

matthew-white (approved):

Makefile
alxndrsn: One is the API, the other is the web interface.

PR #1843, opened 2026-06-02
Dataset access filters

ktuite:
Closes https://github.com/getodk/central/issues/1877

#### Before submitting this PR, please make sure you have:

- [ ] run `make test` and confirmed all checks still pass, or witnessed Github completing all checks with success
- [ ] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1846, opened 2026-06-09
Refactor owner only form attachment code for clarity

ktuite:
This refactors the code around form attachments that generates hashes/etags for dataset-backed form attachments, building on the previous commit to go further in separating concerns.

The key changes are:

1. **Moved dataset OpenRosa metadata fetching out of `form-attachments.js` and into `datasets.js`** as `computeHashes` (batch) and `getDatasetActivity` (single dataset). The `_get` query in form-attachments is now simpler — it just fetches attachments with blob md5s, with no conditional dataset metadata JOINs.

How code is structured and shared in Datasets query module:
```
_fetchActivityStats(all, datasetIds, filterActorId)   ← SQL, returns Map<id, rawStats>
    └── used by:
        ├── getDatasetActivity(datasetId, filterActorId)   ← single dataset, returns rawStats object
        │       └── used in getMetadata for the lastUpdate field
        └── computeHashes(datasetIds, filterActorId)       ← batch, pipes each rawStats through:
                └── _computeHashResult(stats, filterActorId)   ← pure fn, rawStats → { openRosaHash, datasetType, filteredByAuth }
```

3. **Replaced `QueryOptions` as the ownerOnly mechanism** — instead of threading a `QueryOptions` with a special `entities.creatorId` condition through multiple layers, callers now pass a plain `filterActorId` (or `null`). This makes the intent explicit and avoids leaking query-layer abstractions into resource/business logic code.

4. **`Entities.streamForExport` now handles the ownerOnly filter itself** given an `actorId`, rather than relying on the caller to pass the right `QueryOptions`.

5. **Renamed `getOwnerOnlyOptions` → `filteredAccessActorId`** in `forms.js` to reflect what it now returns.

6. **Aligned the REST CSV ETag format with the OpenRosa manifest hash** — the `GET /datasets/:name/entities.csv` ETag is now computed using the same JSON structure (`{ latestEntityCreatedOrUpdated, latestAuditEntry }`) as `getMetadataForOpenRosa`, so the two values agree. Previously they used different hash inputs and would never match, which broke the manifest hash tests that compared them.

Tests all still pass. New tests added for dataset-backed form attachment hash behavior with ownerOnly. This is built on top of and used by: https://github.com/getodk/central-backend/pull/1843

Fetching dataset metadata in a separate query (rather than via JOINs inside the form attachments query) makes each query simpler and more focused. The tradeoff is an extra round-trip, but these are manifest/manifest-item requests that aren't on a hot path.

No intentional behavior changes — this is a pure refactor. Regression risk is around etag/hash values for dataset-backed attachments changing; the new test coverage addresses this.

No.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -728,7 +728,7 @@ const _fetchActivityStats = (all, datasetIds, filterActorId = null) => {
         SELECT
           datasets.id,
           GREATEST(MAX(entities."createdAt"), MAX(entities."updatedAt")) AS "lastEntityUpdate",
-          COUNT(1) AS entities
+          COUNT(*) AS entities
         FROM entities
         JOIN datasets ON datasets.id = entities."datasetId"
         WHERE
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -734,7 +734,7 @@ const getMetadataForOpenRosa = (datasetIds, filterActorId = null) => ({ all }) =
       ${filterActorId == null ? sql`` : sql`
         WHERE
           (NOT datasets."ownerOnly") OR
-          audits.action IN ('dataset.create', 'dataset.update')
+          audits.action = 'dataset.update'
       `}
       GROUP BY datasets.id
     ) AS "openRosaMetadata" ON "openRosaMetadata".id = datasets.id
@@ -758,12 +758,16 @@ const getMetadataForOpenRosa = (datasetIds, filterActorId = null) => ({ all }) =
   `)
     .then(rows => new Map(rows.map(unjoin).map(ds => {
       const metadata = ds.aux.openRosaMetadata;
-      const lastUpdate = metadata.lastEntityUpdate != null && metadata.lastEntityUpdate > metadata.lastAudit
-        ? metadata.lastEntityUpdate
-        : metadata.lastAudit;
-      const lastIso = lastUpdate.toISOString();
+      const latestEntityCreatedOrUpdated = metadata.lastEntityUpdate?.toISOString() ?? null;
+      const latestAuditEntry = metadata.lastAudit?.toISOString() ?? null;
+      const hashInput = JSON.stringify({
+        owner: metadata.filteredByAuth ? filterActorId : undefined,
+        entities: metadata.filteredByAuth ? (metadata.entities ?? 0) : undefined,
+        latestEntityCreatedOrUpdated,
+        latestAuditEntry,
+      });
       return [ds.id, {
-        openRosaHash: md5sum(metadata.filteredByAuth ? `${metadata.entities ?? 0},${lastIso}` : lastIso),
+        openRosaHash: md5sum(hashInput),
         datasetType: ds.approvalRequired ? 'approvalEntityList' : 'entityList',
         filteredByAuth: metadata.filteredByAuth
       }];
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -9,6 +9,7 @@
 
 const { sql } = require('slonik');
 const { extender, insert, QueryOptions, sqlEquals, unjoiner, updater, markDeleted } = require('../../util/db');
+const { Frame, into } = require('../frame');
 const { Dataset, Form, Audit } = require('../frames');
 const { validatePropertyName } = require('../../data/dataset');
 const { difference, isEmpty, isNil, either, reduceBy, groupBy, uniqWith, equals, map, pipe, values, any, prop, toLower, sortBy } = require('ramda');
@@ -19,6 +20,7 @@ const { construct } = require('../../util/util');
 const { sanitizeOdataIdentifier } = require('../../util/util');
 const { isTrue } = require('../../util/http');
 const { PURGE_DAY_RANGE } = require('../../util/constants');
+const { md5sum } = require('../../util/crypto');
 
 
 ////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


lib/model/query/datasets.js
```diff
@@ -711,6 +713,67 @@ const getLastUpdateTimestamp = (datasetId) => ({ maybeOne }) =>
     .then((t) => (t ? t.loggedAt : null));
 
 
+// Returns dataset metadata needed to compute OpenRosa hashes for dataset-backed form attachments.
+// Takes a list of dataset IDs and an optional filterActorId (for ownerOnly datasets).
+// Returns a Map of datasetId -> metadata object with fields:
+//   approvalRequired, filteredByAuth, lastAudit, and (when filterActorId given) lastEntityUpdate, entities
+const getMetadataForOpenRosa = (datasetIds, filterActorId = null) => ({ all }) => {
```
lognaturel: This is not just `forOpenRosa`, right? Can it be `getHashMetadata`?

lib/model/query/datasets.js
```diff
@@ -711,6 +713,67 @@ const getLastUpdateTimestamp = (datasetId) => ({ maybeOne }) =>
...
+    SELECT ${unjoin.fields} FROM datasets
+    LEFT OUTER JOIN (
+      SELECT datasets.id, MAX("loggedAt") AS "lastAudit",
+        ${filterActorId == null ? sql`FALSE` : sql`datasets."ownerOnly"`} AS "filteredByAuth"
+      FROM audits
+      JOIN datasets ON audits."acteeId" = datasets."acteeId"
+      ${filterActorId == null ? sql`` : sql`
+        WHERE
+          (NOT datasets."ownerOnly") OR
+          audits.action = 'dataset.update'
```
lognaturel, line 725: Maybe you could cherry pick commits from #1814 and we close that one since these changes are so intimately connected? @matthew-white that would save you from addressing those merge conflicts.

lib/model/query/datasets.js
```diff
@@ -711,6 +713,67 @@ const getLastUpdateTimestamp = (datasetId) => ({ maybeOne }) =>
...
+      `}
+      GROUP BY datasets.id
+    ) AS "openRosaMetadata" ON "openRosaMetadata".id = datasets.id
+    ${filterActorId == null ? sql`` : sql`
+      LEFT OUTER JOIN (
+        SELECT
+          datasets.id,
+          GREATEST(MAX(entities."createdAt"), MAX(entities."updatedAt")) AS "lastEntityUpdate",
+          COUNT(1) AS entities
```
lognaturel: Why not `COUNT(*)`?

lib/resources/datasets.js
```diff
@@ -152,7 +152,10 @@ module.exports = (service, endpoint) => {
 
     if (options.filter) return csv();
 
-    const serverEtag = md5sum(lastModifiedTime?.toISOString() ?? '1970-01-01');
+    const serverEtag = md5sum(JSON.stringify({
+      latestEntityCreatedOrUpdated: null,
```
lognaturel: Why have this key at all if it's always null?

lib/resources/datasets.js
```diff
@@ -152,7 +152,10 @@ module.exports = (service, endpoint) => {
...
+      latestAuditEntry: lastModifiedTime?.toISOString() ?? null,
```
lognaturel: Why change `'1970-01-01` to `null`?

lognaturel: So `acteeID` on an audit event for all Entity create, update, delete events is the ID of the **Entity List** not the Entity? Unless I'm misunderstanding, that must be the case or this etag needs to be rethought. Is there any way to share code between this and the manifest hash?

lognaturel (commented):
A few comments/suggestions for now!

lib/resources/datasets.js
ktuite: Yes, the `acteeID` for entity events is the Dataset/Entity List actee ID. We don't make actee records for smaller, numerous things like entities and submissions.

lognaturel: Let me make sure I get this right! If EntityA is in EntityListA, then modifying EntityA leads to an audit log entry with actee ID EntityListA? Or no audit log entry?

ktuite: It'll log it as `entity.update.version` with the dataset actee ID. 

Also, the details of the audit entry contain the entity UUID, id, def id, and dataset name: https://github.com/getodk/central-backend/blob/master/lib/model/query/entities.js#L210

lib/model/query/datasets.js
ktuite: The the PR has merge conflicts with master (just based on the test file) so merge conflicts may be unavoidable. But they are small, I will deal with them when it comes up! 

@matthew-white if you have followup ideas here you should just put them in a new PR because I'm about to shift which file this query even lives in! 

```
<<<<<<< entity-property-delete-hash
      it('changes hash after a dataset property is deleted', testServiceFullTrx(async (service) => {
        const [asAlice, asChelsea] = await service.login(['alice', 'chelsea']);
        await createData(asAlice);
        await assignToProject(asAlice, asChelsea, 'formfill');

        // Add an entity property.
        await asAlice.post('/v1/projects/1/datasets/people/properties')
          .send({ name: 'foo' })
          .expect(200);
        const originalHash = await getHash(asChelsea);

        // Delete the property.
        await asAlice.delete('/v1/projects/1/datasets/people/properties/foo')
          .expect(200);
        (await getHash(asChelsea)).should.not.equal(originalHash);
      }));

      it('computes hash based on timestamps and the entity count', testServiceFullTrx(async (service, container) => {
=======
      it('computes hash based on timestamps and the entity count @slow', testServiceFullTrx(async (service, container) => {
>>>>>>> master
        const [asAlice, asChelsea] = await service.login(['alice', 'chelsea']);
        await createData(asAlice);
 ```

File: lib/model/query/datasets.js (hunk 1 above)
Comment (ktuite) [reply lib/model/query/datasets.js]: Good point. It was just used by the openrosa path, but I just restructured the code so that the base entity CSV also uses this same code (now called `Datasets.computeHashes([dsIds])`) and `Datasets.getLastUpdateTimestamp` doesn't exist because it shares code.

How the code is shared:
```
_fetchActivityStats(all, datasetIds, filterActorId)   ← SQL, returns Map<id, rawStats>
    └── used by:
        ├── getDatasetActivity(datasetId, filterActorId)   ← single dataset, returns rawStats object
        │       └── used in getMetadata for the lastUpdate field
        └── computeHashes(datasetIds, filterActorId)       ← batch, pipes each rawStats through:
                └── _computeHashResult(stats, filterActorId)   ← pure fn, rawStats → { openRosaHash, datasetType, filteredByAuth }
```

File: lib/model/query/datasets.js (hunk 3 above)
Comment (ktuite) [reply lib/model/query/datasets.js]: It was `COUNT(1)` before but it could be `COUNT(*)`

Comment (lognaturel) [reply lib/model/query/datasets.js]: Looks like they're identical in modern postgres and `COUNT(*)` is generally recommended!

File: lib/resources/datasets.js (hunk 4 above)
Comment (ktuite) [reply lib/resources/datasets.js]: These changes were just to make this server etag hash match the hash coming out of the form attachment path, but I've changed up the code so this serverETag uses that hash directly. 

We could talk about what should/shoudn't be included over in `lib/model/query/datasets.js`: 
```js
// Private: computes the OpenRosa hash result from a single dataset's activity stats.
const _computeHashResult = (stats, filterActorId) => {
  const latestEntityCreatedOrUpdated = stats.lastEntityUpdate?.toISOString() ?? null;
  const latestAuditEntry = stats.lastAudit?.toISOString() ?? null;
  const hashInput = JSON.stringify({
    owner: stats.filteredByAuth ? filterActorId : undefined,
    entities: stats.filteredByAuth ? (stats.entities ?? 0) : undefined,
    latestEntityCreatedOrUpdated,
    latestAuditEntry,
  });
...
```

lib/model/query/datasets.js
lognaturel: That's exciting!

If the user is a project manager or project viewer, then that should override the "owner only" setting any time the CSV is requested, right? But "owner only" applies to Data Collectors? I imagine all this is handled before we get here but we should make sure there are tests.

ktuite: Yes, and the way that's handled for managers vs. data collectors is with this:

`lib/resources/forms.js`:
```js
// Check auth verbs on project to determine if user might use auth access-based entity filter
const filteredAccessActorId = async (auth, project) => ((await auth.can('entity.list', project))
  ? null
  : auth.actor.get().id);
```

Basically, we used to not do ownerOnly filtering if the user had access to all the entities anyway (through the `entity.list` verb). We still do that, but now the concept is "is there going to be some kind of dataset filtering applied for this user?" and if they are a manager or above, the answer is no, no filtering, give 'em all the entities always.

Then we just pass the filter actor ID around. If it's not set, we don't filter, and if it is set, we look up _on the dataset_ if there's a filter and if it's the ownerOnly for the user properties filter.

ktuite: the filter actor ID is used for 
a) the calculations about last dataset update time, or last action on an entity in the filtered set, and 
b) actually filtering the entity CSV when sent through the form attachment endpoint

lib/model/query/datasets.js
```diff
@@ -698,18 +700,87 @@ const getUnprocessedSubmissions = (datasetId) => ({ all }) =>
...
-  maybeOne(sql`
-  SELECT "loggedAt" FROM audits
-  JOIN datasets ON audits."acteeId" = datasets."acteeId"
-  WHERE datasets."id"=${datasetId}
-  ORDER BY audits."loggedAt" DESC LIMIT 1`)
-    .then((t) => t.orNull())
-    .then((t) => (t ? t.loggedAt : null));
+// Private: fetches audit and entity activity data for a list of datasets.
+// Returns a Map of datasetId -> { approvalRequired, filteredByAuth, lastAudit, lastEntityUpdate, entities }
+const _fetchActivityStats = (all, datasetIds, filterActorId = null) => {
```
lognaturel, line 708: I like the comments you've added! I would find it helpful to see something like "filterActorId is null if dataset filter rules don't apply to the current user's type (e.g. manager)"

lognaturel: Bah, maybe it's not necessary, there's now a good comment in form.js I see so it's not so hard to follow.

lognaturel (approved):
Agreed this is a lot easier to follow! I think we should get it in asap so you can keep building on top.

We probably don't generally need to have changesets for refactors but this one is fairly significant and could have user-facing consequences, particularly around performance. How about something short about reimplementing the manifest hash for Entity Lists?

lognaturel:
Saw you just requested a review from @sadiqkhoja, fine to wait for that! I also think he could review this more carefully once property-based filtering is in.

ktuite:
I just met with Sadiq and briefly described the gist of this PR, but I'm happy to take your ✅ and merge this, @lognaturel!

> Saw you just requested a review from @sadiqkhoja, fine to wait for that! I also think he could review this more carefully once property-based filtering is in.

lognaturel:
Unless @sadiqkhoja sees a possibility for anything show stopping, I think we should merge. With a changeset!

lognaturel:
Not with a changeset! We've only added that to the frontend repo, my apologies.

ktuite:
Adding a more detailed comment but not adding a changeset here. Then I'll merge!

PR #1849, opened 2026-06-10
Migration and API to add dataset access filters

### LINKED ISSUE #1878: API to PATCH dataset and configure dataset access filters alongside existing "owner only" access mode
Allow property-based filters to be configured on datasets so that when a user (App User or Public Link) requests a dataset through a form attachment, it will be a subset of the entire entity list. This is a followup to the `ownerOnly` approach of filtering a dataset, where user only gets access to the entities they created. 

API should allow setting ownerOnly or the new accessFilter, but have them be mutually exclusive. The API should also return information about the access filter in the dataset metadata.

## Depends on
- https://github.com/getodk/central/issues/1876 (design)

## Enables
- https://github.com/getodk/central/issues/1877 (apply filter rules)

ktuite:
Closes https://github.com/getodk/central/issues/1878 and does most of https://github.com/getodk/central/issues/1877
(Applying the access filter to come in a follow-up PR.)

## Summary: 

Adds an `accessFilter` field to dataset metadata responses and the dataset settings PATCH endpoint.

**GET response** — `accessFilter` added when a filter is set:
```json
{ "ownerOnly": true, "accessFilter": { "type": "ownerOnly" } }

{ "ownerOnly": false, "accessFilter": { "type": "property", "rules": [
  { "datasetProperty": "region", "actorProperty": "zone" }
]}}

{ "ownerOnly": false }  // no accessFilter field when no filter
```

**PATCH request body:**
```json
{ "ownerOnly": true }                    // legacy, still works
{ "accessFilter": null }                 // clears all filters
{ "accessFilter": { "type": "ownerOnly" } }
{ "accessFilter": { "type": "property", "rules": [
  { "datasetProperty": "region", "actorProperty": "zone" }
]}}
```

- If both `ownerOnly` and `accessFilter` are sent, `accessFilter` wins silently
- Setting `ownerOnly: true` clears any existing property rules
- `ownerOnly` boolean in DB; property rules in separate `dataset_access_filters` table

`rules` must be a non-empty array; each rule requires `datasetProperty` (a published dataset property name) and `actorProperty` (an actor property name in the same project). Replaces all existing rules atomically.

The legacy `ownerOnly` boolean field continues to work. If both `ownerOnly` and `accessFilter` are sent in the same request, `accessFilter` takes precedence.

-----

### Database schema:

**New table: `dataset_access_filters`**

Stores one row per filter rule, mapping a published dataset property to an actor property.

| Column | Type | Notes |
|---|---|---|
| `datasetId` | integer | FK to dataset (redundant, required for composite FK) |
| `datasetPropertyId` | integer PK | FK to `ds_properties` |
| `actorPropertyId` | integer | FK to `actor_properties`, cascades on delete |

A composite FK on `(datasetId, datasetPropertyId)` → `ds_properties(datasetId, id)` prevents `datasetId` from being inconsistent with `datasetPropertyId`. A unique index on `(datasetPropertyId, actorPropertyId)` ensures each actor property is used at most once per dataset property.

**New view: `actor_dataset_filter`**

Joins `dataset_access_filters` with `ds_properties` and `actor_property_values` to produce one row per `(datasetId, actorId)` pair, containing a `jsonb` filter object (keyed by dataset property name, valued by the actor's value for the mapped actor property).

Also contains an MD5 hash of that object that can be used as an ETag component in incremental entity list downloads. See issue: https://github.com/getodk/central/issues/1880

Integration tests cover the new `accessFilter` field on `PATCH /projects/:id/datasets/:name`, including setting property-based rules, switching to `ownerOnly`, clearing filters, and error cases for invalid input (bad type, empty rules array, missing rule fields). 

The `accessFilter` field is a unified, structured alternative to the bare `ownerOnly` boolean, making it easier to add new filter types in the future without introducing more top-level fields. Adding a separate `propertyFilter` field alongside `ownerOnly` was considered but rejected — it would leave two independent filter fields that could contradict each other.

The legacy `ownerOnly` field is preserved for backward compatibility. `accessFilter` takes precedence if both are sent in the same request.

**Intentional changes:**
- Dataset metadata responses now include an `accessFilter` field when a filter is active. This is additive and should not affect existing clients.
- `PATCH /projects/:id/datasets/:name` accepts a new `accessFilter` body field.
- Setting `accessFilter: { type: 'ownerOnly' }` clears any existing property rules, and setting `accessFilter: { type: 'property', ... }` clears `ownerOnly` if it was set. The two filter types cannot coexist.

**Regression risks:**
- If a client sends both `ownerOnly` and `accessFilter` in the same PATCH request, `accessFilter` silently wins. This could be surprising if a client is unaware of the new field.
- `Datasets.update` is now only called when there are column-level changes to make. If the request only touches `accessFilter` (no `approvalRequired` or `ownerOnly` changes), the dataset row is returned as-is without a DB write and won't make an audit log entry.  This needs to be fixed in a follow up issue: https://github.com/getodk/central/issues/1882 

Yes. `docs/api.yaml` should be updated to document:

1. The `accessFilter` field in the dataset metadata response schema (optional, present when a filter is set).
2. The `accessFilter` field in the `PATCH /projects/{projectId}/datasets/{name}` request body schema.

### DIFF
--- a/lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
+++ b/lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
@@ -1,0 +1,41 @@
+-- Maps a dataset property to an actor property, defining a filter rule:
+-- entities are visible to an actor only if their value for "datasetPropertyId" matches the actor's value for "actorPropertyId".
+CREATE TABLE dataset_access_filters (
+    "datasetId" integer NOT NULL,  -- Redundant with "datasetPropertyId" but needed for the composite FK below to prevent insertion anomalies.
+    "datasetPropertyId" integer NOT NULL PRIMARY KEY,
+    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE
+);
+-- The unique index on ds_properties enables the composite FK, which ensures "datasetId" is always
+-- consistent with "datasetPropertyId" (preventing insertion anomalies).
+CREATE UNIQUE INDEX "dataset_access_filters__unique_for_composite_fk_referent" ON ds_properties USING btree ("datasetId", id);
+ALTER TABLE dataset_access_filters
+    ADD CONSTRAINT "dataset_access_filters__composite_fk" FOREIGN KEY ("datasetId", "datasetPropertyId") REFERENCES ds_properties ("datasetId", id) ON DELETE CASCADE;
+-- Ensures each actor property is used at most once per dataset.
+CREATE UNIQUE INDEX "dataset_access_filters__spend_actorprop_once_per_dsprop" ON dataset_access_filters ("datasetPropertyId", "actorPropertyId");
+-- Index to support FK lookup on actorPropertyId (CASCADE deletes from actor_properties).
+CREATE INDEX ON dataset_access_filters ("actorPropertyId");
+-- Index to support FK lookup on (datasetId, datasetPropertyId) (CASCADE deletes from ds_properties).
+CREATE INDEX ON dataset_access_filters ("datasetId", "datasetPropertyId");
+
+CREATE VIEW actor_dataset_filter AS (
+    WITH filter_as_json AS (
+        SELECT
+            pfilter."datasetId",
+            aprop."actorId",
+            jsonb_object_agg (dsprops."name", aprop."value") AS thefilter
+        FROM
+            dataset_access_filters pfilter
+            INNER JOIN ds_properties dsprops ON (pfilter."datasetPropertyId" = dsprops.id)
+            INNER JOIN actor_property_values aprop USING ("actorPropertyId")
+        GROUP BY
+            (pfilter."datasetId", aprop."actorId")
+    )
+    SELECT
+        *,
+        md5(thefilter::text) AS filterhash  -- Used to make the filter part of an HTTP resource for incremental entity list download.
+    FROM
+        filter_as_json
+);
+
+-- Commented out: add back w/ info about taking a long time
+-- CREATE INDEX "idx_entity_defs_data" on entity_defs using gin (data);
--- a/lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
+++ b/lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
@@ -2,16 +2,15 @@
 -- entities are visible to an actor only if their value for "datasetPropertyId" matches the actor's value for "actorPropertyId".
 CREATE TABLE dataset_access_filters (
     "datasetId" integer NOT NULL,  -- Redundant with "datasetPropertyId" but needed for the composite FK below to prevent insertion anomalies.
-    "datasetPropertyId" integer NOT NULL PRIMARY KEY,
-    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE
+    "datasetPropertyId" integer NOT NULL,
+    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE,
+    PRIMARY KEY ("datasetPropertyId", "actorPropertyId")
 );
 -- The unique index on ds_properties enables the composite FK, which ensures "datasetId" is always
 -- consistent with "datasetPropertyId" (preventing insertion anomalies).
 CREATE UNIQUE INDEX "dataset_access_filters__unique_for_composite_fk_referent" ON ds_properties USING btree ("datasetId", id);
 ALTER TABLE dataset_access_filters
     ADD CONSTRAINT "dataset_access_filters__composite_fk" FOREIGN KEY ("datasetId", "datasetPropertyId") REFERENCES ds_properties ("datasetId", id) ON DELETE CASCADE;
--- Ensures each actor property is used at most once per dataset property.
-CREATE UNIQUE INDEX "dataset_access_filters__spend_actorprop_once_per_dsprop" ON dataset_access_filters ("datasetPropertyId", "actorPropertyId");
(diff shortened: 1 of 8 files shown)


ktuite:
There is definitely room for improvement with the API schema design and probably the code itself, but I wanted to get something concrete out there for us to discuss! Choosing a schema that we can easily extend or replace in the future when we have different rules will be important.

lib/resources/datasets.js
```diff
@@ -33,24 +33,74 @@ module.exports = (service, endpoint) => {
...
-    const updatedDataset = await Datasets.update(dataset, newDataset, autoConvert);
+
+    // Collect all updates into one object to call Datasets.update once
+    const updates = {};
+    if (newDataset.approvalRequired !== undefined)
+      updates.approvalRequired = newDataset.approvalRequired;
+
+    // Handle access filter changes.
+    // accessFilter in the body takes precedence over the legacy ownerOnly field.
+    if (Object.hasOwn(body, 'accessFilter')) {
```
sadiqkhoja, line 64: I am wondering if we should return 400 Bad Request if client sends in both accessFilter and ownerOnly

lib/resources/datasets.js
```diff
@@ -33,24 +33,74 @@ module.exports = (service, endpoint) => {
...
+        // Switch to ownerOnly mode; clear any property rules first
+        await Datasets.deleteAllAccessFilters(dataset.id);
+        updates.ownerOnly = true;
+
+      } else if (accessFilter.type === 'property') {
+        const { rules } = accessFilter;
+        if (!Array.isArray(rules) || rules.length === 0)
+          throw Problem.user.unexpectedValue({ field: 'accessFilter.rules', value: rules, reason: 'rules must be a non-empty array.' });
+        for (const rule of rules) {
+          if (!rule.datasetProperty || !rule.actorProperty)
```
sadiqkhoja, line 82: should we confirm the existence of properties here or in the SQL query? and throw error if either property doesn't exit

test/integration/api/datasets.js
```diff
@@ -6282,6 +6283,195 @@ describe('datasets and entities', () => {
...
+
+          await asAlice.patch('/v1/projects/1/datasets/trees')
+            .send({ approvalRequired: false })
+            .expect(200)
+            .then(({ body }) => {
+              body.accessFilter.should.eql({ type: 'property', rules: [{ datasetProperty: 'region', actorProperty: 'region' }] });
+            });
+        }));
+
+        it('should reject empty rules array', testService(async (service) => {
```
sadiqkhoja, line 6474: how about a test that send duplicate rules like 
```
{ "rules": [
    { "datasetProperty": "region", "actorProperty": "zone1" },
    { "datasetProperty": "region", "actorProperty": "zone1" }
  ]}
```

lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
```diff
@@ -0,0 +1,34 @@
...
+    FROM
+        dataset_access_filters pfilter
+        INNER JOIN ds_properties dsprops ON (pfilter."datasetPropertyId" = dsprops.id)
+        INNER JOIN actor_property_values aprop USING ("actorPropertyId")
+    GROUP BY
+        (pfilter."datasetId", aprop."actorId")
+);
+
+-- Commented out: add back w/ info about taking a long time
+-- CREATE INDEX "idx_entity_defs_data" on entity_defs using gin (data);
```
sadiqkhoja: dead code

lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
```diff
@@ -0,0 +1,34 @@
...
+    "datasetPropertyId" integer NOT NULL PRIMARY KEY,
+    "actorPropertyId" integer NOT NULL REFERENCES actor_properties (id) ON DELETE CASCADE
+);
+-- The unique index on ds_properties enables the composite FK, which ensures "datasetId" is always
+-- consistent with "datasetPropertyId" (preventing insertion anomalies).
+CREATE UNIQUE INDEX "dataset_access_filters__unique_for_composite_fk_referent" ON ds_properties USING btree ("datasetId", id);
+ALTER TABLE dataset_access_filters
+    ADD CONSTRAINT "dataset_access_filters__composite_fk" FOREIGN KEY ("datasetId", "datasetPropertyId") REFERENCES ds_properties ("datasetId", id) ON DELETE CASCADE;
+-- Ensures each actor property is used at most once per dataset property.
+CREATE UNIQUE INDEX "dataset_access_filters__spend_actorprop_once_per_dsprop" ON dataset_access_filters ("datasetPropertyId", "actorPropertyId");
```
sadiqkhoja: I don't think this unique constraint is necessary because `datasetPropertyId` is already PK

sadiqkhoja (approved):
Overall this is looking great, added few comments. They can be done in the follow-up PR as well if you agree with the suggestions.

lib/model/migrations/20260602-01-add-dataset-filters-01.up.sql
ktuite: I've updated the migration to remove some of the unneeded indexes and commented out code.

I also changed the primary key to `PRIMARY KEY ("datasetPropertyId", "actorPropertyId")` to support some of the multi-rule scenarios i listed [here](https://github.com/getodk/central/issues/1876#issuecomment-4683360141). I have some tests of those scenarios for the next PR that applies the filters.

lib/resources/datasets.js
ktuite: That sounds like a good idea to me, I noted it here: https://github.com/getodk/central/issues/1972

PR #1850, opened 2026-06-10
Apply dataset access filter to streaming entity export form attachments

### LINKED ISSUE #1877: Apply dataset access filter rules to form attachment CSV downloads
When an App User or Public Access Link downloads an entity list CSV attachment, filter entities based on the configured rules and the user's assigned property values

Filtering is transparent — no client changes required

## Out of scope
- Applying filters to the entity REST and OData endpoints (separate issue)
- ETag invalidation (separate issue)
- Indexing for query efficiency (separate issue)
- Audit logging (separate issue)
- Reconciling with the existing "owner only" access mode (separate issue)

## Depends on
- https://github.com/getodk/central/issues/1868 (user properties backend)
- https://github.com/getodk/central/issues/1876 (design)
- https://github.com/getodk/central/issues/1878

ktuite:
Closes https://github.com/getodk/central/issues/1877

## Summary

This PR applies dataset access filters to the streaming entity export used for form attachments.

Before this change, this query path already respected `ownerOnly` datasets, but it did not apply dataset property-based access filters.

## Where the filtering happens

The filtering is added in `lib/model/query/entities.js`, in the SQL predicate used when building the entity query for the streaming export path.

It is added alongside the existing actor-scoped dataset checks, so the query now enforces both:

- `ownerOnly` restrictions
- dataset access filters from `dataset_access_filters`

## When the filtering applies

This logic applies when the query is executed with an `actorId`, meaning the export is being evaluated in the context of a specific actor whose dataset access should be restricted.

This is not applied when `actorId` is not set, including the case where the requester is an admin or manager who has access to all entities in a dataset via the `entity.list` project verb.

## How the query works

The new condition does the following:

- if the dataset has no access filters, allow the entity through
- otherwise, require at least one matching access-filter rule for the current actor

More specifically, the query:

- checks whether any rows exist in `dataset_access_filters` for the entity's dataset
- if no rows exist, the entity is allowed
- if filters do exist, it joins:
  - `dataset_access_filters`
  - `ds_properties`
  - `actor_property_values`
- then requires:
  - a filter configured for this dataset
  - an `actor_property_values` row for the current `actorId`
  - a match between the entity property value in `entity_defs.data` and the actor property value

In effect, an entity is returned only if:

- the dataset has no access filters, or
- the actor has at least one configured property-value match for that dataset and entity

If multiple access-filter rules are configured for the dataset, this query treats them as ORed together: an entity is included as soon as one configured rule matches the actor's property value.

The main regression risk is query cost, since this adds additional existence and join checks during entity selection, although it should still be preferable to exporting a larger unfiltered result set.

No. The API behavior is being brought into line with existing dataset access rules rather than introducing a new API surface.

- [ ] run `make test` and confirmed all checks still pass, or witnessed GitHub completing all checks with success

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -931,6 +931,18 @@ ${actorId != null ? sql`
     -- ownerOnly: if the dataset is ownerOnly, restrict to entities created by this actor
     NOT EXISTS (SELECT 1 FROM datasets WHERE datasets.id = entities."datasetId" AND datasets."ownerOnly")
     OR entities."creatorId" = ${actorId}
+  )
+  AND (
+    -- property filters: if the dataset has access filters, restrict to matching entities
+    NOT EXISTS (SELECT 1 FROM dataset_access_filters WHERE "datasetId" = entities."datasetId")
+    OR EXISTS (
+      SELECT 1 FROM dataset_access_filters daf
+      JOIN ds_properties dp ON dp.id = daf."datasetPropertyId"
+      JOIN actor_property_values apv ON apv."actorPropertyId" = daf."actorPropertyId"
+      WHERE daf."datasetId" = entities."datasetId"
+        AND apv."actorId" = ${actorId}
+        AND entity_defs.data ->> dp."name" = apv."value"
+    )
   )` : sql``}
 ${options.orderby ? sql`
   ${odataOrderBy(options.orderby, odataToColumnMap, 'entities.id')}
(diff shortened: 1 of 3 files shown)


sadiqkhoja (approved):
Looks good to me!

Only thing that is bothering me is that whether we should add this filter condition and owner only filter condition to `getEntitiesState` (for integrity URL) 🤔 

If filter condition is modified and user no longer have access to some of the Entities, should we inform Collect that those entities are "delete" ?

ktuite:
> Only thing that is bothering me is that whether we should add this filter condition and owner only filter condition to `getEntitiesState` (for integrity URL) 🤔
> 
> If filter condition is modified and user no longer have access to some of the Entities, should we inform Collect that those entities are "delete" ?

Really good point! Tracked in https://github.com/getodk/central/issues/1971

PR #1851, opened 2026-06-10
Compute etag for user-filtered dataset

### LINKED ISSUE #1880: Invalidate ETag of segmented entity list when filter rules or user property values change
Ensure the ETag on the form attachment CSV download changes when dataset access filter rules are modified or when an App User / Public Access Link's property values change, so that Collect knows to re-download the updated filtered list.

## Scope
- Update ETag computation to incorporate filter rules and relevant user property values
- Trigger re-download on Collect when rules or values change
- Granularity decision per the design issue (e.g. per-user ETag vs. global)

## Depends on
- https://github.com/getodk/central/issues/1876 (design — ETag granularity)
- https://github.com/getodk/central/issues/1877 (filter rules CRUD + apply)

ktuite:
Closes https://github.com/getodk/central/issues/1880

A number of tests for `describe('Etag / open rosa hash with dataset user property filters')`:

1. **hash changes for alice but not app user when entity outside segment is added** — baseline: before/after filter applied, adding an out-of-segment entity changes Alice's hash but not the filtered app user's
2. **hash changes for app user when filter rules are first added** — filter rule itself (no entity change) causes hash change
3. **hash changes for app user when filter rules are removed** (`accessFilter: null`) — removing the rule causes hash change
4. **hash changes for app user when their actor property value changes** — north → south triggers hash change
5. **hash changes for app user when their actor property value is assigned for the first time** — null → north triggers hash change (fail-closed scenario)
6. **two app users with different property values get different hashes** — north user ≠ south user hash
7. **when user has no property set, hash changes when filter is added but stays stable when entities are added outside the empty segment** — fail-closed: filter application changes hash; adding entities outside the (null) segment does not
8. **hash changes when an entity in the app user segment is updated** — in-segment entity update triggers hash change
9. **hash does not change when an entity outside the app user segment is updated** — out-of-segment entity update leaves hash unchanged
10. **hash changes when the filter rule mapping is changed** — swapping `region→region` to `district→district` changes hash
11. **hash changes for filtered app user when a new dataset property is added** — adding a new dataset property (unrelated to the filter rule) changes hash via audit log entry

This small change made possible by https://github.com/getodk/central-backend/pull/1846

Should not change the hash for a user if they receive a segmented entity list and nothing in their segment has changed (and no big operations to the dataset have happened).

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -718,27 +718,29 @@ const getUnprocessedSubmissions = (datasetId) => ({ all }) =>
     .then(map(construct(Audit)));
 
 // Private: fetches audit and entity activity data for a list of datasets.
-// Returns a Map of datasetId -> { approvalRequired, filteredByAuth, lastAudit, lastEntityUpdate, entities }
+// Returns a Map of datasetId -> { approvalRequired, filteredByAuth, lastAudit, lastEntityUpdate, entities, filterRules }
 // filterActorId will be null if dataset filter rules don't apply to the current user's type (e.g. manager),
 //   otherwise it will be used to compute stats on the subset of entities filtered for that user (if that dataset
 //   has an active filter at all).
+// filterRules is a JSON array of { datasetProperty, actorProperty, value } objects for property-filtered datasets,
+//   or null if no property filter rules exist. The value is the actor's current value for that actor property.
 const _fetchActivityStats = (all, datasetIds, filterActorId = null) => {
   if (datasetIds.length === 0) return Promise.resolve(new Map());
 
   const fields = ['filteredByAuth', 'lastAudit'];
-  if (filterActorId != null) fields.push('lastEntityUpdate', 'entities');
+  if (filterActorId != null) fields.push('lastEntityUpdate', 'entities', 'filterRules');
   const unjoin = unjoiner(Dataset, Frame.define(into('openRosaMetadata'), ...fields));
 
   return all(sql`
     SELECT ${unjoin.fields} FROM datasets
     LEFT OUTER JOIN (
       SELECT datasets.id, MAX("loggedAt") AS "lastAudit",
-        ${filterActorId == null ? sql`FALSE` : sql`datasets."ownerOnly"`} AS "filteredByAuth"
+        ${filterActorId == null ? sql`FALSE` : sql`("ownerOnly" OR EXISTS (SELECT 1 FROM dataset_access_filters WHERE "datasetId" = datasets.id))`} AS "filteredByAuth"
       FROM audits
       JOIN datasets ON audits."acteeId" = datasets."acteeId"
       ${filterActorId == null ? sql`` : sql`
         WHERE
-          (NOT datasets."ownerOnly") OR
+          (NOT datasets."ownerOnly" AND NOT EXISTS (SELECT 1 FROM dataset_access_filters WHERE "datasetId" = datasets.id)) OR
           audits.action = 'dataset.update'
       `}
       GROUP BY datasets.id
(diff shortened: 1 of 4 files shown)


lib/model/query/datasets.js
```diff
@@ -754,10 +756,40 @@ const _fetchActivityStats = (all, datasetIds, filterActorId = null) => {
...
       ) AS entity_metadata ON entity_metadata.id = datasets.id
+      LEFT OUTER JOIN (
+        SELECT
+          daf."datasetId",
+          jsonb_agg(
+            jsonb_build_object(
+              'datasetProperty', dp.name,
+              'actorProperty', ap.name,
+              'value', apv.value
+            ) ORDER BY dp.name
```
lognaturel: I don't believe this order is fully deterministic. It would be possible to have [region, region, region1] and [region, region, region2] and without that full order rule I think they could come out in different orders and lead to over-invalidation. Can we order by `dp.name, ap.name, apv.value` to be sure?

lib/model/query/datasets.js
```diff
@@ -754,10 +756,40 @@ const _fetchActivityStats = (all, datasetIds, filterActorId = null) => {
...
+              JOIN actor_property_values apv ON apv."actorPropertyId" = daf."actorPropertyId"
+              JOIN entity_defs ef ON ef."entityId" = entities.id AND ef.current = true
+              WHERE daf."datasetId" = entities."datasetId"
+                AND apv."actorId" = ${filterActorId}
+                AND ef.data ->> dp."name" = apv."value"
+            )
           )
         GROUP BY datasets.id
       ) AS entity_metadata ON entity_metadata.id = datasets.id
+      LEFT OUTER JOIN (
```
lognaturel, line 773: What are the main drivers of making this part of the single big query? Perf? One DB round trip? It feels like it would be much more readable as a separate query.

lib/model/query/datasets.js
```diff
@@ -718,27 +718,29 @@ const getUnprocessedSubmissions = (datasetId) => ({ all }) =>
...
     LEFT OUTER JOIN (
       SELECT datasets.id, MAX("loggedAt") AS "lastAudit",
-        ${filterActorId == null ? sql`FALSE` : sql`datasets."ownerOnly"`} AS "filteredByAuth"
+        ${filterActorId == null ? sql`FALSE` : sql`("ownerOnly" OR EXISTS (SELECT 1 FROM dataset_access_filters WHERE "datasetId" = datasets.id))`} AS "filteredByAuth"
       FROM audits
       JOIN datasets ON audits."acteeId" = datasets."acteeId"
       ${filterActorId == null ? sql`` : sql`
         WHERE
-          (NOT datasets."ownerOnly") OR
+          (NOT datasets."ownerOnly" AND NOT EXISTS (SELECT 1 FROM dataset_access_filters WHERE "datasetId" = datasets.id)) OR
```
lognaturel: As we've discussed, I'd love to see us drop using the audit log at all if the list is filtered. Ideally we could do something like get a representation of all the properties defined for the Entity List instead. That should be much faster, even if it's in a separate query.

That would address https://github.com/getodk/central-backend/pull/1814/changes.

Maybe a candidate for a follow-up PR then?

lib/model/query/datasets.js
```diff
@@ -754,10 +756,40 @@ const _fetchActivityStats = (all, datasetIds, filterActorId = null) => {
         WHERE
           entities."deletedAt" IS NULL
           AND (
+            -- ownerOnly: only consider entities created by this actor
             (datasets."ownerOnly" AND entities."creatorId" = ${filterActorId})
+            OR
+            -- property filters: only consider entities matching this actor's property values
+            EXISTS (
```
lognaturel: This is the part that makes me the most nervous because I think it could be quite a bit more resource-intensive to compute than the unfiltered count and last update. The alternative would be to use the unfiltered values which would over-invalidate with the idea that it's still much better than sending the whole list over the wire. 

I imagine your logic was that this isn't a hot path so it's ok for it to take time and the main thing we're trying to optimize for is end user bandwidth which this will save. Anything else?

lognaturel (commented):
The tests are so great! I think we need a few more with Entity deletion:
* Delete an entity, count and hash should change
* Delete one Entity that matches the filter and add another. The count should match but not the hash.

I agree with you that https://github.com/getodk/central/issues/1969 will still be needed because we don't test the case where there's no filtering.

lib/model/query/datasets.js
ktuite: i was trying to get the active filters + values for that user and the dataset(s) in question, but it would be easy (and i agree more readable) to pull it out and do it separately.

lib/model/query/datasets.js
ktuite: Ah, that was the key i was missing - not using the audit log at all if list is filtered.

lognaturel: Yeah! If we’re only looking at update events that’s just property changes (and filter changes which we capture elsewhere and approval setting change which we don’t care about).

lib/model/query/datasets.js
ktuite: It was how the ownerOnly stats were computed, to actually look at the timestamps of all the entities in the user's set. It does seems like a resource-intensive operation, though I guess the ownerOnly one had a nice index.

For the property-filtered case, it seems to need to actually apply the filter to get the same sort of information. 

I'm not sure what a better way would look like, so yes I'm just hoping that even though it's expensive, it's worth it to save sending the data down the line.

I guess the other way to look at it (maybe this was discussed and i just forgot) is that for the ds-prop filtered entity lists, we should do a less intensive operation for computing the hash, but accept that it may change even when the entities in the user's segment didn't change.

ktuite: I'll get back to all this tomorrow!

lognaturel: > the ownerOnly one had a nice index

Yeah, exactly. The property filter one is going to be very different computationally.

> we should do a less intensive operation for computing the hash, but accept that it may change even when the entities in the user's segment didn't change

Exactly, that would be over-invalidation. We could use the last Entity List audit entry for the whole list like we do for the unfiltered case. It’s hard to reason about because they’re going to favor different scenarios. Maybe we start with this and keep a close watch on Cloud server performance.

lib/model/query/datasets.js
```diff
@@ -717,72 +717,153 @@ const getUnprocessedSubmissions = (datasetId) => ({ all }) =>
...
+// Private: fetches last entity timestamp and count for a single actor, used for ownerOnly datasets
+const _fetchOwnerEntityStats = (all, datasetIds, filterActorId) => {
+  const unjoin = unjoiner(Dataset, Frame.define(into('entityMetadata'), 'lastEntityUpdate', 'entities'));
+  return all(sql`
+    SELECT ${unjoin.fields} FROM datasets
+    LEFT OUTER JOIN (
+      SELECT
+        entities."datasetId" AS id,
+        GREATEST(MAX(entities."createdAt"), MAX(entities."updatedAt")) AS "lastEntityUpdate",
+        COUNT(*) AS entities
```
lognaturel, line 744: Every time I see this `entities` identifier I am confused. Not a big deal but `entity_count` would easier to understand to me since `entities` is also the table name!

lognaturel (approved):
Tests are thorough and I find this a lot easier to follow than when it was a giant query!

We're deliberately over-invalidating in the filtered case for now and are going to see how it performs and decide what to do from there.

I think the most important thing now is to try it for real! 🚀

lib/model/query/datasets.js
ktuite: Yeah it's a bit awkward... We could come back and change it up later if you want!

PR #1854, opened 2026-06-23
Fix webformsEnabled migration

sadiqkhoja:
#### Before submitting this PR, please make sure you have:

- [x] run `make test` and confirmed all checks still pass, or witnessed Github completing all checks with success
- [x] verified that any code from external sources are properly credited in comments or that everything is internally sourced

PR #1858, opened 2026-06-26
Revert #1840 and #1854 and set the default value of DEFAULT_WEBFORMS_ENABLED to true

sadiqkhoja:
As discussed, we don't need to have separate `webformsEnabled` state for draft and published version for now. In the near future we can bring re-revert 🤣 to have that functionality.


lib/util/constants.js
```diff
@@ -15,7 +15,7 @@ const PURGE_DAY_RANGE = config.has('default.taskSchedule.purge')
 
 const DEFAULT_WEBFORMS_ENABLED = config.has('default.webformsEnabled')
```
lognaturel, line 16: We introduced this config to be able to force set Web Forms to default while in dev. Now that it's the default for all new forms, we could remove the setting as well because I don't think we have a need to force Enketo by default. But this could be future cleanup, too!

lognaturel: Actually, let's keep it for this version, it could be a useful escape valve if needed.

lognaturel (approved):
I see that Web Forms is now the default for new Forms. I've gone through and verified that every change is related to the reversion.

lib/model/query/forms.js
```diff
@@ -110,13 +110,13 @@ with sch as
...
+  select nextval(pg_get_serial_sequence('forms', 'id')), sch.id, ${form.xml}, ${def.name}, ${def.hash}, ${def.sha}, ${def.sha256}, ${def.version}, ${def.keyId}, ${form.xls.xlsBlobId || null}, ${def.draftToken || null}, ${def.enketoId || null}, clock_timestamp()
     from sch
   returning *),
 form as
-  (insert into forms (id, "xmlFormId", state, "projectId", "draftDefId", "acteeId", "enketoId", "enketoOnceId", "createdAt") -- webformsEnable is set to null implicitly for new draft in forms table
-  select def."formId", ${form.xmlFormId}, ${form.state || 'open'}, ${project.id}, def.id, ${form.acteeId}, ${form.enketoId || null}, ${form.enketoOnceId || null}, def."createdAt" from def
+  (insert into forms (id, "xmlFormId", state, "projectId", "draftDefId", "acteeId", "enketoId", "enketoOnceId", "createdAt", "webformsEnabled")
+  select def."formId", ${form.xmlFormId}, ${form.state || 'open'}, ${project.id}, def.id, ${form.acteeId}, ${form.enketoId || null}, ${form.enketoOnceId || null}, def."createdAt", ${form.webformsEnabled || DEFAULT_WEBFORMS_ENABLED} from def
```
lognaturel, line 119: Can `form.webformsEnabled` ever have a value when creating a new form through the UI or is this only relevant when using the API directly? This would require changing the form's setting before it's published, right?

sadiqkhoja: actually `form.webformsEnabled ||` is dead code, it can never have a value. Same is the case with other variables like `form.state`, `form.enketoId` and `form.enketoOnceId`

lognaturel: Makes sense. Maybe something to file an issue to clean up? Doesn't seem like a big deal but it's confusing to read.

sadiqkhoja: https://github.com/getodk/central/issues/2025

PR #1863, opened 2026-07-01
Fixes: dont change etag if submission is made to a different form

sadiqkhoja:
Towards https://github.com/getodk/central/issues/2028

Added a test that fails without the fix.

Added the missing where clause in the sql.

None

No

### DIFF
--- a/test/integration/api/odata.js
+++ b/test/integration/api/odata.js
@@ -659,6 +659,27 @@ describe('api: /forms/:id.svc', () => {
             });
           }))));
 
+    it('should update Etag if a new submission is made', testService(async (service) => {
+      const asAlice = await withSubmissions(service, identity);
+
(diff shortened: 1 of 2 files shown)


test/integration/api/odata.js
```diff
@@ -659,6 +659,26 @@ describe('api: /forms/:id.svc', () => {
             });
           }))));
 
+    it('should not update Etag if submission is made to another form', testService(async (service) => {
```
lognaturel, line 683: Feels like there should also be an integration test showing it does change when a submission is made to this form.

lognaturel (approved):
Very unfortunate miss! Hopefully this will ease some pressure on servers.

It feels very logical and safe. Let's do some checks on staging once it's in and see whether we also want to let QA know more about it.

PR #1867, opened 2026-07-04
api/submissions: test limit & offset params

### LINKED ISSUE #1672: submissions REST API needs tests for limit & offset query parameters
Submissions REST API needs tests for limit & offset query parameters.

In the PR for documenting these parameters [it came up](https://github.com/getodk/central-backend/pull/1749#issuecomment-3995568931) that we have them, we have docs for them, but no tests it seems.

alxndrsn:
Closes getodk/central#1672

- [x] ci

Simple test.  More complicated or extensive tests could be added.  Or they can be added when there's a regression.

No impact.

No.

### DIFF
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -3284,7 +3284,7 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff,,
       const asAlice = await service.login('alice');
 
       await Promise.all(Array.from({ length: 100 }, (_, idx) => asAlice.post('/v1/projects/1/forms/simple/submissions')
-        .send(`<data id="simple"><meta><instanceID>instance-${idx}</instanceID></meta>${idx}</data>`)
+        .send(`<data id="simple"><meta><instanceID>instance-${idx}</instanceID></meta></data>`)
         .set('Content-Type', 'text/xml')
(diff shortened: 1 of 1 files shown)


test/integration/api/submissions.js
```diff
@@ -3279,6 +3279,30 @@ one,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff,,
               body[0].should.be.an.ExtendedSubmission();
               body[0].submitter.displayName.should.equal('Alice');
             })))));
+
+    it('should support limit & offset params', testService(async (service) => {
+      const asAlice = await service.login('alice');
+
+      await Promise.all(Array.from({ length: 100 }, (_, idx) => asAlice.post('/v1/projects/1/forms/simple/submissions')
+        .send(`<data id="simple"><meta><instanceID>instance-${idx}</instanceID></meta>${idx}</data>`)
```
matthew-white: What's the second `${idx}` between `</meta>` and `</data>`?

matthew-white (approved):

test/integration/api/submissions.js
alxndrsn: :thinking: I'll check...

alxndrsn: Seems to be happy without it - removed.

PR #1869, opened 2026-07-07
/audits?end: use half-closed date range

### LINKED ISSUE #2052: AuditList: use half-open interval for time interval
In frontend, the `AuditList` component uses a closed range for start & end dates.

As described in #1323, https://github.com/getodk/central-frontend/pull/1697, this is not ideal.

The request is handled:

1. frontend
    1. https://github.com/getodk/central-frontend/blob/fb9761ee2f281fb7c6b8757a5869042bf8c70e5b/apps/central/src/components/audit/list.vue#L76-L85
2. backend
    1. https://github.com/getodk/central-backend/blob/ba45b5ff99a076e0a02dc53c793e62d24fa3610a/lib/resources/audits.js#L14-L16
    2. https://github.com/getodk/central-backend/blob/ba45b5ff99a076e0a02dc53c793e62d24fa3610a/lib/model/query/audits.js#L74-L80

# Observed

```
start <= audit.loggedAt <= end
```

# Expected

```
start <= audit.loggedAt < end+1
```

# Related

*  #1323
* https://github.com/getodk/central-frontend/pull/1697

alxndrsn:
Frontend PR: https://github.com/getodk/central-frontend/pull/1700

Closes getodk/central#2052

- [x] updated a test
- [ ]  tested manually with frontend updates

Time is continuous.

Should prevent confusion if ever something happens in the microseconds before midnight.

Yes.  It is done.


matthew-white:
@alxndrsn to what extent do you feel like this is a breaking change? `end` used to be inclusive, and we documented that, but this PR changes it to be exclusive.

I don't know how likely it is that anyone is using `end`. We know that some users regularly poll /v1/audits, then respond to new audit log events, but I imagine that such an approach would use `start` rather than `end`.

If someone were previously using `end` and specifying the last millisecond of the day like Central Frontend was (in order not to over-include the first millisecond of the next day), then this change would cause them to start missing that last millisecond. Not a particularly likely scenario, but this is why I'm trying to figure out whether this is a breaking change (and/or whether that matters).

One option is that we could change `end` along the lines of this PR, then make note of it as a breaking change in the changelog for v2026.3. The risk/impact feels low, so I'd be open to that approach.

Another option that I think Wicher was contemplating is introducing one or more new query parameters. Something like: keep `end` inclusive, but add a new `end_lt` query parameter that would be exclusive. Or: deprecate `end` and replace it with two new, more explicit query parameters, `end_lt` and `end_lte`.

I also seem to remember that @sadiqkhoja wants to add a new OData endpoint for the audit log and maybe even deprecate the REST endpoint /v1/audits. If we went with that approach, we could use `$filter` similar to the other OData endpoints, which would automatically give users flexibility around `lt` vs. `lte`. Perhaps we could seed the ground for that and introduce OData-style `$filter`ing even now. There's some precedent for that: GeoJSON endpoints aren't OData, yet use `$filter`.

I don't have strong feelings about which direction we go, mostly I just wanted to throw out these thoughts/options to see what you think.

alxndrsn:
@matthew-white Thanks for the comprehensive thoughts!

> I also seem to remember that @sadiqkhoja wants to add a new OData endpoint for the audit log and maybe even deprecate the REST endpoint /v1/audits.

If either of deprecating `/v1/audits` or the switch to `$filter` are likely to happen soon, then I think this PR can be closed, as either would close https://github.com/getodk/central/issues/2052.

Otherwise, I'd vote for making the change and documenting in release notes - as you say, it's unlikely for anyone to be affected by this.

matthew-white:
> If either of deprecating `/v1/audits` or the switch to `$filter` are likely to happen soon, then I think this PR can be closed

Deprecating /v1/audits probably isn't something that's going to happen in this release, since there's no concrete need for an OData endpoint, and there are other priorities.

I had a look at switching to `$filter`, but it's not immediately clear to me how we would reimplement the current `action` filter query parameter within OData `$filter`. The current `action` query parameter does a sort of prefix matching that might not be trivial to reimplement in OData. We could implement `$filter` just for `start`/`end`/`loggedAt`, not `action`, but that feels too much like an awkward in-between place to me. If we're going to implement `$filter`, I think we should implement support for all current query parameters. It'll be too hard to message to users otherwise.

Given that, I'd say let's go ahead with making the change in this PR. I'll take a closer look at the PR now.

docs/api.yaml
matthew-white, line 1: It'd be great if you could:

- Add an entry for .3 to the changelog toward the top of this file
- Make a note within the changelog of this change

docs/api.yaml
```diff
@@ -12265,7 +12265,7 @@ paths:
         example: 2000-01-01z
       - name: end
         in: query
-        description: The timestamp after which log entries are to be filtered out.
+        description: The timestamp from which log entries are to be filtered out.
         schema:
           type: string
         example: 2000-12-31T23:59.999z
```
matthew-white: How about changing this example to the first millisecond of the day?

matthew-white (approved):

docs/api.yaml
alxndrsn: I've updated this + 3 other references to `...999Z`.

docs/api.yaml
alxndrsn: Added.

matthew-white (approved):

PR #1870, opened 2026-07-07
/users?q: make search more user-friendly

### LINKED ISSUE #2053: Improve _Project Roles_ user search
Rewrite of https://github.com/getodk/central/issues/723

On the Project Roles page (e.g. https://dev.getodk.cloud/projects/1/users), user search isn't very helpful:

Current searching for "a" gives one obscure hit:

<kbd>
Image
</kbd>
<br/><br/>

But checking https://dev.getodk.cloud/users, there are two users whose **names and emails both** _start with_ a, which are not even matched.

It might be helpful to add `ILIKE` or fuzzy search to the search functionality at:

https://github.com/getodk/central-backend/blob/ba45b5ff99a076e0a02dc53c793e62d24fa3610a/lib/model/query/users.js#L69-L75

See tests at:

https://github.com/getodk/central-backend/blob/ba45b5ff99a076e0a02dc53c793e62d24fa3610a/test/integration/api/users.js

alxndrsn:
Closes getodk/central#2053

New tests, which pass on this branch.

With `lib/model/query/users.js` from master:

```
  78 passing (3s)
  3 failing

  1) api: /users
       GET
         should return sane email matches:

      AssertionError: expected 1 to be 4
      + expected - actual

      -1
      +4

  2) api: /users
       GET
         should return sane displayName matches:

      AssertionError: expected 1 to be 4
      + expected - actual

      -1
      +4

  3) api: /users
       GET
         should escape backslashes in search string:

      AssertionError: expected 1 to be 2
      + expected - actual

      -1
      +2

```

The search is quite confusing.

Should make life a bit better.  Unlikely to confuse or annoy.

No.

### DIFF
--- a/lib/model/query/users.js
+++ b/lib/model/query/users.js
@@ -71,16 +71,16 @@ ${options.ifArg('q', (q) => sql`
     greatest(word_similarity("displayName", ${q}) + similarity("displayName", ${q}),
       word_similarity(email, ${q}) + similarity(email, ${q})) as score on true`)}
 where ${sqlEquals(options.condition)} and actors."deletedAt" is null
-${options.ifArg('q', (q) => {
-  const qMatcher = `%${q.replaceAll(/[_%\\\\]/g, 'g$&')}%`;
-  return sql`
-    AND (
-      score > 0.5 OR
-      "email"       ILIKE ${qMatcher} ESCAPE 'g' OR
-      "displayName" ILIKE ${qMatcher} ESCAPE 'g'
-    )
-  `;
-})}
+  ${options.ifArg('q', (q) => {
+    const qMatcher = `%${q.replaceAll(/[_%\\\\]/g, 'g$&')}%`;
+    return sql`
+      AND (
+        score > 0.5 OR
+        "email"       ILIKE ${qMatcher} ESCAPE 'g' OR
+        "displayName" ILIKE ${qMatcher} ESCAPE 'g'
+      )
+    `;
+  })}
 order by ${options.ifArg('q', () => sql`score desc,`)} email asc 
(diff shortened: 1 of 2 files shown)


matthew-white:
I feel like @sadiqkhoja has thought about search things before, he might be a good person to review.

lib/model/query/users.js
```diff
@@ -71,7 +71,16 @@ ${options.ifArg('q', (q) => sql`
     greatest(word_similarity("displayName", ${q}) + similarity("displayName", ${q}),
       word_similarity(email, ${q}) + similarity(email, ${q})) as score on true`)}
 where ${sqlEquals(options.condition)} and actors."deletedAt" is null
-  ${options.ifArg('q', () => sql`and score > 0.5`)}
+  ${options.ifArg('q', (q) => {
+    const qMatcher = `%${q.replaceAll(/[_%\\\\]/g, 'g$&')}%`;
```
sadiqkhoja: what if `q` contains letter `g`?

```
> SELECT 'google' ILIKE '%google%' ESCAPE 'g'

?column?|
--------+
false   |
```

sadiqkhoja (approved):
I like this improvement, I myself have suffered from the poor search on user page.

My only concern with this change is that user might just enter a single character like `a` or `e` and will get the full list of users. Maybe that's not too bad since frontend sends the request on RETURN key not on `keyup`.

lib/model/query/users.js
alxndrsn: Good catch - fixed!

alxndrsn: \+ new test case

PR #1872, opened 2026-07-10
ci: run against both postgres 14 AND 18

alxndrsn:
In preparation for https://github.com/getodk/central/issues/1385

- [x] CI!

Alternatives:

* just run a parallel postgres v18 branch, and keep it updated
* drop v14 supprot immediately; not so great while prod is still on postgres v14

There are a few fixes here for minor things which have changed with pg 18.

No effect - this is just for tests.

No effect - this is just tests.


lognaturel:
Good idea 👍

ktuite (approved):
Looks great. Nice touch that the docker dev postgres in the makefile is also getting a new name.

PR #1873, opened 2026-07-10
test/http/preprocessors: test for colons in basic auth password

alxndrsn:
Add test to demonstrate assertion made in https://github.com/getodk/central-backend/pull/1246.

Basic Auth isn't very well specified, and currently this codebase implements its own decoding.  It seems sensible to test for documented behaviours.

A new test.

It seems sensible to test for documented behaviours.

No impact - just a new test to prevent regression.

No.

### DIFF
--- a/test/unit/http/preprocessors.js
+++ b/test/unit/http/preprocessors.js
@@ -135,6 +135,21 @@ describe('preprocessors', () => {
           )).then((context) => {
             context.auth.actor.should.eql(Option.of(new Actor({ displayName: 'test' })));
           })));
+
+      it('should treat additional colons as part of password', () =>
+        hashPassword('password:4alice').then((hashed) =>
+          Promise.resolve(authHandler(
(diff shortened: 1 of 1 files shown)


test/unit/http/preprocessors.js
```diff
@@ -135,6 +135,21 @@ describe('preprocessors', () => {
           )).then((context) => {
             context.auth.actor.should.eql(Option.of(new Actor({ displayName: 'test' })));
           })));
+
+      it('should treat additional colons as part of password', () =>
```
alxndrsn:
```suggestion
      it('should handle colons-in-password as part of password (not username)', () =>
```

matthew-white:
@alxndrsn is this PR ready for review? It looks like it may be ready or at least that it's close, so I thought I'd ask. I'd be happy to take a look once it's ready.

matthew-white (approved):

PR #1876, opened 2026-07-15
sentry: ignore known REPL issue

### LINKED ISSUE #2057: [backend] REPL crashes are reported to sentry
https://getodk.sentry.io/issues/7597650885/

```
TypeError: Cannot read properties of undefined (reading 'length')
    at [_multilineMove] (node:internal/readline/interface:1104:36)
    at [_moveUpOrHistoryPrev] (node:internal/readline/interface:1161:27)
    at [_ttyWrite] [as _ttyWrite] (node:internal/readline/interface:1510:37)
    at REPLServer.self._ttyWrite (node:repl:1001:9)
    at ReadStream.onkeypress (node:internal/readline/interface:284:20)
    at ReadStream.emit (node:events:509:28)
    at ReadStream.emit (node:domain:489:12)
    at emitKeys (node:internal/readline/utils:371:14)
    at emitKeys.next (<anonymous>)
    at ReadStream.onData (node:internal/readline/emitKeypressEvents:64:36)
    at ReadStream.emit (node:events:509:28)
    at ReadStream.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:563:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:514:3)
    at Readable.push (node:internal/streams/readable:394:5)
    at TTY.onStreamRead (node:internal/stream_base_commons:189:23)
```
Looks like https://github.com/nodejs/node/issues/59431

### Reproduction

1. paste a multiline, long string into the repl, and execute it
2. press ⬆️ a few times until crash occurs

Hard to provide a clearer repro, as the exact commands can't be reached in the REPL history (due to this bug).

Something like the output from:
```js
console.log(
  JSON.stringify(
    Array(500).fill("A".repeat(1000)),
    null,
    2,
  )
)
```

### Suggested resolution

REPL crashes are infrequent, so their reports are interesting to receive in Sentry.

If possible, add handling for this specific bug in https://github.com/getodk/central-backend/blob/master/lib/bin/repl.js.

alxndrsn:
Closes getodk/central#2057

- [x] ci
- [x] tested on dev.getodk.cloud

Also considered trying to do something with `uncaughtException` handler in `repl.js`.  This seems more direct, but there are loads of issues with it.  Could also invalidate the user's history once it's corrupt, or prompt them to, but those both seem like a lot of work for a rare case, which should be reasonably well understood.

No impact expected - should only affect Sentry reporting, and only for a very specific case.

No.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -206,7 +206,7 @@ const _init = (config) => { // eslint-disable-line no-shadow
       const error = hint.originalException;
       if (error == null) return sanitizedEvent; // we aren't sure why there isn't an exception; pass through.
 
-      if (process.argv[1]?.match(/lib\/bin\/repl(.js)?$/)) {
+      if (process.argv[1]?.match(/lib\/bin\/repl(\.js)?$/)) {
         // Ignore REPL crashes caused by corrupted history.  See:
         // * https://github.com/getodk/central/issues/2057
         // * https://github.com/nodejs/node/issues/59431
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -205,7 +205,7 @@ const _init = (config) => { // eslint-disable-line no-shadow
       const error = hint.originalException;
       if (error == null) return sanitizedEvent; // we aren't sure why there isn't an exception; pass through.
 
-      if (process.argv[1]?.match(/lib\/bin\/repl(\.js)?$/)) {
+      if (event.tags?.entrypoint === 'repl') {
         // Ignore REPL crashes caused by corrupted history.  See:
         // * https://github.com/getodk/central/issues/2057
         // * https://github.com/nodejs/node/issues/59431
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -202,7 +202,6 @@ const _init = (config) => { // eslint-disable-line no-shadow
         sanitizedEvent.extra.duration = Date.now() - sanitizedEvent.extra.startTimestamp;
       }
 
-      // only file the event if it is a bare exception or it is a true 500.x Problem.
       const error = hint.originalException;
       if (error == null) return sanitizedEvent; // we aren't sure why there isn't an exception; pass through.
 
@@ -215,6 +214,7 @@ const _init = (config) => { // eslint-disable-line no-shadow
             error.stack.includes('_moveUpOrHistoryPrev')) return null;
       }
 
+      // only file the event if it is a bare exception or it is a true 500.x Problem.
       if ((error.isProblem !== true) || (error.httpCode === 500)) return sanitizedEvent; // throw exceptions.
       return null; // we have a user-space problem.
     }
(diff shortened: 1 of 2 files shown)


lib/external/sentry.js
```diff
@@ -205,6 +205,16 @@ const _init = (config) => { // eslint-disable-line no-shadow
       // only file the event if it is a bare exception or it is a true 500.x Problem.
```
matthew-white: I feel like this comment is now disconnected from the line of code it applies to. Maybe it applied to this block of code as a whole, but it seems most relevant to L208/L218 in particular:

```js
if ((error.isProblem !== true) || (error.httpCode === 500)) return sanitizedEvent; // throw exceptions.
```

What do you think about moving this comment further below now?

lib/external/sentry.js
```diff
@@ -205,6 +205,16 @@ const _init = (config) => { // eslint-disable-line no-shadow
...
       const error = hint.originalException;
       if (error == null) return sanitizedEvent; // we aren't sure why there isn't an exception; pass through.
+
+      if (process.argv[1]?.match(/lib\/bin\/repl(.js)?$/)) {
```
matthew-white:
```suggestion
      if (process.argv[1]?.match(/lib\/bin\/repl(\.js)?$/)) {
```

lib/external/sentry.js
```diff
@@ -205,6 +205,16 @@ const _init = (config) => { // eslint-disable-line no-shadow
...
+        // Ignore REPL crashes caused by corrupted history.  See:
+        // * https://github.com/getodk/central/issues/2057
+        // * https://github.com/nodejs/node/issues/59431
+        if (error instanceof TypeError &&
```
matthew-white, line 212: How about using more `&&`s instead of nesting `if`s?

matthew-white (approved):

lib/external/sentry.js
matthew-white: Like replacing the second `if` with an `&&` instead (hopefully this suggestion is making sense 😅)

matthew-white: Also not something I feel super strong about, just a suggestion

alxndrsn: I quite like the current structure as it separates REPL issues from express server issues.  I wouldn't be surprised if there are more REPL-specific issues that need filtering in future.  I think the comments should be improved before merging though, to make this clearer.

Also if it's not happening already, the REPL should be setting a Sentry tag (or similar) to make sure it's easy to identify/filter out REPL-triggered reports in Sentry.

matthew-white: > if it's not happening already, the REPL should be setting a Sentry tag (or similar)

I like this idea! I sort of doubt that it's happening already.

alxndrsn: probably neater than checking `process.argv`...

lib/external/sentry.js
alxndrsn: :+1: moved

lib/external/sentry.js
alxndrsn: I've added a tag, and check for that instead of `process.argv`.

alxndrsn:
* updated to use a tag
* tried writing automated tests, but it's quite horrible trying to simulate interactive terminal sessions within automated tests
* tested manually locally and in live envs

PR #1877, opened 2026-07-17
Add size field to blobs and compute from content length (and backfill)

ktuite:
Part of closing https://github.com/getodk/central/issues/1895 

I was first considering just computing the size of every blob with `octet_length(content)` because I believe it just looks up `bytea` metadata and isn't an expensive query. 

But I wanted to also track the size for blobs that were uploaded to S3, so I went with this other approach of adding a `size` column and populating it on insert. Blobs that have already been moved to S3 won't have size information. Blobs that are later uploaded but match a blob on S3 also won't go through this insert path so they also wont have size information. 

The only place the blob size is returned in the API is through form attachments because I think that's the only place it's currently needed..

Tests and trying it with frontend.

The issue [mentioned](https://github.com/getodk/central/issues/1895#issuecomment-4455644381) some other ways of capturing the content size but we can get it pretty easily from the content itself as it comes in. Those comments might be more about backfilling S3 blobs, but the overall consensus is that that doesn't seem worth it, just handling regular blobs and new blobs is good enough. 

Touches blob upload path but I don't think it will hurt anything for users.

Made small change to API docs in this PR.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -45,6 +45,7 @@ info:
 
     **Changed**:
     - [GET /audits](/central-api-system-endpoints/#getting-audit-log-entries) now uses less-than when filtering with the `end` parameter.
+    - [Form Attachment](/central-api-form-management/#listing-form-attachments) objects now include a `size` field (in bytes) when a file has been uploaded. The field is absent for dataset-linked attachments and empty slots, and may be `null` for blobs uploaded before this version.
 
     ## ODK Central v2026.2
 
@@ -3940,6 +3941,7 @@ paths:
                 exists: true
                 blobExists: true
                 datasetExists: false
+                size: 4096
                 hash: 'd41d8cd98f00b204e9800998ecf8427e'
                 updatedAt: 2018-03-21T12:45:02.312Z
   /v1/projects/{projectId}/forms/{xmlFormId}/attachments/{filename}:
@@ -4419,6 +4421,7 @@ paths:
                 exists: true
                 blobExists: true
                 datasetExists: false
+                size: 4096
                 hash: 'd41d8cd98f00b204e9800998ecf8427e'
                 updatedAt: 2018-03-21T12:45:02.312Z
   /v1/projects/{projectId}/forms/{xmlFormId}/draft/attachments/{filename}:
@@ -5029,6 +5032,7 @@ paths:
                 exists: true
                 blobExists: true
                 datasetExists: false
+                size: 4096
                 hash: 'd41d8cd98f00b204e9800998ecf8427e'
                 updatedAt: 2018-03-21T12:45:02.312Z
   /v1/projects/{projectId}/forms/{xmlFormId}/versions/{version}/attachments/{filename}:
@@ -13033,11 +13037,16 @@ components:
         blobExists:
           type: boolean
           example: true
-          description: Whether the server has the file or not.
+          description: True only when a file has been uploaded for this attachment slot. False when no file has been uploaded, or when the attachment is linked to a Dataset instead of a file.
         datasetExists:
           type: boolean
           example: true
           description: Whether attachment is linked to a Dataset.
+        size:
+          type: integer
+          nullable: true
+          example: 4096
+          description: The size of the uploaded file in bytes. Only present when `blobExists` is `true`. May be `null` for files on external blob store uploaded before version 2026.3.
         updatedAt:
           type: string
           example: '2018-03-21T12:45:02.312Z'
--- a/lib/model/frames.js
+++ b/lib/model/frames.js
@@ -132,7 +132,7 @@ Form.Attachment = class extends Frame.define(
       datasetExists: this.datasetId != null
     };
     if (this.updatedAt != null) data.updatedAt = this.updatedAt;
-    if (this.blobId != null) data.size = this.aux.blob?.size; // computed from octet_length(content) on blob table
+    if (this.blobId != null) data.size = this.aux.blob?.size;
     return data;
   }
 };
--- a/lib/model/frames.js
+++ b/lib/model/frames.js
@@ -113,10 +113,13 @@ FieldKey.Extended = Frame.define('lastUsed', readable);
 const { Form } = require('./frames/form');
 Form.Attachment = class extends Frame.define(
   table('form_attachments'),
-  'formId',                             'formDefId',
-  'blobId',                             'datasetId',
-  'name',         readable,             'type',       readable,
-  'updatedAt',    readable,
+  'formId',
+  'formDefId',
+  'blobId',
+  'datasetId',
+  'name', readable,
+  'type', readable,
+  'updatedAt', readable,
   fieldTypes(['int4', 'int4', 'int4', 'int4', 'text', 'text', 'timestamptz'])
 ) {
   forApi() {
@@ -129,6 +132,7 @@ Form.Attachment = class extends Frame.define(
       datasetExists: this.datasetId != null
     };
     if (this.updatedAt != null) data.updatedAt = this.updatedAt;
+    if (this.blobId != null) data.size = this.aux.blob?.size; // computed from octet_length(content) on blob table
     return data;
   }
 };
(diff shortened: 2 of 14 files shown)


lib/model/migrations/20260717-01-add-blob-size.js
matthew-white, line 1: How about adding one or more tests of this migration? There are a lot of great tests below, but nothing that demonstrates the backfill aspect of this. Especially the case where `blobs.size` can't be backfilled and becomes `NULL`.

lib/model/query/blobs.js
```diff
@@ -19,8 +19,8 @@ const ensure = (blob) => ({ oneFirst }) => oneFirst(sql`
     SELECT id FROM blobs WHERE (sha, "contentType") = (${blob.sha}, ${blob.contentType || DEFAULT_CONTENT_TYPE})
   ),
   ensured AS (
-    INSERT INTO blobs (sha, md5, content, "contentType") VALUES (
-      ${blob.sha}, ${blob.md5}, ${sql.binary(blob.content)}, ${blob.contentType || sql`DEFAULT`}
+    INSERT INTO blobs (sha, md5, content, "contentType", size) VALUES (
+      ${blob.sha}, ${blob.md5}, ${sql.binary(blob.content)}, ${blob.contentType || sql`DEFAULT`}, ${blob.content.length}
```
matthew-white, line 23: `blob.content.length` is definitely the number of bytes, not the number of characters, right? I feel like `length` in JavaScript can be surprising. Maybe worth adding a test along those lines?

test/assertions.js
```diff
@@ -217,7 +217,7 @@ should.Assertion.add('ExtendedForm', function() {
 should.Assertion.add('FormAttachment', function() {
   this.params = { operator: 'to be a Form Attachment' };
 
-  Object.keys(this.obj).should.be.a.subsetOf([ 'name', 'type', 'blobExists', 'datasetExists', 'exists', 'hash', 'updatedAt' ]);
+  Object.keys(this.obj).should.be.a.subsetOf([ 'name', 'type', 'blobExists', 'datasetExists', 'exists', 'size', 'hash', 'updatedAt' ]);
```
matthew-white, line 220: How about asserting that `size` is a number? I see a lot of other type checks below.

lib/model/frames.js
```diff
@@ -129,6 +132,7 @@ Form.Attachment = class extends Frame.define(
       datasetExists: this.datasetId != null
     };
     if (this.updatedAt != null) data.updatedAt = this.updatedAt;
+    if (this.blobId != null) data.size = this.aux.blob?.size;
```
matthew-white, line 135: This is making me realize that we'll be adding `size` to the API response for form attachments, but not other blobs (in particular, not submission attachments). That sounds fine to me, as it's all we need for the issue. But I'm wondering what you think of the value of adding `size` for submission attachments as well. Maybe that could be done in a follow-up PR? (Or maybe there are higher-priority items. 😅)

docs/api.yaml
matthew-white, line 1: It looks like there's now a merge conflict on this file.

lib/model/query/blobs.js
ktuite: Yes, this `content` is always a buffer (from Blob.fromStream/fromBuffer/fromFile) so this is the byte length. Adding a test of this.

lib/model/frames.js
ktuite: Let's wait to add it later when we need it! The data will be already in the DB and it'll be easy to expose.

lib/model/query/form-attachments.js
```diff
@@ -129,11 +129,11 @@ const autoLinkWithDataset = (form, datasetId) => async ({ FormAttachments, Datas
 // GETTERS
 
 // This unjoiner pulls md5 from blob table (if it exists) and adds it to attachment frame
-const _unjoinMd5 = unjoiner(Form.Attachment, Frame.define(into('blob'), 'md5'));
+const _unjoinMd5 = unjoiner(Form.Attachment, Frame.define(into('blob'), 'md5', 'size'));
```
matthew-white: The name here (referencing `Md5`) feels slightly outdated now that the `unjoiner` is for both `md5` and `size`. How about something like `_unjoinBlobs` instead? I think the code comment above could be updated as well.

test/integration/other/blobs.js
```diff
@@ -150,4 +150,22 @@ describe('blob query module', () => {
...
+      .set('Content-Type', 'application/xml')
+      .send(testData.forms.withAttachments)
+      .expect(200);
+
+    await asAlice.post('/v1/projects/1/forms/withAttachments/draft/attachments/goodone.csv')
+      .set('Content-Type', 'text/csv')
+      .send(bufferContent)
+      .expect(200);
+
+    const { size } = await container.one(sql`SELECT size FROM blobs LIMIT 1`);
```
matthew-white: How about using the API instead to fetch the `size`? I think it's nice to use the API where possible compared to querying the database directly.

Maybe it's even possible that the POST response will return the `size`?

test/assertions.js
```diff
@@ -227,6 +227,7 @@ should.Assertion.add('FormAttachment', function() {
   exists.should.equal(blobExists || datasetExists);
   (hash != null).should.equal(blobExists);
   if (hash != null) hash.should.be.an.md5Sum();
+  if (this.obj.size != null) this.obj.size.should.be.a.Number();
```
matthew-white: How about also asserting that `size` only exists if `blobExists` is true?

```suggestion
  if (this.obj.size != null) {
    this.obj.size.should.be.a.Number();
    this.obj.blobExists.should.be.true();
  }
```

test/integration/api/forms/versions.js
matthew-white, line 1: (Not sure which file to add this comment to.) Do you think we do enough to test that `size` is `null`/missing for a dataset attachment? I think we probably do, but I thought I'd ask. We probably implicitly test that `size` is missing when making assertions about dataset attachments without referencing `size`.

lib/model/query/blobs.js
matthew-white: I think the tests would be extra reassuring if they referenced specific numbers rather than the `length` property. One idea I had is a test that specifically uses a Unicode character (maybe an emoji) that is just a single character long, but multiple bytes.

matthew-white (approved):
Love all the test cases!

lib/model/query/blobs.js
matthew-white: > I think the tests would be extra reassuring if they referenced specific numbers rather than the `length` property.

I know some of them do already, but maybe the ones that currently reference `length` could reference specific numbers instead. Just my two cents.

lib/model/query/blobs.js
matthew-white, line 1: I just thought of an interesting case. What happens in this sequence:

- You have an existing Central server that uses S3.
- You have a form attachment on the server. Its database blob `content` will be `NULL`, so it will end up with a `NULL` `size` after the migration.
- You upgrade your server to the latest version of Central, running the migration.
- You upload the same form attachment to a new form.
  - Will the blob (1) update with the size that's just been uploaded and is newly available (2) result in a conflict/error (3) not change at all (but also not error)? My guess is (1), but I think an extra test along these lines would be interesting.

test/integration/other/blobs.js
ktuite: ah yes, the POST response does return size

test/integration/api/forms/versions.js
ktuite: I think the existing tests (in datasets.js, for example) cover this because if size were to be coming through, those existing tests would have issues. I tried adding a new test and also couldn't figure out a good place for it, and it didn't feel like it was doing anything new.

lib/model/query/blobs.js
ktuite: If you upload the same blob, to the same attachment or a new attachment of a new form, the md5 hash will match that of an existing blob and the blob wont change. That means that (3) will happen -- the size of the file won't get computed and won't be available at all (unless you deleted and purged the blob and _then_ reuploaded).

ktuite: I've added a test showing this case.

PR #1889, opened 2026-07-23
lint: replace airbnb-base config with eslint:recommended

alxndrsn:
Ref https://github.com/getodk/central/issues/1090

```sh
git reset && \
git reset --hard && \
git checkout master && \
make node_modules && \
git checkout remove-dep-airbnb-config && \
mv .eslintrc.json thisbranch.eslintrc.json && \
git checkout master -- .eslintrc.json && \
diff \
     <(npx eslint --no-eslintrc -c .eslintrc.json           --print-config lib/constants.js | jq --sort-keys .rules) \
     <(npx eslint --no-eslintrc -c thisbranch.eslintrc.json --print-config lib/constants.js | jq --sort-keys .rules)
```

Removes an unsupported dependency, while ensuring no effective change in eslint config.

No effect.

No.


ktuite (approved):
Looks great! Glad to get that dependency out of there!

PR #1896, opened 2026-07-28
query/Entities: reduce fields fetched for export

### LINKED ISSUE #2092: `Entities.streamForExport()`: don't SELECT unused columns
The results from this query are consumed exclusively in `lib/data/entity.js`:

1. `streamEntityCsv()`
2. `streamEntityCsvAttachment()`
3. `streamEntityOdata()`

For each of these uses, confirm which of the selected columns is actually used.

Consider results with respect to:

* common unused columns
* use-specific unused columns

alxndrsn:
In reality, there don't seem to be many fields which are only used for specific cases, and they're not big fields:

* `entity.uuid`
* `entity.conflict`
* `actor.id`

Closes getodk/central#2092

All the tests pass.

Less fetch from DB = less overhead.

The `extends F` approach is "neat" as in succinct, but a bit horrible.

If the approach is accepted, then it can be applied in various other places too.

Should make things a little faster.  There's a risk it will also exclude columns that are actually used!

No.


ktuite (approved):
Seems fine to try this and see if it makes things a little bit faster. I like the idea of being able to be more selective with the fields.

PR #1900, opened 2026-07-29
oidc: document choice of flow

alxndrsn:

It's just a comment.  Hopefully the reviewer will read it, accept its accuracy, and check this box:

- [x] :wave:

Resolves the TODO, which was reasonable; hopefully accurately.

No effect.

No.

### DIFF
--- a/lib/resources/oidc.js
+++ b/lib/resources/oidc.js
@@ -11,8 +11,8 @@
 /* eslint-disable no-use-before-define */
 
 // OpenID Connect auth handling using Authorization Code Flow with PKCE.
-// As of January 2025, this is the current recommended flow per:
-// https://datatracker.ietf.org/doc/html/rfc9700#name-best-practices
+// As of January 2025, this is the current recommended flow.
+// See: https://datatracker.ietf.org/doc/html/rfc9700#name-best-practices
 
 const { createHash } = require('node:crypto');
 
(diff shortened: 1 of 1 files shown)


lib/resources/oidc.js
```diff
@@ -11,7 +11,8 @@
 /* eslint-disable no-use-before-define */
 
 // OpenID Connect auth handling using Authorization Code Flow with PKCE.
-// TODO document _why_ auth-code-flow, and not e.g. implicit flow?
+// This is the current recommended flow per:
```
alxndrsn:
```suggestion
// As of January 2025, this is the current recommended flow per:
```

lib/resources/oidc.js
```diff
@@ -11,7 +11,8 @@
 /* eslint-disable no-use-before-define */
 
 // OpenID Connect auth handling using Authorization Code Flow with PKCE.
-// TODO document _why_ auth-code-flow, and not e.g. implicit flow?
+// As of January 2025, this is the current recommended flow per:
+// https://datatracker.ietf.org/doc/html/rfc9700#name-best-practices
```
alxndrsn:
```suggestion
// As of January 2025, this is the current recommended flow.
// See: https://datatracker.ietf.org/doc/html/rfc9700#name-best-practices
```

sadiqkhoja (approved):
have read the comment not the doc 👋

PR #1901, opened 2026-07-29
remove unused dependency: pg-query-stream

alxndrsn:
* not used directly in this project
* not used indirectly by dependencies

Noticed while looking into https://github.com/getodk/central/issues/1090

- [ ] checked `package-lock.json` updates
- [ ] searched for dependency references in `node_modules` directory
- [ ] ci

It's not used, so why include it?

No effect.

No.


alxndrsn:
Incorrect!  This is required by `knex`, but only listed in `devDependencies`.

PR #1904, opened 2026-08-06
update test dependency: nock

alxndrsn:
Ref getodk/central#1090

One Enketo test changes response status from 204 No Content to 209.  nock no longer seems to pass through the supplied body.

See: https://github.com/nock/nock/issues/3000

- [x] ran unit tests locally
- [x] ci

No other approaches considered.

No effect.

No.


test/unit/external/enketo.js
```diff
@@ -74,7 +74,7 @@ describe('external/enketo', () => {
     it('should throw a Problem if the Enketo response code is unexpected', () => {
       enketoNock
         .post('/enketoPath/api/v2/survey/all')
-        .reply(204, {});
+        .reply(209, {});
```
ktuite, line 77: I'm thrown off by this 209. I see how 204 doesnt work anymore, but maybe this was supposed to become 409.

ktuite (approved):

test/unit/external/enketo.js
alxndrsn: It could be nearly anything actually!  I was trying to avoid the checks for specific error messages at https://github.com/getodk/central-backend/blob/f900ffb070b2abe98de1ed0eb8aa8c8fccf5d441/lib/external/enketo.js#L38-L45 by keeping an "OK" response code, but one which wouldn't be expected.  But it's true that right now, pretty much any code except `200`, `201`, `405` could be used.

alxndrsn: Some options:

1. update the response code to a non-specific 4xx error
2. update the response code to a non-specific 5xx error
3. update the test name:
```diff
-  it('should throw a Problem if the Enketo response code is unexpected', () => {
+  it('should throw a Problem if the Enketo response code is OK, but unexpected', () => {
```

Any preference?

ktuite: got it. just knowing that is in intentional, it's fine to leave it as is.

PR #1905, opened 2026-08-06
update test dependency: streamtest

alxndrsn:
Ref getodk/central#1090

- [x] ran unit tests locally
- [x] ran integration tests locally
- [x] ci

Keeps test changes minimal, while updating the lib.

No effect - just test code.

No.

### DIFF
--- a/test/unit/data/attachments.js
+++ b/test/unit/data/attachments.js
@@ -1,5 +1,5 @@
 const appRoot = require('app-root-path');
-const streamTest = require('../../util/streamtest');
+const streamTest = require(appRoot + '/test/util/streamtest');
 const { zipStreamToFiles } = require(appRoot + '/test/util/zip');
 const { PartialPipe } = require(appRoot + '/lib/util/stream');
 const { streamAttachments } = require(appRoot + '/lib/data/attachments');
--- a/test/unit/data/attachments.js
+++ b/test/unit/data/attachments.js
@@ -1,5 +1,5 @@
 const appRoot = require('app-root-path');
-const streamTest = require('streamtest').v2;
+const streamTest = require('../../util/streamtest');
 const { zipStreamToFiles } = require(appRoot + '/test/util/zip');
 const { PartialPipe } = require(appRoot + '/lib/util/stream');
 const { streamAttachments } = require(appRoot + '/lib/data/attachments');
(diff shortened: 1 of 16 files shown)


test/unit/data/attachments.js
```diff
@@ -1,5 +1,5 @@
 const appRoot = require('app-root-path');
-const streamTest = require('streamtest').v2;
+const streamTest = require('../../util/streamtest');
```
ktuite: Should this be like this?
```suggestion
const streamTest = require(appRoot + '/test/util/streamtest');
```

test/unit/http/endpoint.js
```diff
@@ -2,7 +2,7 @@ const should = require('should');
 const { EventEmitter } = require('events');
 const { Transform } = require('stream');
 const { createRequest, createResponse } = require('../../util/node-mocks-http');
-const streamTest = require('streamtest').v2;
+const streamTest = require('../../util/streamtest');
```
ktuite, line 5: Same here, don't understand difference between `../../` and `appRoot` when other unit test files use the other pattern.

test/unit/model/container.js
```diff
@@ -1,6 +1,6 @@
 const should = require('should');
 const { queryModuleBuilder, injector, withDefaults } = require('../../../lib/model/container');
-const streamTest = require('streamtest').v2;
+const streamTest = require('../../util/streamtest');
```
ktuite, line 3: Another `../../`

ktuite (approved):
Wondering about inconsistent ways of importing streamTest, but otherwise seems ok to preserve tests but move to new version.

test/unit/data/attachments.js
alxndrsn: Yeah, generally I'm trying to keep these consistent with surrounding code - some use appRoot, and some use relative paths...

test/unit/http/endpoint.js
alxndrsn: Just copy/paste failure.  Reviewing all & fixing.

alxndrsn: This particular test file doesn't use `appRoot`, so leaving this one as-is.

test/unit/model/container.js
alxndrsn: Consistent with surrounding code; leaving as-is.

ktuite (approved):

PR #1906, opened 2026-08-06
test: standardise app-root-path import name

alxndrsn:
Current master stats:

```
$ git grep -h "require('app-root-path')" | sort | uniq -c
      6 const appPath = require('app-root-path');
     75 const appRoot = require('app-root-path');
```

Current code is a bit annoying when copy/pasting imports amongst test files.

- [x] ci

* current code is a bit annoying when copy/pasting imports amongst test files
* consistency is more convenient

No effect.

No.


ktuite (approved):
ahh, feels refreshing

PR #1909, opened 2026-08-06
Add "viewAs" query param to OData and REST entities endpoints

### LINKED ISSUE #1792: Add "View as [app user]" filter to Entities
On the Entities page for a dataset, allow project managers to preview the entity list as a specific App User would see it, with that user's dataset access filter applied.

- Selector to pick an App User to "view as"
- Entity list re-queries with the selected user's filter applied
- Indication that a "view as" filter is active
- "View as" for Public Access Links is tracked separately

[Figma](https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=4242-642&t=PMDCYeAiF6LkLuOa-4)

### LINKED ISSUE #1879: Extend dataset access filter application to entity REST and OData endpoints
Extend dataset access filter rules so they also apply when querying entities via the REST and OData endpoints. This enables the "View as [app user]" / "View as [public access link]" preview feature in the frontend.

## Scope
- Apply configured filter rules to entity REST endpoint responses when the request is made on behalf of (or impersonating) a specific App User or Public Access Link
- Same for the OData entity endpoints
- API shape for "view as" — query parameter, header, or other (per design decision)

## Depends on
- https://github.com/getodk/central/issues/1876 (design — which API surface backs "View as")
- https://github.com/getodk/central/issues/1877 (filter rules CRUD + apply)
- Enables: #1792 and the PAL "View as" frontend issue

ktuite:
Closes https://github.com/getodk/central/issues/1879

Add `?viewAs=<actorId>` query parameter to entities endpoints (both OData and REST) to filter the list to only what would be shown to that actor if the entities were fetched as a CSV form attachment. 

This will emulate
- `ownerOnly` filters (for app users, public links, _and_ web data collectors)
- actor property filters (app users and public links)

If the actor ID of a manager, viewer, or admin is provided (someone who can view all entities), this filter will return all entities.

This filter can be applied with other filters in the OData endpoint, e.g. `$filter` and `$search` to further filter the entities for a specific user.

Tests.

I have a few questions to discuss:
- is `viewAs` a good query parameter name?
- is it actually necessary for the REST endpoint to have this? it doesn't have any other filters. But since this one isn't odata, it seems ok to add it to the REST endpoint. 
- is it weird to have this non-OData query param alongside the OData parameters? 

Gives more options to users.

API docs are updated as part of the PR.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -9090,7 +9090,7 @@ paths:
           This is intended for previewing what a particular App User or Public Link would receive when downloading the Dataset via a linked Form Attachment.
         schema:
           type: number
-        example: "42"
+        example: 42
       responses:        
         "200":
           description: OK
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -44,7 +44,7 @@ info:
     ## ODK Central v2026.3
 
     **Added**:
-    - [Entities OData](/central-api-odata-endpoints/#data-document-1) now supports a `viewAs` query parameter to preview the filtered Entity list for a specific Actor.
+    - [Entities OData](/central-api-odata-endpoints/#odata-dataset-service) and [GET Entities Metadata](/central-api-entity-management/#entities-metadata) now support a `viewAs` query parameter to preview the filtered Entity list for a specific Actor.
 
     **Changed**:
     - [GET /audits](/central-api-system-endpoints/#getting-audit-log-entries) now uses less-than when filtering with the `end` parameter.
@@ -9063,7 +9063,7 @@ paths:
       - Entity Management
       summary: Entities Metadata
       description: |-
-        This endpoint returns a list of the Entities of a Dataset. Please note that this endpoint only returns metadata of the entities not the data. If you want to get the data of all entities then please refer to [OData Dataset Service](/central-api-odata-endpoints/#odata-form-service)
+        This endpoint returns a list of the Entities of a Dataset. Please note that this endpoint only returns metadata of the entities not the data. If you want to get the data of all entities then please refer to [OData Dataset Service](/central-api-odata-endpoints/#odata-dataset-service)
         You can provide `?deleted=true` to get only deleted entities.
       
       operationId: entitiesMetadata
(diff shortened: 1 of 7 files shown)


docs/api.yaml
```diff
@@ -9079,6 +9082,15 @@ paths:
...
         example: people
+      - name: viewAs
+        in: query
+        description: |-
+          If provided, filters the response to only the Entities that the specified Actor would be able to see, based on the Dataset's access filter (`ownerOnly` or property rules). The value must be a numeric Actor ID.
+
+          This is intended for previewing what a particular App User or Public Link would receive when downloading the Dataset via a linked Form Attachment.
+        schema:
+          type: number
+        example: "42"
```
sadiqkhoja: nit:

```suggestion
        example: 42
```

ktuite:
We decided in the thursdsay team meeting that a non-OData (without the `$`) `viewAs` query parameter was fine. In the future, we could explore API changes that let a superuser view a request on behalf of a lower-access user.

test/integration/api/odata-entities.js
```diff
@@ -894,6 +894,388 @@ describe('api: /datasets/:name.svc', () => {
...
+        await asAlice.get(`/v1/projects/1/datasets/people.svc/Entities?viewAs=${appUser.id}&$search=entity+1`)
+          .expect(200)
+          .then(({ body }) => {
+            body.value.length.should.eql(1);
+            body.value[0].label.should.eql('north entity 1');
+          });
+      }));
+    });
+  });
+
```
sadiqkhoja, line 1313: how about a test to assert that (soft) deleted entities are return when `viewAs` is used even when `deletedAt ne null` is passed to the filters

sadiqkhoja (approved):

ktuite:
I'm realizing that the `viewAs` on the REST endpoint is fine, but where it's possibly needed more is on the CSV download. That's where filter and search is actually used outside of OData. The REST endpoint may not even need `viewAs` at all because it doesn't have those other filters, either.

matthew-white:
What about the GeoJSON endpoints out of curiosity? Right now they support `$filter`.

ktuite:
I'm going to merge this first part and then do the followup work for this other issue:
- https://github.com/getodk/central/issues/2162

PR #1910, opened 2026-08-10
postgres: support root CA cert if supplied to PGSSLROOTCERT

### LINKED ISSUE #2109: [backend] configure db connection with PGSSLROOTCERT if supplied
Should be able to configure with `ssl.ca`: https://node-postgres.com/features/ssl#self-signed-cert

Currently requires either:

1. `NODE_TLS_REJECT_UNAUTHORIZED=0` or
2. `NODE_EXTRA_CA_CERTS=path-to-pem`

May need to be tested with either:

1. `PGSSLMODE=verify-full`, or
2. `PGSSLMODE=require`

Origin of the issue: https://getodk.slack.com/archives/C01TKJ7TG7J/p1785312311227589

### LINKED ISSUE #2115: [backend] cron jobs broken in environments with postgres SSL enabled
### task: `purge`

https://getodk.sentry.io/issues/7641561073/

```
ConnectionError: self-signed certificate in certificate chain
    at Object.createConnection (/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js:55:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at ? (/usr/odk/lib/task/purge.js:14:19)
    at ? (/usr/odk/lib/task/task.js:62:18)
```

### task: `s3.getUploadCount()`

https://getodk.sentry.io/issues/7641633121

```
ConnectionError: self-signed certificate in certificate chain
    at Object.createConnection (/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js:55:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at getUploadCount (/usr/odk/lib/task/s3.js:24:26)
    at ? (/usr/odk/lib/task/s3.js:64:17)
    at ? (/usr/odk/lib/task/task.js:62:18)
```

### task: `process-backlog`

https://getodk.sentry.io/issues/7641416413/

```
ConnectionError: self-signed certificate in certificate chain
    at Object.createConnection (/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js:55:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at ? (/usr/odk/lib/model/query/entities.js:760:18)
    at ? (/usr/odk/lib/task/task.js:62:18)
```

### task: `run-analytics`

https://getodk.sentry.io/issues/7641489540/

```
ConnectionError: self-signed certificate in certificate chain
    at Object.createConnection (/usr/odk/node_modules/slonik/dist/src/factories/createConnection.js:55:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at ? (/usr/odk/lib/task/task.js:62:18)
    at ? (/usr/odk/lib/task/task.js:62:18)
```

alxndrsn:
Closes getodk/central#2109
Closes https://github.com/getodk/central/issues/2115

- [x] new ci job

* This approach is more targetted than adding the root CA for all of NodeJS via `NODE_EXTRA_CA_CERTS`.
* An alternative would be to make the code changes inside the `slonik` lib.  There are complications associated with that, as it's an old fork, but it does at least currently build: https://github.com/alxndrsn/getodk-legacy-slonik/
* Another alternative _might_ be to use `pg-connection-string`'s `uselibpqcompat=true` mode, but monkey-patching `pg.Pool` would likely still be necessary to avoid slonik's injection of `rejectUnauthorized: false` into `pg.ssl` config: https://github.com/alxndrsn/getodk-legacy-slonik/blob/135fa07b4d11e49dd626d5d2d0e01d35cef00879/src/factories/createPoolConfiguration.ts#L31-L36
* Tests might be more readable if fully refactored into data-driven format, with a table of:

`PGSSLMODE` | `PGSSLMODE` | expected error
---|---|---
... | ... | ...

There's a risk that some deployers are using SSL with less secure options.  Extra work could be done to exhausitvely support all different values for `PGSSLROOTCERT` at https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS.

No.


.github/workflows/db-ssl.yml
```diff
@@ -1,12 +1,16 @@
-name: Full Standard Test Suite
+name: Database Migrations
```
alxndrsn:
```suggestion
name: Database SSL Checks
```

lib/external/slonik.js
```diff
@@ -25,7 +59,7 @@ const slonikPool = (config) => {
       createNumericTypeParser(),
       timestampTypeParser,
       timestamptzTypeParser
-    ]
+    ],
```
alxndrsn:
```suggestion
    ]
```

sadiqkhoja:
What do you think about this alternative:

```diff
 const slonikPool = (config) => {
   setlibpqEnv(config);
-  return createPool('postgres://', {
+  return createPool(`postgres://${process.env.PGSSLROOTCERT ? `?sslrootcert=${process.env.PGSSLROOTCERT}` : ''}`, {
     captureStackTrace: false,
     maximumPoolSize: config.maximumPoolSize ?? 10,
     typeParsers: [
```

I think it's simpler and does the job.

What I understand is that the root cause of the issue is that [node-postgresql doesn't support PGSSLROOTCERT](https://github.com/brianc/node-postgres/issues/2723). So to solve that we can pass `sslrootcert=<path>` query parameter to the connection string and `pg-connection-string` will happily set the root cert for us even without other information in the string like host, user, password - those will be read via env var.

_PS: Whichever solution we go with `knex-migrator.js` also needs to be updated._

alxndrsn:
> I think it's simpler and does the job.

:+1: yes, that would be much simpler.

Note that `PGSSLROOTCERT` should be escaped when injecting into the connection string.  E.g. 

```js
const connectionUrl = new URL('postgres://');

const { PGSSLROOTCERT } = process.env;
if(PGSSLROOTCERT) u.searchParams.set('sslrootcert', PGSSLROOTCERT);

return createPool(connectionUrl.toString(), {
```

> _PS: Whichever solution we go with `knex-migrator.js` also needs to be updated._

:eyes: If that's the case, then there's a gap in the new test job.  That should be addressed before changing `knex-migrator.js`.

alxndrsn:
All locations where `setlibpqEnv()` is called should also be reviewed:

* `lib/bin/restore.js`
* `lib/external/slonik.js`
* `lib/model/knex-migrator.js`
* `lib/model/knexfile.js`

alxndrsn:
> I think it's simpler and does the job.

It's simpler.  Does it override `rejectUnauthorized: true`, which `slonik` sets internally for all connections?

sadiqkhoja:
what I have observed is that when `PGSSLMODE` is set to `verify-full` or `verify-ca` then value of `rejectUnauthorized` has no impact.

alxndrsn:
> what I have observed is that when `PGSSLMODE` is set to `verify-full` or `verify-ca` then value of `rejectUnauthorized` has no impact.

Interesting, because when I disable the override of `pg.Pool` like so:

```diff
-const { Pool } = pg;
-pg.Pool = function SslAwarePool(_cfg) {
-  const cfg = clone(_cfg);
-
-  const { PGSSLMODE, PGSSLROOTCERT } = process.env;
-
-  console.log('slonik', {
-    PGSSLMODE,
-    PGSSLROOTCERT,
-  });
-
-  if (PGSSLROOTCERT) {
-    // See: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES
-    if (!['verify-ca', 'verify-full'].includes(PGSSLMODE))
-      throw new Error('Incompatible values for env vars: PGSSLMODE, PGSSLROOTCERT.');
-
-    cfg.ssl = {
-      ca: readFileSync(PGSSLROOTCERT),
-      rejectUnauthorized: true,
-    };
-  }
-
-  return new Pool(cfg);
-};
-
 const slonikPool = (config) => {
   setlibpqEnv(config);
-  return createPool('postgres://', {
+  const { PGSSLROOTCERT } = process.env;
+  return createPool(`postgres://${PGSSLROOTCERT ? `?sslrootcert=${PGSSLROOTCERT}` : ''}`, {
     captureStackTrace: false,
     maximumPoolSize: config.maximumPoolSize ?? 10,
     typeParsers: [
```

then I see the following test results:

```
  db connection
    1) should fail to create pool if PGSSLROOTCERT is set without PGSSLMODE
    2) should fail to create pool if PGSSLROOTCERT is set with PGSSLMODE="nonsense"
    3) should fail to create pool if PGSSLROOTCERT is set with PGSSLMODE="require"
    4) should fail to create pool if PGSSLROOTCERT is set with PGSSLMODE="disable"
    5) should fail to create pool if PGSSLROOTCERT is set with PGSSLMODE=""
    ✔ should fail to create pool if cert file is missing
    ✔ should fail to create pool if cert file is a directory
    6) should fail at first query if cert file is a file, but not a cert
    ✔ should not connect to db if expected env vars are not set
    ssl mode: verify-ca
      ✔ should not connect to db if no root cert is supplied
      7) should not connect to db if a different root cert is supplied
      ✔ should connect to db if PGSSLMODE and PGSSLROOTCERT are supplied
    ssl mode: verify-full
      ✔ should not connect to db if no root cert is supplied
      8) should not connect to db if a different root cert is supplied
      ✔ should connect to db if PGSSLMODE and PGSSLROOTCERT are supplied


  7 passing (91ms)
  8 failing
```

alxndrsn:
@sadiqkhoja Can you clarify how the connection string approach should be implemented so that tests pass?

sadiqkhoja:
I was wrong about this: 

> when PGSSLMODE is set to verify-full or verify-ca then value of rejectUnauthorized has no impact."

`rejectUnauthorized` is option passed to tls.connect by `pg` package and once it is set by slonik there is no cleaner way to unset or update it.

So we have no other option than monkey-patching `Pool` or `pg-connection-string.parse` or update our `slonik` fork.

sadiqkhoja (approved):

PR #1911, opened 2026-08-10
Fixes getodk/central#2111: added dataset id in the where clause

### LINKED ISSUE #2111: Entities can be bulk deleted even if name of the dataset in the endpoint is wrong
Consider the following test:

```
it('should not delete entities that belong to a different dataset than the one in the URL', testService(async (service) => {

      // Create two datasets in the same project
      await asAlice.post('/v1/projects/1/datasets')
        .send({ name: 'people' })
        .expect(200);

      await asAlice.post('/v1/projects/1/datasets')
        .send({ name: 'trees' })
        .expect(200);

      // Create entities in the 'people' dataset
      await asAlice.post('/v1/projects/1/datasets/people/entities')
        .send({
          entities: [
            { uuid: '12345678-1234-4123-8234-123456789abc', label: 'Person 1' },
            { uuid: '12345678-1234-4123-8234-123456789def', label: 'Person 2' }
          ],
          source: { name: 'people.csv', size: 100 }
        })
        .expect(200);

      // Create entities in the 'trees' dataset
      await asAlice.post('/v1/projects/1/datasets/trees/entities')
        .send({
          entities: [
            { uuid: '12345678-1234-4123-8234-aaaaaaaaaaa1', label: 'Tree 1' },
            { uuid: '12345678-1234-4123-8234-aaaaaaaaaaa2', label: 'Tree 2' }
          ],
          source: { name: 'trees.csv', size: 100 }
        })
        .expect(200);

      // Try to delete 'people' entities by calling bulk-delete on 'trees' dataset
      await asAlice.post('/v1/projects/1/datasets/trees/entities/bulk-delete')
        .send({
          ids: ['12345678-1234-4123-8234-123456789abc', '12345678-1234-4123-8234-123456789def']
        })
        .expect(200)
        .then(({ body }) => {
          // Should return 0 because these UUIDs don't belong to the 'trees' dataset
          body.count.should.equal(0);
        });

    }));
  });
```

Even though entities' UUID belong to people dataset and `trees` is given in the URL, request goes through.

sadiqkhoja:
Closes getodk/central#2111

Added bunch of tests

Adding `dataset.id` is sufficient to fix the bug because `dataset.id` is unique across the application. 

Also added the same for bulk restore endpoint.

None

No.

### DIFF
--- a/lib/model/query/entities.js
+++ b/lib/model/query/entities.js
@@ -1133,8 +1133,11 @@ inner join
 order by actors."displayName" asc`)
   .then(map(construct(Actor)));
 
-const deleteByIds = (uuids) => ({ allFirst }) => allFirst(sql`
-  UPDATE entities SET "deletedAt" = now() WHERE "deletedAt" IS NULL AND uuid = ANY (ARRAY[${sql.array(uuids, 'uuid')}])
+const deleteByIds = (uuids, dataset) => ({ allFirst }) => allFirst(sql`
+  UPDATE entities SET "deletedAt" = now()
+  WHERE "deletedAt" IS NULL
+    AND uuid = ANY (ARRAY[${sql.array(uuids, 'uuid')}])
+    AND "datasetId" = ${dataset.id}
   RETURNING uuid
 `);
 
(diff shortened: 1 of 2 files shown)


ktuite (approved):
Looks great.

Just clarifying for our future selves that the first test never passed -- a project manager could not delete entities from a project they did not have access to. They could, however, delete entities from the wrong project if they had access to both. 

I wanted to ask for a test of that but then I realized it is covered under the existing tests, like "should not delete entities that belong to a different project than the one in the URL".  This is a good set of tests that captures the core issue.

sadiqkhoja:
correction: first test never failed.

PR #1912, opened 2026-08-14
log-upgrade: suppress output

### LINKED ISSUE #2125: log-upgrade prints a disconcerting null when everything is probably OK
The `null` in the service logs is a little disconcerting. Could it print out something more helpful or nothing at all?

```
generating local service configuration..
waiting for PostgreSQL to become connectable to...
running migrations..
logging server upgrade..
null   <--- 😥
starting cron..
using 4 worker(s) based on available memory (2023972864)..
starting server.
```

The code that's being run is [node ./lib/bin/log-upgrade](https://github.com/getodk/central/blob/72c2bd63785fa1a913a235e7e7c8fbdb019653f2/files/service/scripts/start-odk.sh#L63) in start-odk.sh.

alxndrsn:
Add option to `task.run()` to prevent default of printing JSON-serialised return value to stdout.

Closes getodk/central#2125

- [x] ci

Also:

```
$ node lib/bin/log-upgrade.js 
```
(there was no output)

```
$ SENTRY_TAGS='{"version.central":"c1","version.client":"c2"}' node lib/bin/log-upgrade.js 
attempted to log Sentry transaction in development:
{ op: 'task', name: 'log-upgrade', duration: 54.322198000000014 }
```

Seems helpful anyway to have an option to suppress output from tasks.

Should decrease their alarm reported in getodk/central#2125.

No.

### DIFF
--- a/lib/bin/log-upgrade.js
+++ b/lib/bin/log-upgrade.js
@@ -27,4 +27,4 @@ try {
   // and logUpgrade task wont run
 }
 
-if (version) run(logUpgrade({ version, server, client }));
+if (version) run(logUpgrade({ version, server, client }), { silent: true });
--- a/lib/task/task.js
+++ b/lib/task/task.js
@@ -22,6 +22,7 @@ const config = require('config');
 const container = require('../model/container');
 const Problem = require('../util/problem');
 const Option = require('../util/option');
+const { noop } = require('../util/util');
 const { slonikPool } = require('../external/slonik');
 const jsonSerialize = require('../util/json-serialize');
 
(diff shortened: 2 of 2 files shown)


sadiqkhoja (approved):
This is fine. But I would have updated `lib/task/log-upgrade.js` to return the result of the action to make consistent with other tasks. 

With this PR `start-odk.sh` would log "logging server upgrade.." but it's not necessary that an log entry would be created in Audits table. My preference would be to see a message telling me something like "nothing to log" / "server upgrade logged"

alxndrsn:
Thanks for the review :+1:

> This is fine. But I would have updated `lib/task/log-upgrade.js` to return the result of the action to make consistent with other tasks.

Can you clarify what you'd expect `lib/task/log-upgrade.js` to return?

> With this PR `start-odk.sh` would log "logging server upgrade.." but it's not necessary that an log entry would be created in Audits table.

Isn't this already the case?

sadiqkhoja:
> Can you clarify what you'd expect lib/task/log-upgrade.js to return?

In case of failure `nothing to log, server is up to date` and in case of success `logged server upgraded to [new version] in the Audits`

> Isn't this already the case?

yes, adding above messages would make stdout nicer though.

alxndrsn:
> > Can you clarify what you'd expect lib/task/log-upgrade.js to return?
> 
> In case of failure `nothing to log, server is up to date` and in case of success `logged server upgraded to [new version] in the Audits`
> 
> > Isn't this already the case?
> 
> yes, adding above messages would make stdout nicer though.

:+1: please file issues.

PR #1913, opened 2026-08-14
Fix entity submission parsing to allow fields named label to be saved

### LINKED ISSUE #1970: Form field with name `label` and `save_to` bind does not populate Entity property
Reported by user: https://forum.getodk.org/t/a-property-named-label-in-the-xlsform-cannot-be-saved-to-an-entity/58253

I can reproduce with fields `/data/label` and `/data/grp/label` but not `/data/label2`.

Form definition: [label_to_property.xlsx](https://github.com/user-attachments/files/28815031/label_to_property.xlsx)
On staging server: https://staging.getodk.cloud/projects/11/forms/label_to_property/submissions

I tried the form in Collect and I can see that the property is populated by Collect offline. I've verified that the field is correctly populated in Central, its value is just not saved to the Entity.

ktuite:
Closes https://github.com/getodk/central/issues/1970

Found the bug where any form field named `"label"` (because it was expecting things like`/meta/entity/label`) was written to the entity label, even if it was associated with a dataset property. 

Reproducing the bug with a test and then fixing it. 

I think flipping the if statement like that is fine. It will write the any valid property now and not get confused by properties from a field named "label".

Fixes a bug for users.

No.

### DIFF
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -161,11 +161,10 @@ const submissionXmlToEntityData = (structurals, entityFields, xml) => {
 
         const entity = last(entities[datasetId]);
 
-        if (name === 'label')
-          entity.system.label = textBuffer;
-        else
+        if (propertyName != null)
           entity.data[propertyName] = textBuffer;
-
+        else if (name === 'label')
+          entity.system.label = textBuffer;
       }
     }
   }, { xmlMode: true, decodeEntities: true });
--- a/lib/data/entity.js
+++ b/lib/data/entity.js
@@ -152,7 +152,7 @@ const submissionXmlToEntityData = (structurals, entityFields, xml) => {
 
       if (structural == null && entityField != null) {
 
-        const { name, propertyName, datasetId } = entityField;
+        const { propertyName, datasetId } = entityField;
 
         if (!entities[datasetId]) {
           // It's possible to see entity properties before the entity block
@@ -163,7 +163,7 @@ const submissionXmlToEntityData = (structurals, entityFields, xml) => {
 
         if (propertyName != null)
           entity.data[propertyName] = textBuffer;
-        else if (name === 'label')
+        else // entityField with name === 'label' is the only other possibility
           entity.system.label = textBuffer;
       }
     }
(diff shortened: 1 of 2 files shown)


lib/data/entity.js
```diff
@@ -161,11 +161,10 @@ const submissionXmlToEntityData = (structurals, entityFields, xml) => {
 
         const entity = last(entities[datasetId]);
 
-        if (name === 'label')
-          entity.system.label = textBuffer;
-        else
+        if (propertyName != null)
           entity.data[propertyName] = textBuffer;
-
+        else if (name === 'label')
```
matthew-white: Before, this was an `if`/`else`, but now it's an `if`/`else if`. Is there a case that's not covered by the `if` and `else if`? If there isn't a third case, how about reverting to `if`/`else`? Or how about adding an `else` that throws an error and says, "something unexpected happened"?

test/unit/data/entity.js
```diff
@@ -307,6 +307,59 @@ describe('extracting and validating entities', () => {
...
+                            </meta>
+                        </data>
+                    </instance>
+                    <instance id="things" src="jr://file-csv/things.csv"/>
+                    <bind nodeset="/data/grp/label" type="string" entities:saveto="denomination"/>
+                    <bind nodeset="/data/thing" type="string"/>
+                    <bind nodeset="/data/thing_note" readonly="true()" type="string"/>
+                    <bind nodeset="/data/meta/entity/@id" readonly="true()" type="string"/>
+                    <setvalue ref="/data/meta/entity/@id" event="odk-instance-first-load" value="uuid()"/>
+                    <bind nodeset="/data/meta/entity/label" calculate=" /data/grp/label " readonly="true()" type="string"/>
```
matthew-white: I don't totally understand this `calculate`. I thought `/data/meta/entity/label` and `/data/grp/label` were supposed to be different. Isn't the `calculate` setting them to be the same?

matthew-white (approved):
I'm not super familiar with `submissionXmlToEntityData()`, but this approach looks reasonable to me. The test is convincing.

test/unit/data/entity.js
ktuite: This came from a user's form, but it's not relevant to this bug/test so I'll simplify the form XML to just highlight having a field named "label" mapped to a property not named label.

PR #1916, opened 2026-08-19
test/api/projects: test dataset/extended metadata relationship

alxndrsn:
Add explicit test linking datasets to extended metadata.

Ref getodk/central#2032

Looking into getodk/central#2032, there didn't seem to be any explicit link in the tests between the `X-Extended-Metadata` header and inclusion of dataset information.

Could continue to not test it.

No effect - just a new test.

No.


garethbowen (approved):
Sweet as.

PR #1917, opened 2026-08-19
GET projects/:id: add option to include verbs only

### LINKED ISSUE #2032: Loading a webform does a request for extended project metadata which is slow
We use a user's verbs on the form to see what they're allowed to do. We can't get that information just from fetching the XML because a user who can preview a form may not be allowed to fill it.

Currently getting the verbs requires doing an extended project metadata request which is slow. Could we introduce an endpoint to get just the verbs?

Related: https://github.com/getodk/central/issues/1318

CC @garethbowen

alxndrsn:
Closes https://github.com/getodk/central/issues/2032

- [x] existing tests
- [x] new tests

- [x] confirm if using query parameters is idiomatic
- [x] confirm if chosen query parameter is acceptable

Adds an extra way to query the API, but shouldn't affect existing endpoints or users.

Yes, probably.

- [x] update docs

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -2642,7 +2642,7 @@ paths:
         example: "16"
       - name: verbs
         in: query
-        description: If set to `true`, will return allowed verbs for the current user on this project.  When extended metadata is requested, this parameter is ignored, and verbs are always included.
+        description: If set to `true`, will return allowed verbs for the current actor on this project.  When extended metadata is requested, this parameter is ignored, and verbs are always included.
         schema:
           type: boolean
         example: "true"
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -2640,6 +2640,12 @@ paths:
         schema:
           type: number
         example: "16"
+      - name: verbs
+        in: query
+        description: If set to `true`, will return allowed verbs for the current user on this project.  When extended metadata is requested, this parameter is ignored, and verbs are always included.
+        schema:
+          type: boolean
+        example: "true"
       responses:
         "200":
           description: 'This is the Extended Metadata response, if requested via the
--- a/lib/resources/projects.js
+++ b/lib/resources/projects.js
@@ -40,7 +40,7 @@ module.exports = (service, endpoint) => {
     Projects.getById(params.id, queryOptions)
       .then(getOrNotFound)
       .then((project) => auth.canOrReject('project.read', project))
-      .then((project) => ((queryOptions.extended === true)
+      .then((project) => ((queryOptions.extended === true || queryOptions.argData.includeVerbs === 'true')
         ? auth.verbsOn(project).then((verbs) => Object.assign({ verbs }, project.forApi()))
         : project))));
 
--- a/lib/resources/projects.js
+++ b/lib/resources/projects.js
@@ -40,7 +40,7 @@ module.exports = (service, endpoint) => {
     Projects.getById(params.id, queryOptions)
       .then(getOrNotFound)
       .then((project) => auth.canOrReject('project.read', project))
-      .then((project) => ((queryOptions.extended === true || query.includeVerbs === 'true')
+      .then((project) => ((queryOptions.extended === true || query.verbs === 'true')
         ? auth.verbsOn(project).then((verbs) => Object.assign({ verbs }, project.forApi()))
         : project))));
 
--- a/lib/resources/projects.js
+++ b/lib/resources/projects.js
@@ -36,11 +36,11 @@ module.exports = (service, endpoint) => {
     auth.canOrReject('project.create', Project.species)
       .then(() => Projects.create(Project.fromApi(body)))));
 
-  service.get('/projects/:id', endpoint(({ Projects }, { auth, params, queryOptions }) =>
+  service.get('/projects/:id', endpoint(({ Projects }, { auth, params, query, queryOptions }) =>
     Projects.getById(params.id, queryOptions)
       .then(getOrNotFound)
       .then((project) => auth.canOrReject('project.read', project))
-      .then((project) => ((queryOptions.extended === true || queryOptions.argData.includeVerbs === 'true')
+      .then((project) => ((queryOptions.extended === true || query.includeVerbs === 'true')
         ? auth.verbsOn(project).then((verbs) => Object.assign({ verbs }, project.forApi()))
         : project))));
 
--- a/test/integration/api/projects.js
+++ b/test/integration/api/projects.js
@@ -504,7 +504,7 @@ describe('api: /projects', () => {
                 ]);
               }))))));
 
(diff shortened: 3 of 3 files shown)


alxndrsn:
@garethbowen see failing tests for potential API for providing project information + verbs, but without dataset information.

garethbowen:
If it's not too much effort it would be good to check for the presence of the other extended fields too, ie:

```
  "appUsers": 4,
  "forms": 7,
  "lastSubmission": "2018-04-18T03:04:51.695Z",
  "datasets": 2,
  "lastEntity": "2023-04-18T03:04:51.695Z",
```

alxndrsn:
> If it's not too much effort it would be good to check for the presence of the other extended fields too

Added.

lib/resources/projects.js
```diff
@@ -40,7 +40,7 @@ module.exports = (service, endpoint) => {
     Projects.getById(params.id, queryOptions)
       .then(getOrNotFound)
       .then((project) => auth.canOrReject('project.read', project))
-      .then((project) => ((queryOptions.extended === true)
+      .then((project) => ((queryOptions.extended === true || queryOptions.argData.includeVerbs === 'true')
```
matthew-white: Not 100% sure about this, but my sense is that it's more common to use `query` rather than `queryOptions.argData`. E.g., the GET /projects endpoint above uses both `queryOptions` and `query`. That said, when I search the codebase, I do see examples of `queryOptions.argData` (especially for the `deleted` query parameter for some reason). So I think `queryOptions.argData` is valid as well. After all, you're just using it to do an equality check.

lib/resources/projects.js
matthew-white, line 1: From the PR description:

> confirm if using query parameters is idiomatic

I think so. :+1: I can't think of a more idiomatic pattern. E.g., the GET /v1/projects endpoint uses query parameters to decide whether to add a list of forms and/or datasets to the response.

Most endpoints don't offer fine-grained control over which fields to return. Usually it's just the extended metadata header. But I think query parameters are normal for where that does come up.

I also prefer a query parameter over adding an entirely new endpoint. :+1:

It'd definitely be good to document the query parameter though.

matthew-white, line 1: > confirm if chosen query parameter is acceptable

I kind of like `verbs=true` over `includeVerbs=true`. E.g., GET /v1/projects supports `forms=true` and `datasets=true`, not `includeForms=true` and `includeDatasets=true`. I don't have strong feelings about this though.

matthew-white (approved):

lib/resources/projects.js
alxndrsn: I didn't have a preference for `query` vs `queryOptions.argData`, but got to `argData` via introspection, because `queryOptions` was already in use in this function.

History implies that the intention of `argData` was that we shouldn't be using `query` at all:

https://github.com/getodk/central-backend/commit/c1f5cc02ba03649842eea963e6c1d511b7d8d904

```
commit c1f5cc02ba03649842eea963e6c1d511b7d8d904
Author: Issa Tseng <maintainer@heyissa.com>
Date:   Wed Jun 12 19:37:15 2019 -0700

    refactor: create QueryOptions arg input system.
    
    * the QueryOptions preprocessor now picks up all query args and puts
      them on the queryOptions object automatically.
    * BUT they are not actually enabled by default. they are put in an inert
      holding bucket. only after queryOptions.allowArgs(...someArgs) is
      called are they accepted as valid input.
    * this is so that the API resource layer has a say in which parameters
      should apply at which times.
...
```

OTOH, I haven't called `allowArgs(...)` at all, and it's not used much across the repo, so maybe this is all wildly out of date?

I **think** the intention is that `.argData` is not used directly; the first direct use is 4 years later in https://github.com/getodk/central-backend/commit/376e2bb3ea4894531138d8989d166b43a42632e5.  Maybe direct access should be disabled, and either `.allowArgs(...).args` and then `.args` should be used, or `query` used directly?

lib/resources/projects.js
alxndrsn: :+1: switched to `.verbs`

lib/resources/projects.js
alxndrsn: I've switched to using `query.verbs`; seems to work fine.

alxndrsn:
@matthew-white your suggestions have been included, and docs have been updated.  Please review.

lib/resources/projects.js
matthew-white: > I **think** the intention is that `.argData` is not used directly

That's my guess as well. I don't have a concrete memory along those lines though.

> the first direct use is 4 years later

It does stand out to me that direct access to `argData` is pretty limited in the codebase. Whether or not it's an anti-pattern, it's certainly not a common pattern.

> Maybe direct access should be disabled, and either `.allowArgs(...).args` and then `.args` should be used, or `query` used directly?

I like this idea. :+1:

- If you're trying to sanitize `argData` before passing it to something that expects an object: `.allowArgs()` is probably the tool for the job.
- If you're doing a simple check of a single query parameter: `query` is probably good enough.
- In neither case is direct access to `argData` needed.

Definitely feel free to file an issue along these lines.

docs/api.yaml
```diff
@@ -2640,6 +2640,12 @@ paths:
         schema:
           type: number
         example: "16"
+      - name: verbs
+        in: query
+        description: If set to `true`, will return allowed verbs for the current user on this project.  When extended metadata is requested, this parameter is ignored, and verbs are always included.
```
matthew-white: I would say "actor" instead of "user", since e.g. app users could technically retrieve this information. So maybe "authenticated actor" instead of "current user".

matthew-white (approved):

docs/api.yaml
matthew-white: I guess app users are arguably a kind of user, so not the best example. 😅 I just think that "actor" will make it clear that it's not limited to web users. Public link tokens could also be used to retrieve this information.

matthew-white: > Public link tokens could also be used to retrieve this information.

Though it'd be strange for a public link token to grant project-level access. 🤔 I'm not sure that public links even get the `project.read` verb by default. Anyway, I think I still slightly prefer "actor", since it's more generic, but I guess in practice, it probably is just web users.

lib/resources/projects.js
alxndrsn: :+1: Filed at https://github.com/getodk/central/issues/2152.

docs/api.yaml
alxndrsn: Fixed.

alxndrsn:
@garethbowen please confirm if this works when integrated with frontend.

garethbowen (approved):
Verified this works as needed for web-forms. NB I only tested functionality not performance.

PR #1918, opened 2026-08-21
Add logging with count to db migration to add blob size

### LINKED ISSUE #2135: Log details in blob file size migration
Add logging to migration in https://github.com/getodk/central-backend/pull/1877

ktuite:
Closes https://github.com/getodk/central/issues/2135

Running it.

This migration counts the number of blobs (with content) and console logs that. We discussed just logging that the migration was running, but thought the count info would be useful.

Very old db migrations using knex relied on the knex-specific output, but this migration just uses `db.raw`. The result from `raw` is direct from the database so it should still be usable even if we switch away from knex in the future. 

Provides more information about potentially long-running migrations. 

No.


lognaturel (approved):
Message looks good, thanks!

PR #1919, opened 2026-08-26
Check for case-insensitive uniqueness on actor props

### LINKED ISSUE #2161: Custom properties that only differ on letter case
Central doesn't allow users to create two entity properties that only differ on letter case (example: `foo` and `FOO`). However, right now it allows users to create two custom properties on a project that only differ on letter case. This issue is to prevent the creation of custom properties like that.

Wherever the user can create custom properties on a project, Central should check for an existing property that is the same except for letter case:

- Custom properties page
- App user create modal
- Public link create modal

ktuite:
Closes https://github.com/getodk/central/issues/2161 

* Renames problem `409.17` from `propertyNameConflict` to `propertyNameConflictInForm` in the code to make it specific to property name conflicts within the form upload/publish flow.
* Introduces a new problem `409.24` called `uniquenessViolationCaseMismatch`
* Updates `ActorProperties.create` function and SQL to be able to check for case-insensitive duplicates and return either the existing `409.3` (exact match) problem or the new `409.24` problem.
* Updates the code for creating Dataset properties via the API to be a little more self-contained, use a pattern from the actor-properties query, and to use the new `409.24` problem when in the case-mismatch duplicate situation.
* Updates dataset property API tests to distinguish between `409.3` and `409.24`
* Adds more dataset property API tests that use draft properties and deleted properties and ensure the right problem is returned. 

Tests.

Discussed a lot in code review! When tying to reuse the 409.17 problem (also about properties that match but not on case) we realized how much that is tied into the form+dataset flow. Decided to introduce a special problem for the case-mismatch situation that can be used for actor properties and dataset properties and isn't too tied to properties. 

Prevents users from naming properties `FOO` and `Foo` to avoid future confusion. Will use the casing of the first property entry. 

Small docs change made.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -2266,7 +2266,7 @@ paths:
       description: |-
         Registers a new actor property name for this project. Once registered, the property can be set on any App User or Public Link within the project via `PATCH`.
 
-        Property names follow the same rules as Entity property names: they must be non-empty, may contain letters, digits, hyphens, and underscores, and the name `displayName` (case-insensitive) is reserved.
+        Property names follow the same rules as Entity property names: they must be non-empty, may contain letters, digits, hyphens, and underscores, and the name `displayName` (case-insensitive) is reserved. Property names must be unique within a project and the uniqueness check is case-insensitive, so `region` and `Region` cannot coexist.
 
         Requires `actor_property.update` permission.
       operationId: createActorProperty
--- a/lib/model/query/actor-properties.js
+++ b/lib/model/query/actor-properties.js
@@ -2,15 +2,25 @@ const { sql } = require('slonik');
 const { map } = require('ramda');
 const { ActorProperty } = require('../frames');
 const { construct } = require('../../util/util');
+const Problem = require('../../util/problem');
 
 
 // Creates a new actor property name for a project.
-const create = (project, name) => ({ one }) =>
-  one(sql`
+// Enforce case-insensitive uniqueness without a separate round-trip.
+const create = (project, name) => ({ maybeOne }) =>
+  maybeOne(sql`
 INSERT INTO actor_properties ("projectId", "name")
-VALUES (${project.id}, ${name})
+SELECT ${project.id}, ${name}
+WHERE NOT EXISTS (
+  SELECT 1 FROM actor_properties
+  WHERE lower(name) = lower(${name}) AND "projectId" = ${project.id}
+)
 RETURNING *`)
-    .then(construct(ActorProperty));
+    .then((result) => {
+      if (!result.isDefined())
+        throw Problem.user.uniquenessViolation({ table: 'actor_properties', fields: [ 'name', 'projectId' ], values: [ name, project.id ] });
+      return construct(ActorProperty)(result.get());
+    });
 
 create.audit = (project, name) => (log) => log('actor_property.create', project, { name });
 
--- a/lib/model/query/actor-properties.js
+++ b/lib/model/query/actor-properties.js
@@ -19,7 +19,7 @@ const create = (project, name) => async ({ maybeOne, one }) => {
 
   // Returns a special problem if name conflicts with existing name on casing
   if (conflict.isDefined())
-    throw Problem.user.propertyNameConflict({ duplicateProperties: [{ current: conflict.get().name, provided: name }] });
+    throw Problem.user.uniquenessViolationCaseMismatch({ field: 'name', current: conflict.get().name, provided: name });
 
   // Could return a uniqueness violation if name matches exactly.
   return one(sql`
--- a/lib/model/query/actor-properties.js
+++ b/lib/model/query/actor-properties.js
@@ -6,21 +6,28 @@ const Problem = require('../../util/problem');
 
 
 // Creates a new actor property name for a project.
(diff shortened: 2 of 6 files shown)


ktuite:
After discussion in the team meeting, I used a special problem for when a property name conflicts with an existing name but with a different capitalization. 

This problem existed already but was only used in the context of datasets and properties created through forms. I generalized the wording of that problem and then used it for both actor properties and dataset properties. There was already a check for this kind of duplicate name in datasets, but it returned the regular uniqueness constraint violation.

lib/model/query/datasets.js
```diff
@@ -214,12 +214,13 @@ const _getDuplicatePropertyName = (property) => sql`
     AND "deletedAt" IS NULL
 `;
```
ktuite: If we add `name != ${property.name}` can this function be more explicitly about checking for duplicate names where the case does NOT match? That would match the new actor property check above and make the following code less confusing. Is this function used other places?

ktuite: Hopefully existing tests cover combinations of properties in various states of: draft, deleted and recreated, case mismatch... 

What if you create "region" and then delete it and then add "REGION" via API? 

Or what if you care "region", delete it, make a draft form with "REGION" and then add "REGION" via API? That should work.. what about adding "Region" then?

Basically add more tests where forms and API calls interact more.

lib/model/query/datasets.js
```diff
@@ -214,12 +214,13 @@ const _getDuplicatePropertyName = (property) => sql`
...
 
-// eslint-disable-next-line no-unused-vars
 const createPublishedProperty = (property, dataset) => async ({ all, maybeOne }) => {
 
   const duplicateProperty = await maybeOne(_getDuplicatePropertyName(property));
   if (duplicateProperty.isDefined()) {
-    throw Problem.user.uniquenessViolation({ table: 'ds_properties', fields: [ 'name', 'datasetId' ], values: [ duplicateProperty.get().name, dataset.id ] });
+    const currentPropertyName = duplicateProperty.get().name;
+    if (currentPropertyName !== property.name)
```
ktuite: A possible alternative is to add an `else` here and throw the 409.3 problem again.

lib/util/problem.js
```diff
@@ -250,7 +250,10 @@ const problems = {
 
     datasetNameConflict: problem(409.16, ({ current, provided }) => `A dataset named '${current}' exists and you provided '${provided}' with the same name but different capitalization.`),
 
-    propertyNameConflict: problem(409.17, () => 'This form attempts to create new Entity properties that match with existing ones except for capitalization.'),
+    propertyNameConflict: problem(409.17, ({ duplicateProperties }) => {
```
ktuite: Renaming to `formPropertyNameConflict` or something and then adding a new problem (`409.24`) that is specifically about name case-insensitive mismatch that has nothing to do with forms or properties! It can still return current and provided.

lib/model/query/actor-properties.js
```diff
@@ -2,15 +2,32 @@ const { sql } = require('slonik');
...
+  const conflict = await maybeOne(sql`
+    SELECT name FROM actor_properties
+    WHERE lower(name) = lower(${name})
+    AND name != ${name}
+    AND "projectId" = ${project.id}
+    LIMIT 1`);
+
+  // Returns a special problem if name conflicts with existing name on casing
+  if (conflict.isDefined())
+    throw Problem.user.propertyNameConflict({ duplicateProperties: [{ current: conflict.get().name, provided: name }] });
```
ktuite: Use 409.24 here

test/integration/api/datasets.js
```diff
@@ -529,8 +529,8 @@ describe('datasets and entities', () => {
           })
           .expect(409)
           .then(({ body }) => {
-            body.code.should.equal(409.3);
-            body.message.should.match(/A resource already exists with name,datasetId/);
+            body.code.should.equal(409.17);
```
ktuite: Will become 409.24
Add more tests

test/integration/api/datasets.js
```diff
@@ -4428,7 +4428,7 @@ describe('datasets and entities', () => {
             .set('Content-Type', 'application/xml')
             .expect(409)
             .then(({ body }) => {
-              body.message.should.match(/This form attempts to create new Entity properties that match with existing ones except for capitalization/);
```
ktuite, line 4431: revert the changes for just forms.

ktuite (commented):

matthew-white (approved):

ktuite:
Another note about frontend dataset property creation...

Check the 409 error messages on dataset property creation. Add a new message that gets translated when it sees the new non 409.3 problem.

docs/api.yaml
matthew-white, line 1: How about adding this to the API changelog as well?

matthew-white: (Random thought I just had)

ktuite: We document very few of the 409s already so I don't think it's necessary...

matthew-white: But this fixes a user-facing bug, right? I think that's worth documenting in the changelog. Otherwise this change won't appear anywhere in release notes.

PR #1920, opened 2026-08-26
Add "viewAs" query param to entities CSV download and geojson endpoints

### LINKED ISSUE #1792: Add "View as [app user]" filter to Entities

### LINKED ISSUE #2162: Add "viewAs" filter to entities CSV download and geojson endpoints
This other [issue](https://github.com/getodk/central/issues/1879) called for adding `?viewAs=<actor_id>` to the entities OData and REST endpoints but it didn't capture some other places that need the filter.

The `viewAs` filter needs to be added to
- entity csv download
- entities geojson endpoint

The entities CSV download also needs to be filtered by this "viewAs" query param so it can support downloading a subset of the entities. For example, when a search filter and a "view as" filter is specified, the download button should count the number of matching entities and download only those.
Image

The map view should also only show the entities matching the "viewAs" filter.

ktuite:
Closes https://github.com/getodk/central/issues/2162

Adds `viewAs` query parameter to entities csv download and entities geojson endpoint. These let Frontend use this viewAs filter when showing entities on map or preparing a segmented CSV download. 

Tests. Will be testing it with a frontend branch making use of it, too.

I think this approach is what's needed.

Frontend filters work better! 

Should probably come back and add docs.

### DIFF
--- a/lib/resources/datasets.js
+++ b/lib/resources/datasets.js
@@ -181,7 +181,7 @@ module.exports = (service, endpoint) => {
     return success;
   }));
 
-  service.get('/projects/:projectId/datasets/:name/entities.csv', endpoint(async ({ Datasets, Entities, Projects }, { params, auth, query }, request, response) => {
+  service.get('/projects/:projectId/datasets/:name/entities.csv', endpoint(async ({ Auth, Actors, Datasets, Entities, Projects }, { params, auth, query }, request, response) => {
     const project = await Projects.getById(params.projectId).then(getOrNotFound);
     await auth.canOrReject('entity.list', project);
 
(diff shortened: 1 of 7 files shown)


lib/resources/datasets.js
```diff
@@ -190,16 +190,23 @@ module.exports = (service, endpoint) => {
     const dataset = await Datasets.get(params.projectId, params.name, true, true).then(getOrNotFound);
     const properties = await Datasets.getProperties(dataset.id);
 
+    let filterActorId = null;
+    if (query.viewAs != null) {
+      const actor = await Actors.getById(parseInt(query.viewAs, 10)).then(getOrNotFound);
```
sadiqkhoja, line 195: I thinking if we should really throw 404 if actor is not found, maybe it would be better to return empty list; in other words we would be saying that provided actorId can't see anything.

this applies to other endpoints as well.

sadiqkhoja (approved):

lib/resources/datasets.js
ktuite: I was wondering about that, too. There is some code on the frontend that gets the viewAs actor ID from the URL and sends it to backend, possibly before it has a chance to check if it's in the list of existing actors that frontend has, and it seems weird for it to 404 when it could just return an empty list. 

I will change this so it doesn't 404 when the actor is not found.

ktuite: Actually, because I need to get the Actor from the database to check its verbs, it's more work to identify an invalid actor ID and then either return a custom empty result, or pass the not-valid actor ID along to go through the database query, than it is to just 404 on the actor.

```
const targetCanList = await Auth.can(actor, 'entity.list', dataset);
```

I'm not going to make that change after all and just leave it as returning 404 for a missing actor (for all the endpoints where `viewAs` can be provided.)

PR #1922, opened 2026-08-30
client-audits: order exported entries by start

### LINKED ISSUE #1243: Order client audit log entries consistently in export
The following test fails infrequently. I've seen this failure locally, and it's also been seen on CircleCI:

```
  1) api: /forms/:id/submissions
       .csv.zip GET
         should return worker-processed consolidated client audit log attachments:

      AssertionError: expected 'instance ID,event,node,start,end,latitude,longitude,accuracy,old-value,new-value\none,e,/data/e,2000-01-01T00:11,,,,,hh,ii\none,d,/data/d,2000-01-01T00:10,,10,11,12,gg,\none,c,/data/c,2000-01-01T00:03,2000-01-01T00:04,7,8,9,ee,ff\none,b,/data/b,2000-01-01T00:02,2000-01-01T00:03,4,5,6,cc,dd\none,a,/data/a,2000-01-01T00:01,2000-01-01T00:02,1,2,3,aa,bb\ntwo,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff\ntwo,g,/data/g,2000-01-01T00:05,2000-01-01T00:06,-3,-4,,cc,dd\ntwo,f,/data/f,2000-01-01T00:04,2000-01-01T00:05,-1,-2,,aa,bb\n' to be 'instance ID,event,node,start,end,latitude,longitude,accuracy,old-value,new-value\none,a,/data/a,2000-01-01T00:01,2000-01-01T00:02,1,2,3,aa,bb\none,b,/data/b,2000-01-01T00:02,2000-01-01T00:03,4,5,6,cc,dd\none,c,/data/c,2000-01-01T00:03,2000-01-01T00:04,7,8,9,ee,ff\none,d,/data/d,2000-01-01T00:10,,10,11,12,gg,\none,e,/data/e,2000-01-01T00:11,,,,,hh,ii\ntwo,f,/data/f,2000-01-01T00:04,2000-01-01T00:05,-1,-2,,aa,bb\ntwo,g,/data/g,2000-01-01T00:05,2000-01-01T00:06,-3,-4,,cc,dd\ntwo,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff\n'
      + expected - actual

       instance ID,event,node,start,end,latitude,longitude,accuracy,old-value,new-value
      +one,a,/data/a,2000-01-01T00:01,2000-01-01T00:02,1,2,3,aa,bb
      +one,b,/data/b,2000-01-01T00:02,2000-01-01T00:03,4,5,6,cc,dd
      +one,c,/data/c,2000-01-01T00:03,2000-01-01T00:04,7,8,9,ee,ff
      +one,d,/data/d,2000-01-01T00:10,,10,11,12,gg,
       one,e,/data/e,2000-01-01T00:11,,,,,hh,ii
      -one,d,/data/d,2000-01-01T00:10,,10,11,12,gg,
      -one,c,/data/c,2000-01-01T00:03,2000-01-01T00:04,7,8,9,ee,ff
      -one,b,/data/b,2000-01-01T00:02,2000-01-01T00:03,4,5,6,cc,dd
      -one,a,/data/a,2000-01-01T00:01,2000-01-01T00:02,1,2,3,aa,bb
      +two,f,/data/f,2000-01-01T00:04,2000-01-01T00:05,-1,-2,,aa,bb
      +two,g,/data/g,2000-01-01T00:05,2000-01-01T00:06,-3,-4,,cc,dd
       two,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff
      -two,g,/data/g,2000-01-01T00:05,2000-01-01T00:06,-3,-4,,cc,dd
      -two,f,/data/f,2000-01-01T00:04,2000-01-01T00:05,-1,-2,,aa,bb

      at Assertion.fail (node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
      at test/integration/api/submissions.js:1799:51
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at async Object.createConnection (node_modules/slonik/dist/src/factories/createConnection.js:96:18)
```

I ran `make test` on CircleCI [200 times](https://github.com/getodk/central-backend/issues/588#issuecomment-1235669602), and this error came up just once (in [this](https://app.circleci.com/pipelines/github/getodk/central-backend/943/workflows/223477a5-ee43-4c3a-8964-1e37893a206a/jobs/1696/artifacts) build). The test failure is infrequent, but since it's been seen on a variety of machines, I thought I'd go ahead and file an issue. I think it's interesting that it only seems to be this test that has this ordering issue.

dylanpulver:
Closes getodk/central#1243

The export orders only by `submission_defs."createdAt", id`. Every audit row of a submission shares that key, so their order relative to each other is unconstrained and follows physical row order, which an UPDATE or VACUUM FULL reshuffles. That is why it has been intermittent since 2022 rather than constant.

It also shows up for users: an attachment the worker has not consolidated yet is parsed inline and always comes out in file order, so one submission can export in two different orders before and after processing.

Submit, `exhaust`, rewrite the rows in reverse physical order, export. Master returns `e,d,c,b,a`, the ordering from the issue. Patched returns `a,b,c,d,e`. The fixture uses three different orderings for file position, event name and `start`, so it also fails if the sort key is `event` or `node`.

`test/integration/api/submissions.js`, same machine and session: baseline 296 passing, patched 297, the extra being this test. `make lint` clean.

This is what @matthew-white described on the issue. `start` is `text` holding epoch milliseconds, so it sorts lexicographically. Every value from 2001-09-09 to 2286 is 13 digits, and over 200k realistic timestamps lexicographic and numeric order agreed exactly. Casting to a number is not safe, since non-numeric values already exist in this column.

#### How does this change impact users? What are the regression risks?

**Correcting this description: an earlier version said the plan and sort method were unchanged. That was wrong, and @alxndrsn is right to flag the cost.** On PostgreSQL 17.9 with 1M audit rows, master rides `submission_defs_createdat_id_index` and performs no sort at all (~480ms). With the added key the planner switches to hash joins plus an external merge sort (~1810ms), about 3.8x. An index on `client_audits("blobId", start)` is ignored by the planner, and raising `work_mem` makes it worse, so it is not tunable away.

Sorting per submission in the application keeps the fast index-ordered plan. Rows for one submission arrive contiguously and the blob path already buffers them, so it adds no new memory profile. Measured ~11ms against the ~1330ms the database sort adds. Happy to move it there, or to drop this in favour of relaxing the test expectations if you would rather not fix the ordering at all.

Not fixed: entries sharing an identical `start` are still unordered.

#### Does this change require updates to the API documentation?

No.

---

AI usage: this PR was prepared with Claude Code. The repro, the mutation checks and the timings were run against a local build rather than asserted.

### DIFF
--- a/lib/model/query/client-audits.js
+++ b/lib/model/query/client-audits.js
@@ -35,7 +35,7 @@ select client_audits.*, blobs.id AS "blobId", blobs.s3_status, blobs.content, bl
   where ${odataFilter(options.filter, odataToColumnMap)}
     and current=true
     and (form_defs."keyId" is null or ${sqlInArray(sql`form_defs."keyId"`, keyIds)})
-  order by submission_defs."createdAt" asc, submission_defs.id asc`)
+  order by submission_defs."createdAt" asc, submission_defs.id asc, client_audits.start asc`)
   .then(dbStream => streamBlobs(s3, dbStream));
 
 module.exports = { createMany, existsForBlob, streamForExport };
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -2441,6 +2441,49 @@ two,h,/data/h,2000-01-01T00:06,2000-01-01T00:07,-5,-6,,ee,ff,,
           .then(() => container.oneFirst(sql`select count(*) from client_audits`)
             .then((count) => { count.should.equal(8); })))));
 
+    // The event names, the order of the rows in the file, and the start
+    // timestamps are all deliberately different orderings here, so that this
+    // test can only pass if the export is actually ordered by start.
+    const unorderedAudit = `event,node,start,end
+a,/data/a,1700000000003,1700000000004
+b,/data/b,1700000000005,1700000000006
+c,/data/c,1700000000001,1700000000002
+d,/data/d,1700000000009,1700000000010
+e,/data/e,1700000000007,1700000000008
+`;
+
+    it('should order client audit log entries by start, whatever order the rows are stored in', testService(async (service, container) => {
+      const asAlice = await service.login('alice');
+      await asAlice.post('/v1/projects/1/forms?publish=true')
+        .set('Content-Type', 'application/xml')
+        .send(testData.forms.clientAudits)
+        .expect(200);
+      await asAlice.post('/v1/projects/1/submission')
+        .set('X-OpenRosa-Version', '1.0')
+        .attach('audit.csv', Buffer.from(unorderedAudit), { filename: 'audit.csv' })
+        .attach('xml_submission_file', Buffer.from(testData.instances.clientAudits.one), { filename: 'data.xml' })
+        .expect(201);
+      await exhaust(container);
+
+      // Rewrite the rows so that their physical order in the table is the
+      // reverse of the order they were inserted in. Any UPDATE, VACUUM FULL or
+      // restore-from-dump can do this, which is why the export order was not
+      // reproducible before.
+      await container.run(sql`
+        with removed as (delete from client_audits returning *)
+        insert into client_audits ("blobId", event, node, start, "end")
+          select "blobId", event, node, start, "end" from removed order by start desc`);
+
+      const result = await httpZipResponseToFiles(asAlice.get('/v1/projects/1/forms/audits/submissions.csv.zip'));
+      result.files.get('audits - audit.csv').should.equal(`instance ID,event,node,start,end,latitude,longitude,accuracy,old-value,new-value,user,change-reason
+one,c,/data/c,1700000000001,1700000000002,,,,,,,
+one,a,/data/a,1700000000003,1700000000004,,,,,,,
+one,b,/data/b,1700000000005,1700000000006,,,,,,,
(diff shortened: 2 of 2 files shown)


lib/model/query/client-audits.js
```diff
@@ -35,7 +35,7 @@ select client_audits.*, blobs.id AS "blobId", blobs.s3_status, blobs.content, bl
   where ${odataFilter(options.filter, odataToColumnMap)}
     and current=true
     and (form_defs."keyId" is null or ${sqlInArray(sql`form_defs."keyId"`, keyIds)})
-  order by submission_defs."createdAt" asc, submission_defs.id asc`)
+  order by submission_defs."createdAt" asc, submission_defs.id asc, client_audits.start asc`)
```
alxndrsn, line 38: This change has performance implications.  If ordering is not currently guaranteed, an alternative approach might be to relax the test expectations.

PR #1923, opened 2026-08-31
docker-postgres: extract make targets as script

alxndrsn:
* simplifies and unifies massive make targets
* prepares for fixing bug with waiting (https://github.com/getodk/central/issues/2141)

- [x] tested locally
- [x] tested locally with ssl
- [x] ci

* make syntax is even more arcane than bash
* the ssl vs. non-ssl scripts were similar, but hard to share implementations in make

No impact - just dev/test code.

No.

### DIFF
--- a/test/bin/docker-postgres.sh
+++ b/test/bin/docker-postgres.sh
@@ -3,7 +3,7 @@ set -e
 set -u
 set -o pipefail
 
-log() { echo >&2 "[start-postgres] $*"; }
+log() { echo >&2 "[docker-postgres] $*"; }
 
 imageName=odk-central-backend-dev-postgres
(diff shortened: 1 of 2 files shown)


test/bin/docker-postgres.sh
```diff
@@ -0,0 +1,94 @@
...
+  sudo -k
+
+  log "SSL certs setup completed OK."
+fi
+
+log "Attempting to start existing container..."
+if docker start "$imageName"; then
+  log "(Re)started OK."
+  exit
+fi
```
sadiqkhoja: it would be nice to check existence of database and create it if not there.

sadiqkhoja (approved):

PR #1926, opened 2026-09-01
Check that dataset and actor properties exist for access filter

### LINKED ISSUE #1972: Improve dataset access filter config API
Some comments from reviewing https://github.com/getodk/central-backend/pull/1849 were suggestions for improvements to the API:

1. Return 400 Bad Request if client sends in both accessFilter and ownerOnly

2. Confirm existing of properties in the SQL or here? Throw error if property doesn't exist? Currently, these requests return 200 but don't set the filter.

```js
it.only('should reject if property is not valid', testService(async (service) => {
  await createDataset(asAlice, 1, 'trees', ['region']);
  await asAlice.post('/v1/projects/1/actor-properties').send({ name: 'region' }).expect(200);

  await asAlice.patch('/v1/projects/1/datasets/trees')
    .send({ accessFilter: { type: 'property', rules: [{ datasetProperty: 'region', actorProperty: 'doesNotExist' }] } })
    .expect(400);

  await asAlice.patch('/v1/projects/1/datasets/trees')
    .send({ accessFilter: { type: 'property', rules: [{ datasetProperty: 'doesNotExist', actorProperty: 'region' }] } })
    .expect(400);
}));
```

ktuite:
Closes https://github.com/getodk/central/issues/1972

Checks the provided dataset and actor properties (from the access filter rules) against actual dataset properties and actor properties without just skipping over them. 

Test.

Adds an extra check of property existence within the Dataset.setAccessFilters query. 

Will give a better error if a user provides (via api) a property that doesn't exist instead of silently accepting and ignoring it. 

We don't document these failure/problem cases that much in the API docs so I don't think it's necessary to add.

### DIFF
--- a/lib/model/query/datasets.js
+++ b/lib/model/query/datasets.js
@@ -1079,23 +1079,19 @@ const getListWithProperties = (projectId, ids) => async ({ all }) => {
 
 // Adds or updates a single access filter rule for a dataset.
 // datasetPropertyName and userPropertyName are names (strings).
-const setAccessFilter = (datasetId, datasetPropertyName, userPropertyName) => ({ maybeOne }) =>
-  maybeOne(sql`
-WITH ds_prop AS (
-  SELECT id FROM ds_properties
-  WHERE "datasetId" = ${datasetId} AND "name" = ${datasetPropertyName} AND "publishedAt" IS NOT NULL
-), actor_prop AS (
-  SELECT ap.id FROM actor_properties ap
-  JOIN datasets d ON d."projectId" = ap."projectId"
-  WHERE d.id = ${datasetId} AND ap."name" = ${userPropertyName}
-), inserted AS (
-  INSERT INTO dataset_access_filters ("datasetId", "datasetPropertyId", "actorPropertyId")
-  SELECT ${datasetId}, ds_prop.id, actor_prop.id FROM ds_prop, actor_prop
-  ON CONFLICT ("datasetPropertyId", "actorPropertyId") DO NOTHING
-  RETURNING *
-)
-SELECT * FROM inserted`)
-    .then((rows) => rows.map((row) => new Dataset.AccessFilter(row)));
+// If a non-existent actor or dataset property is passed in, this will silently do nothing.
+// This function is called from setAccessFilters, which pre-checks that properties exist.
+const _setAccessFilter = (datasetId, datasetPropertyName, userPropertyName) => ({ run }) =>
+  run(sql`
+INSERT INTO dataset_access_filters ("datasetId", "datasetPropertyId", "actorPropertyId")
+SELECT ${datasetId}, dp.id, ap.id
+FROM ds_properties dp
+JOIN actor_properties ap ON ap."projectId" = (SELECT "projectId" FROM datasets WHERE id = ${datasetId})
+WHERE dp."datasetId" = ${datasetId}
+  AND dp."name" = ${datasetPropertyName}
+  AND dp."publishedAt" IS NOT NULL
+  AND ap."name" = ${userPropertyName}
+ON CONFLICT ("datasetPropertyId", "actorPropertyId") DO NOTHING`);
 
 // Returns all access filter rules for a dataset as DatasetAccessFilter instances.
 const getAccessFilter = (datasetId) => ({ all }) =>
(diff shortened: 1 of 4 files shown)


lib/model/query/datasets.js
```diff
@@ -1116,11 +1116,19 @@ const deleteAllAccessFilters = (datasetId) => ({ run }) =>
...
 // Passing an empty array deletes all rules.
-const setAccessFilters = (datasetId, rules) => async ({ Datasets }) => {
-  await Datasets.deleteAllAccessFilters(datasetId);
+const setAccessFilters = (dataset, rules) => async ({ ActorProperties, Datasets }) => {
+  const datasetPropertyNames = new Set((await Datasets.getProperties(dataset.id)).map(({ name }) => name));
+  const actorPropertyNames = new Set((await ActorProperties.getAllForProject(dataset.projectId)).map(({ name }) => name));
+
+  for (const { datasetProperty, actorProperty } of rules) {
+    if (!datasetPropertyNames.has(datasetProperty) || !actorPropertyNames.has(actorProperty))
+      throw Problem.user.accessFilterError({ reason: 'Access filter references a property that does not exist.' });
```
sadiqkhoja: how about we provide more details in the error, something like this:

```
if (!datasetPropertyNames.has(datasetProperty))
    throw Problem.user.accessFilterError({ reason: `Dataset property '${datasetProperty}' does not exist.` });
  if (!actorPropertyNames.has(actorProperty))
    throw Problem.user.accessFilterError({ reason: `Actor property '${actorProperty}' does not exist.` });
```

or we collect all the errors in case both properties are missing or multiple rules are provided. That would help API client to know all the problems in one go.

This can certainly be a followup improvement.

lib/model/query/datasets.js
```diff
@@ -1116,11 +1116,19 @@ const deleteAllAccessFilters = (datasetId) => ({ run }) =>
 // Atomically replaces all access filter rules for a dataset.
 // rules is an array of { datasetProperty, actorProperty } objects.
 // Passing an empty array deletes all rules.
-const setAccessFilters = (datasetId, rules) => async ({ Datasets }) => {
-  await Datasets.deleteAllAccessFilters(datasetId);
+const setAccessFilters = (dataset, rules) => async ({ ActorProperties, Datasets }) => {
+  const datasetPropertyNames = new Set((await Datasets.getProperties(dataset.id)).map(({ name }) => name));
+  const actorPropertyNames = new Set((await ActorProperties.getAllForProject(dataset.projectId)).map(({ name }) => name));
```
sadiqkhoja: I would put these two calls in `Promise.all`

lib/model/query/datasets.js
```diff
@@ -1116,11 +1116,19 @@ const deleteAllAccessFilters = (datasetId) => ({ run }) =>
...
+  for (const { datasetProperty, actorProperty } of rules) {
+    if (!datasetPropertyNames.has(datasetProperty) || !actorPropertyNames.has(actorProperty))
+      throw Problem.user.accessFilterError({ reason: 'Access filter references a property that does not exist.' });
+  }
+
+  await Datasets.deleteAllAccessFilters(dataset.id);
   for (const { datasetProperty, actorProperty } of rules) {
     // eslint-disable-next-line no-await-in-loop
-    await Datasets.setAccessFilter(datasetId, datasetProperty, actorProperty);
+    await Datasets.setAccessFilter(dataset.id, datasetProperty, actorProperty);
```
sadiqkhoja: wondering if we should lock the properties when we fetch them for validation. It is theoretically possible that by the time `setAccessFilter` is executed those properties are deleted. 

This can also be later enhancement.

lib/model/query/datasets.js
sadiqkhoja: if `ds_prop.id` or `actor_prop.id` is null then does database throw exception since "datasetPropertyId" and "actorPropertyId" are not-null? if yes then can we leverage existing `postgresErrorToProblem` to return appropriate error response?

sadiqkhoja (approved):

lib/model/query/datasets.js
ktuite: I updated the query to be a bit clearer but basically no, the "SELECT ... FROM" will have zero rows if a property doesn't exist and it will not throw an exception but also not insert anything.

PR #1930, opened 2026-09-04
Remove request body properly

alxndrsn:
Instead of removing request bodies in `sanitizeEventRequest()`, use the standard configuration of the Sentry HTTP integration to achieve the same result.

- [ ] existing tests which still pass after change, but didn't when stripping was removed from `sanitizeEventRequest()` but not added to `HttpIntegration` config?
- [ ] new tests which failed when stripping was removed from `sanitizeEventRequest()` but not added to `HttpIntegration` config?

Uses the official approach, so should also apply to other event sources beyond errors.

See docs for `beforeSend()`: https://docs.sentry.io/platforms/javascript/guides/node/configuration/options/#beforeSend

Vs:

* ...TODO...

No expected effect, except better privacy and perhaps occasional improved performance.

No.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -102,10 +102,6 @@ const sanitizeEventRequest = (event) => {
     if (event.request.headers && event.request.headers['x-action-notes'])
       event.request.headers['x-action-notes'] = null;
 
-    // clear out request body data for all endpoints
-    if (event.request.data)
-      event.request.data = null;
-
     // remove sensitive info from URL via query string or path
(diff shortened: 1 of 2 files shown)


alxndrsn:
As Sentry is mocked in integration tests, testing this in the `standard-e2e` suite with a mocked Sentry might be the most reliable approach.  However, in this case it may be hard to trigger unhandled errors which are forwarded to Sentry.

alxndrsn:
Not sure this is correct.

PR #1931, opened 2026-09-04
sentry: exclude request bodies from breadcrumbs

alxndrsn:
See:

* https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/#modifying-default-integrations
* https://docs.sentry.io/platforms/javascript/guides/connect/configuration/integrations/http/#maxincomingrequestbodysize

Related:

* https://github.com/getodk/central-backend/pull/1928
* https://github.com/getodk/central-backend/pull/1929

Nothing - just read the docs #YOLO 🤘

Considered adding tests, but there are currently no tests in this repo for real Sentry output:

* integration tests run with a mocked/faked Sentry implementation.
* e2e tests do not make any assertions about Sentry calls

Hopefully no bad effect, but improved privacy and performance.

No.

### DIFF
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -191,7 +191,7 @@ const _init = (config) => { // eslint-disable-line no-shadow
     environment: process.env.SENTRY_ENVIRONMENT || 'production',
     integrations: defaults => [
       ...defaults.filter(i => !disabledIntegrations.includes(i.name)),
-      
+
       // Modified default integrations
       Sentry.httpIntegration({
         maxIncomingRequestBodySize: 'none',
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -192,6 +192,8 @@ const _init = (config) => { // eslint-disable-line no-shadow
     integrations: defaults => [
       ...defaults.filter(i => !disabledIntegrations.includes(i.name)),
 
+      Sentry.dedupeIntegration(),
+
       // Modified default integrations
       Sentry.httpIntegration({
         maxIncomingRequestBodySize: 'none',
--- a/lib/external/sentry.js
+++ b/lib/external/sentry.js
@@ -192,6 +192,14 @@ const _init = (config) => { // eslint-disable-line no-shadow
     integrations: defaults => [
       ...defaults.filter(i => !disabledIntegrations.includes(i.name)),
       Sentry.dedupeIntegration(),
+
+      // Modified default integrations:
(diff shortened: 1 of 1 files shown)


lib/external/sentry.js
```diff
@@ -192,6 +192,14 @@ const _init = (config) => { // eslint-disable-line no-shadow
     integrations: defaults => [
       ...defaults.filter(i => !disabledIntegrations.includes(i.name)),
       Sentry.dedupeIntegration(),
+
+      // Modified default integrations:
+      // See: https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/#modifying-default-integrations
+
+      // See: https://docs.sentry.io/platforms/javascript/guides/connect/configuration/integrations/http/
+      Sentry.httpIntegration({
+        ignoreIncomingRequestBody: 'none',
```
lognaturel: I think this conflates two different options. I don't see any evidence of `ignoreIncomingRequestBody` being able to be set to a string. It looks from https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/http/#ignoreincomingrequestbody that it takes a callback that returns a boolean so `ignoreIncomingRequestBody: () => true` seems like an option here.

But I think `maxIncomingRequestBodySize: 'none'` is clearer and probably what the intent was.

lognaturel (dismissed):

lognaturel:
@matthew-white and @alxndrsn are you ok with me pushing the change I proposed to this branch?

lognaturel:
@matthew-white I went ahead and made it so in case you have time for a quick review (or maybe @sadiqkhoja)? I read the docs and noticed that the dedupe integration is documented as being on by default. When I removed it explicitly, our test failed, though (nice!). So I brought it back. Could be a different library version, I didn't dig any deeper.

I removed some of the comments linking out to docs because the usage is pretty clear I think.

I spend a little bit of time seeing whether I could test that the option is set based on the existing tests for configured integrations. However, I found that there's no direct way to access the config for individual options so I backed out.

I changed the PR title because it used to say "for span and breadcrumbs." There was no attempt to change outgoing HTTP request logging (spans) and I think that was out of scope anyway. This focuses on incoming requests only.

matthew-white (approved):
LGTM. We can always adjust later if need be depending on what we see in Sentry.

lib/external/sentry.js
alxndrsn: Thanks for catching this :+1:

PR #1924, opened 2026-08-31
Wait for postgres properly

alxndrsn:
This PR contains work foundational to:

* https://github.com/getodk/central/issues/1385

---

Closes https://github.com/getodk/central/issues/2141
Closes https://github.com/getodk/central/issues/2194

#### What has been done to verify that this works as intended?

- [x] ran locally
- [x] ci
- [x] ran a few hundred times in ci without issue, with both pg14 and pg18
- [x] ran 2,000 times in ci without issue, using both pg14 and pg18
    - [x] https://github.com/getodk/central-backend/actions/runs/34943311766
    - [x] https://github.com/getodk/central-backend/actions/runs/34943311829
    - [x] https://github.com/getodk/central-backend/actions/runs/34945552174
    - [x] https://github.com/getodk/central-backend/actions/runs/34945552350
    - [x] https://github.com/getodk/central-backend/actions/runs/34945580935
    - [x] https://github.com/getodk/central-backend/actions/runs/34945580939
    - [x] https://github.com/getodk/central-backend/actions/runs/34945595783
    - [x] https://github.com/getodk/central-backend/actions/runs/34945596015
    - [x] https://github.com/getodk/central-backend/actions/runs/34945818909
    - [x] https://github.com/getodk/central-backend/actions/runs/34945818944

(Each run is 200 attempts.  Jobs/workflows were retried when failures were due to docker image download or docker API 502.)

#### Why is this the best possible solution? Were any other approaches considered?

Waits better than before.  Should be quick on quick machines, and slow on slower ones.

#### How does this change impact users? Describe intentional behavior changes from code updates. What are the regression risks?

No immediate effect, but this approach can (and likely should) be applied in the `central` repo to prevent issues for self-hosters.

#### Does this change require updates to the API documentation? If so, please update docs/api.yaml as part of this PR.

No.

PR #1932, opened 2026-09-04
ci: output postgres logs

alxndrsn:
Helpful for debugging workflow failures.

Used in https://github.com/getodk/central-backend/pull/1924 to help debug intermittent postgres startup issues.

Not including these logs until they are needed?  But in the case of https://github.com/getodk/central-backend/pull/1924, there were occasional failures which could not be recreated easily on demand.  This meant significant added time caused by configuring logs later, and then waiting for a failure to recur.

No effect - just CI.

No.


lognaturel (approved):
Consider adding `|| true` so it doesn't look like it failed if previous steps failed.

PR #1937, opened 2026-09-09
Additional API docs for Central 2026.3

### LINKED ISSUE #2166: Update API docs for v2026.3
This PR is to track changes that we need to make to the API docs for v2026.3. If a PR makes a change to the API and simultaneously changes the API docs, then that change doesn't need to be listed here.

⚠️ If there is a breaking change to the API, an issue should be filed with pyODK.

- [ ] https://github.com/getodk/central-backend/pull/1919#discussion_r3908251351

ktuite:
Closes https://github.com/getodk/central/issues/2166

The main API changes this release were
- adding `?viewAs=<actorId>` on entity endpoints to show filtered entity lists
- form attachment file size
- case-insensitive property uniqueness for actor properties
- audit log end range

Except for the last note in this PR, everything else was documented as part if its original PR.


matthew-white (approved):
The changes here look good to me. :+1:

PR #1938, opened 2026-09-09
Decode form version in submission XML

### LINKED ISSUE #2219: Error when sending a form with special character in form version
#### Problem description
Error when sending a form with special character in form version
#### URL of the page
https://test.getodk.cloud/projects/749/forms/styletest/submissions
#### Steps to reproduce the problem
1. Open form from URL above
2. Tap on New submission
3. Try to send a form
#### Screenshot

Image

#### Expected behavior
It should be possible to send a form with special character in form version
#### Central version shown in version.txt
versions:
bc46bbcb80bcb7a0cc04eb57aaeecd3590414671 (v2026.2.4-24-gbc46bbc)
 0000000000000000000000000000000000000000 client (v2026.3.0-beta.0)
 a61b0acbc7d5d8e3bc381523e78e108e60fddfdb server (v2026.2.0-67-ga61b0acb)
#### Browser version
Firefox 152.0, chrome 138.0.7204.168 
#### Around when did you see the problem (in UTC)?
8

matthew-white:
This PR closes getodk/central#2219. It updates `Submission.fromXml()` to decode the form `version` string found in the submission XML.

New tests.

I tried to do something similar to what was done in #550.

At some point, we could consider decoding more systematically. However, the goal of this PR is just to decode `version`.

I think it should be safe to merge this PR even though regression testing is in progress. It should only come up for users whose form `version` string contains encoded characters.

### DIFF
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -289,7 +289,7 @@ describe('api: /submission', () => {
     it('should decode the form version', testService(async (service) => {
       const asAlice = await service.login('alice');
 
-      // Adds a version string of <&> to XML after encoding it
+      // Adds a version string of <&@> to XML after encoding it
       const addVersion = (xml) =>
         xml.replace(/(<data id="simple")>/, '$1 version="&lt;&amp;&#64;&gt;">');
 
--- a/test/integration/api/submissions.js
+++ b/test/integration/api/submissions.js
@@ -286,10 +286,12 @@ describe('api: /submission', () => {
           .then(({ formDefId }) => one(sql`select * from form_defs where id=${formDefId}`))
           .then((formDef) => { formDef.version.should.equal('two'); }))));
 
+    // getodk/central#2219
     it('should decode the form version', testService(async (service) => {
       const asAlice = await service.login('alice');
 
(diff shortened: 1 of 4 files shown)


test/integration/api/submissions.js
```diff
@@ -286,6 +286,44 @@ describe('api: /submission', () => {
...
+
+      // Create a submission to the new version of the form.
+      await asAlice.post('/v1/projects/1/submission')
+        .set('X-OpenRosa-Version', '1.0')
+        .attach(
+          'xml_submission_file',
+          Buffer.from(addVersion(testData.instances.simple.one)),
+          { filename: 'data.xml' }
+        )
+        .expect(201);
```
matthew-white, line 319: Without the change to lib/, this test fails here: a 404 is returned instead.

matthew-white:
I'm tagging multiple people for review, but I'm only looking for a single checkmark. @sadiqkhoja wrote #550, so he would be a good person to review. However, I think @lognaturel could review as well if we're wanting to try to get this change on staging tonight for QA verification.

lognaturel (approved):
Wonderful, thanks for adding the @ too!

PR #1940, opened 2026-09-17
OpenRosa formList: use versioned paths

### LINKED ISSUE #2251: Form list `downloadUrl` and `manifestUrl` are consistent between form versions
#### Problem description

The Open Rosa form list response that Central serves for projects returns the same `downloadUrl` and `manifestUrl` for a given form ID for every version. This creates a race condition if a form update occurs between a client fetching the form list and actually downloading the form: the form downloaded by the client might actually be an update of the version in the form list response meaning the metadata of the file downloaded won't match the form listing.

This creates even more problems when a form has media files - in the scenario above, the manifest obtained could be out of date or newer (depending on when it's fetched) by the time the form file is downloaded, resulting in the client fetching incorrect or missing media files.

#### Steps to reproduce the problem

1. Create a project with a form and app user
2. Fetch the form list for the app user
3. Update the form
4. Fetch the form list again

#### Expected behavior

It's not specified in the [Form List API spec](https://docs.getodk.org/openrosa-form-list/#elements-within-xform), but I'd expect `downloadUrl` and `manifestUrl` to be unique to a form ID **AND** version. 

The `downloadUrl` and `manifestUrl` requests cannot be made "atomically" together, so if they are consistent between versions, there is always a chance that the responses are for different versions.

#### Central version shown in version.txt

```
65a2cd573423f7ab4bae962a838702dac5a634fb (v2026.3.0-3-g65a2cd5)
 89a69159eb1081dc6383d057138440a4f4dc505b client (v2026.3.0-6-g89a69159e)
 94f3b82e034d2aa3a07b56573e994c7479490f0a server (v2026.3.0)
```

matthew-white:
Closes getodk/central#2251.

Updated tests.

I followed the recommendations in the issue, as well as [discussion](https://getodk.slack.com/archives/C01TKJ7TG7J/p1789676099196999) in Slack.

I don't view this change as being particularly risky.


matthew-white:
S3 e2e tests are currently failing, but I don't think that's related to this change. I've filed an issue about that at getodk/central#2255.

lib/formats/openrosa.js
```diff
@@ -47,16 +48,18 @@ const formListTemplate = template(200, `<?xml version="1.0" encoding="UTF-8"?>
       <name>{{def.name}}{{^def.name}}{{xmlFormId}}{{/def.name}}</name>
       <version>{{def.version}}</version>
       <hash>md5:{{def.hash}}</hash>
-      <downloadUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}.xml</downloadUrl>
+      <downloadUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}{{^draft}}/versions/{{urlSafeVersion}}{{/draft}}.xml</downloadUrl>
```
matthew-white, line 51: An alternative I thought about was to define a `versionPath` for each form in `formList()` below:

```html
<downloadUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}/{{versionPath}}.xml</downloadUrl>
```

```js
versionPath: data.draft ? 'draft' : url`versions/${form.def.version || Form.EmptyVersion}`
```

However, when I did that, mustache encoded the `/` after `versions` as `&#x2F;`. I could prevent that by using `{{{ }}}` instead of `{{ }}`. However, I hesitated to do that, since I saw that `urlSafeXmlFormId` used `{{ }}`, i.e., it's both URL-encoded and encoded for XML.

When I read about `encodeURIComponent()`, I do feel like it'd be safe to use `{{{ }}}`. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#description), `encodeURIComponent()` "escapes all characters **except**:"

```
A–Z a–z 0–9 - _ . ! ~ * ' ( )
```

I don't think any of those characters need to be encoded for XML in this context. `'` might need to be encoded in an XML attribute, but not here. In other words, given that it's URL-encoded, it's also effectively encoded for XML.

I'd be happy to take that alternative approach if it seems preferable. However, I also think the current approach is reasonable.

lib/model/query/forms.js
```diff
@@ -593,7 +593,7 @@ const versionJoinCondition = (defVersion, enketoId) => {
 
   if (defVersion == null) throw new Error('Must request a specific defVersion or defVersion class');
   else if (defVersion === Form.WithoutDef) throw new Error('No JOIN should be performed if WithoutDef');
-  else if (defVersion === '___') return versionJoinCondition('');
```
matthew-white, line 596: I searched the codebase for `___`, and this was the only current usage in lib/.

matthew-white (commented):
Adding notes to facilitate async code review.

lib/formats/openrosa.js
```diff
@@ -47,16 +48,18 @@ const formListTemplate = template(200, `<?xml version="1.0" encoding="UTF-8"?>
...
       {{#aux.openRosa.hasAttachments}}
-      <manifestUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}/manifest</manifestUrl>
+      <manifestUrl>{{{domain}}}{{{basePath}}}/forms/{{urlSafeXmlFormId}}{{#draft}}/draft{{/draft}}{{^draft}}/versions/{{urlSafeVersion}}{{/draft}}/manifest</manifestUrl>
```
lognaturel, line 53: In the draft case, we allow modifications to the form without a change to the version so adding the version to the URL doesn't always prevent the race condition issues we're trying to address, right? I think that's fine but want to double check my understanding.

matthew-white: Yes, that sounds right to me. If a new form definition is uploaded for the draft, that won't change the `<downloadUrl>`. If a form attachment is updated as is allowed for form drafts, that won't change the `<manifestUrl>`. We have specialized OpenRosa endpoints for form drafts. They don't include the form version in the URL.

https://github.com/getodk/central-backend/blob/94f3b82e034d2aa3a07b56573e994c7479490f0a/lib/resources/forms.js#L488-L510

matthew-white: The form `draftToken` (the token after /test/) changes once the draft is published, right? But I think you're right that the form version can change without the URL changing as the user iterates on the draft.

matthew-white: I was thinking more about this comment on the issue:

> I'd expect `downloadUrl` and `manifestUrl` to be unique to a form ID **AND** version.

One idea: we could add an optional `?version` query parameter to the draft endpoints for the form XML and form manifest. If Backend sees the query parameter, it can verify that the version string of the current form draft matches the query parameter and return an error if not.

But if the draft-related race condition isn't a big deal, I'd be just as happy to leave things as-is.

docs/api.yaml
```diff
@@ -10174,15 +10180,15 @@ paths:
                     <name>basic</name>
                     <version></version>
                     <hash>md5:a64817a5688dd7c17563e32d4eb1cab2</hash>
-                    <downloadUrl>https://your.odk.server/v1/projects/7/forms/basic.xml</downloadUrl>
-                    <manifestUrl>https://your.odk.server/v1/projects/7/forms/basic/manifest</manifestUrl>
+                    <downloadUrl>https://your.odk.server/v1/projects/7/forms/basic/versions/___.xml</downloadUrl>
```
lognaturel: What do you think about adding a `version` value of either `v1` or `20260920` above and update all the URLs accordingly? I think that would be much more helpful to understand what's going on just looking at the example.

lib/formats/openrosa.js
lognaturel: Unless I'm missing something, nothing we do to include version for drafts will help because we allow changing the actual contents without changing the version in the draft case.

We currently only allow 1 draft form per project so that greatly reduces the chance of the weird race conditions. Let's leave things as they are. If/when we have a way to have a draft project with multiple forms we may want to require that each change to the form def or attachments results in a version change.

lognaturel (approved):

docs/api.yaml
matthew-white: Done in 521a1cadace1e70dc8135cbae0e4b029a6b4cbef

PR #1941, opened 2026-09-18
test/e2e/s3: replace minio with pgsty/silo

### LINKED ISSUE #2255: S3 e2e tests: pull access denied for minio/minio
https://github.com/getodk/central-backend/actions/runs/35280126866/job/105399883052

```
Unable to find image 'minio/minio:latest' locally
docker: Error response from daemon: pull access denied for minio/minio, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

Run 'docker run --help' for more information
make: *** [Makefile:66: fake-s3-server-persistent] Error 125
```

Initially, I thought this was the same as #2193. However, there are a couple of differences, so I thought I should file a separate issue:

1. The error seems consistent, not intermittent
2. The specific error response is different: "pull access denied" as opposed to "connection reset by peer"

CC @alxndrsn

alxndrsn:
Closes https://github.com/getodk/central/issues/2255

- [x] CI
- [x] ran locally
- [ ] checked that data stored from `minio/minio` is still available after switch to `pgsty/silo`

Drop-in replacement of unpublished minio docker image (https://hub.docker.com/r/minio/minio/).

No effect - just test code.

No.


matthew-white:
@alxndrsn just FYI, I see that `oidc-integration-test` is failing here in a similar way as it did for me in #1940. I copied some of the logs from your failure to getodk/central#2257.

Makefile
```diff
@@ -48,16 +48,16 @@ dev-s3: fake-s3-accounts base
 	NODE_CONFIG_ENV=s3-dev npx nodemon --watch lib --watch config lib/bin/run-server.js
 
 # default admin credentials: minioadmin:minioadmin
-#   See: https://hub.docker.com/r/minio/minio/
+#   See: https://hub.docker.com/r/pgsty/silo/
 # MINIO_KMS_SECRET_KEY, MINIO_KMS_AUTO_ENCRYPTION enable encryption - this changes how s3 ETags are generated.
 #   See: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html
-#   See: https://github.com/minio/minio/discussions/19012
+#   See: https://github.com/pgsty/silo/discussions/19012
```
matthew-white: Probably this link should stay the same, right? The new one is a 404.

Makefile
```diff
@@ -48,16 +48,16 @@ dev-s3: fake-s3-accounts base
...
 S3_SERVER_ARGS := -p 127.0.0.1:9000:9000 -p 127.0.0.1:9001:9001 \
 		-e MINIO_ROOT_USER=odk-central-dev \
 		-e MINIO_ROOT_PASSWORD=topSecret123 \
 		-e MINIO_KMS_AUTO_ENCRYPTION=on \
 		-e MINIO_KMS_SECRET_KEY=odk-minio-test-key:QfdUCrn3UQ58W5pqCS5SX4SOlec9sT8yb4rZ4zK24w0= \
-		minio/minio server /data --console-address ":9001"
+		pgsty/silo:RELEASE.2026-09-03T13-18-01Z server /data --console-address ":9001"
```
matthew-white, line 60: You mentioned on the issue:

<blockquote>

## Recommendation

Switch to Silo.

Pros:

&hellip;

Cons:

* looks AI-heavy

Con can be mitigated by version pinning.
</blockquote>

How often would we upgrade Silo? Or would we try to stay on this pinned version for as long as possible?

I feel like we usually try to keep our dependencies up-to-date. I hear what you're saying though about it being AI-heavy. Mostly I'm just curious to hear more about the plan going forward. Do you feel like there are any risks associated with staying on a pinned version?

Makefile
alxndrsn:
```suggestion
#   See: https://github.com/minio/minio/discussions/19012
```

Makefile
alxndrsn: > How often would we upgrade Silo?

Infrequently?

> Do you feel like there are any risks associated with staying on a pinned version?

It's a balance.  The hope is that an old version with a serious security issue would get removed from the docker repos, whereas staying on `latest` introduces the risk of a rogue release.  The ideal is updating with a cooldown period, but I don't think docker supports this natively.  Dependabot may be an option for automating these updates with a cooldown.

Makefile
alxndrsn: :+1: reverted

Makefile
matthew-white: What do you think about updating it every time that Backend dependencies are updated before regression testing, so about four times a year? I could add it to getodk/central#2237 as a to-do item and note that the cooldown period should be respected.

It stands out to me that I don't think we consider a cooldown period for other Docker images in dev:

- There doesn't seem to be any version-pinning for sphinx:

https://github.com/getodk/central-backend/blob/b3c50006b3002563bcb050fc3a3994f011748f86/odk-docs.dockerfile#L1

- For postgres, we just pin to v14, nothing more specific:

https://github.com/getodk/central-backend/blob/b3c50006b3002563bcb050fc3a3994f011748f86/test/bin/docker-postgres.sh#L77

So part of me thinks that we should start pinning and regularly updating those dependencies as well. Does that sound right? E.g., we could update Postgres in dev whenever we update it in prod.

Or is a cooldown period more important for Silo compared to e.g. Postgres?

alxndrsn: I agree that PostgreSQL should match prod.

I think Sphinx should also be pinned.

I don't have strong opinions about when Silo or Sphinx should be updated.

matthew-white (approved):

Makefile
matthew-white: > I agree that PostgreSQL should match prod.

I've made a note of this to-do at getodk/central#2236.

> I think Sphinx should also be pinned.

I've filed an issue about this at getodk/central#2259.

> I don't have strong opinions about when Silo or Sphinx should be updated.

For now, I've added them to the checklist at getodk/central#2237. I think it's a good default to update them with every release, but let's see if anyone can think of a different approach.

PR #1942, opened 2026-09-18
Order forms returned with dataset metadata

matthew-white:
This PR sorts the `linkedForms` and `sourceForms` returned to users as part of dataset metadata.

Along with #1943, this PR closes getodk/central#2257. By sorting these lists of forms, this PR addresses the last remaining test failure of getodk/central#2257:

```
  1) datasets and entities
       listing and downloading datasets
         projects/:id/datasets/:name GET
           should not return a linked form that has been deleted:

      AssertionError: expected Array [ 'consumeDatasets2', 'consumeDatasets' ] to equal Array [ 'consumeDatasets', 'consumeDatasets2' ] (at '0', A has 'consumeDatasets2' and B has 'consumeDatasets')
      + expected - actual

       [
      -  "consumeDatasets2"
         "consumeDatasets"
      +  "consumeDatasets2"
       ]
```

Here is the failing assertion from that test:

https://github.com/getodk/central-backend/blob/94f3b82e034d2aa3a07b56573e994c7479490f0a/test/integration/api/datasets.js#L1992-L1997

`linkedForms` and `sourceForms` are surfaced in Frontend, so there is some (hopefully positive) user impact. I think it's nice to sort these forms for users. In other words, this PR was prompted by getodk/central#2257, but it also has the nice effect of sorting the forms listed in the `Infonav` component shown on entity list pages.

#### What has been done to verify that this works as intended?

Tests continue to pass.

#### Why is this the best possible solution? Were any other approaches considered?

I used different sorting strategies for `linkedForms` vs. `sourceForms`:

`sourceForms`: I just added `ORDER BY name` to the query. `_getSourceForms()` is only used in `Datasets.getMetadata()`.

`linkedForms`: I thought about adding `ORDER BY name` to `_getLinkedForms()`. However, that function is also used in `Datasets.canReadForOpenRosa()`, which is called by the entity integrity URL. `canReadForOpenRosa()` doesn't need the forms to be sorted, so I hesitated to always order within `_getLinkedForms()`. I thought about adding an argument to `_getLinkedForms()` for whether to sort or not, but then I noticed that we have this function:

https://github.com/getodk/central-backend/blob/94f3b82e034d2aa3a07b56573e994c7479490f0a/lib/model/query/datasets.js#L405-L409

[...]

PR #1943, opened 2026-09-18
getRelatedDatasetsOfForm: order results consistently

### LINKED ISSUE #2257: Intermittent test failures: dataset query module
https://github.com/getodk/central-backend/actions/runs/35305752041/job/105484579618

```
  2506 passing (4m)
  5 pending
  2 failing

  1) datasets and entities
       dataset query module
         getLinkedDatasets
           should not return deleted properties:

      AssertionError: expected Array [ 'trees', 'people' ] to equal Array [ 'people', 'trees' ] (at '0', A has 'trees' and B has 'people')
      + expected - actual

       [
      -  "trees"
         "people"
      +  "trees"
       ]
      
      at Assertion.fail (node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (test/integration/api/datasets.js:11007:29)
      at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
      at async Object.createConnection (node_modules/slonik/dist/src/factories/createConnection.js:97:18)

  2) datasets and entities
       dataset query module
         getRelatedDatasetsOfForm
           should return target dataset for a form that only writes to a dataset:

      AssertionError: expected Array [ 'age', 'first_name' ] to equal Array [ 'first_name', 'age' ] (at '0', A has 'age' and B has 'first_name')
      + expected - actual

       [
      -  "age"
         "first_name"
      +  "age"
       ]
      
      at Assertion.fail (node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (test/integration/api/datasets.js:11030:69)
      at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
      at async Object.createConnection (node_modules/slonik/dist/src/factories/createConnection.js:97:18)



make[1]: *** [Makefile:123: test-integration] Error 2
make[1]: Leaving directory '/home/runner/work/central-backend/central-backend'
make: *** [Makefile:22: test-oidc-integration] Error 2
```

matthew-white:
Along with #1942, this PR closes getodk/central#2257. It sorts the results from the following methods of the `Datasets` query module:

- `Datasets.getLinkedDatasets()`
- `Datasets.getTargetDatasetsAndProperties()`
- `Datasets.getRelatedDatasetsOfForm()`

Tests pass (hopefully now more consistently).

To try to figure out what the result order should be, I searched the codebase for usage of the methods. They have limited usage:

- `Datasets.getLinkedDatasets()` is just called by `Datasets.getRelatedDatasetsOfForm()`
- `Datasets.getTargetDatasetsAndProperties()` is just called by `Datasets.getRelatedDatasetsOfForm()`
- `Datasets.getRelatedDatasetsOfForm()` is just called by the endpoint to delete a form, in order to log details in the audit log:

https://github.com/getodk/central-backend/blob/94f3b82e034d2aa3a07b56573e994c7479490f0a/lib/resources/forms.js#L379-L385

Given that, I feel like the exact order doesn't matter too much and that the most important thing is to make it consistent.

I first sort by dataset (by dataset name), then by property. These `Datasets` methods seem to return both published and draft datasets, but I don't incorporate that in the order. When sorting by property, I tried to match the order in `Datasets.getProperties()`. However, I don't order by `form_fields.order`, as `form_fields` isn't currently part of the queries. I don't feel like the order matters enough for it to be worth joining in `form_fields`. Here's the order in `Datasets.getProperties()`:

https://github.com/getodk/central-backend/blob/94f3b82e034d2aa3a07b56573e994c7479490f0a/lib/model/query/datasets.js#L574

I think the risk of regression is low.

I don't think we need to document this order in the API docs.


matthew-white:
S3 e2e tests are failing, but that's expected given getodk/central#2255.

matthew-white:
I've tagged both @sadiqkhoja and @alxndrsn for code review, but I'm only looking for a single checkmark.

alxndrsn (approved):
> **Were any other approaches considered?**
> ...
> Given that, I feel like the exact order doesn't matter too much and that the most important thing is to make it consistent.

One alternative that isn't mentioned is sorting the results in the frontend.

Adding sorting to datasets queries might might degrade performance further.  This decision could be revisited if it proves problematic.

matthew-white:
Frontend does some limited sorting today (for the project list and the form list). It could certainly do more if need be.

Though on the flip sides, in some cases, I wish Frontend did more pagination, which would point toward Backend doing the sorting. E.g., the form list can get really long, so I think pagination there would be nice.

In this particular case, the list only appears in audit log details. Currently, Frontend renders those very minimally, just displaying the raw JSON. Yet I think it's still nice for the order within the audit log details to be deterministic.

I agree that we can revisit the sorting here if it becomes problematic. It's only done as part of audit logging during form deletion. Form deletion is hopefully not the most common operation, so I think it's less important that it's quite as optimized as, say, getting the project list or getting the list of entity lists. But if there's a noticeable slowdown, we could definitely revisit.

PR #1944, opened 2026-09-18
Increase timeout for oidc-integration-test

matthew-white:
I haven't done much work in Backend for a bit, but as I've been pushing PRs over the past day, I've noticed that `oidc-integration-test` fails somewhat frequently due to timeout. It ends up getting canceled. If I rerun it enough times, it eventually passes. However, rather than rerunning, I think it'd be better to increase the timeout.

Tests continue to pass (hopefully more consistently).

Is it unexpected that it's taking longer than before? If so, that's something we could investigate. (I'd be happy to file an issue about that.) However, even if that's the case, I think it'd be nice to go ahead and increase the timeout for now. That way, it won't be necessary to rerun OIDC tests for unrelated PRs.

I increased the timeout from 6 minutes to 9, just adding 50%. I could try a smaller number instead if that'd be preferable.

No risk of regression; just a change to tests.


alxndrsn (approved):
> I've noticed that `oidc-integration-test` fails somewhat frequently due to timeout

I've had a quick look at https://github.com/getodk/central-backend/actions/workflows/oidc-integration.yml, and I agree it's happening a lot!

There was a big jump when postgres logs were added in https://github.com/getodk/central-backend/pull/1924 - it can take ~1 minute for the logs to be output.

I'm up for increasing the timeout, but I think also it would help to remove postgres logs from the full-test-suite jobs.

alxndrsn:
> I think also it would help to remove postgres logs from the full-test-suite jobs.

@matthew-white I've filed https://github.com/getodk/central-backend/pull/1945 to address this.

matthew-white:
I've approved #1945, thank you for that PR! ✅

Should I still merge this PR, or should we see if #1945 is sufficient?

One thing that stands out to me is that standard-suite.yml has a much higher timeout than oidc-integration.yml (20 minutes vs. 6):

https://github.com/getodk/central-backend/blob/b2596c68d65272262aae7c187290fd4f172e5374/.github/workflows/standard-suite.yml#L9

standard-suite.yml does do some things that oidc-integration.yml doesn't, e.g., linting and unit tests, but those things are pretty fast. The slowest part is running integration tests, and both workflows do that.

So part of me thinks that increasing the timeout here would bring oidc-integration.yml more in line with standard-suite.yml. Then again, the timeout in standard-suite.yml is probably higher than it needs to be.

matthew-white:
I'm going to go ahead and merge. I don't think there's much risk to increasing the timeout. It brings `oidc-integration` in line with `standard-suite`. I see this PR as being consistent with #994, where the idea was:

> I've set the timeouts as approximately double the time I've seen for recent ✅ green builds.

`oidc-integration` often takes 5 minutes or more, so if anything, the increase here might be too low.

